PR #566, opened 2014-12-16
#1596 90 above

vsmalladi:


src/encoded/upgrade/donor.py
```diff
@@ -19,6 +20,7 @@ def donor_0_2(value, system):
 
 
 @upgrade_step('mouse_donor', '2', '3')
+@upgrade_step('human_donor', '2', '3')
```
lrowe: This applies the mouse_donor upgrade step removing sex from all human_donors...

PR #578, opened 2015-01-07
1794 multi object forms (not enabled)

davisagli:
This has the same changes as the 1794-multi-object-forms branch to allow for editing child objects in a form, but doesn't actually link to the forms for any types except page and image.

### DIFF
--- a/src/encoded/calculated.py
+++ b/src/encoded/calculated.py
@@ -88,6 +88,12 @@ class CalculatedProperties(object):
     def __init__(self):
         self.category_cls_props = {}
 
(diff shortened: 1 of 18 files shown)


davisagli:
I just updated this so it shows the forms for users in the forms group (ticket 2552). The forms permission is assigned to group.forms on the root ACL and checked there so we don't have to index it in elasticsearch; maybe there's a better way?

lrowe:
Looks good. Better than the alternative of indexing the forms permission, this way adding a user to the forms group will have immediate effect. Hopefully this will all clean up once I'm able to move ES lookup into the traverser giving us real views.

PR #590, opened 2015-01-22
make sure user's session has a csrf token as soon as they log in

davisagli:
Make sure the user's session has a CSRF token once they've logged in. I think this fixes http://redmine.encodedcc.org/issues/2527

Although, as a side note, I'm puzzled why the /login view appears to not require a CSRF token in the browser but test_login_no_csrf in test_persona.py shows it failing without a CSRF token.

### DIFF
--- a/src/encoded/persona.py
+++ b/src/encoded/persona.py
@@ -112,6 +112,7 @@ def login(request):
         request.session['user_properties'] = {}
         request.response.headerlist.extend(forget(request))
         raise LoginDenied()
+    request.session.get_csrf_token()
     request.session['user_properties'] = embed(request, '/current-user', as_user=userid)
     request.response.headerlist.extend(remember(request, 'mailto.' + userid))
     return request.session


lrowe:
The javascript retries login on a CSRF failure: https://github.com/ENCODE-DCC/encoded/blob/master/src/encoded/static/components/mixins.js#L191

This works because on CSRFTokenError refresh_session is called: https://github.com/ENCODE-DCC/encoded/blob/master/src/encoded/validation.py#L19

I see this issue too, but I can't quite work out why its happening as I thought on first edit you would need to have a csrf token...

davisagli:
Hmm, I'm not seeing that retry happen though. In the FF network panel I only see one request to /login, and it doesn't include an X-CSRF-Token request header, but does succeed and set a session cookie.

Before this PR, request.get_csrf_token is only called in the session view (I can't see where that is used apart from tests?), the logout view, and the refresh_session helper...so it makes sense that the first validation would fail.

lrowe:
Right, worked out what's happening. I was expecting that first login POST to be checked by the CSRF tween at https://github.com/ENCODE-DCC/encoded/blob/master/src/encoded/renderers.py#L175

But at that point the login namespace is 'persona' (based on the persona assertion itself) rather than 'mailto' (based on the session) so the user is not challenged. Now, I think that's ok from a CSRF perspective because the persona assertion is tied to our domain...

That means your fix is the correct one.

lrowe:
Actually, that's not right. We do need the CSRF protection for the persona login attempt too. Fixed in master now.

davisagli:
Yep, looks good. Thanks!

PR #592, opened 2015-01-22
add test for validation of child add permission

davisagli:


src/encoded/tests/test_post_put_patch.py
```diff
@@ -196,6 +196,25 @@ def test_put_object_adding_child(content_with_child, testapp):
     assert len(res.json['reverse']) == 2
 
 
+def test_submitter_put_object_adding_disallowed_child(content_with_child, submitter_testapp):
+    from ..contentbase import LOCATION_ROOT
+    root = submitter_testapp.app.registry[LOCATION_ROOT]
+    root['testing-link-sources'].__acl__ = ()
+    from pyramid.security import Allow
+    root['testing-link-targets'].__acl__ = ((Allow, 'group.submitter', 'edit'),)
```
lrowe: As `app` is session scoped, we need to be careful about changing stuff on fixtures. Use the pytest monkeypatch fixture like in https://github.com/ENCODE-DCC/encoded/blob/master/src/encoded/tests/test_post_put_patch.py#L133. http://pytest.org/latest/monkeypatch.html

src/encoded/tests/test_post_put_patch.py
```diff
@@ -196,6 +196,25 @@ def test_put_object_adding_child(content_with_child, testapp):
     assert len(res.json['reverse']) == 2
 
 
+def test_submitter_put_object_adding_disallowed_child(content_with_child, submitter_testapp):
+    from ..contentbase import LOCATION_ROOT
+    root = submitter_testapp.app.registry[LOCATION_ROOT]
```
lrowe: The `root` fixture gives you the same object.

src/encoded/tests/test_post_put_patch.py
davisagli: Thanks, was wondering about that. Updated.

PR #597, opened 2015-01-27
2518 Pipeline UI with 1400 and master merged in

forresttanaka:
Includes a fix for the Jest test failure, caused by a bug in the software component. This might be the first time a Jest test detected a software bug, which is very cool.


src/encoded/schemas/file.json
```diff
@@ -388,6 +390,39 @@
...
+            "title": "Analysis steps" ,
+            "description": "The runs of the step the file was generated from.",
+            "type": "array",
+            "items": {
+                "title": "Analysis step",
+                "description": "The run instance of the step the file was generated from.",
+                "comment": "For steps, see analysis_step_run.json (and analysis_step.json) - may be virtual",
+                "type": "string",
+                "linkTo": "analysis_step_run"
+            }
```
lrowe: @vsmalladi surely a file can only ever be generated by exactly one analysis_step_run? And can't the pipeline be found through the analysis_step_run? (We may want to address this after merging this lot.)

vsmalladi: @lrowe This is to address fantom analysis_step_run that generate files that we don't capture from uniform pipeline but are runs. An example is the methylation filtering step, which we don't capture the filtered fastq's but they are the inputs into the alignments step.

As for the pipeline, ya we can remove that from file schema I think. Was an artifact of an old version of the pipeline.

vsmalladi: @lrowe Actually the reason that file has pipeline is because any given analysis_step_run can be part of multiple "pipelines" if that step doesn't change but a downstream step does. Its metadata mostly for us, but might prove useful.

lrowe: @vsmalladi this should be done like the file.dataset / experiment.related_files. A file has a single analysis_step_run which generated it. Other analysis_step_runs may reference it as a property of the anlysis_step_run and should not require PATCHing the file. You can then get the pipeline from file.analysis_step_run.workflow_run.pipeline

src/encoded/schemas/analysis_step.json
```diff
@@ -0,0 +1,123 @@
...
+    ],
+    "properties": {
+        "schema_version": {
+            "default": "1"
+        },
+        "name": {
+            "title": "Name",
+            "description": "Unique name of the analysis step",
+            "type": "string",
+            "pattern": "^[a-z0-9\\-\\_]+"
```
lrowe: This pattern needs anchoring to the end of the string too ($)

src/encoded/schemas/pipeline.json
```diff
@@ -0,0 +1,172 @@
...
+    ],
+    "properties": {
+        "schema_version": {
+            "default": "1"
+        },
+        "name": {
+            "title": "Name",
+            "description": "Unique name of the software package, lower cased version of title.",
+            "type": "string",
+            "pattern": "^[a-z0-9\\-\\_]+",
```
lrowe: Pattern needs anchoring to end of string here too.

src/encoded/schemas/quality_metric.json
```diff
@@ -0,0 +1,94 @@
...
+    "properties": {
+        "schema_version": {
+            "default": "1"
+        },
+        "step": {
+            "title": "Analysis Step",
+            "description": "Reference to analysis step run in pipeline",
+            "type": "string",
+            "linkTo": "analysis_step_run"
+        },
```
lrowe: Should this maybe be renamed to 'step_run' to avoid confusion with 'analysis_step'?

src/encoded/schemas/file.json
vsmalladi: @lrowe workflow_run will only be instatiated for pipelines that we run. Should figure out a better way for lab pipelines then otherwise we could have fantom workflow_run's as well. Will talk to Ben tommorow.

src/encoded/schemas/quality_metric.json
vsmalladi: Ya we that might be better. Adding to list of things to fix.

src/encoded/schemas/pipeline.json
vsmalladi: @lrowe Thanks for catching this.

lrowe:
Merging as is, fixes to go in a new branch.

PR #612, opened 2015-01-29
2614 audit embed level

cricketsloan:
All done except for the antibody_characterization.  It did not seem to be the solution to Esther's issue


lrowe:
This was 2612

PR #616, opened 2015-01-29
2518 Pipeline updates

forresttanaka:
Merges in Venkat’s 1400-file-fix along with UI changes for that. Also includes loading in specific version of dagre-d3.


src/encoded/schemas/file.json
```diff
@@ -417,17 +417,12 @@
             "type": "string",
             "linkTo": "pipeline"
         },
```
lrowe: @vsmalladi Do we still want "pipeline" here? We can now get to it via `step_run.workflow_run.pipeline`, though it looks like analysis_step_run does not require "workflow_run" at the moment.

PR #623, opened 2015-02-02
2431 antibody char attachment req

estherc:
#2431 Make attachments a requirement in characterizations.


lrowe:
Closing as merged earlier version without upgrade.

PR #631, opened 2015-02-05
#1400 pipeline fixes and added suggetions

vsmalladi:


src/encoded/types/pipeline.py
```diff
@@ -62,6 +63,15 @@ class AnalysisStepRun(Item):
...
 
+    @calculated_property(schema={
+            "title": "Pipeline",
+            "type": "string",
+            "linkTo": "pipeline"
+        })
+    def pipeline(self, request, workflow_run=None):
+        if workflow_run is not None:
+            return request.embed(workflow_run, '@@object').get('pipeline')
+
```
lrowe: @vsmalladi Doesn't this calculated property belong in file.py?

vsmalladi: @lrowe Based on our previous conversation and since we are implementing workflow_runs for lab pipelines I thought it best to just keep the calculated property at the analysis_step_run.

vsmalladi:
@lrowe Updated code to calculate pipeline at file level.

PR #633, opened 2015-02-06
2518 graph updates for rc3

forresttanaka:
* Released and unreleased files are now merged into one graph.
* The graph isn’t displayed if no file in the experiment has a step_run.


forresttanaka:
One more change I forgot:

* Date added in the graph detail panel always displays as YYYY-MM-DD regardless of how it’s entered in the data. That fixes the long ISO 8601-formatted dates showing up.

PR #644, opened 2015-02-19
Add audit for host_organism comparison

estherc:
#2693 add audit to compare antibody_lot.host_organism in IP vs. control


estherc:
Let's take this out, it turns out we probably won't need this particular audit. I did write others but will make a new ticket for them.

PR #645, opened 2015-02-20
fix grunt watch

davisagli:
- upgrade browserify and watchify for some fixes when using 'external' deps
- don't export a require function for deps included in the browser bundle;
  it's unnecessary and somehow makes requiring jquery go into endless recursion
  if the bundle was built with watchify

Even with watchify the rebuilds are a bit slow, probably because it's a large bundle that still has to be run through minifyify on each change. We could probably fix this by splitting react and other vendor deps out to another bundle so that we're only rebuilding our own app as we edit the source.


lrowe:
I've added ``require: ['react']`` back to master as that is required for the React Chrome dev tools plugin to work.

davisagli:
Oh, right.

Shoot, that triggers the same error I was seeing: after "grunt watch" notices a change and rebuilds, I get a "Uncaught RangeError: Maximum call stack size exceeded" error in the console.

davisagli:
Instead can we just set window.React = React in browser.js?

davisagli:
Yeah, that seems to avoid the error and satisfy the dev tools.

davisagli:
9a1438a8b9a3243918226d57111c0fcf94fc234b

PR #653, opened 2015-02-26
828 batch download

nikhilRP:


src/encoded/views/download.py
```diff
@@ -0,0 +1,154 @@
...
+            for f in row['files']:
+                if 'files.file_format' in param_list:
+                    if f['file_format'] not in param_list['files.file_format']:
+                        continue
+                f['href'] = request.host_url + f['href']
+                for prop in file_attributes:
+                    value = ''
+                    if prop.split('.')[1] in f:
+                        value = f[prop.split('.')[1]]
+                    if prop.split('.')[1:][0] == 'accession':
```
lrowe: Instead of ``prop.split('.')[1:][0]`` just use ``prop.split('.')[1]``.

src/encoded/views/download.py
```diff
@@ -0,0 +1,154 @@
...
+            exp_data_row = []
+            new_data = []
+            for arr in dict_generator(row):
+                new_data.append(arr)
+            for column in header:
+                if not _tsv_mapping[column].startswith('files'):
+                    path = _tsv_mapping[column].split('.')
+                    temp = []
+                    for ld in new_data:
+                        if sorted(path) == sorted(ld[:-1]):
```
lrowe: Sorting seems wrong here, order matters in a path.

src/encoded/views/download.py
```diff
@@ -0,0 +1,154 @@
...
+    ('Experiment date released', ('date_released')),
+    ('Project', ('award.project')),
+    ('File format', ('files.file_format')),
+    ('Size', ('files.file_size')),
+    ('Output type', ('files.output_type')),
+    ('Lab', ('files.lab')),
+    ('md5sum', ('files.md5sum')),
+    ('href', ('files.href')),
+    ('Assembly', ('files.assembly'))
+])
```
lrowe: The extra parens here are unnecessary, just use ``('Assembly', 'files.assembly')``. Though maybe we want to use the same format as schema columns?

src/encoded/views/download.py
```diff
@@ -0,0 +1,154 @@
...
+            else:
+                yield pre + [key, value]
+    else:
+        yield indict
+
+
+@view_config(route_name='metadata', request_method='GET', renderer='tsv')
+def metadata_tsv(context, request):
+
+    param_list = parse_qs(request.matchdict['search_params'].encode('utf-8'))
```
lrowe: Any reason not to just use request.params here and supply them as the query string?

src/encoded/renderers.py
```diff
@@ -315,3 +319,28 @@ def reload_process(process):
...
+            if ct == response.default_content_type:
+                response.content_type = 'text/tsv'
+
+        fout = io.BytesIO()
+        writer = csv.writer(fout, delimiter='\t')
+
+        writer.writerow(value.get('header', []))
+        writer.writerows(value.get('rows', []))
+
+        return fout.getvalue()
```
lrowe: I don't think it's worthwhile to make a renderer here as the functionality isn't really generic. Better just to return a Response object from the view instead. That way you can use ``response.body_file`` as the file to write to.

src/encoded/views/download.py
nikhilRP: For joining multiple fields as single field. 
have to implement it. Few lines of code
Eg: age + units

src/encoded/renderers.py
nikhilRP: Ok. Will move it download.py

src/encoded/views/download.py
nikhilRP: This will change. 
I have to rewrite this part to support path with more than 2 fragments.

eg: "files.platform.name"

src/encoded/views/download.py
nikhilRP: Yeah I have to modify

```
dict_generator
```

to have paths in order and remove sorting but as of now I don't have any conflicts.
Can you think of a path which will be a problem

nikhilRP:
@lrowe don't merge this yet I will make few changes and we can have a look

src/encoded/static/components/search.js
```diff
@@ -474,7 +476,7 @@ var AuditMixin = audit.AuditMixin;
     var Target = module.exports.Target = React.createClass({
         mixins: [PickerActionsMixin, AuditMixin],
         render: function() {
-            var result = this.props.context;
+            var result = this.props.context;4
```
lrowe: Typo?

nikhilRP: @lrowe fixed it.

PR #662, opened 2015-03-03
Upgrade React, etc

davisagli:
Updates:
- react to 0.12
- react-forms to latest master (unfortunately the release candidates that are compatible with react 0.12 don't actually work)
- react-bootstrap to 0.15.1

This definitely needs some manual testing, especially on IE 8 which I don't have handy.


davisagli:
Rebased on current master, now passes

PR #665, opened 2015-03-04
Update file.py

estherc:
#2710 - fix audit so that it checks to see if the linked target has investigated_as control to exclude control expts from the audit

### DIFF
--- a/src/encoded/audit/file.py
+++ b/src/encoded/audit/file.py
@@ -73,7 +73,7 @@ def audit_file_controlled_by(value, system):
     if value['dataset'].get('assay_term_name') not in ['ChIP-seq', 'RAMPAGE', 'CAGE']:
         return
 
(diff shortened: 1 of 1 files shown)


src/encoded/audit/file.py
```diff
@@ -73,7 +73,7 @@ def audit_file_controlled_by(value, system):
     if value['dataset'].get('assay_term_name') not in ['ChIP-seq', 'RAMPAGE', 'CAGE']:
         return
 
-    if 'target' in value['dataset'] and value['dataset']['target'].get('investigated_as') == 'Control':
+    if 'target' in value['dataset'] and 'control' in value['dataset']['target'].get('investigated_as'):
```
lrowe: This should say ``.get('investigated_as', [])`` as the default of get is ``None`` and ``'control' in None`` is an error.

PR #688, opened 2015-03-24
2560 lib pair end

cricketsloan:
Undoing the connection between lib.paired_ended and rep.paired_ended


lrowe:
Merged #689 instead.

PR #690, opened 2015-03-25
Fix various form regressions

davisagli:
I set out to fix issue 2748 and found a number of regressions in form functionality, mostly side effects of our changes to work with the newer react-forms library. Here are fixes and some tests.

- Don't try to load the form until persona auth is complete
- Fix bug where re-rendering the form generated a new form schema which ended up triggering an update to the form value, causing child objects to always be sent even when not edited.
- Fix bug where object picker no longer showed the text filter if no search results were found (fixes 2748)
- Fix showing form validation errors from server (react-forms changed the format it expects)
- Fix bug where the schema_version field was not disabled
- Fix bug where editing a child object did not send its id to the server, so the server tried to create a new child instead.
- Add executable feature specs

### DIFF
--- a/src/encoded/static/components/fetched.js
+++ b/src/encoded/static/components/fetched.js
@@ -117,8 +117,11 @@ var FetchedData = module.exports.FetchedData = React.createClass({
             }, this);
         }
 
-        if (!this.props.loadingComplete || !params.length) {
+        if (!params.length) {
             return null;
+        }
+        if (!this.props.loadingComplete) {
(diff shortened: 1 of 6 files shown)


davisagli:
Having trouble figuring out why that one test fails on chrome on travis, but works locally.

lrowe:
The travis tests run on Python 3.4 not 2.7, so that error message will be ``'bogus'...`` instead of ``u'bogus'...``

davisagli:
Ah, of course!

PR #697, opened 2015-03-26
Update file.json

cricketsloan:
adding in hdf5 and gff to the enum and the mapping


src/encoded/schemas/file.json
```diff
@@ -473,6 +475,7 @@
         "fastq": ".fastq.gz",
         "gff": ".gff.gz",
         "gtf": ".gtf.gz",
+        "hdf5": ".hdf5.gz",
```
lrowe: I think .h5 is the preferred extension, see Pac Bio (http://www.pacb.com/devnet/code.html) and some parts of the spec (http://www.hdfgroup.org/HDF5/doc/UG/UG_frame10Datasets.html). Also I think hdf5 is not gzipped as compression is applied to the internal data chunks within the file. (http://www.hdfgroup.org/HDF5/faq/compression.html)

PR #722, opened 2015-04-28
#2726 added project as required to award

vsmalladi:


lrowe:
Hmm, I think all the awards already have a project, perhaps we can avoid the upgrade? If not it needs adding to the list of DEFAULT_COLLECTIONS for upgrade.py

vsmalladi:
We can avoid the upgrade then. I just need to back out changes regarding the upgrade.

PR #740, opened 2015-05-08
Update library.py

estherc:
Fixed casing mismatch between upgrade and enum option in fragmentation_method


lrowe:
For consistency with the other enums in library.json we should probably use the lowercase "n/a".

PR #745, opened 2015-05-13
This fixes "limit=all" ES timeout. "highlighting" search term is slow…

nikhilRP:
…ing down the query run time

### DIFF
--- a/src/encoded/views/search.py
+++ b/src/encoded/views/search.py
@@ -334,8 +334,10 @@ def search(context, request, search_type=None):
         query['sort'] = get_sort_order()
         query['query']['match_all'] = {}
         del query['query']['query_string']
+        del query['highlight']
     elif len(doc_types) != 1:
         del query['query']['query_string']['fields']
+        del query['highlight']
 
     # Setting filters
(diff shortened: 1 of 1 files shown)


src/encoded/views/search.py
```diff
@@ -334,8 +334,10 @@ def search(context, request, search_type=None):
         query['sort'] = get_sort_order()
         query['query']['match_all'] = {}
         del query['query']['query_string']
+        del query['highlight']
     elif len(doc_types) != 1:
         del query['query']['query_string']['fields']
+        del query['highlight']
```
lrowe: Shouldn't query['highlight'] instead be deleted when ``size > 25`` to catch the limit=all requests?

nikhilRP: We want to keep the highlight behavior consistent irrespective of the size of the results.
If user see "highlights" for "limit=25" and doesn't for bigger "limits" then there will be a issue.

Though I will work around to improve performances in future release I think this should be good for now.

nikhilRP:
@lrowe  highlighting only works if limit<=25

PR #756, opened 2015-05-14
adding in the new field

cricketsloan:
I am confused why the second of three build tests on travis does not work.


lrowe:
Closing pending further discussion.

PR #761, opened 2015-05-15
BDD tests for trackhubs

nikhilRP:


nikhilRP:
@lrowe merge this for now. I will rewrite in python very soon until then these will help test trackhubs

PR #815, opened 2015-06-09
Mapping boolean fields as string for now

nikhilRP:


nikhilRP:
@lrowe I reverted boolean fields back to string because of the issue mentioned in the ticket below.

https://3006-boolean-ecdcbc3-nikhil.demo.encodedcc.org

PR #831, opened 2015-06-17
Revert default schema value

cricketsloan:
In large merge I lost change of default from 1 to 2.  So the upgrades did not happen


cricketsloan:
This is only updating the version default to engage the upgrade

PR #836, opened 2015-06-19
#30 Enable Persona

forresttanaka:
This is a temporary UI to enable Persona login/logout. It appears in a little window-relative tab in the upper right of the header area.

As part of #29, I’m refactoring the navigation components to be more general, and I’ll implement the navigation you see in the upper right of the Tam/Selina UI document as a proposed final UI, including Persona login.

Sorry about the temporary PR confusion. I realized I used the ENCODED branch naming “standard” (which doesn’t actually exist) instead of the ClinGen one (which does), so I renamed it which closed the previous PR.


forresttanaka:
Wrong repo

PR #857, opened 2015-07-03
3037: add rudimentary user view including access key management

davisagli:
also includes:
- access keys and lab are now expanded in the user page view if user has view_details permission
- adds a link to the current user's profile in the user menu
- adds an impersonate-user view for use in tests and by admins

### DIFF
--- a/src/encoded/persona.py
+++ b/src/encoded/persona.py
@@ -122,7 +122,7 @@ def login(request):
 def logout(request):
     """View to forget the user"""
     request.session.get_csrf_token()
-    request.session['user_properties'] = {}
+    request.session.invalidate()
     request.response.headerlist.extend(forget(request))
     if asbool(request.params.get('redirect', True)):
         raise HTTPFound(location=request.resource_path(request.root))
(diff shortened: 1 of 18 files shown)


src/encoded/persona.py
```diff
@@ -146,3 +147,15 @@ def session(request):
...
+
+
+@view_config(route_name='impersonate-user', request_method='GET',
+             permission='impersonate')
+def impersonate_user(request):
+    """As an admin, impersonate a different user."""
+    request.session.get_csrf_token()
+    userid = request.params['userid']
+    request.session['user_properties'] = request.embed('/current-user', as_user=userid)
+    request.session['disable_persona'] = True
```
lrowe: I think we need to make the logout view call session.invalidate() so that the disable_persona key is unset. (Currently it just clears session['user_peroperties'].)

We also need to consider whether this is safe to be a GET given it is not CSRF protected.

davisagli: Good point. I'm mostly on vacation this week, so will take a look at this next week if not before.

PR #863, opened 2015-07-14
3059 signin move to footer

forresttanaka:
For Release 29.


davisagli:
@lrowe heads up that this will conflict with my changes in #857 since I also added a way for extra user actions to be supplied in the session.

lrowe:
@davisagli I think we might be ok as the changes don't actually overlap. This branch changes the code path that displays the 'Sign In' link, a few lines above the full user menu code that your branch changes.

PR #873, opened 2015-07-16
Update analysis_step.py

cricketsloan:
adding the post upgrade changes to the upgrade

### DIFF
--- a/src/encoded/upgrade/analysis_step.py
+++ b/src/encoded/upgrade/analysis_step.py
@@ -77,3 +77,22 @@ def analysis_step_2_3(value, system):
 
     # http://redmine.encodedcc.org/issues/3074
     del value['software_versions']
+    
+    # http://redmine.encodedcc.org/issues/3074 note 16 and 3073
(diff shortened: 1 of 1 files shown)


src/encoded/upgrade/analysis_step.py
```diff
@@ -77,3 +77,22 @@ def analysis_step_2_3(value, system):
...
+                            'rampage-grit-peak-calling-step-v-1-1'
+                            ]:
+        value['status'] = 'deleted'
+
+    if value.get('name') == 'lrna-pe-rsem-quantification-v-1':
+        value['parents'] = ['/analysis-steps/lrna-pe-star-alignment-step-v-1/',
+                            '/analysis-steps/index-rsem-v-1-0/']
+    elif value.get('name') == 'lrna-se-rsem-quantification-step-v-1':
+        value['parents'] = ['/analysis-steps/lrna-se-star-alignment-step-v-1/',
+                            '/analysis-steps/index-rsem-v-1-0/']
```
lrowe: In an upgrade, we operate on the 'raw' object, so 'parents' should use the uuids of those items.

PR #891, opened 2015-07-26
Use sentence case on antigen description

forresttanaka:


lrowe:
We decided against this in the end.

PR #912, opened 2015-08-14
3038: s3 attachments

davisagli:
This is not quite ready to merge but I've made good progress and wanted to give @lrowe a chance to look it over.

If the blob_bucket setting is set, we use an s3-backed BlobStorage implementation; otherwise the old rdb-backed one. The s3-backed storage falls back to the rdb-backed one for stored files with no bucket name stored.

Outstanding questions:

- Is there a reliable way to tell when we're behind a proxy and should use X-Accel-Redirect? The existing code in types/file.py checked for an Origin header, which wasn't there when I was testing access via the dev nginx. The X-Forwarded-For header set by nginx is swallowed by the PasteDeploy prefix middleware. For now I am comparing the port from the Host header to the SERVER_PORT to see if they differ, but that feels hacky.
- What buckets should we use in production and for demo servers? I've been using 'davisagli-test' for testing.

Remaining tasks:

- [x] calculate md5 checksum (ticket 478)
- [x] upgrade script to move existing files

### DIFF
--- a/src/encoded/commands/migrate_attachments_aws.py
+++ b/src/encoded/commands/migrate_attachments_aws.py
@@ -1,0 +1,73 @@
+"""\
+Move attachment blobs to S3.
+
+"""
+import copy
+import logging
+import transaction
+from pyramid.paster import get_app
+from pyramid.threadlocal import manager
+from pyramid.testing import DummyRequest
+from contentbase.interfaces import (
+    BLOBS,
+    DBSESSION
+)
+from contentbase.storage import PropertySheet
+
+EPILOG = __doc__
+
+logger = logging.getLogger(__name__)
+
+
+def run(app):
+    root = app.root_factory(app)
+    dummy_request = DummyRequest(root=root, registry=app.registry, _stats={})
+    manager.push({'request': dummy_request, 'registry': app.registry})
+    session = app.registry[DBSESSION]()
+    blob_storage = app.registry[BLOBS]
+
+    for sheet in session.query(PropertySheet).filter(PropertySheet.name == 'downloads'):
+        # Copy the properties so sqlalchemy realizes it changed after it's mutated
+        properties = copy.deepcopy(sheet.properties)
+        download_meta = properties['attachment']
+        if 'bucket' not in download_meta:
+            # Re-writing the blob while the S3BlobStorage is in use
+            # will move it to S3.
+            data = blob_storage.get_blob(download_meta)
+            blob_storage.store_blob(data, download_meta)
+            sheet.properties = properties
+            logger.info('Updated %s' % sheet.sid)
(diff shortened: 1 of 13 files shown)


lrowe:
> Is there a reliable way to tell when we're behind a proxy and should use X-Accel-Redirect? The existing code in types/file.py checked for an Origin header, which wasn't there when I was testing access via the dev nginx. The X-Forwarded-For header set by nginx is swallowed by the PasteDeploy prefix middleware. For now I am comparing the port from the Host header to the SERVER_PORT to see if they differ, but that feels hacky.

It's safe to assume that we'll always be behind an Nginx proxy. The checking of the Origin header is to avoid a redirect response to CORS XHR requests which browsers do not currently support.

> What buckets should we use in production and for demo servers? I've been using 'davisagli-test' for testing.

I've created `encoded-blobs` under the production account and `encoded-blobs-dev` under the dev account. I'm added bucket permissions to the `encoded-files-upload` user so we can use the same aws profile (access keys) as we do for uploading files to `encode-files` / `encoded-files-dev` (yes, different spellings on that pair.)

davisagli:
> It's safe to assume that we'll always be behind an Nginx proxy. The checking of the Origin header is to avoid a redirect response to CORS XHR requests which browsers do not currently support.

If we're assuming there's always a proxy, what's the reason for doing redirects at all?

Thanks for adding the buckets.

Currently struggling to get the build running since mock requires setuptools > 17.1. Any idea why it's "upgrading" to setuptools 15 even though I specified 18.1?

lrowe:
> If we're assuming there's always a proxy, what's the reason for doing redirects at all?

For blobs it makes sense to always proxy. Some of our external files are 60GB so proxying is not unproblematic. (We are however in the process of commissioning a beefy 10GbE proxy at Stanford to benefit from cheaper bandwidth egress charges so those files will probably end up proxied too.)

davisagli:
I think this is ready for review.

src/encoded/commands/migrate_attachments_aws.py
```diff
@@ -0,0 +1,73 @@
...
+        # Copy the properties so sqlalchemy realizes it changed after it's mutated
+        properties = copy.deepcopy(sheet.properties)
+        download_meta = properties['attachment']
+        if 'bucket' not in download_meta:
+            # Re-writing the blob while the S3BlobStorage is in use
+            # will move it to S3.
+            data = blob_storage.get_blob(download_meta)
+            blob_storage.store_blob(data, download_meta)
+            sheet.properties = properties
+            logger.info('Updated %s' % sheet.sid)
```
lrowe: Should we calculate the md5sum here? Existing blobs will be missing it.

PR #931, opened 2015-09-16
3200 3202 replicates ui

hitz:
Resolves 3200 and 3202.  I have a branch for 3199 that is branched off of this, but not sure it will get done by Mon/Tue.


lrowe:
The JavaScript tests (`npm test`) are failing here. I suspect it's because the javascript test fixtures in https://github.com/ENCODE-DCC/encoded/tree/master/src/encoded/static/components/testdata/file now need a biological_replicates property.

hitz:
I tested for existence of biological_repliates and npm test passes for me.

PR #937, opened 2015-09-23
3199 replicates plural graph

hitz:
Today I learned:  If you have a graph with a replicate, the graphy part of the graph are actually a sub graph of the replicate note.


hitz:
Closing to create PR from v32-hotfix

PR #970, opened 2015-10-12
Update library.json

cricketsloan:
Remove GSE from option


cricketsloan:
We have no data in the field atm.

PR #979, opened 2015-10-14
2787 qc compound

lrowe:


forresttanaka:
I don’t know what state this is in. Internet fell apart here. Gitter fell apart. Things might be back to normal; I don’t know.

PR #992, opened 2015-10-26
1415 Links to references

forresttanaka:
The biosample, experiment, dataset, human donor, and mouse donor pages can now display links to their PMID: and doi: references. When they appear, they link to the appropriate ENCODE publication page which itself has links to PubMed or source page as appropriate.

The references get displayed as their own line item in the summary panel of these pages.

### DIFF
--- a/src/encoded/types/donor.py
+++ b/src/encoded/types/donor.py
@@ -12,7 +12,7 @@ from .base import (
 
 class Donor(Item):
     base_types = ['Donor'] + Item.base_types
-    embedded = ['organism']
+    embedded = ['organism', 'references']
     name_key = 'accession'
     rev = {
         'characterizations': ('DonorCharacterization', 'characterizes'),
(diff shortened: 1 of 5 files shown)


src/encoded/types/donor.py
```diff
@@ -12,7 +12,7 @@
 
 class Donor(Item):
     base_types = ['Donor'] + Item.base_types
-    embedded = ['organism']
+    embedded = ['organism', 'references']
```
lrowe: The subtypes of Donor which override embedded need to have this change made too. Taking FlyDonor as an example, it can be cleaned up to first copy Donor's embeds.

Currently:
```python
    embedded = ['organism', 'constructs', 'constructs.target']
```
Should change to:
```python
    embedded = Donor.embedded + ['constructs', 'constructs.target']
```

forresttanaka:
@lrowe Cool; thanks. I have a somewhat urgent ClinGen thing to work on today so I’ll get back to this later.

forresttanaka:
The Fly and Worm schemas don’t include the references mixin. Should I really add Donor.embedded for fly and worm?

lrowe:
In that case it belongs on the HumanDonor and MouseDonor objects rather than the base Donor class.

forresttanaka:
Makes sense.

PR #999, opened 2015-10-29
2978 Experiment file graph filtering

forresttanaka:
Ticket [2978](http://redmine.encodedcc.org/issues/2978):

Display a menu with all files’ assemblies and annotations above the graph. Choosing an assembly and annotation displays only those matching files and any files they derive from — other files do not display.

This also includes more general graphing fixes for issues documented in the ticket.


.cache/v/cache/lastfailed
```diff
@@ -0,0 +1 @@
+{}
```
lrowe: This looks like something from pytest... We need to add .cache to .gitignore and not check it in.

forresttanaka:
Got it; I removed the .cache directory and added .cache/ to .gitignore.

Oddly, a local bdd test failed twice in a row, saying that www.uniprot.org wasn’t in the URL after clicking the link. But clicking the link manually definitely worked. Will monitor the remote Travis CI test.

PR #1017, opened 2015-11-12
Snp search

lrowe:


src/encoded/static/components/search.js
```diff
@@ -890,9 +902,10 @@ var AuditMixin = audit.AuditMixin;
             var facetdisplay = context.facets.some(function(facet) {
                 return facet.total > 0;
             });
+            console.log(searchBase);
             return (
                 <div>
-                    {facetdisplay ?
+                    {context['total'] ?
```
lrowe: I suspect this is the result of a bad merge. We need to work out if any of the other changes to this file are necessary now that we do not include the browser.

CC @forresttanaka

PR #1038, opened 2015-11-24
Remove pipeline dependency

forresttanaka:
Removes circular dependency causing software.js modules to not load.

### DIFF
--- a/src/encoded/static/components/software.js
+++ b/src/encoded/static/components/software.js
@@ -3,7 +3,6 @@ var React = require('react');
 var globals = require('./globals');
 var navbar = require('./navbar');
 var search = require('./search');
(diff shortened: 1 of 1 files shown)


src/encoded/static/components/software.js
```diff
@@ -125,20 +123,6 @@ var Software = module.exports.Software = React.createClass({
...
-        var context = this.props.context;
-        return (
-            <div>
-                <h3>Pipelines using software {context.title}</h3>
-                <PipelineTable {...this.props} softwareId={context['@id']}/>
-            </div>
-        );
-    }
-});
-
```
lrowe: Note that you can resolve a circular dependency by moving the `require(...)` from the module level to the function in which you use it.

PR #1039, opened 2015-11-24
remove links to matrix search

davisagli:
@hitz doesn't want the matrix search linked yet. Ideally we probably want this change applied to v35 but not to master.


lrowe:
We can just create a branch for v36 with a revert of this commit to avoid master and v35 branches from diverging just yet.

davisagli:
Yes, good.

David Glick
(mobile)

PR #1045, opened 2015-11-30
Remove treatment classification mixin

estherc:
#3473 removed treatment_classification from treatment_time_series mixins

### DIFF
--- a/src/encoded/schemas/treatment_concentration_series.json
+++ b/src/encoded/schemas/treatment_concentration_series.json
@@ -26,7 +26,7 @@
         }
     },
     "facets": {
-        "related_datasets.assay_term_name": {
+        "assay_term_name": {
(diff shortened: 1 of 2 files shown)


src/encoded/schemas/treatment_concentration_series.json
```diff
@@ -26,7 +26,7 @@
         }
     },
     "facets": {
-        "related_datasets.assay_term_name": {
+        "assay_term_name": {
```
lrowe: Is this change to facets intentional?

estherc: Yes, assay_term_name should be calculated in treatment_concentration_series and that should be what goes into the facet, not digging into related_datasets. All the other series should also have this in their schemas

PR #1064, opened 2015-12-11
2595 biosample links

forresttanaka:
The ```source``` and ```rnai.source``` objects of a biosample could have a ```title``` property but no ```url``` property. Previous code didn’t take this into account and put an ```<a>``` there regardless, causing a link that didn’t go anywhere. This branch changes this to check for ```url``` properties, and simply display the ```title``` without the link if no ```url``` property exists.


hitz:
curious about some data issue, commmented in redmine.

PR #1071, opened 2015-12-16
3063 array unique items

estherc:
#3063 - enforce unique values in array properties


hitz:
I think this should be OK to merge.  @lrowe maybe you can take a look?

lrowe:
For anything including upgrades I would log into the demo machine and check the apache log of the first indexing run to see if there are any upgrade / audit errors introduced.

PR #1085, opened 2016-01-07
splitting ERROR into WARNING and ERROR depending on the status of the experiment

gabdank:


hitz:
I don't understand what is going on here, nor am I sure that we should split WARNING/ERROR by experiment status.    I guess this is for: http://redmine.encodedcc.org/issues/3482 but it's not clearly specified...

gabdank:
Esther commented, that may it is not an ERROR for a proposed experiment not to have files associated with the replicates. I agreed that may be it is not an ERROR - but the lab should be WARNED that replicates have no files. Cricket agreed - so I made a fix to 3482.

hitz:
Well, I will confirm with Cricket that this is an acceptable pattern.
However, the code is a little repetitive; you should probably define level as a variable and have it set by the if value['status'] in ... 
That way if the message(s) or any metadata field change you won't have to change it in 2 places.

gabdank:
I can fix it. I confirmed it with Cricket - prior to implementing the FIX.

gabdank:
Removed duplication.

PR #1087, opened 2016-01-08
3556 ReferenceEpigenome search result biosample_term_name

forresttanaka:
Search results with ReferenceEpigenome objects now display:

Reference epigenome series in _biosample_term_name_

…for the linked title. This makes it the same as the other Series dataset search results.


forresttanaka:
Per [Cricket’s Redmine note], this PR is for 36rc2 if it ever exists. Otherwise for Release 37.

PR #1096, opened 2016-01-12
making integer division (to calculate read depth) in cases of paired ended sequencing in ChIP-seq experiments

gabdank:


hitz:
I merged this but deciding if it needs a new ticket for QA.

PR #1100, opened 2016-01-15
Fix bad frame name

lrowe:
Return a 404 Not Found instead returning None causing a later error to be thrown.


hitz:
There was no ticket for this, it was just an theoretical fix?

lrowe:
I noticed an error in the apache error log, but had not filed a bug.

PR #1126, opened 2016-01-27
3586 3578 3594 3603

gabdank:
3605 - support Infinity
3586 - use ontology in part_of audit
3578 - duplicated NTR messages
3594 - spearman correlation STD fix

### DIFF
--- a/src/encoded/audit/biosample.py
+++ b/src/encoded/audit/biosample.py
@@ -78,19 +78,6 @@ def audit_biosample_human_no_model_organism_properties(value, system):
         return
 
 
-@audit_checker('biosample', frame=['object'])
-def audit_biosample_term_id(value, system):
-    if value['status'] in ['deleted', 'replaced', 'revoked']:
-        return
-    biosample_prefix = value['biosample_term_id'].split(':')[0]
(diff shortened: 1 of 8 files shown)


src/encoded/audit/biosample.py
```diff
@@ -346,18 +346,18 @@ def audit_biosample_part_of_consistency(value, system):
...
         if (term_id in ontology) and (part_of_term_id in ontology):
-            ontology_child = ontology[term_id]
+            # ontology_child = ontology[term_id]
             ontology_parent = ontology[part_of_term_id]
             if 'name' in ontology_parent and ontology_parent['name'] == 'multi-cellular organism':
                 return
-
+            '''
             if 'organs' in ontology_child and 'organs' in ontology_parent:
                 child_organs = ontology_child['organs']
```
hitz: I don't think you should comment out this code unless you plan on bringing it back very soon.

hitz:
You commented out a chunk of an audit -- I think this is only OK if you plan to actually bring this code back.  Otherwise, let git remember what you wrote, because comments like this tend to stay FOREVER.

gabdank:
It is part of #3586 I am going to keep working on for release 38

PR #1136, opened 2016-02-03
3319 multi attachments

strattan:
Includes changes to loadxl to loop through multiple attachment properties.  That is exercised by the new inserts and documents.


hitz:
@lrowe  can you review the changes to contentbase/attachment.py (if @strattan hasn't already run them by you

src/encoded/schemas/quality_metric.json
```diff
@@ -55,7 +56,8 @@
                     "title": "MIME type",
                     "type": "string",
                     "enum": [
-                        "text/plain"
+                        "text/plain",
+                        "image/png"
```
lrowe: Should every quality_metric allow pngs or should this be decided per type?

strattan:
It was Laurence who made those … I just watched.
Seth

On Feb 2, 2016, at 4:32 PM, Ben Hitz <notifications@github.com<mailto:notifications@github.com>> wrote:


@lrowe<https://github.com/lrowe> can you review the changes to contentbase/attachment.py (if @strattan<https://github.com/strattan> hasn't already run them by you

src/encoded/schemas/quality_metric.json
strattan: Yea I thought it would be more flexible that way.  Limiting it to just text seems equally arbitrary to limiting it to just png.  With both (and the optional size parameters), you get the flexibility of both.

But I may be overlooking other considerations?
Seth

On Feb 2, 2016, at 4:35 PM, Laurence Rowe <notifications@github.com<mailto:notifications@github.com>> wrote:


In src/encoded/schemas/quality_metric.json<https://github.com/ENCODE-DCC/encoded/pull/1136#discussion_r51661735>:

> @@ -55,7 +56,8 @@
>                      "title": "MIME type",
>                      "type": "string",
>                      "enum": [
> -                        "text/plain"
> +                        "text/plain",
> +                        "image/png"


Should every quality_metric allow pngs or should this be decided per type?

lrowe: Mixins are just a convenience for constructing the final schemas. If there are only a few quality metrics that support image, we should leave the default be text and override the attachment property only on those. If it's common then making it generic like this is fine. Being overly restrictive about the attachment type probably doesn't buy us much.

strattan: I see.  Yes explicitly stating the type in the schema (rather than getting it from the mixin) is better.  The way I have it with the enum in the mixin you could end up with a particular QC object type that relies on the mixin for the mime type, and get some instances of the object with “attachment” properties as text and some as image.  That’s probably not good because consumers of that object type wouldn’t necessarily know what mime type to expect, just looking at the schema.

In fact, for ChIP, all but one of the QC objects have image attachments.  But those are (or will be soon) named something other than “attachment” anyway.

So, yes, keeping this just text is fine and I now see why you did it this way in the first place.

On Feb 2, 2016, at 5:01 PM, Laurence Rowe <notifications@github.com<mailto:notifications@github.com>> wrote:


In src/encoded/schemas/quality_metric.json<https://github.com/ENCODE-DCC/encoded/pull/1136#discussion_r51663992>:

> @@ -55,7 +56,8 @@
>                      "title": "MIME type",
>                      "type": "string",
>                      "enum": [
> -                        "text/plain"
> +                        "text/plain",
> +                        "image/png"


Mixins are just a convenience for constructing the final schemas. If there are only a few quality metrics that support image, we should leave the default be text and override the attachment property only on those. If it's common then making it generic like this is fine. Being overly restrictive about the attachment type probably doesn't buy us much.

PR #1148, opened 2016-02-10
3635 wgbs rep corr qc metric

tdreszer:
Please pull this simple change, which adds a single schema: cpg_correlation_quality_metric.  This object type is needed for posting WGBS experiments.


tdreszer:
Okay, I see I was missing the test/data/inserts/cpg_correlation_quality_metric.json file.

PR #1180, opened 2016-02-22
Update check_files.py

cricketsloan:
In order to actually check the files, we need to give chromInfo


cricketsloan:
Laurence and Ben,

I think Laurence will need to pull this into his changes as well.  Ben, can you tell me where to deploy this now.  I thought you changed the location?

PR #1185, opened 2016-02-24
3671 Trackhubs Search UI

forresttanaka:
Search pages now include a button with a drop-down menu for applicable track hubs. This button gets disabled if more than 500 search results exist.


hitz:
This will get merged as part of 3793

PR #1193, opened 2016-02-29
3112 Check files aws

lrowe:


hitz:
Pretty sure we can merge this whenever, but waiting for aditi to try it.

PR #1195, opened 2016-03-01
Matrix facet improvements -- see http://redmine.encodedcc.org/issues/3739

davisagli:
* Better formatting when there are 0 results
* Show filters that aren't in the facet list, like we do for standard search
* Include audit facets for users with search_audit permission


hitz:
I would love to merge this, but sadly I cannot because it has changes from 3563-boto3 and those break the buildout somewhat.  Not 100% on how to exclude the commits, maybe just merge master to this branch and re-PR?   I will see if I get to this tonight, since it's mostly my fault

hitz:
that seemed to clean it up, will merge when travis CI passes

davisagli:
@hitz I rebased on master (sorry, didn't realize you were going to try merging master right away; I've overwritten that)

PR #1204, opened 2016-03-03
Ontology part of

gabdank:
BEN, this branch changes the ontology.json file it works with.
I have uploaded it to S3, and modified the buildout.cfg


hitz:
There needs to be a ticket and a demo for QA for this.

gabdank:
There is a ticket () and demo was alive and QA.
http://redmine.encodedcc.org/issues/3586

PR #1220, opened 2016-03-08
3330 dataset files

forresttanaka:
Fixes empty ucsc dataset pages so that their files get displayed. A couple potential undefined variables also fixed.

Fails Travis, but I think we need a fix from Idan to make that work.


hitz:
Tests actually failing...

hitz:
I have merged Idan's stuff... new PR?  There seem to be conflicts now.

PR #1231, opened 2016-03-14
MAD changed to be DCC_ACTION

gabdank:
MAD = Median instead of Mean


gabdank:
I launched a demo, it was checked by Kath

PR #1234, opened 2016-03-16
3631 Enable Editing Menu for QC Objects

forresttanaka:


hitz:
This looks like it has pretty print debug code ...

forresttanaka:
Oh yeah. I’ll clean that up.

PR #1241, opened 2016-03-21
modified read depth audit to treat CRISPR like shRNA knockdown followed by RNA-seq

gabdank:


hitz:
hmmm this is having a bdd test timeout as well.

PR #1264, opened 2016-04-06
3342 avoid partial matches

Bek:


hitz:
I don't think we want to merge with the loggin in search.py... I think it's fine in the create_mapping.py.
Do we need the logging now?

Bek:
removed logging, demo is in sync with the updates

PR #1288, opened 2016-04-18
3892 Search UI Improvements

forresttanaka:


hitz:
Remove import pprint from line 570 src/encoded/search.py

forresttanaka:
pprint removed from search.py.

PR #1294, opened 2016-04-21
Add raw alignments to output_type

strattan:


strattan:
Passes local schema tests.

strattan:
https://3453-raw-mapping-output-type-bb75d08-jseth.demo.encodedcc.org/_indexer

I'll be testing the new accessioning code against that machine.

PR #1300, opened 2016-04-25
"Unfiltered" is better than "raw"

strattan:


strattan:
Local schema tests pass.
https://3453-raw-mapping-output-type-acf32e0-jseth.demo.encodedcc.org

PR #1305, opened 2016-04-28
Modernize javascript build stack

davisagli:
This modernizes various aspects of our build system, including:
- Use webpack instead of browserify. Webpack includes support for automatically breaking bundles into multiple chunks and serving them with a hash-based filename so they can be cached without preventing new changes from taking effect. It also restores our ability to watch for changes and update the bundles on the fly during development.
- Use gulp instead of grunt (mostly because then I could borrow configuration from another project)
- Update to babel 6 and a recent version of jest.

I also removed our compatibility shims for IE8...at least the ones I know about.

The automated tests are passing but this needs to be tested carefully in multiple browsers. Also I imagine @forresttanaka and @kilodalton want to have a look.


hitz:
So to re-create JSbundle/CSS manually I run bin/gulp insead of bin/grunt?

davisagli:
Run "npm run build" or "npm run dev" like it says in the readme (which are aliases for gulp with a couple different configurations).

David Glick
(mobile)

PR #1315, opened 2016-05-04
4018 update ontology

jahilton:


hitz:
Jason - this has a conflict, can you work with someone to resolve it.  Either me, Bek or Idan.  Probably not a big deal.

hitz:
These are real test failures, I cannot pass this.   Check with Idan I think.

PR #1344, opened 2016-05-13
4044-adding wal-e logs to crontab for back up at midnight PST

gto389:
added the wal-e logs to the crontab for midnight upload to S3.


hitz:
This needs to be gone over with Karen, possibly should be user postgres.   We need to install this on 42 first.

gto389:
@hitz temporarily closing until finished testing to verify functionality.

PR #1375, opened 2016-06-06
No persona

j1z0:
remove persona login and replaced with basic web auth.


hitz:
I am going to try to merge this into snowflakes branch, it should be cleaner.  Do you know what's up with the behave tests?   I would expect the persona tests to fail, but the behave tests don't even start?

j1z0:
hmmm…I must have broken something when I merged… let me have a look…

j1z0:
@hitz  

Ah ok I see, the secure environment variables, is SAUCE_USERNAME and SAUCE_ACCESS_KEY are not avaliable in a pull request.  I guess that makes since if you don't want random people to have access to that stuff.  So you'll just have to run those tests locally I think.

hitz:
This also happened with a doc update from @detrout; yes the sauce-labs keys are not available.  So what I did for her is to create a branch in my repo, merge yours on it, and the PR that.

That seems super clunk, we should revisit.

detrout:
Wait, how does a merging branch have anything to do with your testing API tokens? That seems a bit odd.

j1z0:
testing api tokens are stored as encrypted variables in travis.  Travis doesn’t export those encrypted variables for pull requests / forks as a security features.  Thus for the PR they don’t get exported and the tests can’t connect to sauce and things go downhill from there… :(

detrout:
I see. So for third party pull requests, you really need to manually merge them so a trusted user triggers the test

hitz:
This has been dealt with in https://github.com/ENCODE-DCC/snovault/pull/6

PR #1399, opened 2016-06-24
adding the travisCI reporting for slack now with more encryption

gto389:
Okay, the key is now encrypted for travis ci


hitz:
going to let travis pass again before I merge this

PR #1402, opened 2016-06-27
4115 checkfiles comments

gabdank:


hitz:
Idan - I think there maybe some tab/space mixing in lines 295-305.    Possibly you made an edit with vi or something on the deployed server?

Can you double check; maybe I am just misreading the git PR diff.   Even though the code "works", we have to be vigilant about tabs.

gabdank:
I think now the tabs/spaces are ok

PR #1422, opened 2016-07-07
4252 tech reps

gabdank:


hitz:
Are you sure files.json should have a property technical_replicates in the SCHEMA .json if it also has a calculated property?

PR #1425, opened 2016-07-08
4257 biosample audits

gabdank:


gabdank:
Ben I think the change you see here is the only change needed.
When I merged master and launched local instance it is working as expected.

PR #1450, opened 2016-07-26
4180 platforms

gabdank:


gabdank:
Due to "permissive" link recognition the following string is recognized as link:
HiSeq 2000*/2500, Illumina Genome Analyzer II/e/*x

Forrest is going to work on restricting link recognition (ticket #4315)

PR #1459, opened 2016-07-28
4262 Fix Replicate and Files columns in report view

forresttanaka:
Removed Files column. Replicate column now displays `@id`s of replicates.


hitz:
Branch conflicts, I think with Idan's previous branch.

PR #1485, opened 2016-08-15
Mitochondria blacklisted regions and input normalized peaks in file o…

aditin14:
…utput types, blacklisted regions has open permissions


aditin14:
Un-PRing this since we need more discussion on adding input normalized peaks

PR #1490, opened 2016-08-17
4253-modern-chip-qc

estherc:
#4253 audits for QC from the modERN ChIP-seq pipeline


hitz:
Do we really have no other way to special case this other than != /lab/kevin-white??

if bam_file['lab'] not in ['/labs/encode-processing-pipeline/', '/labs/kevin-white/']:

PR #1497, opened 2016-08-22
4295 scss fix

forresttanaka:
Fixes what I think is a bug in the Sass plugin for webpack. Caused not_compliant audits to not receive their styles.


hitz:
I don't understand why "non_compliant" ONLY needs to be double quoted.  Shouldn't all the parameters?

forresttanaka:
I don't understand either. But that's the only one that causes a problem. The Sass syntaz says none of them should be quoted.

forresttanaka:
You could close this if you like. https://github.com/ENCODE-DCC/encoded/pull/1503 includes this fix along with the build fixes.

PR #1504, opened 2016-08-25
4365 Add a missing embed to display biosample target protocol documents in experiments

forresttanaka:
Redmine #4365 note #9


hitz:
Interesting that we are adding MORE embedding...

PR #1509, opened 2016-08-29
fixing audit in experiment

gabdank:
fixing double report on error in experiment audit of assay_term_name/id


gabdank:
The audit without this fix will report sometimes two errors instead of only one. It will report 'NTR assay' and 'not present in ontology' error instead of only reporting 'NTR assay'.

PR #1518, opened 2016-09-06
3719 qc metrics for dnase

tdreszer:
Changes/additions required for all the Dnase flora of qc metrics.


hitz:
I am asking @estherc to look at this to make sure the property names are kosher.

PR #1522, opened 2016-09-08
Adding mappability targets and mappable regions as output types to fi…

aditin14:
…le.json

Tim approves


estherc:
Commented on #4451 - I don't think "mappability regions" is needed as an output_type and "mappability targets" is an ambiguous name.

PR #1533, opened 2016-09-19
resurrected seq. align, moved to reference with seq. uniqueness. Remo…

aditin14:
…ved mappability options


src/encoded/schemas/file.json
```diff
@@ -854,6 +851,11 @@
         "primer sequence": "reference",
         "spike-in sequence": "reference",
         "reference": "reference",
+<<<<<<< HEAD
```
hitz: You have bad merge spinach here... how did this pass tests?

PR #1563, opened 2016-10-10
4501 auditing geo

gabdank:


src/encoded/audit/experiment.py
```diff
@@ -1904,6 +1904,24 @@ def audit_experiment_gtex_biosample(value, system):
     return
 
 
+@audit_checker('Experiment', frame=['object'])
```
hitz: Idan - I agree this does what you ask it to... but there is no guarantee that having a GEO dbxref means it was actually submitted...   I guess it's close enough

gabdank: I know, but I don't see another way of doing that. Except comparing with the list of submitted by me experiments. I think this way we will become aware of anything not submitted. And once we don't have any more audits of this type - we can make a swap comparing with the list of already submitted by me.

PR #1564, opened 2016-10-10
4422 second attempt

gabdank:
implemented gating of new illumina read names, read1 and read2 mixes, read_length conflicts etc.

### DIFF
--- a/checkfiles/checkfiles.py
+++ b/checkfiles/checkfiles.py
@@ -12,6 +12,7 @@ import json
 import sys
 from shlex import quote
 import subprocess
+import re
 from urllib.parse import urljoin
 
 EPILOG = __doc__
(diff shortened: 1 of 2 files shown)


checkfiles/checkfiles.py
```diff
@@ -183,6 +184,180 @@ def check_format(encValData, job, path):
...
+    errors = job['errors']
+    result = job['result']
+
+    sequence_pattern = re.compile('[ACTGN.]+')
+    read_name_pattern = re.compile(
+        '^(@[a-zA-Z\d]+[a-zA-Z\d_-]*:[a-zA-Z\d-]+:[a-zA-Z\d_-]' +
+        '+:\d+:\d+:\d+:\d+[\s_][12]:[YXN]:[0-9]+:([ACNTG\+]*|[0-9]*))$'
+    )
+
+    special_read_name_pattern = re.compile(
```
gabdank: didn't want to call it 'broad_read_name_pattern'

checkfiles/checkfiles.py
```diff
@@ -183,6 +184,180 @@ def check_format(encValData, job, path):
...
+        # signatures
+        uniqueness_flag = True
+        signatures_for_comparison = set()
+        is_UMI = False
+        if 'flowcell_details' in item and len(item['flowcell_details']) > 0:
+            for entry in item['flowcell_details']:
+                if 'barcode' in entry and entry['barcode'] == 'UMI':
+                    is_UMI = True
+                    break
+        if is_UMI is True:
```
gabdank: will change

checkfiles/checkfiles.py
```diff
@@ -183,6 +184,180 @@ def check_format(encValData, job, path):
...
+                                    unzipped_fastq_path
+
+
+def process_read_lengths(read_lengths_dict,
+                         lengths_list,
+                         submitted_read_length,
+                         read_count,
+                         threshold_percentage,
+                         errors_to_report):
+    expected_read_lengths = [
```
gabdank: the current decision is to check whether 95% of read lengths in the file are close to reported read length in the metadata. Since we allow deviation of +-2bp in other places (length standards) I used this cutoff here as well - if it will not work we can narrow/widen acordingly

gabdank:
Aditi wants to be able to use old checkfiles version - so lets not merge this one for now. The branch can not handle WGBS files, so until I fix it - we should wait. Aditi will use the branch for "small" fastqs.

PR #1586, opened 2016-11-09
4517 display mapped read lengths

forresttanaka:
A somewhat unrelated change in this branch is that I removed the `rawArrayTableColumns` object for table display, because I had replaced it with the RawFileTable component before, so `rawArrayTableColumns` wasn’t used anymore. The other changes to this branch are small and pretty straightforward. The `hide` property of `SortTable` tables hides the column if a condition returns truthy. In this case, if all file.mapped_read_length are undefined (avoids filtering out file.mapped_read_length===0), then hide the column because we have nothing to show.


hitz:
Sorry forrest this has a conflict with your previous branch.

PR #1610, opened 2016-11-15
4576 visibility button for trackhubs

tdreszer:
This fixes the visibility button for searches, making it dependent upon at least one released experiment.  The visibility button on matrix results does is not dependent on released experiments (unfortunately).

### DIFF
--- a/src/encoded/search.py
+++ b/src/encoded/search.py
@@ -458,15 +458,15 @@ def format_results(request, hits, result=None):
 
     if frame in ['embedded', 'object']:
         for hit in hits:
-            if not any_released and hit['_source'][frame]['status'] == 'released':
+            if not any_released and hit['_source'][frame].get('status','released') == 'released':
                 any_released = True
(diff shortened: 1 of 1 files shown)


src/encoded/search.py
```diff
@@ -786,7 +795,7 @@ def search(context, request, search_type=None, return_generator=False):
 
     # Stream response using chunked encoding.
     # XXX BeforeRender event listeners not called.
-    app_iter = iter_long_json('@graph', graph, **result)
+    app_iter = iter_long_json('@graph', graph, result)
```
tdreszer: **result treats the contents of result as if it were individual parameters, but result treats it as a single object.  The effects on this function is that the individual parameters can be returned with a yield, thus passing the memory out of the function and potentially relieving the memory burden earlier.  However, result is not large AND I need to change one parameter of result based upon the contents of iterator (@graph exp list).  The memory efficiency is not needed but the memory access late in the function IS needed.

PR #1631, opened 2016-12-01
Recreating checkfiles

gabdank:


gabdank:
I had to redo the re-factoring because of a bug I managed to introduce in the last cycle
This branch is for ticket 4639

PR #1634, opened 2016-12-03
3126 replace react-forms with our own components

davisagli:
(ticket 3126)


forresttanaka:
## Overall Style
* The code in this PR isn’t AirBnB compliant, but most of our code base isn’t as it’s quite a process. Would be nice if new modules were though, especially for something like form.js which is key and has been largely rewritten. The eslintrc.json file enforces AirBnB (with exceptions) in Sublime and VSCode.
* I largely didn’t review forms.js very closely even though it’s so key and nearly completely rewritten, because not having any comments in the key parts of the file just makes it tough to go through in a reasonable time.

## Operation
* Really like that bad-input feedback is immediate instead of only after clicking Save.
* Glad we can get rid of the webpack preloader.

Will add some inline comments.

src/encoded/static/components/blocks/fallback.js
```diff
@@ -16,25 +16,56 @@ var FallbackBlockView = React.createClass({
...
 });
 
+var fallbackSchema = {
+    type: 'object',
+    formInput: <JSONInput />
+}
+
+var FallbackBlockEdit = module.exports.FallbackBlockEdit = React.createClass({
+    render: function() {
+        var {schema, value, ...props} = this.props;
```
forresttanaka: Doesn’t seem `props` is used.

src/encoded/static/components/blocks/fallback.js
```diff
@@ -16,25 +16,56 @@ var FallbackBlockView = React.createClass({
...
+        this.setState({ error });
+        if (!error) {
+            this.props.onChange(value);
+        }
     }
 });
 
+var fallbackSchema = {
+    type: 'object',
+    formInput: <JSONInput />
```
forresttanaka: Is this the only place JSONINput is called? I’m not sure how any props would ever get passed to it.

src/encoded/static/components/form.js
```diff
@@ -1,20 +1,29 @@
...
 var offset = require('../libs/offset');
 var fetched = require('./fetched');
-var ga = require('google-analytics');
+var inputs = require('./inputs');
+var jsonschema = require('jsonschema');
 var _ = require('underscore');
 
+var v = new jsonschema.Validator();
+v.attributes.pattern = function validatePattern(instance, schema, options, ctx) {
+  if (typeof instance != 'string') return;
```
forresttanaka: Some places indent 2 and some 4 in this file.

src/encoded/static/components/form.js
```diff
@@ -23,152 +32,205 @@ var filterValue = function(value) {
...
 
-});
+    getInitialState: function() {
+        // counter used to make sure that we get new keys
+        // once an item is removed
+        return { generation: 0 };
+    },
+
+    render() {
+        var { path, value, schema, onAdd, onRemove, ...props } = this.props;
```
forresttanaka: `props` isn’t used.

src/encoded/static/components/layout.js
```diff
@@ -54,13 +54,12 @@ var BlockEditModal = React.createClass({
 
     render: function() {
         var blocktype = globals.blocks.lookup(this.props.value);
-        var schema = blocktype.schema();
         var BlockEdit = blocktype.edit || FallbackBlockEdit;
         return (
-            <Modal actuator={this.props.actuator}>
+            <Modal ref="modal" actuator={this.props.actuator}>
```
forresttanaka: * I’m not clear on the reason for adding the `ref` to BlockEditModal and Modal, and then rendering when the component mounts. I’m wondering if this reveals a weakness in the Modal component itself (or possibly my out-of-date comments in modal.js)?

davisagli:
Thanks for the good feedback. I'll merge master, add some comments, review code style, and follow up on the other comments in the next few days.

davisagli:
eslint wasn't running successfully because I didn't have eslint-config-airbnb installed. I will add it to the npm devDependencies of `encoded`

src/encoded/static/components/blocks/fallback.js
davisagli: If the form schema specifies a React element as a property's `formInput`, then the `Field` component in form.js clones that element with the necessary props.

src/encoded/static/components/layout.js
davisagli: I don't think this a problem with the Modal component. It's a feature of the BlockEditModal, to automatically open the settings modal for a new block as soon as it is dragged onto the layout.

davisagli:
@forresttanaka I think this is ready for another look. I've cleaned up the style and added lots of comments.

forresttanaka:
I don’t see anything I would change. Looks really cool! I jotted down some notes and a question that can be ignored if preferred:

* That jsonschema package seems really cool. Had wondered if something that simple-to-use exists.
* Seems like way more functionality than the amount of code, which is very cool.
* In forms.js showErrors(), the two setStates in a row have a different effect than one? componentDidUpdate gets called twice?
* Didn’t know about Object.assign. Nice way to avoid underscore there.

davisagli:
I looked at a few JSON schema validation libraries; it's nice to know there are other options if this one doesn't work out. (It also isn't a ton of code if we need to fork it and make changes.)

That's actually a fair question about the two calls to setState; I'm not sure if React would debounce them and only call componentDidUpdate once, or not. In testing it did seem to work for making sure that we scroll to where the error is displayed.

I used Object.assign out of habit but I think it's actually not available without a polyfill in some browsers and the recommended way is to use object spread properties, i.e. foo = { ...obj1, ...obj2, otherProp } (that is ES7 and also requires enabling the correct plugin for babel)

PR #1645, opened 2016-12-20
4747 checking bed files

gabdank:


gabdank:
Renamed the variable and added clarification in comment

PR #1647, opened 2017-01-04
4753 GM search results crash Javascript

forresttanaka:
Was using regular 'context' instead of 'results'. We don’t actually have nicely formatted GM search results because we haven’t speced it — but it shouldn’t _crash_.


forresttanaka:
Error in display as a part of AirBnB compliance. I’ll fix that and resubmit.

PR #1650, opened 2017-01-04
4688 Location of "genetic modifications" in pages is off

forresttanaka:
Mostly just a matter of shuffling existing code around to reorder. Also added just a touch of AirBnB compliance to genetic_modifcation.js, but not all-out.


hitz:
Branch has conflicts, probably with something I just merged a second ago.

PR #1654, opened 2017-01-05
4378 annotation boost

gabdank:


hitz:
This change is rejected until we upgrade elasticsearch and re-do search/mapping

PR #1655, opened 2017-01-05
4723 file signature fix

gabdank:


checkfiles/checkfiles.py
```diff
@@ -278,7 +278,9 @@ def process_old_illumina_read_name_pattern(read_name,
             if (flowcell.find('-') != -1 or
                flowcell.find('_') != -1):
                 flowcell = 'TEMP'
-            lane_number = arr[1]
+            lane_number = '0'
+            if arr[1].isdigit():
+                lane_number = arr[1]
```
hitz: What is it supposed to do if the 2nd field ISN'T an integer?    Just assume it's 0?

Can you list the known cases (as comments in the code) that apply in all these situations for known signatures?

gabdank: added comments, not sure I am listing all cases, but I am explaining this specific choice to put zero in lane number

PR #1656, opened 2017-01-05
4716 Identify QC types in file graphs

forresttanaka:
Most of this implementation is straightforward, mostly involving the mapping of QC metric type to a two-digit abbreviation.

The hover-over tooltip constituted a whole ’nother matter. I had initially thought to overlay HTML tooltips based on the coordinates of each tooltip, but taking into account horizontal scrolling and zooming, and that the tooltips wouldn’t scale as the graph was zoomed made that a non-starter for me. More on that in a bit.

So I implemented these as SVG elements in the graph. I had initially rendered the graph without tooltips, and had Javascript re-render the graph with one tooltip when it detected a hover. This failed because by the time the graph was rendered, it was possible to lose the mouseLeave event if the user moved the mouse quickly.

So I now render all tooltips but with the “display:none” CSS property. When CSS detects a hover, it changes the property to “display:block.” This works perfectly in Chrome. Unfortunately, because I render these tooltips with no visibility, IE11, Edge, and Safari say they have zero dimensions, which means I can’t position the tooltips properly. FireFox just crashes when I try to measure the display:none tooltips.

So if we want this to work everywhere, I think we have to go with my first idea. I think it’s really clunky, but it would work everywhere. I’m thinking if this is important, we can look at that in the future.


hitz:
branch has conflicts

forresttanaka:
Since I’ll be PRing 4338 shortly, and 4716 (among others not PRed yet) have 4338 merged in, I suggest merging 4338 first, then this one. Not that it technically makes a difference, but emotionally it does for me.

hitz:
I did as you said, and merged 4338 first but now there is a conflict...

forresttanaka:
Working on resolving conflicts…

PR #1657, opened 2017-01-06
4543 Reference epigenome donor diversity property and display

forresttanaka:
The heart of this branch is in types/dataset.py which calculates a new ReferenceEpigenome property. This property now gets displayed on individual ReferenceEpigenome pages, search results, and search facets.


forresttanaka:
Jason says to hold on this for the moment until he gets a commit in.

forresttanaka:
These Travis CI runs both had transient errors. Since I‘m redoing this, there was no point in having them run anyway.

forresttanaka:
I have a separate branch — 4543-re-frontend — that needs QA and a decision on which way’s the right way to go.

PR #1661, opened 2017-01-09
4743: Add separate ENCODE4 viewing group

davisagli:


hitz:
I think these are still real test failures...

davisagli:
@hitz passing now

PR #1665, opened 2017-01-11
4651 Award Page

forresttanaka:
We get a new component file with this branch: award.js which closely follows the general layout of other regular components. It however does include charting code for the three doughnut displays. One might hope it shares a lot of this code with the home page, but it does not, as I think the home page needs refactoring to eventually use _this_ code once some of the home page’s special cases can be worked out, mostly for computational predictions.

The other important piece to this has little code itself: a new component to display Awards in a standard way across all pages that show them, including taking into account the logged-in state of the user. This resides in typeutils.js which I also made AirBnB compliant.

I have one small embedding change in Award so that it has access to the lab object.


forresttanaka:
Resolving conflicts…

forresttanaka:
Resolving Travis CI failures…

src/encoded/static/components/app.js
```diff
@@ -70,6 +71,50 @@ const biosampleTypeList = [
     'in vitro differentiated cells',
     'induced pluripotent stem cell line'
 ];
+const assayTermNameList = [
```
forresttanaka: It’s so that colors stay consistent, like with the other color arrays.

forresttanaka: I suppose I could do a search of experiments and grab the assay_title array from facets. Have to think about when to do this and how to store it.

forresttanaka: Starting a new instance from a branch off this branch. When drawing the charts, it does a `/search/?type=Experiment&field=@id` and picks through the result to find the assay_title facet, and uses its contents to generate colors.

forresttanaka:
Switching development to 4651-award-page-colors branch, which is a small development branch off of 4651-award-page-linked. Will PR that branch instead.

PR #1667, opened 2017-01-12
4543 ReferenceEpigenome donor diversity

forresttanaka:
The main change for this branch now that this is a front-end only change is the addition of the `donorDiversity` function in objectutils.js. It performs basically the same function as Jason’s Python version, though in a different way, collecting all relevant replicates into one place, then searching for donors that way. It seems longer than Jason’s Python version, but I have a lot of comments (and no, I didn’t strip out my comments and compare lines).

I put this function into objectutils.js so that both ReferenceEpigenome displays and ReferenceEpigenome searches can call it.

I did keep Jason’s change to the ReferenceEpignome audit.


forresttanaka:
Working on conflicts

PR #1700, opened 2017-02-06
4652 news sorting

forresttanaka:
This is purely a back-end change. Because news items need to get sorted differently from all other searches, I’ve added a news_search boolean that’s true if “Page” is the only `type` in the query string, and the `news` query string boolean is true.

If news_search is true then I want searches not to use do_scan because the do_scan searching stops sorting by date from working. So now do_scan only doesn’t get set to true if news_search is true.


hitz:
I have to think about this, this is a pretty serious hack job.

hitz:
We decided to punt this.

PR #1710, opened 2017-02-11
4815 CRISPRi

jahilton:


hitz:
The travis test is just a timeout on 2.7, ignoring.

PR #1715, opened 2017-02-14
4796 award links

forresttanaka:
As we now want to be able to display disabled awards to non-admin users, I now set a custom STATUS_ACL for Award items to allow this. By default for Items, only “current” displays to non-admin users.

I also change the embedding of award.pi.lab so that none of the dataset-derived types request it be embedded — instead the Dataset type embeds it so that all dataset-derived types get the same award.pi.lab embedded.

Added a way for dataset-derived React components to detect the admin status of the current user, getting that information from the session_properties context property. There were some unused “loggedIn” local variables that I didn’t use, and are now deleted. The admin status is needed because whether the use can view an award or not depends only on their admin status, not their logged-in status.


hitz:
This is a little clunky (at the python)  but I like that it provides a case study in how to customize statuses.
I am going to ask Bek to review this as well.

hitz:
@Bek maybe since Users already have a special ACL we should just extend the special case to that and have disabled="viewable" as the default?

Bek:
Changes look good. We should be careful about making all disabled objects viewable. We currently make disabled objects viewable only to admins by default. I am not sure which other object types should be viewable to everyone besides the Award.

hitz:
I think the ideas is that only Awards and Users are even set to "disabled" but that's not secured in the code.

PR #1716, opened 2017-02-14
4786 matrix data source

forresttanaka:
Now getting the biosample_type from the experiment’s calculated property instead of the experiment’s replicate for the home page. This better matches up with what we see on matrix pages.


hitz:
what is the ${var} syntax? Probably you've been using it for months and I haven't noticed.

forresttanaka:
Been using that for a handful of months. These are JS template literals. AirBnB doesn’t allow string concatenation using arithmetic (var1 + var2), and instead require template literals (`${var1}${var2}`). Babel fortunately knows about it.

PR #1727, opened 2017-02-16
4828 File page documents and QC panels

forresttanaka:
* As we now display file format specifications (documents) on the file page, yet `file.file_format_specifications` isn’t embedded and I was averse to adding embedding, I now request each document in `file_format_specifications` as a search. This meant adding a `columns` property to the document.json schema so that I could get the data to display in the File Format Specifications panel as a list of documents.
* This also meant refactoring the function I added recently called `requestFiles` in objectutils.js. It took an array of file @ids and does a /search/ GET request to get an array of those files that it returns as a JS promise. It’s now called `requestObjects` and takes a URL to use as a search, and `requestFiles` now calls this with a file URL. I use `requestObjects` to get the array of `file_format_specification` documents.
* Because the file page needs file-download links, I moved the code for file-download links out of filegallery.js and into the shared objectutils.js file. This allows file download links on the file page to have the restricted-file tooltip appear just like it does on the file graph/tables.
* The code to display the quality metrics dialog used to be in filegallery.js, and I’ve moved that to a new quality_metric.js file so that the modal can be displayed on both the file table/graph pages as well as the file page.
* As we talked about, the quality metrics panel now does a GET on /profiles/ to get the schemas for the quality metrics objects. It uses this to map QC property names to human-readable titles. It also uses the GenericQualityMetric schema to know what properties to _not_ display, like @id, uuid, etc. I used to use a static array for this.


forresttanaka:
Will work out the conflicts.

PR #1728, opened 2017-02-22
4871 Datasets with hidden graphs showing info buttons

forresttanaka:
Datasets that should never show graphs were still showing the information buttons on the file tables, and clicking them brought up no modal. That’s because datasets without graphs were still generating graphs — just not showing them.

This change makes datasets that don’t show graphs not generate them either, so their information buttons don’t appear because they’re based on files that actually appear in a generate graph.

If there’s a v53rc2, this should go in. If not, it should be OK to wait for v54 because it just shows buttons that don’t actually work, but has no other consequences.


cricketsloan:
I walked through this with forrest.

PR #1729, opened 2017-02-22
3734 Filter out experiments with statuses removed, revoked, deleted

forresttanaka:


cricketsloan:
This is a bit odd since we really want to just fix the data errors here.  However, the audits still show the data errors, and this code prevents those errors from showing to the public.

PR #1730, opened 2017-02-23
Update annotation.json

cricketsloan:
adding new annotation types


cricketsloan:
DAC has more annotation types for us

PR #1733, opened 2017-02-24
Visualization blob cache priming was slow because the request object …

tdreszer:
One line change to make the peak_indexer process use elasticsearch when building vis blob (trackHub) cache.


hitz:
I think my plan is to hot fix this v53, once it's all stable.  Not sure though.  Is there a demo?

tdreszer:
The demo is the v52rc3 machines.  I have modified some additional things on
it for debugging, so it isn't as clean as master + one line.

PR #1738, opened 2017-03-08
4835 donor exps

forresttanaka:
This ticket includes both [4885](http://redmine.encodedcc.org/issues/4885) and [4835](http://redmine.encodedcc.org/issues/4835). If 4885 gets merged first, then this only includes the changes for donor experiments. If this (4835) gets merged first, then 4885 won't need to be PRed.

Ostensibly, this branch simply adds another panel that triggers a GET request with this search:

`/search/?type=Experiment&replicates.library.biosample.donor.uuid={donor.uuid}`

The rest of the notes here come from changes for 4885.

Because the donor pages have been getting more sophisticated, I moved all the donor code out of biosample.js into a new donor.js file just to I can have an easier time digesting them.

I also changed loadxl.py so that donors can refer to themselves while loading local test data. This now applies to `parents`, `children`, `siblings`, `fraternal_twin`, `identical_twin`, though we only use `parents` and `children` here.

I had to change the human_donor.json schema to add a `columns` property so that searches of parent and child donors return the data we need.

I added a `encodedURI` function to globals.js that’s like the built-in Javascript encodeURI function, but adds escapes for the at-sign and exclamation point. This avoids a 301 redirect, and I’d like to start using this in more places. There's also a built-in Javascript encodeURIComponent but it goes way too far, encoding things like slashes and equals signs which is good when a URL is part of a query string, but not so much when the URL is the URL.

Lastly, I added some `parents` and `children` test data.


forresttanaka:
4835 had some updates based on feedback, and those updates are now merged into this branch.

PR #1742, opened 2017-03-09
3402 mapping chip

gabdank:


gabdank:
The branch contains audit changes that deal with Seth's addition of Mapping Pipeline to ChIP-seq.
To function properly, the instance has to include Seth's changes, without them ChIP-seq experiments would be partially not audited.

PR #1744, opened 2017-03-09
removing audit checking for consistency between assay_term_name and id since id is calculated property

gabdank:


hitz:
ooh me like code deleted from audits.

PR #1760, opened 2017-03-16
4513 ab status clean

estherc:
#4513 - Major refactoring and cleanup of ab status calculation code


src/encoded/audit/antibody_lot.py
```diff
@@ -32,72 +32,51 @@ def audit_antibody_missing_characterizations(value, system):
         return
```
hitz: This would be a little tighter if it checked if ANY of the elements of the array are 'control'

`if ‘control’ in [ t.get(‘investigated_as’) for t in value[‘targets’] ]:`

*should* work (untested)

src/encoded/audit/antibody_lot.py
```diff
@@ -32,72 +32,51 @@ def audit_antibody_missing_characterizations(value, system):
...
+             'Primary characterization not reviewed and pending review of a secondary characterization.',
+             'Primary characterization not reviewed.',
+             'Pending review of primary and secondary characterizations.',
+             'Pending review of primary characterization and awaiting a compliant secondary characterization.',
+             'Pending review of primary characterization and secondary characterization not reviewed.',
+             'Pending review of primary characterization.']:
             biosample = lot_review['biosample_term_name']
-            if biosample == 'not specified':
+            if biosample == 'any cell type or tissue':
                 biosample = 'one or more cell types/tissues.'
```
hitz: biosample variable is never used, is it?

src/encoded/types/antibody_lot.py
```diff
@@ -158,151 +162,166 @@ def lot_reviews(characterizations, targets, request):
         target_organisms[organism] = target['@id']
     target_organisms['all'] = tmp
```
hitz: It would be a little more succinct to do:
`target_organisms = { 'all' : [] } `
and then append to that list in the targets loop

src/encoded/types/antibody_lot.py
```diff
@@ -158,151 +162,166 @@ def lot_reviews(characterizations, targets, request):
...
-            'targets': sorted(targets),  # Copy to prevent modification of original data
-            'status': 'characterized to standards with exemption' if is_control else 'awaiting characterization',
-            'detail': 'IgG does not require further characterization.' if is_control else 'No characterizations submitted for this antibody lot yet.'
-        }]
+        return [base_review]
 
     review_targets = set()
-    primary_chars = []
-    secondary_chars = []
+    char_organisms = dict()
```
hitz: I think the usual is to declare:
`char_organisms = {}`
and
`primary_chars = []`

src/encoded/types/antibody_lot.py
```diff
@@ -158,151 +162,166 @@ def lot_reviews(characterizations, targets, request):
...
         # target_organism list. If not, we'll need to indicate that they characterized an
         # organism not in the antibody_lot.targets list so it'll have to be reviewed and
         # added if legitimate.
         review_targets.add(target['@id'])
-
+        char_organisms[characterization['@id']] = target['organism']
         # Split into primary and secondary to treat separately
         if 'primary_characterization_method' in characterization:
             primary_chars.append(characterization)
+
```
hitz: Should maybe antibody_lot have 2 arrays (calculated props?), one for primary and one for secondaries?
I guess UI treats them similarly.

src/encoded/types/antibody_lot.py
```diff
@@ -158,151 +162,166 @@ def lot_reviews(characterizations, targets, request):
...
-                      target_organisms):
+                      char_organisms):
+
+    # We have primary characterizatons
+    char_reviews = {}
+    lane_organism = None
+    key = None
+    for primary in primary_chars:
+        # The default base review
+        base_review = {
```
hitz: It confuses me that this is named the same as the base_review (default no characterizations) in line 67.

src/encoded/types/antibody_lot.py
```diff
@@ -158,151 +162,166 @@ def lot_reviews(characterizations, targets, request):
...
+            for lane_review in primary.get('characterization_reviews', []):
+                # Get the organism information from the lane, not from the target since
+                # there are lanes. Build a new base_review using the lane information
+                base_review = dict()
+                lane_organism = lane_review['organism']
+
+                base_review['biosample_term_name'] = 'any cell type or tissue' \
+                    if is_histone_mod else lane_review['biosample_term_name']
+                base_review['biosample_term_id'] = 'NTR:99999999' \
+                    if is_histone_mod else lane_review['biosample_term_id']
```
hitz: it seems like for histones you could just ignore the biosample?  I guess it's expected to be non-null.

src/encoded/types/antibody_lot.py
```diff
@@ -158,151 +162,166 @@ def lot_reviews(characterizations, targets, request):
...
-                    base_review['detail'] = 'Awaiting one or more compliant primary characterization(s).'
-                else:
-                    # Only in progress, not reviewed, pending dcc review secondaries should be left
-                    base_review['status'] = 'awaiting characterization'
-
+            # This primary characterization has characterization_reviews
+            for lane_review in primary.get('characterization_reviews', []):
+                # Get the organism information from the lane, not from the target since
+                # there are lanes. Build a new base_review using the lane information
+                base_review = dict()
```
hitz: `base_review = {}`

PR #1768, opened 2017-03-22
4902 gtex audits

gabdank:


hitz:
why are we commenting out audits?  You can always bring the version back from git?

gabdank:
We will for sure want to run these audits once in a month or so - I didn't want to remove them if I know that I will use them in the near future.
If you still think I should delete them, I will.

PR #1769, opened 2017-03-24
4905 kevin files

gabdank:


hitz:
Again, I am not sure you should comment out audits and tests, should we use git for this.
Maybe it would be too hard to remember / track what version.

gabdank:
deleted the comments. commit message includes the release number hopefully making it easier to find the change later

hitz:
What I want is a comment in the code that says:

# def audit_experiment_missing_processed_files(value, system): Removed from v44
# def test_audit_experiment_modERN_control_missing_files(testapp,... Removed from v44

gabdank:
done

PR #1776, opened 2017-03-29
3950 Sorting external accessions in file lists

forresttanaka:
External accessions were actually sorting randomly. I added a file object sorting routine to the process and reference file tables so that all external accessions get sorted after accessions, and the external accessions get sorted case-insensitively.


hitz:
Shouldn't this sort fuction be pulled out into some utility module rather than repeated?
also seems that accession (and possibly title) could be a parameter.

forresttanaka:
Yeah, I should have pulled out the sort function to begin with; it’s a separate function now. I didn’t see an advantage to making the accession and/or title be parameters themselves, since this sorting function can really only sort files in a specific way, I didn’t see an advantage to making it more general in some way, with more parameters. Maybe I’m misunderstanding.

forresttanaka:
I’ll fix the conflict. Let me know if anything else should change.

hitz:
well now it conflicts... I guess I thought it could be used for all possible column sorts.

forresttanaka:
<SortTable> handles the all-possible-column-sort case. The caller handles sorting for exceptions, and file.title makes an exception. Anyway, hopefully Travis CI passes twice with the conflict resolved.

PR #1781, opened 2017-03-31
4373 talens

gabdank:


hitz:
This is rejected; these should all be migrated to hg19 via batch upgrade and the enum NOT extended.  The schema change is fine.

I will leave the PR open; Idan can just re-push new commits.

gabdank:
removed GRCh37

PR #1787, opened 2017-04-05
4533 biodalliance

tdreszer:
The demo for this is just coming up and indexing:
https://4533-biodalliance-bec301b-tdreszer.demo.encodedcc.org


src/encoded/static/components/search.js
```diff
@@ -1,111 +1,111 @@
...
-
-var DbxrefList = dbxref.DbxrefList;
-var statusOrder = globals.statusOrder;
-var SingleTreatment = objectutils.SingleTreatment;
-var AuditIndicators = audit.AuditIndicators;
-var AuditDetail = audit.AuditDetail;
-var AuditMixin = audit.AuditMixin;
-var DropdownButton = button.DropdownButton;
-var DropdownMenu = dropdownMenu.DropdownMenu;
+import React from 'react';
```
hitz: This file is way too big, needs to be broken up.

src/encoded/static/components/genome_browser.js
```diff
@@ -0,0 +1,432 @@
+import React from 'react';
+
+
+const maxFilesBrowsed = 5; // Maximum number of files to browse
```
hitz: I don't like this that low.   Maybe it could guess based on vertical size of screen

src/encoded/static/components/genome_browser.js
```diff
@@ -0,0 +1,432 @@
...
+const domainName = 'https://www.encodeproject.org';
+const dummyFiles = [
```
hitz: What is a dummyFile?  Should this be test data?  Anyway, shouldn't be here.

hitz: I was working on a branch of peak_indexer that had a few local bed files for testing purposes.

src/encoded/static/components/genome_browser.js
```diff
@@ -0,0 +1,432 @@
...
+    {
+        file_format: 'bigBed',
+        output_type: 'peaks',
+        accession: 'ENCFF847CBY',
+        href: '/files/ENCFF847CBY/@@download/ENCFF847CBY.bigBed',
+    },
+];
+
+
+function rAssemblyToSources(assembly, region) {
```
hitz: All these defaults should be moved to a local JSON configuration file.   Not sure in what directory it would go or what the best way of loading JSON data in but @forresttanaka will know.

src/encoded/static/components/genome_browser.js
```diff
@@ -0,0 +1,432 @@
...
+        // Extract only bigWig and bigBed files from the list:
+        let files = this.props.files.filter(file => file.file_format === 'bigWig' || file.file_format === 'bigBed');
+        files = files.filter(file =>
+                    ['released', 'in progress', 'archived'].indexOf(file.status) > -1
+                );
+
+
+        // Make some fake file objects from "test" just to give the genome browser something to
+        // chew on if we're running locally.
+        files = !this.context.localInstance ?
```
hitz: I get why you need this to debug but needs to be moved to test suite.

src/encoded/static/components/genome_browser.js
```diff
@@ -0,0 +1,432 @@
...
+            (limitFiles ? files.slice(0, maxFilesBrowsed - 1) : files)
+        : dummyFiles;
+
+        const browserCfg = rAssemblyToSources(assembly, this.props.region);
+
+        this.browserFiles = [];
+        let domain = `${location.protocol}//${location.hostname}`;
```
hitz: Is this just a big hack to get around @@download for demo instances?

src/encoded/static/components/genome_browser.js
```diff
@@ -0,0 +1,432 @@
...
+            domain = domainName;
+        }
+        files.forEach((file) => {
+            const trackLabels = this.makeTrackLabel(file);
+            if (file.file_format === 'bigWig') {
+                this.browserFiles.push({
+                    name: trackLabels.shortLabel,
+                    desc: trackLabels.longLabel,
+                    bwgURI: `${domain}${file.href}`,
+                    style: [
```
hitz: I feel like file_format -> glyph config should be handled externallly as well.
Maybe ideally meshed with config for vizualize.py in python code

src/encoded/static/components/genome_browser.js
```diff
@@ -0,0 +1,432 @@
...
+
+        const files = !this.context.localInstance ? this.props.files.slice(0, maxFilesBrowsed - 1) : dummyFiles;
+        if (this.browser && files && files.length) {
+            let domain = `${location.protocol}//${location.hostname}`;
+            if (domain.includes('localhost')) {
+                domain = domainName;
+            }
+            files.forEach((file) => {
+                const trackLabels = this.makeTrackLabel(file);
+                if (file.file_format === 'bigWig') {
```
hitz: Why is this same (style) config repeated 50 lines up?

src/encoded/static/components/genome_browser.js
```diff
@@ -0,0 +1,432 @@
...
+                    name: trackLabels.shortLabel,
+                    desc: trackLabels.longLabel,
+                    bwgURI: `${domain}${file.href}`,
+                    style: [
+                        {
+                            type: 'default',
+                            style: {
+                                glyph: 'HISTOGRAM',
+                                HEIGHT: 30,
+                                BGCOLOR: 'rgb(166,71,71)',
```
hitz: don't hard code RGB colors

src/encoded/static/components/genome_browser.js
```diff
@@ -0,0 +1,432 @@
...
+                        ],
+                    });
+                }
+            });
+        }
+    },
+
+    makeTrackLabel: function (file) {
+        const datasetAccession = file.dataset.split('/')[2];
+        // Unreleased files are not in visBlob so get default labels
```
hitz: should maybe the label be colored/marked for unreleased?
This also gets back to "how do we make track labels a calculated property of files" (although *functionally* they could be a property of Dataset)

src/encoded/static/components/search.js
forresttanaka: I wouldn’t mind moving the type-specific search rendering functions to their own files (Experiment search result component to experiment.js) as we do with the newer object search results.

src/encoded/static/components/genome_browser.js
forresttanaka: That could be done (wouldn’t update after resizing the browser without a high-CPU-load updates). Maybe make a dropdown for the user to select?

src/encoded/static/components/genome_browser.js
forresttanaka: This mechanism is probably all out of date. That was for running locally it would just pick some files on production to use.

src/encoded/static/components/genome_browser.js
forresttanaka: This might be related to dummyFile and possibly no longer needed.

src/encoded/static/components/genome_browser.js
forresttanaka: Do you mean the console.log of the assembly? Yeah that should be deleted.

src/encoded/static/components/genome_browser.js
forresttanaka: It’s a hack to load production files if we’re running on localhost.

src/encoded/static/components/genome_browser.js
forresttanaka: Without giving it a _whole_ lot of thought, that does sound like a good idea. Not sure how visualize.py handles it.

src/encoded/static/components/genome_browser.js
forresttanaka: That just has to be factored out. Like with home-page charts, you have to do the same thing on initial draw as well as update. I factored that out on the home page, but haven’t touched this one since being in desperation mode.

PR #1789, opened 2017-04-05
issue 4911: disable indexing of page layout.rows

davisagli:
This avoids elasticsearch errors while trying to index pages that have a different layout structure than the elasticsearch mapping that was dynamically created based on the first pages indexed. It does so by disabling indexing of the layout altogether. That should be fine; we care about indexing the content of blocks, but they are in a separate field (because we anticipated this problem!)

Note that this depends on the changes in the `explicit-mapping` branch of `snovault`. https://github.com/ENCODE-DCC/snovault/pull/28 should be merged first, then a 0.24 release of snovault should be made and this PR updated to use it.


hitz:
there is something up with this in snovault. I am holding this back

hitz:
OK, merging master in seems to have cleared up indexing issues.

PR #1794, opened 2017-04-06
4653 /news/ page

forresttanaka:
* The core of this branch involves the changes to search.py. Because we now have `/news/` as a new search path, I added this path as a new route here, and a corresponding function, `news`.
* I made significant modifications to the existing `set_filters` function which solely used the `request` parameter to get information (from the query string) to make the query. Unlike any previous search functions (`search`, `matrix`, and `report`), `news` sets search parameters that don’t exist in the query string: `type=Page`, `news=true`, and `status=released`. I added the `static_items` parameter to pass static search parameters to add to `result` along with whatever was actually in the query string. Instead of calling `request.params.items()` each time it’s needed, I call it once and add our static items to it, and use this along with `static_items` to build the all_terms dictionary.
* When generating the filters in the search results, I only do this for the terms that actually appear in the query string — not for our static terms because these can’t ever be filtered out for the news search page. That’s why I put the existing filtering code underneath `if (field in qs_fields):` as qs_fields only includes fields in the query string while `fields` contains those plus the static ones for news.
* While `report` and `matrix` both act as wrappers for `search`, `news` does its own search, mostly because of the static terms issue that `search` never had to deal with, and because we don’t ever use scan mode because we only ever display the most recent 25 for the news search page even if you include `limit=all` in the query string. This is just to avoid the sorting issue for the moment (and likely forever unless we put in some front-end mechanism to display all news items sorted). Though it generally follows what `search` does, it’s more specific and simple.
* On the front end, I added a news.js component. In a way it’s a more specific and simple form of search.js, but it also includes a kind of plug-in method for facets. The `DefaultFacet` component gets used to display facets that work like the search.js ones, where multiple facet items can be selected which change the search. But I also added a much more complex `DateFacet` component that overrides the default one for the `month_released` facet. It not only displays the facet items from the search results, it also fabricates items for displaying a single entire year’s facets, and for clearing all date specifications from the query string. The fabricated ones for a single entire year generate a query string parameter for a new `year_released` search term that I added as a calculated property in page.py.


hitz:
I am not merging this yet because I think Forrest found a bug?

PR #1796, opened 2017-04-06
4973 assembly audit

gabdank:


src/encoded/audit/file.py
```diff
@@ -148,8 +148,8 @@ def audit_file_assembly(value, system):
     if 'derived_from' not in value:
         return
     for f in value['derived_from']:
-        if 'assembly' in f:
-            if f['assembly'] != value['assembly']:
+        if f.get('assembly') and value.get('assembly') and \
```
hitz: I think you write this something like:

if f.get('assembly', 'file assembly missing') != value.get('assembly', 'derived assembly missing'):

Actually I am unsure what is supposed to happen if one or more of the assemblies is Null.
As written it, it ignores unless they both have assemblies.   I suppose that's fine.

gabdank: it resolves audit error caused by my assumption that a file hat is derived from something will always have an assembly (turns out to be sometimes untrue)

PR #1800, opened 2017-04-10
4985 Javascript crash when clicking form Browse button

forresttanaka:
Some new biodalliance-related code used a new `assemblies` prop to the `ResultTable` component. While regular /search/ results add this prop, form code doesn’t, and really shouldn’t have to. This change makes it so we check for their existence before using it.

Also fixed some incorrect prop types that caused React warnings.


forresttanaka:
Hotfixing this into v55rc1 shouldn't be a problem if you’d like.

PR #1808, opened 2017-04-12
4927 run type dependency

gabdank:


gabdank:
We have 5 files on production failing validation because of the introduced dependency between run_type and paired_end information. We are reaching out to the corresponding labs to resolve these.

PR #1814, opened 2017-04-18
4975 Add strand specificity to metadata.tsv

forresttanaka:
The changes are pretty simple.


src/encoded/batch_download.py
```diff
@@ -140,6 +141,8 @@ def make_cell(header_column, row, exp_data_row):
         if column == 'replicates.library.biosample.post_synchronization_time' and len(temp):
             if len(c_value):
                 temp[0] = temp[0] + ' + ' + c_value[0]
+        elif column == 'replicates.library.strand_specificity':
```
hitz: This seems wrong.   But the only alternative is to change schema from Bool to text string or add calcuated property.  So I am checking with wranglers.

forresttanaka:
Will close this PR and make changes from the Wrangler meeting and resubmit.

PR #1818, opened 2017-04-19
4755 React 15 — at last!

forresttanaka:
# Crucial Changes
## setProps deprecation and removal
The React `setProps` method has been removed from React 15. We have used this in various places in mixins.js, largely for changing properties sent to `<App>` when logging in. These properties have been replaced by React states, and the `setProps` commands have been replaced with the eternal setState method. I originally made this change for clincoded when I updated that to React 0.14, and most of the same processes happen for encoded, though with some cleanup of issues that didn’t affect clincoded.

## Removal (mostly) of mixins
React mixins have been _practically_ though not _technically_ deprecated, so I have _almost_ removed them from encoded. A conspicuous user of mixins was audits, and I have replaced those with HOCs which act a little like Python decorators. I've renamed past users of the audit mixin to add the word Component to them (e.g. `<Experiment>` becomes `<ExperimentComponent>`) and then `<Experiment>` becomes assigned to `<ExperimentComponent>` wrapped in the new auditDecor HOC (“Decor” short for “decorator” — a nomenclature I might or might not abandon).

One minor usage of a mixin still exists just because I didn’t get around to it. A later update to get rid of it shouldn’t be very difficult.

## Removal of mixins.js
Along with the eradication of mixins was the removal of mixins.js that housed HistoryAndTriggers and Auth0. I found HOCs wouldn’t cut it partially because only `<App>` used these mixins, and because these mixins called each others methods — a model not fitting in with HOCs (and I’d argue not fitting in with any concept at all). All their functionality has moved into `<App>`.

# Less Crucial Changes
## ES6 styling
All new modules now use React ES6 styling, and I’ve converted some of the smaller existing modules to that styling. This also involves converting some components to stateless ones where appropriate. They use a syntax that looks odd at first, but is easy to get used to, and is easier overall. I had thought about converting the entire code base to ES6 styling for this PR, but currently only a small part uses the new styling and it took a long time just to get to that point, so this will have to be a gradual process. Once we're basically there, I’d like to experiment with JSDoc commenting which Visual Studio Code now has direct support for.

# Problem
I remember David said something about React 15 breaking something, but now I can’t find where he said that.

Note: I see this branch lags a bit behind the last PR merge, but that seems to have been just a documentation change so this should be fine.


forresttanaka:
I don’t think there’s a good way to merge this, but a tolerable way would be to merge everything else in the queue first. That will almost definitely cause merge conflicts with this PR. I’ll fix those and push, then you can merge this.

forresttanaka:
Added some changes from @davisagli I forgot to merge in.

hitz:
I merge the 2 JS tickets, so you can resolve conflicts now.  @kilodalton and I are going to review this together,  so it's not likely to get merged before thursday 4/27.

forresttanaka:
Merge plan sounds good. Resolving conflicts now, and will push when done.

kilodalton:
Holding temporarily while @forresttanaka checks a JS see on a instance build from his branch:
image

Note: error seen in console while logged in on demo instance

src/encoded/static/components/blocks/richtext.js
```diff
@@ -57,7 +59,7 @@ var RichTextBlockView = module.exports.RichTextBlockView = React.createClass({
 
     render: function() {
         return (
-            <div contentEditable={this.context.editable} dangerouslySetInnerHTML={{__html: this.state.value.body}} />
+            <div contentEditable={this.context.editable} dangerouslySetInnerHTML={{__html: this.state.value.body}} ref={(comp) => { this.domNode = comp; }} />
```
forresttanaka: Yes.

src/encoded/static/components/app.js
```diff
@@ -60,7 +78,7 @@ const projectList = [
     'Roadmap',
     'modENCODE',
     'modERN',
-    'GGR'
+    'GGR',
```
forresttanaka: JSON doesn’t allow it. AirBnB requires it if you have each value on its own line. That applies to arrays and objects.

src/encoded/static/components/app.js
```diff
@@ -69,277 +87,967 @@ const biosampleTypeList = [
...
+        } catch (e) {
+            console.warn('session parse err %o', session);
         }
     }
-});
+    return session || {};
+}
+
+
+function recordServerStats(serverStats, timingVar) {
```
forresttanaka: recordServerStats is in the mixins.js file in master. Had to move it since I removed mixins.js.

src/encoded/static/components/app.js
```diff
@@ -69,277 +87,967 @@ const biosampleTypeList = [
...
+                    console.log('changed state: %s', stateKey);
+                }
+            });
+        }
+
+        const xhr = this.state.contextRequest;
+        if (!xhr || !xhr.xhr_end || xhr.browser_stats) {
+            return;
+        }
+        const browserEnd = 1 * new Date();
```
forresttanaka: Yeah the `1 *` is one recommended way to convert from string to integer. Strings can’t be subtracted.

src/encoded/static/components/app.js
```diff
@@ -69,277 +87,967 @@ const biosampleTypeList = [
...
+    handleClick(event) {
+        // https://github.com/facebook/react/issues/1691
+        if (event.isDefaultPrevented()) {
+            return;
+        }
+
+        let target = event.target;
+        const nativeEvent = event.nativeEvent;
+
+        // SVG anchor elements have tagName == 'a' while HTML anchor elements have tagName == 'A'
```
forresttanaka: I guess so. Laurence put that in.

src/encoded/static/components/app.js
```diff
@@ -69,277 +87,967 @@ const biosampleTypeList = [
...
+        // data-trigger links invoke custom handlers.
+        const dataTrigger = target.getAttribute('data-trigger');
+        if (dataTrigger !== null) {
+            event.preventDefault();
+            this.trigger(dataTrigger);
+            return;
+        }
+
+        // Ensure this is a plain click
+        if (nativeEvent.which > 1 || nativeEvent.shiftKey || nativeEvent.altKey || nativeEvent.metaKey) {
```
forresttanaka: I haven’t looked at this line. It’s part of master.

src/encoded/static/components/audit.js
```diff
@@ -1,168 +1,350 @@
...
+}
+
+
+// Determine whether audits should be displayed or not, given the audits object from a data object,
+// and an optional session to determine logged-in state. It follows this logic:
+//
+// if audits exist
+//     if logged out
+//         if only one audit level and it's INTERNAL_ACTION
+//             return do display audits
```
forresttanaka: For [ENCSR878KIY](https://www.encodeproject.org/experiments/ENCSR878KIY/) I see “Biological replicates with identical biosample” which is an INTERNAL_ACTION only visible when you’re logged in.

src/encoded/static/components/award.js
```diff
@@ -129,29 +128,9 @@ function createDoughnutChart(chartId, values, labels, colors, baseSearchUri, nav
 
 
 // Display and handle clicks in the chart of labs.
-const LabChart = React.createClass({
```
forresttanaka: That’s a future goal. The main thing stopping it right now is Computational Predictions which have _just_ enough different to make that a non-obvious case so far. I’m sure there’s a way, but I haven’t had a chance to really nail that down yet.

src/encoded/static/components/collection.js
```diff
@@ -76,28 +78,57 @@ class Data {
...
 
-var RowView = function (props) {
-    var row = props.row;
-    var id = row.item['@id'];
-    var tds = row.cells.map( function (cell, index) {
-        if (index === 0) {
-            return (
-                <td key={index}><a href={row.item['@id']}>{cell.value}</a></td>
-            );
+const RowView = (props) => {
```
forresttanaka: This uses the React 15 stateless component mechanism. The other part of this involves not crashing when objects appear in the schema’s “columns” property. See https://github.com/ENCODE-DCC/encoded/commit/722e95b66bebd956ebec53b1dd34023cb3a93502

src/encoded/static/components/fetched.js
```diff
@@ -1,50 +1,42 @@
...
-    },
+    }
 
-    componentWillReceiveProps: function (nextProps, nextContext) {
+    componentWillReceiveProps(nextProps, nextContext) {
         if (!this.state.fetchedRequest && nextProps.url === undefined) return;
         if (this.state.fetchedRequest &&
             nextProps.url === this.props.url &&
-            nextContext.session === this.context.session) return;
+            _.isEqual(nextContext.session, this.context.session)) return;
```
forresttanaka: This problem was what caused a two-week (at least) delay. The existing code compared whether nextContext.session referred to the same object in memory as this.context.session. This is a _very_ poor assumption, and no longer ever true under React 15. _.isEqual compares whether the two objects have the same contents.

src/encoded/static/components/form.js
```diff
@@ -602,7 +607,10 @@ const Field = module.exports.Field = React.createClass({
         } else if (schema.type === 'integer' || schema.type === 'number') {
             input = <input type="number" {...inputProps} />;
         } else {
-            input = <input type="text" {...inputProps} />;
+            if (name === 'antigen_description') {
+                console.log('PROPS: %o', inputProps);
```
forresttanaka: Yeah that’s debugging code. Whoops.

src/encoded/static/components/globals.js
```diff
@@ -116,16 +117,17 @@ module.exports.unreleased_files_url = function (context) {
         "archived",
         "content error",
     ].map(encodeURIComponent).join('&status=');
-    return '/search/?limit=all&type=file&dataset=' + context['@id'] + file_states;
+    return '/search/?limit=all&type=File&dataset=' + context['@id'] + file_states;
```
forresttanaka: Yeah. I don’t really know what to say on that one.

src/encoded/static/components/inputs/object.js
```diff
@@ -20,27 +21,30 @@ var openLinksInNewWindow = function(e) {
...
 
-var SearchBlockEdit = React.createClass({
+var SearchBlockEdit = createReactClass({
     render: function() {
         var styles = {maxHeight: 300, overflow: 'scroll', clear: 'both' };
         return (
-            <div className="well" style={styles} onClick={openLinksInNewWindow}>
+            <div
+                className="well" style={styles} onClick={openLinksInNewWindow}
+                ref={(comp) => { this.domNode = comp; }}
```
forresttanaka: `this.getDOMNode()` isn’t a thing in React 15. You have to use a `ref` callback function to assign the rendered DOM node to an object property that you then use once the component mounts.

src/encoded/static/components/inputs/object.js
```diff
@@ -83,7 +87,7 @@ var ObjectPicker = module.exports.ObjectPicker = React.createClass({
         var previewUrl = url;
         var searchUrl = '/search/' + this.state.search;
         var actions = [
-            <button className="btn btn-primary" onClick={this.handleSelect}>Select</button>
+            <button key={1} className="btn btn-primary" onClick={this.handleSelect}>Select</button>
```
forresttanaka: No — this refers to a React key. Seems very strange usage though. I must have put it in to fix a problem, but not one I currently understand. I’ll try this locally.

src/encoded/static/components/report.js
```diff
@@ -246,22 +247,29 @@ var Table = module.exports.Table = React.createClass({
...
+        const columnPaths = Object.keys(this.props.columns);
         return (
             <div style={{display: 'inline-block', position: 'relative'}}>
                 <a className={'btn btn-info btn-sm' + (this.state.open ? ' active' : '')} href="#" onClick={this.toggle} title="Choose columns"><i className="icon icon-columns"></i> Columns</a>
                 {this.state.open && <div style={{position: 'absolute', top: '30px', width: '230px', backgroundColor: '#fff', padding: '.5em', border: 'solid 1px #ccc', borderRadius: 3, zIndex: 1}}>
                     <h4>Columns</h4>
-                    {_.mapObject(this.props.columns, (column, path) => <div onClick={this.toggleColumn.bind(this, path)} style={{cursor: 'pointer'}}>
-                        <input type="checkbox" checked={column.visible} /> {column.title}
-                    </div>)}
+                    {columnPaths.map((columnPath) => {
```
forresttanaka: No; that `.map()` is an ES6 thing. The problem here was the old code was giving an object to React to render multiple components. That doesn’t work in React 15 — multiple components _have_ to be rendered with an array, and `.map()` returns an array.

src/encoded/static/components/search.js
```diff
@@ -1,27 +1,33 @@
-import React from 'react';
-import cloneWithProps from 'react/lib/cloneWithProps';
-import cx from 'react/lib/cx';
-import url from 'url';
-import _ from 'underscore';
-import queryString from 'query-string';
-import { DropdownButton } from '../libs/bootstrap/button';
+'use strict';
+var React = require('react');
```
forresttanaka: That’s fine for now. Master took search.js to ES6 after 4755-react-15 branched off, and the merges from master kept the 4755-react-15 code as much as possible.

src/encoded/static/components/fetched.js
```diff
@@ -1,50 +1,42 @@
...
-        session: React.PropTypes.object
-    },
-
-    getDefaultProps: function() {
-        return {type: 'json'};
-    },
-
-    getInitialState: function () {
-        return {
+import React from 'react';
```
forresttanaka: I think the main problem is these components aren’t documented in any way anywhere, and took me forever to understand them. `<Param>` supplies parameters to the rendering component that receives the GET request, so it all really makes sense. This is probably a good one for [Redmine 5064](http://redmine.encodedcc.org/issues/5064).

src/encoded/static/components/inputs/object.js
forresttanaka: I kept it the way it is after some experiments. This looks weird, but you need it to avoid key warnings from React. I think we can find a more “sensical” way to do this, but `PickerActions` in search.js likely needs modification for that.

forresttanaka:
This demo with all but the last commit to remove the warning came up fine, fully indexed https://4755-react-15-5c2dd0147-fytanaka.demo.encodedcc.org/

PR #1833, opened 2017-05-02
5005 Enable Quick View for datasets

forresttanaka:
Instead of hard-coding “/experiments/” under the `visualize` function, I get the request’s path and insert it into the visualization URL so that it adapts to whatever’s being searched. Also removed the temporary checks for datasets in Javascript.


hitz:
This must have conflicted with 5004?

PR #1835, opened 2017-05-03
4966 Page the table of files deriving from the displayed one

forresttanaka:
This branch replaces the code driving the “Files deriving from {current file accession}” table. Now, it displays a maximum number of 50 files (easily changed with a constant if needed). If more than 50 files exist in this table, you get a pager component in the table’s footer.

This branch’s main goal reduces the size of the GET request to get the data for all the files that derive from the current one. It does this by first requesting a search of just barely enough information on _all_ the files that derive from this one, and sorts them. If needed, it breaks this list of files into pages, then does a non-minimal search of just the files that fit on the page (50 maximum) by specifying each file’s `@id` in the URL and displaying them in the table along with a new `<Pager>` component. It also saves the minimal list of all the files deriving from the current one in an object variable.

Each time the user clicks forward or back on the Pager, I retrieve the next page of file `@id`s from the minimal list and do another GET request to fill out the table with a new set of files — except…

I also keep within the `<DerivedFiles>` component a cache of each page’s file data as an object variable, just so that if the user clicks back and forth between a handful of pages, it won’t send GET requests each time — it just displays what lives in the cache. I set the limit of the cache arbitrarily to 10 pages. The algorithm chooses the page to kick out of the cache based on the farthest page from the current one. If pages 5 through 14 lived in the cache and the user requested page 4, page 14 would get kicked out and eventually garbage collected hopefully. If they requested page 15, page 5 would get kicked out. Pages get cached when the user clicks _away_ from them. When the page loads, they view page 0 with an empty cache. They click the “right” button and page 0 goes into the cache and they view page 1.

The new `<Pager>` component exists in a new pager.js file in the libs directory — there in the hope it can be used by other encoded components, and even other projects.

The `requestFiles` function has existed for a handful of releases. `requestSearch` debuts here — not that it does much — mostly reducing the promise rigamarole for an arbitrary search query string.


src/encoded/static/libs/bootstrap/pager.js
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
```
hitz: why is this in libs/bootstrap/?  (*looks at code*)  I guess the salient question is - why do we have bootstrap as a part of libs/ anyway?  Is is the set of components that bootstrap-scss applies to?

PR #1837, opened 2017-05-03
4845 d nase pipeline upgrade

tdreszer:
This is needed for adding hotspot1 scoring to the current DNase pipelines.


hitz:
The PR should probably say that this is 
a) a schema change
b) an upgrade script.

PR #1845, opened 2017-05-04
678 aliases regex

estherc:
#678 tighter regex for aliases and upgrade for existing aliases


hitz:
I think this travis BDD failure is not a bug in code or config.  I am going to run locally to be 100% sure.

hitz:
passed locally, seems to have been a timing error on travis.ci.

forresttanaka:
The Saucelabs video shows it failing either when displaying the file table I think, and other times failing when the edit form for an antibody took a _few minutes_ to come up, which was too long for Travis CI. None of this ever fails for me locally.

kilodalton:
Merged RM 678

PR #1856, opened 2017-05-11
update README

hitz:
add current mac osx install instructions.


stuartmiyasato:
Not sure if this is the right place to leave comments for updating this file, but someone can correct me...

In Step 2, the command "brew tap homebrew/versions" gives me this output:

> bash-3.2$ brew tap homebrew/versions
> Warning: homebrew/versions was deprecated. This tap is now empty as all its formulae were migrated.

What should I run instead?

Also this is minor, but "brew install python3" is indicated in both Step 1 and Step 2.  I assume this is redundant and can be deleted from Step 2?

PR #1860, opened 2017-05-17
5098 page HTML editing icon crashes Javascript

forresttanaka:
@davisagli made this fix before rc1 and told me, and I told him I got it merged into my React 15 branch. However, for whatever reason, I had not.

The issue involves the use of the React context object throughout the form.js module, combined with the encoded modal module. When a component that used React context properties while in a modal, the context properties ended up existing, but undefined. This change allows that context to come through using an undocumented React feature (undocumented aside from a [Tweet](https://twitter.com/dan_abramov/status/774591045980024833?lang=en) from [@dan_abramov](https://twitter.com/dan_abramov?lang=en)).

Procedures to reproduce and test in [Redmine #5098](http://redmine.encodedcc.org/issues/5098).


kilodalton:
@hitz @forresttanaka Also for this PR... Should the change to buildout be included?

forresttanaka:
Copying the comment from #1861 PR for future clarity: @kilodalton The buildout change should _not_ go in with either this or the #5097 PR. I had merged in 5079-maxchild-1 thinking that was needed for demos to finish indexing, but it wasn’t because the maxchild change had already rolled into master . I mentioned to @hitz that I should roll that change back out of this and the other branch, but he said it might not matter.

forresttanaka:
I feel like it’ll be easiest for everyone if I back out the buildout.cfg change, so I’ll do that now for this branch.

kilodalton:
@forresttanaka Thanks for the fix.  Because this is a mostly undocumented approach (in the wild), we should keep an eye on it in case it can use a different method down the road.

kilodalton:
Merged RM 5098

PR #1861, opened 2017-05-17
5097 Facets with no terms appearing

forresttanaka:
Redmine #4970 caused me to remove the all-zero terms suppression of facets because it caused the “not” facets to disappear. It _seemed_ (for reasons that I’m not yet clear about) that master had removed that condition in v55, and it got put back in my React 15 branch. But it seems we had two conflicting goals that must be resolved here: 1) Suppress these facets when terms are zero 2) Show these facets when terms are zero but they happen to be “not” facets.

The obvious solution is simply to add a condition, including checking for the terminal “!” in the facet title, so that it gets displayed even though it has no terminals — a condition that represses facets that _don’t_ have the terminal “!.”


kilodalton:
@hitz @forresttanaka can you confirm that this branch should have a change to buildout?

image

I would like a confirmation before merge.

forresttanaka:
@kilodalton The buildout change should _not_ go in with either this or the #5098 PR. I had merged in 5079-maxchild-1 thinking that was needed for demos to finish indexing, but it wasn’t because the maxchild change had already rolled into master . I mentioned to @hitz that I should roll that change back out of this and the other branch, but he said it might not matter.

forresttanaka:
I feel like it’ll be easiest for everyone if I back out the buildout.cfg change, so I’ll do that now for this branch.

kilodalton:
Merged RM 5097

PR #1862, opened 2017-05-17
4998 Fix assumption of replicate library’s existence

forresttanaka:
This was fixed and merged earlier, but it appeared again in v56rc2 because of a similar problem in a different spot. The original issue happened while grouping paired files by replicate/library. That was fixed for v56rc1.

Why this didn’t appear at this spot until now, I’m not sure. The code did shift position between v56rc1 and v56rc2, but it’s the same code run under what I _think_ are the same circumstances.

The fix involves simply checking for the existence of a replicate and its library before retrieving the library’s accession.


kilodalton:
Merged 4998

PR #1863, opened 2017-05-18
5101 DCC Internal Audits not rendered regardless of login state

forresttanaka:
## Background

This problem stemmed from modules not adapting to the new audit mechanism quite as readily as I had thought. With the Javascript mixin-shaming that the React team has engaged in, I had converted the audit module from an extensive mixin system to use High-Order Components (HOCs) instead.

Javascript mixins have the convenient feature of assimilating all the characteristics of the components they’re used in, including React props and states, and React context objects. They were able to use the encompassing component’s importing of _encoded_ login session information to get the user’s current login state.

With the transition to HOCs, this mechanism no longer works. The audit module’s wrapping of the components that use it means they share no props, state, nor React context. This left the audit module without that login information, making it think the user never logged in, and therefore never rending the DCC Internal audits.

## Fix

Now each component using audits has the responsibility to import the current user’s login information the `<App>`’s React context object, and explicitly pass it to the audit module and pass it in the `options` parameter of both `auditDetail` and `auditIndicators`.

### DIFF
--- a/src/encoded/static/components/antibody.js
+++ b/src/encoded/static/components/antibody.js
@@ -27,6 +27,10 @@ var {DocumentsPanel, Document, DocumentPreview, DocumentFile} = doc;
 
 
 var LotComponent = createReactClass({
+    contextTypes: {
+        session: PropTypes.object, // Login information from <App>
+    },
+
     render: function() {
         var context = this.props.context;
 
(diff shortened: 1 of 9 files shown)


kilodalton:
Merged RM 5101

PR #1864, opened 2017-05-18
bump version

hitz:
Might be worthwhile to check this with a local buildout, cd into develop/snovault and do a git status.


kilodalton:
Merged snovault 0.25 version change.

PR #1866, opened 2017-05-25
4744 step status

gabdank:
modifying analysis step statuses to match the statuses of other objects. For now we are leaving virtual - since it is heavily used and we haven’t come up with a solution to it. 
"waiting", --> in progress
"running", --> in progress
"finished", --> released
"error", --> deleted

So the branch includes schema change and an upgrade to the analysis_step object


kilodalton:
Merged RM 4744

PR #1868, opened 2017-05-30
5091 file audit inherit

gabdank:
removing duplicated and unnecessary audit_inherit in file.py under types. There is no need to inherit audits form lab/award coming not directly from file, but from replicate.award, replicate.lab etc.


kilodalton:
Merged RM 5091

PR #1869, opened 2017-05-30
5093 biosample overhaul

gabdank:
removing duplicated and unnecessary audit_inherit in biosample.py under types. There is no need to inherit audits form lab/award coming not directly from biosample, but from donor.characterizations.award, donor.donor_documents.lab etc.


kilodalton:
Merged RM 5093

PR #1871, opened 2017-05-31
5081 file schema

gabdank:
Adding “revoked_detail” to file.json and moving 4 file output types into historical types, that could not be longer submitted by ordinary users


kilodalton:
Merged RM 5081

PR #1872, opened 2017-06-01
5111 update accessions

hitz:
encoded/schema_formats.py contains some jsonschema.FormatChecker objects that are supposed to validate particular fields (type: accession, type: uuid, type: gene_name, etc.).    This module was not getting imported at the right time and the FormatChecker was not picking them up.

This PR includes tests for the operational formatcheckers (accessions and uuid) and with Laurence's help I managed to get the imports in the right sequence.

Note that the change to file.json is just to allow that file to be loaded, as it had an invalid accession!

To test this you can try to patch/put an existing accession; it must  correspond to the RE: '^ENC(FF|SR|AB|BS|DO|LB|PL)[0-9][0-9][0-9][A-Z][A-Z][A-Z]$' in production or ^TST(FF|SR|AB|BS|DO|LB|PL)[0-9][0-9][0-9]([0-9][0-9][0-9]|[A-Z][A-Z][A-Z])$' in a demo or test.

Note that the object type is not checked so you can change an FF to an AB.

Similarly all uri types in JSON schema must be global (://).   The gene_name and target_label Checkers are not implemented, they always return True (require a source for gene names)

### DIFF
--- a/src/encoded/__init__.py
+++ b/src/encoded/__init__.py
@@ -1,7 +1,6 @@
 from future.standard_library import install_aliases
 install_aliases()  # NOQA
-import encoded.schema_formats # needed to import before snovault to add FormatCheckers
-import base64
+import encoded.schema_formats # needed to import before snovault to add FormatCheckersimport base64
 import codecs
 import json
 import netaddr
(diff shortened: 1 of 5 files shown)


kilodalton:
Merged RM 5111

PR #1873, opened 2017-06-02
(5132) modifying derived_from audit to accept derivation from revoked deleted files for revoked deleted files

gabdank:
currently missing derived_from audit is triggered for revoked/deleted/replaced files, when they are derived from revoked/deleted and replaced files. This behavior started after application of audits on all Files (without dependency on their status). To fix this situation and not trigger the audit - it was modified to not flag deleted/revoked and replaced files that are derived from these statuses.


kilodalton:
Merged RM 5132

PR #1875, opened 2017-06-05
Add historical reference published_by enum

estherc:
#5131 new enum added to publication.published_by


kilodalton:
Merged RM 5131

PR #1876, opened 2017-06-06
4924 no file idan

gabdank:
introducing new file schema boolean filed: “no_file_available” that would be true for files not present on the portal. True value in this filed will make md5sum and file_size properties not required. check files was modified to skip the check of files with no_file_available = true.


kilodalton:
Merged RM 4924

PR #1877, opened 2017-06-06
4962 pipeline graph input and output nodes

forresttanaka:
* `PipelineComponent.assembleGraph` now has a large section missing. If it ever had value (and I’m not clear it ever did), it absolutely lost it with the reduced embedding, as it was checking for properties of objects that became simple `@id` strings with the embedding change. Removing it had no apparent effect on graph drawing, and because it was gated on properties that no longer exist, it really _couldn’t_ have an effect on graph drawing — it was dead code.
* I didn’t get to the pipeline.js file for Redmine 5065, so I updated it for AirBnB and ES6 in this branch.
* In pipeline.js, you’ll see a large change for the AnalysisStep function. This mostly moves this function from farther down in the file, as AirBnB prefers not to take advantage of hoisting.
* Because I converted pipeline.js from from old to AirBnb/ES6 style, it’s probably easier to review changes from [this commit](https://github.com/ENCODE-DCC/encoded/commit/4af92bf8b1bc76c690c8989b0ab7af0b3bdb8306#diff-7b9810fcabab452f65a8a0849c234630) instead of comparing to master.
* The key changes for pipeline graph drawing come in three parts:
    * Adding nodes and edges for each analysis step’s output_file_types
    * Adding edges from a node either to its parents output_file_types that match this node’s input_file_types, or edges directly to its parents if none of this node’s input_file_types match its parent’s output_file_types.
    * Adding nodes for input_file_types and edges to analysis steps that don’t have parents.


kilodalton:
Merged 4962

PR #1878, opened 2017-06-06
5128 GM characterizations have blank lab data

forresttanaka:
The primary change for this branch is in genetic_modification.py. The problem was document lab, submitted_by, and award weren’t embedded for biosamples_modified.documents and donors_modified.documents, so the front end had no data to work with. Those properties are now embedded for GeneticModification objects.

I slipped in a second change because I found a bug with Genetic Modification characterization displays. Characterization attachments for genetic modifications used a much older version of the code to display characterizations, and they don’t display correctly. I removed that copy and now have genetic modifications use the same version of AttachmentPanel from doc.js that all other attachment displays use (primarily quality metrics). Currently, I could find no genetic modifications on production that had characterizations, so this bug had never appeared before.

This change involves removing the GM-specific version of AttachmentPanel, and importing the one from doc.js.


kilodalton:
Merged RM 5128

PR #1879, opened 2017-06-06
4173 Matrix See More / See All buttons

forresttanaka:
The <Matrix> component now has a React state with an object for keeping track of which biosample types in the matrix have been expanded by the user, all initialized to false, and only existing if enough biosample term names exist under that biosample type to warrant a “See more” link.

Click handlers for the individual See More buttons and the See All button comprise the other big change. The individual See More click handler toggles the appropriate boolean state in the object that tracks each biosample type. The See All click handler not only toggles the state for expanding all biosample types, it also clears the states for all the individual See All buttons so that collapsing See All forgets what individual biosample types were expanded before.

The only other changes, aside from styling changes in SCSS, involved rendering the new See More and See All buttons.


kilodalton:
Merged RM 4173

PR #1880, opened 2017-06-07
5049 alt accs regex

gabdank:
tightening the schema for all the objects with accessions. any object with alternate_accessions will now be checked for correct accession format in the list of alternate accessions. For example file will be enforced to have ENCFF or TSTFF entries in the list.


kilodalton:
@gabdank You need to pull resolve the conflicts with `src/encoded/schemas/changelogs/file.md` before the the PR can be merged.

gabdank:
conflict resolved

gabdank:
@kilodalton conflict resolved (repeating the original message referring to you directly)

kilodalton:
@hitz can you please review this one, especially with regards to the many version bumps even though the schema itself is not changed in any other way and the data objects are accordingly bumped with no data change that I can see. Please comment compared to the policy expressed in:
https://github.com/ENCODE-DCC/snovault/blob/master/src/snowflakes/docs/schema-changes.md#updating-an-existing-schema
"The schema version is updated, requiring an upgrade of the schema, when the objects stored in the database will not validate and POST to the updated schema."

hitz:
This is a corner case where the schema .json technically didn't change byte-wise but changed functionally for users.    Specifically annotation.json mixes in alternate_accession from dataset.json.

The version needs to bump to flag users that there was a change.

kilodalton:
Merged RM 5049

PR #1881, opened 2017-06-07
5126 Forms embedded in other forms don’t drop down.

forresttanaka:
As part of the React 15 conversion, I changed the way abort handling gets tracked by having `<Param>` (which does GET requests) keep track of whether a GET request had been aborted by setting a flag — but didn’t provide a way to clear that flag. For the most part, this had no consequence because aborts aren't that common, and when they happen they usually clear themselves up (e.g. aborting a page load to go to another page). But when dropping down a form, an abort always happens, and because this new flag doesn’t get cleared, Param just hangs.

I changed to a simpler and [React-recommended way](https://facebook.github.io/react/blog/2015/12/16/ismounted-antipattern.html) (granted, recommended without enthusiasm) — Param now just keeps a flag of whether it’s mounted in the DOM or not. When aborts happen, Param gets unmounted, which causes the new flag to get cleared, and it’s ready for the next GET request.


kilodalton:
Merged RM 5126

PR #1882, opened 2017-06-07
5033 Negation facets feature

forresttanaka:
* Removed the `id` attribute from each term’s <li> component and <a> component, as it served no purpose, and because any selected terms would would all share the same “selected” `id` attribute — a violation of [the HTML spec](https://www.w3.org/TR/2011/WD-html5-20110525/elements.html#the-id-attribute). Had to make the corresponding change to the CSS.
* The Python changes involve getting rid of the fake “not” facets. When users put != operators into the query string, search.py would add these terms into the search results so that they’d get displayed at the bottom of the facet list. As we don’t show those anymore, it’s easier to stop generating those fake facets than it would have been to filter them out on the front end. Had to make a corresponding change to the pyramid tests that checked for the existence of these fake facets.
* As I made substantial changes to search.js, and as I haven’t yet gotten to search.js in 5065-airbnb-es6, I decided to convert search.js to AirBnB/ES6 here. This clutters up the diffs for this branch, so you can use [this diff](https://github.com/ENCODE-DCC/encoded/compare/66b7a30aa69a2eab425a10d5f1cf07e76519e8c2...98d9cfd43e75dbd9db5ac1bf6883285f73a765d5#diff-86d4cdcf5d7d21f7efd7ef21384c7f0b) instead which compares search.js from just after the conversion to the final version in this branch.
* You can see I _started_ to delve a little bit into [JSDoc](http://usejsdoc.org) in search.js. Not yet sure how sold on it I am.
* One large change involves the existing `termSelected` function. It used to check whether the given facet term is selected based on the query string. If it was, it returned the string to unselect the term. If it wasn’t, it returned null. Because `termSelected` now also checks whether a term is negated based on the query string, it now returns an object that the caller can extract using ES6 [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) (which probably looks too familiar for comfort for Python programmers). It now returns the existing `selected` string, a boolean `negated` that’s true if `selected` is for a “not” facet term, and `exists` which is a boolean that’s true if this is an exists facet term (only publication.json has this for one term currently).
* `<Term>` gets a lot of changes because its whole job is to display one facet term, so it needed to become aware of rendering negated terms.
* Other changes are pretty straightforward and small, along with chipping away at adding some badly needed comments.


kilodalton:
Merged RM 5033

PR #1883, opened 2017-06-07
5146 circ seq

gabdank:
Adding 5 new types of assays:                 
"genotype phasing by HiC",
"Circulome-seq", "whole genome sequencing assay", "BruUV-seq", "Mint-ChIP-seq” and updating the ontology.json with relevant information


kilodalton:
Merged RM 5146

PR #1884, opened 2017-06-07
5151 Remove biodalliance login restriction

forresttanaka:
The back end controlled whether the Quick View button could render or not based on the user’s logged-in state, so the primary change happened in dataset.py which now doesn’t check the logged-in state before offering the Quick View data in the `visualize` calculated property.

search.js also checked the logged-in state to decide whether to draw the tabbed search results or not. This check was similarly removed so that it always displays when requested regardless of logged-in state.

### DIFF
--- a/src/encoded/static/components/search.js
+++ b/src/encoded/static/components/search.js
@@ -1012,10 +1012,6 @@ const ResultTable = search.ResultTable = createReactClass({
 
     childContextTypes: { actions: PropTypes.array },
 
-    contextTypes: {
-        session: React.PropTypes.object,
-    },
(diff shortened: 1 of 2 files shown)


kilodalton:
@forresttanaka Please fix merge conflicts in `src/encoded/static/components/search.js`

forresttanaka:
Conflict resolved, and tested in a demo.

forresttanaka:
@kilodalton @hitz Conflict resolved, and tested in a demo.

kilodalton:
Merged RM 5151

PR #1887, opened 2017-06-07
3966 controlled audit

gabdank:
simplifying the audit code of file.controlled_by it will no longer flag controlled_by files with different replicates and paired end files without pairs. Implementation of the checks was wrong in the first place and is not used by DCC - so we decided to remove it, until new check specification comes and the audit code is rewritten


kilodalton:
Merged RM 3966

PR #1889, opened 2017-06-08
5096 bio schema

gabdank:
Biosample schema overhaul and alternate accessions regex.

Any object with alternate accessions is required to follow regex that ensures the alternate accessions specified are correct.
Comments and definitions added to many schema fields in biosample. TALENs were removed from biosample schema. Multiple dependency changes in biosample - to ensure model organism samples and human bio samples are treated differently, and properties such as harvest_date are applied only to specific biosample_types. derived_from field was renamed to be originated_from to solve elastic search problem with same property name in files.

TALEN objects are no longer available for general submission (import_only permissions) because we are getting rid of these and migrating the metadata to genetic modification.


gabdank:
@hitz @kilodalton PR submitted for biosample overhaul

kilodalton:
Merge RM 5096

PR #1890, opened 2017-06-12
5070 Fix search import/export for forms

forresttanaka:
Editing search fields in forms caused Javascript errors. I had updated search.js to AirBnB/ES6 and updated most of the references to it, but forgot about files in components/blocks. This updates components/blocks/search.js to import things the correct way, and also now exporting ResultTable from search.js as this file needs it.


kilodalton:
Merged RM 5070 Fix search import/export for forms #1890

PR #1892, opened 2017-06-13
5094 donor

estherc:
#5094 - Removed encode and roadmap-epigenomics from regex


kilodalton:
Merged RM 5094 donor / PR 1892

PR #1895, opened 2017-06-14
5033 change View All buttons from exclusion to inclusion filters

forresttanaka:
The “View all” buttons at the bottom of the Experiment and Biosample tables filtered on statuses we _didn’t_ want to see, but this caused the resulting search result page to display those facet terms in bright red for those statuses we _don’t_ want to highlight. This latest change makes it so that the “View all” buttons instead filter on statuses we _do_ want to see, so they appear in normal blue in the resulting status facets.

Later, Cricket wants to redo the statuses, and I might want to parameterize the “View all” buttons so we can customize the statuses to values appropriate for the kinds of objects we’re displaying.


kilodalton:
Merged RM 5033 change View All buttons from exclusion to inclusion filters / PR 1895

PR #1896, opened 2017-06-15
4924 backfilling no_file_available

gabdank:
finishing the addition of no_file_available property in File.json
 - backfilling the no_file_available to be False


kilodalton:
Merged RM 4924 backfilling no_file_available / PR 1896

PR #1897, opened 2017-06-20
5190 aliases

gabdank:
added andrew-fire to the aliases namespaces


gabdank:
@kilodalton I have checked the RC2 after your patch and it seems to be working as expected

kilodalton:
Merged RM 5190 aliases / PR 1897

PR #1899, opened 2017-06-23
ENCD-3031 RM4707 srna eclip visdefs

aditin14:
Stranded signal files for eCLIP and unstranded signals for small RNA could not be visualized on UCSC. I changed the vis_defs file to accept both stranded and unstranded files for both assay types. 

Test by visualizing a small RNA experiment with unstranded signals, such as ENCSR519NFO (in general, Manuel Garber's GGR RNA-seq), and an eCLIP with stranded signals such as ENCSR384KAN, on UCSC. 

http://redmine.encodedcc.org/issues/4707


kilodalton:
Merged ENCD-3031 RM4707

PR #1900, opened 2017-06-26
ENCD-3193 (RM4924) Add "virtual" file status

estherc:
ENCD-3193 (RM4924) Add "virtual" file status
Add Jason's post-release file patch to the code base so that it's documented.


kilodalton:
Merged ENCD-3193 (RM4924)

PR #1901, opened 2017-06-26
ENCD-3370 (RM5178) cre

gabdank:
changed annotation type   "candidate regulatory regions" into 
"candidate regulatory elements"


kilodalton:
Merged ENCD-3370 (RM5178)

PR #1902, opened 2017-06-26
ENCD-3387 (RM5200) fixing dataset linkTo in File.json

gabdank:
fixing in File.json the dataset linkTo property, it was a list without a real reason, so I turned it back into a single link instead.

### DIFF
--- a/src/encoded/schemas/file.json
+++ b/src/encoded/schemas/file.json
@@ -894,7 +894,7 @@
             "description": "The experiment or dataset the file belongs to.",
             "comment": "Required. See dataset.json for available identifiers.",
             "type": "string",
-            "linkTo": ["Dataset"]
+            "linkTo": "Dataset"
         },
         "replicate": {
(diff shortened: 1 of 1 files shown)


kilodalton:
@idan @aditin14 Please see the question on ENCD-3387 about the code change here

gabdank:
@kilodalton the answer is no. When I modified the File.json in release 56  I started from list of different dataset types and ended up with one abstract Dataset in the list, the right thing to do was to stop using list , but I didn't - so I am fixing that now. File still can belong to only one Dataset - so no schema version change nor upgrades are needed.

kilodalton:
@gabdank Thanks for the explanation. Typically a change to a pre-existing data object from an array to something that is not an array (or vice versa) needs extra changes. We'll go ahead based on your comments here.

kilodalton:
Merged ENCD-3387 (RM5200)

PR #1903, opened 2017-06-26
ENCD-3296 (RM5065) Make all code AirBnB compliant

forresttanaka:
Got rid of all eslint warnings from static/components and its subdirectories. I didn't touch the Javascript files directly under static/libs though because I think these might take some exploration. I should be able to convert them directly, but I have to see if any get called directly from node which would require some adjustment.

One of the issues that caused rewriting of code to some extent involved the ban on binding parameters to callbacks. We didn’t use this a _lot_, but it was handy to allow click handlers to handle clicks on multiple similar components. To get around this ban, I usually had to write a new component that took the existing component’s click handler as a property, as well as an identifier for the thing that was clicked. This new component would then call the parent’s click handler with that identifier, getting around this problem, and around the problem that was the reason for the ban: code duplication in memory — one copy of the same click handler for each bound item.

static/libs/registry.js had to bend the rules a little bit. Methods in classes that don’t refer to `this` should be made “static” methods. But registry happens to rely on its methods to _not_ be static, so I put in a comment rule to tell eslint not to worry about that rule.


kilodalton:
Merged ENCD-3296 (RM5065)

PR #1906, opened 2017-06-28
ENCD-3237 (RM4987) Tighten analysis_step.name regex and add calcprop analysis_step_version.name

estherc:
Schema and upgrade changes to improve consistency and display of analysis_step and analysis_step_version names and version numbers


kilodalton:
Merged ENCD-3237 (RM4987)

PR #1907, opened 2017-06-28
ENCD 3297 dnase

gabdank:
changed the threshold for reporting DNase read length inconsistency and excluded DNase experiments from mismatched platforms audit


kilodalton:
Merged ENCD-3297

PR #1909, opened 2017-06-28
ENCD-3324 (RM5102) loadxl

gabdank:
updated loadxl.py script to allow loading of bio samples independent of the order dictated by host property and also experiments and files dictated by ‘supersedes’ property


kilodalton:
Merged ENCD-3324 (RM5102)

PR #1912, opened 2017-06-28
ENCD-3413 file

gabdank:
Fixing file.json dependencies to allow submission of file with no status specified even if no_file_available is FALSE


kilodalton:
Merged ENCD-3413

PR #1914, opened 2017-06-29
ENCD-3282 5041 donors

gabdank:
updating statuses in donors and biosamples (removing preliminary and proposed statuses)

added dependency for mutagen, that will prevent human donors to specify mutagens

added comments in the schema and updated changelogs


kilodalton:
Merged ENCD-3282 RM5041

PR #1915, opened 2017-06-29
ENCD-3353 (RM5153) Display new external_ids array

forresttanaka:
A new external_ids array was added to the donor schema. This update now displays external_ids on donor pages as dbxrefs. It also removed the “External references” display as that will be phased out in the data eventually — in a sense, “External references” has becoming “Donor external IDs.”

As a semi-related part of this, CGC updated their URLs that we use for dbxrefs, so I updated that in globals.js as well.


kilodalton:
Merged ENCD-3353 (RM5153)

PR #1920, opened 2017-07-06
ENCD-3431 Restore tags code

forresttanaka:
Not sure why I deleted the tag display code with the AirBnB update. It’s back now, and badder than ever — it now uses an arrow function callback as AirBnB requires.

### DIFF
--- a/src/encoded/static/components/biosample.js
+++ b/src/encoded/static/components/biosample.js
@@ -569,6 +569,21 @@ const Construct = (props) => {
                     </div>
                 : null}
 
+                {context.tags.length ?
+                    <div data-test="tags">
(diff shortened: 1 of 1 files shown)


forresttanaka:
Hang on — there’s a used-to-be-a-warning-but-now-is-a-crash problem here. I’ll fix and update the PR.

forresttanaka:
This is caused by documents no longer being embedded within a construct, so this has probably existed as a problem for a while. I can fix it by doing a GET request, but it’s a big enough job that I’m going to close this PR and resubmit after that’s done. I’ll update the [ticket](https://encodedcc.atlassian.net/browse/ENCD-3431).

PR #1939, opened 2017-07-20
ENCD-3464 Remove timezone from local user test data

forresttanaka:
Including timezone now causes server startup errors. PRing this before Travis CI is done because I need to catch the bus before it’ll be done.


kilodalton:
Merged ENCD-3464

PR #1940, opened 2017-07-21
ENCD-3351 Use production React version with production builds

forresttanaka:
`npm run build` now includes the production version of React instead of the development version we’ve always been using. This reduces its size and increases its performance somewhat, at the cost of potentially making production/demo builds more difficult to detect bugs on — we’ll have to see.

I also took this opportunity to make our webpack.config.js AirBnB/ES6 compliant. I still use commonJS instead of ES6 imports because node executes this script directly, and node still only uses commonJS.


hitz:
> This reduces its size and increases its performance somewhat, at the cost of potentially making production/demo builds more difficult to detect bugs on 

Shouldn't this be a flag in deploy.py to toggle dev on or off?   We could default either way (to dev React vs production React) but there should be a switch.

hitz:
ok as long as there is a mechanism.

PR #1949, opened 2017-08-07
ENCD-3459 safari hang

forresttanaka:
Not sure this is a great fix, but it does work and I understand why it works. What I don’t understand is why this is a problem in the first place. It seems to imply to me that React has a bug, but I don’t even know how to characterize it to put into a report. During debugging for this, I move two experiments — one that showed the problem and one that didn’t — closer and closer to each other in their characteristics, hoping one would either start showing the problem or one would stop showing it, but even though I had them basically identical, one still showed the problem and the other didn’t.

So let’s just go with this for now. Again, I understand how the fix completely prevents the problem. I just don’t know why doing things the normal way shows this problem in Safari and IE11.

### DIFF
--- a/src/encoded/static/components/sorttable.js
+++ b/src/encoded/static/components/sorttable.js
@@ -199,11 +199,20 @@ export class SortTable extends React.Component {
         this.state = {
             sortColumn, // ID of currently sorting column
             reversed: false, // True if sorting of current sort column is reversed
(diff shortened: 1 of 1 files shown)


hitz:
>Not sure this is a great fix, but it does work and I understand why it works.
Sounds pretty great to me!

PR #1950, opened 2017-08-07
ENCD-3439 Award page revamp

forresttanaka:
Original PR notes by @chloetanlimco:

src/encoded/batch_download.py

Datetime -- find current date
u = types[0] -- find the type of Object (e.g. Experiment, Annotation, etc.)
format_header(seq) -- fix url to put in header
Changed request.response.content_disposition for file name
src/encoded/static/components/award.js

Used Moment library to sort dates in Cumulative Line Chart
Used SortTable to create a table of milestones
Function generateQuery from Home Page - used to determine selection of genus buttons
Conditional if statement in drawDonutCenter to remove text from Cumulative Line Chart and Status Experiments Chart
Changed the rectangular legend icons to circular icons
Change baseSearchUri so that when the chart is clicked, genus button selection will be accounted for in the resulting url
createBarChart needed for stacked bar chart
Set the specific chart.options.onClick.baseSearchUri since Experiments, Annotations, Biosamples, and AntibodyLot have different search terms for genus
Created AntibodyChart and BiosampleChart
Array of existing genera that has been converted from the array of existing species is made in order to identify which genus buttons to display for each award
ExperimentDate - sort dates in order and then fill in the missing ones, remove everything before the start date of the award
AffiliatedLabsArray to display list of Affiliated Labs in former Description Panel
src/encoded/test_batch_download.py

Tsv file includes “Experiment Report,” the date, and “.tsv” so we need to make sure that each file includes “Experiment Report” and “.tsv” because the date changes


src/encoded/static/components/award.js
```diff
@@ -113,7 +141,11 @@ function createDoughnutChart(chartId, values, labels, colors, baseSearchUri, nav
                         if (activePoints[0]) { // if click on wrong area, do nothing
                             const clickedElementIndex = activePoints[0]._index;
                             const term = chart.data.labels[clickedElementIndex];
-                            navigate(`${baseSearchUri}${globals.encodedURIComponent(term)}`);
+                            if (chart.options.onClick.baseSearchUri) {
```
hitz: I don't understand what this IF statement is doing.  How does chart.options.onClick.baseSearchURi get set, and why would it differ from ${baseSearchUri}???

src/encoded/static/components/award.js
```diff
@@ -409,24 +737,481 @@ class StatusChart extends React.Component {
...
+    let isogenicArray;
+    let anisogenicArray;
+    const unreplicatedLabel = [];
+    let unreplicatedDataset = [];
+    const isogenicLabel = [];
+    let isogenicDataset = [];
+    const anisogenicLabel = [];
+    let anisogenicDataset = [];
+
+    if (experiments && experiments.facets && experiments.facets.length) {
```
hitz: this whole StatusData method is pretty tortured but I blame the data model, not you.  maybe we need a bigger datastructure  instead of 3 differernt named ones?

src/encoded/static/components/award.js
```diff
@@ -474,67 +1319,177 @@ const ChartRenderer = (props) => {
             searchData[chartCategory].statuses = (statusFacet && statusFacet.terms && statusFacet.terms.length) ? statusFacet.terms : [];
         }
     });
+    // If there are experiements, then the corresponding species are added to the array of species
```
hitz: This code is very repetitive, I think the common elements should be refactored into a function or class.

src/encoded/static/components/award.js
```diff
@@ -409,24 +737,481 @@ class StatusChart extends React.Component {
...
+
+    updateChart(chart) {
+        const { experiments, unreplicated, isogenic, anisogenic, linkUri, award, objectQuery } = this.props;
+        const data = StatusData(experiments, unreplicated, isogenic, anisogenic); // Array of datasets and labels
+        const replicatelabels = ['unreplicated', 'isogenic', 'anisogenic'];
+        const colors = replicatelabels.map((label, i) => statusColorList[i % statusColorList.length]);
+        const unreplicatedDataset = data[2];
+        const isogenicDataset = data[3];
+        const anisogenicDataset = data[4];
+        if (unreplicatedDataset.some(x => x > 0)) {
```
hitz: this doesn't seem quite right either.   LIke instead of 3 variables for the different types of Datasets, use an object with a property "datasetType"?    It's quite possible I am missing something though.

src/encoded/static/components/award.js
```diff
@@ -409,24 +737,481 @@ class StatusChart extends React.Component {
...
+                anisogenicDataset.splice(j, 0, 0);
+            }
+        }
+    } else {
+        anisogenicDataset = [0, 0, 0, 0, 0, 0, 0, 0];
+    }
+    return ([labels, unreplicatedLabel, unreplicatedDataset, isogenicDataset, anisogenicDataset]);
+}
+
+class ControlsChart extends React.Component {
```
hitz: Do we need 6 different Components for different Charts?  Could we create a BlahChart class that each of them inherits from?

There is:
LabChart
CategoryChart
AntibodyChart
BiosampleChart
ControlsChart
StatusExperimentChart
StatusChart

Really they only change the fetch URL and maybe some configuration, right?  Possibly I am misunderstanding some ReactJS stuff.

src/encoded/static/components/award.js
```diff
@@ -1,13 +1,15 @@
...
+import moment from 'moment';
 import { Panel, PanelHeading, PanelBody } from '../libs/bootstrap/panel';
 import DataColors from './datacolors';
 import { FetchedData, Param } from './fetched';
 import * as globals from './globals';
 import { ProjectBadge } from './image';
 import { PickerActions } from './search';
 import StatusLabel from './statuslabel';
-
+import { SortTablePanel, SortTable } from './sorttable';
```
forresttanaka: Tiny request: I’m trying to keep npm_modules sorted by filename, then '../libs/*' sorted by filename, and finally './*' files sorted by filename, so could you move this line above the './statuslabel' line?

src/encoded/static/components/award.js
forresttanaka: It gets set when the chart gets updated because of user input, but this is a highly unusual way to save data, though technically it works. At the very least you need to comment it like crazy because adding properties to a function is highly unusual. I’m not sure why you needed this, making the baseSearchUri sometimes unused.

src/encoded/static/components/award.js
```diff
@@ -126,11 +158,163 @@ function createDoughnutChart(chartId, values, labels, colors, baseSearchUri, nav
     });
 }
 
+function createBarChart(chartId, unreplicatedLabel, unreplicatedDataset, isogenicDataset, anisogenicDataset, colors, labels, replicatelabels, baseSearchUri, navigate) {
```
forresttanaka: Given that there are so many parameters, could you add comments to this function like we have for createDoughnutChart? In JSDoc format would be even better, since I’ll be gradually moving us in that direction. See https://github.com/ENCODE-DCC/encoded/blob/ENCD-3439/src/encoded/static/components/globals.js#L118-L128

src/encoded/static/components/award.js
```diff
@@ -126,11 +158,163 @@ function createDoughnutChart(chartId, values, labels, colors, baseSearchUri, nav
...
+                if (anisogenicDataset.some(x => x > 0)) {
+                    datasets[1].label = 'anisogenic';
+                    datasets[1].data = anisogenicDataset;
+                    datasets[1].backgroundColor = colors[1];
+                }
+            } else if (unreplicatedDataset.every(x => x === 0) && isogenicDataset.every(x => x === 0) && anisogenicDataset.some(x => x > 0)) {
+                datasets[0].label = 'anisogenic';
+                datasets[0].data = anisogenicDataset;
+                datasets[0].backgroundColors = colors[0];
+            }
```
forresttanaka: While technically fine, I think there’s too many “magic” numbers here with the 0, 1, and 2 array indices. I’d prefer to use three different object variables, and then combine them into an array to pass in the `data` object when you create the chart below.

src/encoded/static/components/award.js
forresttanaka: This function is going to be tough to maintain without way more commenting. Describe what you’re doing at each major step.

src/encoded/static/components/award.js
forresttanaka: That’s very possibly true. chartjs can make it tough to reuse code, but I asked Chloe to see if she could combine these more or not.

forresttanaka: This returns an array of arrays. To avoid the use of “magic” numbers in places that call StatusData, I’d like to see returning an object so that the different data can be extracted by key instead of with magic numbers.

src/encoded/static/components/award.js
```diff
@@ -543,90 +1498,621 @@ ChartRenderer.propTypes = {
...
+        const dateSubmittedFacet = experiments.facets.find(facet => facet.field === 'date_submitted');
+        dateReleasedArray = (monthReleasedFacet && monthReleasedFacet.terms && monthReleasedFacet.terms.length) ? monthReleasedFacet.terms : [];
+        dateSubmittedArray = (dateSubmittedFacet && dateSubmittedFacet.terms && dateSubmittedFacet.terms.length) ? dateSubmittedFacet.terms : [];
+    }
+
+    // Use Moment to format arrays of submitted and released date
+    const standardreleasedTerms = dateReleasedArray.map((term) => {
+        const standardDate = moment(term.key, ['MMMM, YYYY', 'YYYY-MM']).format('YYYY-MM');
+        return { key: standardDate, doc_count: term.doc_count };
+    });
```
forresttanaka: This is a matter of taste, but I don’t think you should put the type of the variable in the variable name. Instead of `dateReleasedArray` I’d prefer releasedDates, for example.

src/encoded/static/components/award.js
```diff
@@ -543,90 +1498,621 @@ ChartRenderer.propTypes = {
...
 
-const Award = (props) => {
-    const { context } = props;
-    const statuses = [{ status: context.status, title: 'Status' }];
+    const standardsubmittedTerms = dateSubmittedArray.map((term) => {
+        const standardDate = moment(term.key, ['MMMM, YYYY', 'YYYY-MM']).format('YYYY-MM');
+        return { key: standardDate, doc_count: term.doc_count };
+    });
```
forresttanaka: Since these two loops do the same thing, I would implement them just once as an independent function, and pass dateReleasedArray and dateSubmittedArray in two different calls to that function. That function can be nested inside this function to keep the upper variable name scope cleaner, or you can make it its own independent function in the global space if you prefer.

src/encoded/static/components/award.js
```diff
@@ -543,90 +1498,621 @@ ChartRenderer.propTypes = {
...
+        return 0;
+    });
+    const sortedsubmittedTerms = standardsubmittedTerms.sort((termA, termB) => {
+        if (termA.key < termB.key) {
+            return -1;
+        } else if (termB.key < termA.key) {
+            return 1;
+        }
+        return 0;
+    });
```
forresttanaka: Same thing here. These two sorts do the same thing, and can be implemented just once as as function.

src/encoded/static/components/award.js
```diff
@@ -543,90 +1498,621 @@ ChartRenderer.propTypes = {
...
+        const endDate = moment(sortedsubmittedTerms[j + 1].key);
+        monthsubmitteddiff = endDate.diff(startDate, 'months', false);
+        if (monthsubmitteddiff > 1) {
+            for (let i = 0; i < monthsubmitteddiff; i += 1) {
+                fillsubmittedDates.push({ key: startDate.add(1, 'months').format('YYYY-MM'), doc_count: 0 });
+            }
+        }
+    }
+    fillsubmittedDates.push(sortedsubmittedTerms[sortedsubmittedTermsLength - 1]);
+
```
forresttanaka: Same here; these two loops do the same thing.

src/encoded/static/components/award.js
```diff
@@ -543,90 +1498,621 @@ ChartRenderer.propTypes = {
...
+        }
+    });
+
+
+    // Create an array of dates
+    const date = Object.keys(deduplicatedreleased).map((term) => {
+        label = term;
+        return label;
+    });
+
```
forresttanaka: This callback can be implemented as `….map(term => term);`

forresttanaka:
Finished reviewing. I guess if you “Add single comment” they don’t show up on this page, so you’ll have to look at the PR changes. Generally, I think the main “theme” about the requested changes is DRY — Don’t Repeat Yourself.

src/encoded/static/components/award.js
forresttanaka: This is highly unusual, but nothing technically wrong with it. I say we go with it in the PR for now.

PR #1951, opened 2017-08-07
ENCD 3466 bot

gabdank:
re-submitting PR


gabdank:
The one that is running now is 0.15
(STARTING Checkfiles version 1.15 (status=uploading): with 16 processes  at 2017-08-07 14:21:36.721357)
The 0.15 was to evaluate 90% instead of 95% of matching read lengths in FASTQs.

PR #1956, opened 2017-08-09
Encd 3412 audit matrix

kjain21:
Aggs first gets the query for the normal audit category rows and then gets updated with the no audits row. The no audits row is a nested query that is built through the for loop. After this, it checks if you are logged in, and if you are, it updates aggs with the appropriate Internal Action query.

The summarize_buckets method is to organize the normal audit category data into a usable format. The summarize_no_audits method is to organize the no audit category data into a usable format. Initially I tried putting both of these in one method, but because the no audit category data is nested, it is too different from the audit category data. Ultimately it was a lot simpler to do them as 2 separate methods.

After these methods, the rest of the code just formats the results in a way that is readable by auditmatrix.js


src/encoded/search.py
```diff
@@ -1118,3 +1120,389 @@ def news(context, request):
...
+    # matrix (dictionary): 'matrix' object within the Elasticsearch matrix search results,
+    #        containing all the data to render the matrix and its headers, but not the facets that
+    #        appear along the top and bottom.
+    # x_buckets (dictionary): 'x' object within the Elasticsearch matrix search results, containing
+    #        the headers that give titles to each column of the chart, as well as summary counts
+    #        we don't currently use on the front end.
+    # outer_bucket (list): search result bucket containing the term from which to group by.
+    # grouping_fields (list): List of audit categories that is hard coded above.
+    
+    def summarize_buckets(matrix, x_buckets, outer_bucket, grouping_fields):
```
hitz: dont' define functions inside object methods (view)

src/encoded/search.py
```diff
@@ -1118,3 +1120,389 @@ def news(context, request):
...
+    #        the headers that give titles to each column of the chart, as well as summary counts
+    #        we don't currently use on the front end.
+    # outer_bucket (list): search result bucket containing the term from which to group by.
+    # grouping_fields (list): List of no audit labels in order that is hard coded above. Allows
+    #        for recursion through the nested query.
+    # aggregations (list): same as outer_bucket. Allows for aggregations to be modified so that 
+    #        the nested query that allows for each level of no audits can become separate rows
+    #        under the overall no audits row.
+
+    def summarize_no_audits(matrix, x_buckets, outer_bucket, grouping_fields, aggregations):
```
hitz: move outside of view

src/encoded/search.py
```diff
@@ -1118,3 +1120,389 @@ def news(context, request):
...
+        # Loop by recursion through grouping_fields until we get the terminal no audit field. So
+        # get the initial no audit field in the list and save the rest for the recursive call.
+        group_by = grouping_fields[0]
+        grouping_fields = grouping_fields[1:]
+
+        # If there are still items in grouping_fields, then loop by recursion until there is
+        # nothing left in grouping_fields.
+        if grouping_fields:
+            summarize_no_audits(matrix, x_buckets, outer_bucket[group_by], grouping_fields, aggregations)
```
hitz: are you sure this won't loop infinitely if given bad data?

src/encoded/static/components/auditmatrix.js
```diff
@@ -0,0 +1,435 @@
...
+            const noAuditKey = ['no errors, compliant, and no warnings', 'no errors and compliant',
+                'no errors', 'no audits'];
+            // For each group, compare against the key arrays above and format yGroups so that
+            // it has the same order as the keys and each group in yGroups has the correct title.
+            let orderIndex = 0;
+            let rowIndex = 0;
+            const tempYGroups = [];
+            const tempNoAudits = [];
+            while (orderIndex < orderKey.length) {
+                yGroups.forEach((group) => {
```
hitz: my syntax highlighter complains that you are creating functions within a loop.  @forresttanaka what do you think?

src/encoded/static/components/auditmatrix.js
```diff
@@ -0,0 +1,435 @@
...
+                    if (group.key === orderKey[orderIndex]) {
+                        if (group.key === 'no_audits') {
+                            while (rowIndex < noAuditKey.length) {
+                                group.audit_label.buckets.forEach((row) => {
```
hitz: here too - my syntax highlighter complains that you are creating functions within a loop.  @forresttanaka what do you think?

src/encoded/search.py
forresttanaka: This comes from David, and has been in matrix since the beginning. I’m doing the same thing for target matrix.

src/encoded/search.py
forresttanaka: This also comes from David and has been in matrix since the beginning. It can’t loop forever because it sends ever-smaller lists to the recursive call, and stops when that list becomes one item.

src/encoded/search.py
```diff
@@ -1118,3 +1120,389 @@ def news(context, request):
...
+    facets = [(field, facet) for field, facet in schema['facets'].items() if
+              field in matrix['x']['facets'] or field in matrix['y']['facets']]
+
+    # Display all audits if logged in, or all but INTERNAL_ACTION if logged out
+    for audit_facet in audit_facets:
+        if search_audit and 'group.submitter' in principals or 'INTERNAL_ACTION' not in audit_facet[0]:
+            facets.append(audit_facet)
+
+    # To get list of audit categories from facets
+    audit_field_list_copy = []
```
forresttanaka: I worry that the risk of changing this at this stage won’t have a matching reward.

src/encoded/static/components/auditmatrix.js
forresttanaka: There seems to be [debate](https://github.com/eslint/eslint/issues/5044) about this rule, and they did put in a change to reduce the number of false positives, but all the way back in January 2016.

I still think this warning is wrong because, for the inner loop with the same warning, you can get rid of the warning simply by not using `rowIndex` inside the loop — but that reference is needed. I’m betting you can get around this error by writing this quadruple-nested loop in a different way, but that would take some thinking through.

I recommend having @kjain21 just disable this warning. She told me she’ll see if she can come up with another way to do this code, and disable the warning if not.

forresttanaka:
I would like to acknowledge that @kjain21 has commented her Python code — virtually unprecedented.

Bek:
Functions working with ES queries and the returned results need to have tests to so we know whether the upgrade changes the results.

src/encoded/search.py
hitz: man I never noticed...

PR #1961, opened 2017-08-16
ENCD-3436 Updates/corrections for award pages

forresttanaka:
Many of the changes are just slight source-code styling changes, and very small ones. The main change here was reversing the released and submitted variables so that each one receives the right data.

The change for the cumulative chart not only involved changing the colors to avoid the transparent ones, which also meant changing the order of the data given to chart.js so that the submitted chart wouldn’t hide the released chart now that they’re not transparent. That’s why you see the change to the array of objects to `new Chart`.

Also in this branch is the addition of a “Date submitted” display in the experiment page summary box, and a reformatting of the date to display in MMMM D, YYYY format instead of the existing YYYY-MM-DD format. The data hasn’t changed — just the display.

### DIFF
--- a/src/encoded/static/components/award.js
+++ b/src/encoded/static/components/award.js
@@ -1655,6 +1655,7 @@ const ExperimentDate = (props) => {
             const standardDate = moment(term.key, ['MMMM, YYYY', 'YYYY-MM']).format('YYYY-MM');
             return { key: standardDate, doc_count: term.doc_count };
         });
+
         // Sort arrays chronologically
         const sortedTerms = standardTerms.sort((termA, termB) => {
             if (termA.key < termB.key) {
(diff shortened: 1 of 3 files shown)


src/encoded/static/components/award.js
```diff
@@ -1926,14 +1930,14 @@ class CumulativeGraph extends React.Component {
...
                     datasets: [{
-                        label: 'Date Submitted',
-                        data: submitteddatavalue,
-                        backgroundColor: 'rgba(54, 162, 235, 0.2)',
-                    },
-                    {
                         label: 'Date Released',
                         data: releaseddatavalue,
-                        backgroundColor: 'rgba(75, 192, 192, 0.2)',
+                        backgroundColor: '#604a7b',
```
hitz: shouldn't these colors be mapped in a module somewhere, rather than hard coded.   This is just a future refactoring suggestion.

PR #1963, opened 2017-08-18
ENCD-3512 pipeline assay_term_names update

forresttanaka:
The `assay_term_name` string property no longer exists in pipeline objects. Instead, `assay_term_names` exists and is an array of strings. This led to these changes:

1. The Pipeline page summary panel now shows “Assays” instead of “Assay,” and it shows a comma-separated list of `assay_term_names` elements.
1. Pipeline object search results also displayed `assay_term_name` in the little summary. It now displays a list of `assay_term_names` comma separated.
1. The breadcrumbs at the top of the pipeline page now shows the assay_term_names elements separated by a plus sign, as AntibodyLot pages have done for a while. Clicking it brings up a search page with all the displayed assay_term_names selected in the facets. If no assay_term_names are available, then no breadcrumbs get displayed at all.

### DIFF
--- a/src/encoded/static/components/pipeline.js
+++ b/src/encoded/static/components/pipeline.js
@@ -363,12 +363,15 @@ class PipelineComponent extends React.Component {
         const context = this.props.context;
         const itemClass = globals.itemClass(context, 'view-item');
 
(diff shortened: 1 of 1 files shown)


forresttanaka:
Kath suggested possibly hot fixing v60rc2 with this. That should work fine, or we can delay this until v61.

PR #1971, opened 2017-08-28
ENCD 3493 condense audits

gabdank:
control read depth audits moved from File to Experiment.
All File and Experiment audits were “condensed” into one entry function - lowering the amount of embeddings done in “frames” of the audits


src/encoded/audit/file.py
```diff
@@ -302,57 +322,13 @@ def audit_file_controlled_by(value, system):
...
-        paired_with = context.get_rev_links('paired_with')
-        if len(paired_with) > 1:
-            detail = 'Paired end 1 file {} paired_with by multiple paired end 2 files: {!r}'.format(
-                value['@id'],
-                paired_with,
-            )
-            yield AuditFailure('multiple paired_with', detail, level='ERROR')
-            return
-
 '''
```
hitz: Can we remove this big comment?  It has it's own audit_checker/frame so it's not even compatible.
You could leave a bread crumb to a branch/tag/jira issue that has it incase it needed to be ressurected.

src/encoded/audit/file.py
```diff
@@ -7,33 +7,42 @@
 )
 from .standards_data import pipelines_with_read_depth
```
hitz: add #no qa to the end of this line to prevent import warning

src/encoded/audit/file.py
gabdank: DONE

src/encoded/audit/file.py
gabdank: This import was unused - so I removed it

gabdank:
@hitz, I think I have resolved the issues you raised

The functions dictionary and reorganization of functions into util/audit will happen in the next round unless you think they have to happen now

PR #1974, opened 2017-08-29
ENCD 3490 s3 checkfiles

gabdank:
checkfiles version changed to be 1.18

* cloud-config now creates executable file /etc/rc.local that would restart the checkfiles execution after reboot
* checkfiles runnig command in cloud-config was updated
a file with the script was added to the checkfiles filder (the script would run checkfiles with interval of 8 hours)
* checkfiles would accept --bot-token to post to the slack channel should not fail on transport error (S3 connectivity) it will just skip the file
* checkfiles should not fail on timeout reaching the portal , it will just skip the file
* upload failed status would be assigned to files that were not found in S3, but their upload credentials expiration period is over


checkfiles/checkfiles.py
```diff
@@ -716,109 +716,112 @@ def check_file(config, session, url, job):
...
         file_stat = os.stat(local_path)
     except FileNotFoundError:
-        errors['file_not_found'] = 'File has not been uploaded yet.'
-        if job['run'] < job['upload_expiration']:
-            job['skip'] = True
+        if job['run'] > job['upload_expiration']:
+            errors['file_not_found'] = 'File has not been uploaded yet.'
+        job['skip'] = True
+        return job
+    except OSError:
```
hitz: when do we get OSError, can we document that here?

checkfiles/checkfiles.py
```diff
@@ -1019,15 +1030,18 @@ def run(out, err, url, username, password, encValData, mirror, search_query, fil
         if not dry_run:
             patch_file(session, url, job)
 
+        if not job.get('skip'):
```
hitz: I think you need out.flush() and err.flush() after each file.   This will prevent the logged output from buffering.

checkfiles/checkfiles.py
gabdank: Added comments

checkfiles/checkfiles.py
gabdank: DONE

gabdank:
@hitz, I think I have resolved the issues you raised

PR #1980, opened 2017-09-07
ENCD-3517 deleted files

forresttanaka:
The actual functional change for this ticket is trivial: adding “deleted” and “revoked” to the `FileGalleryRendererComponent.inclusionStatuses` array, which lets them be part of the inclusion checkbox. The other changes were to eliminate the statuses from the search GET request so that the back end just gives us whatever it gives us depending on your privileges. This also eliminated the last reason for the globals.js Jest test.


hitz:
Me like deleted code.

PR #1981, opened 2017-09-08
Encd 3504 genetic mods ui

estherc:
This is a whopper of a PR, combining all the schema changes, upgrades, tests, changelog updates, manually migrated data and Forrest's corresponding UI changes related to GMs. GMs, constructs, RNAis, TALEs and CRISPR objects should have one or more required properties made import_only so we do not get any new submissions for the rollout. The GM upgrade test is commented out because it could not be made to work. GM dependencies are described in the comments. URLs were made into a mixin and mixed into donors. Upgrade only covers existing GM objects that are linked to existing CRISPR and TALE objects, all other migrations (construct, RNAi) have to be done manually.


src/encoded/schemas/genetic_modification.json
```diff
@@ -488,7 +488,6 @@
                 "properties": {
                     "name": {
                         "type": "string",
-                        "default": "",
```
hitz: name is required with a default?  of ""?  That seems not super useful.

src/encoded/types/biosample.py
```diff
@@ -306,20 +306,20 @@ def model_organism_donor_modifications(self, request, donor=None):
         "description": "All genetic modifications made in either the donor and/or biosample.",
         "type": "array",
         "items": {
-            "title": "Applied modification",
+            "title": "All applied modification",
```
hitz: grammar on this is a little odd?

src/encoded/types/biosample.py
```diff
@@ -306,20 +306,20 @@ def model_organism_donor_modifications(self, request, donor=None):
...
             "description": "Genetic modification made in either the donor and/or biosample.",
             "coment": "See genetic_modification.json for available identifiers.",
             "type": "string",
             "linkTo": "GeneticModification",
         }
-    }, define=True)
```
hitz: what does define=True do here?

estherc:
Those issues are all fixed in the latest commits.

PR #1985, opened 2017-09-18
Add Rob Spitale lab to json schema validation namespaces

detrout:
We tried to add an alias for rob spitales lab during  our site visit, and I saw that his name wasn't in the list of valid  alias prefixes.


hitz:
Replace by PR #1986

PR #1986, opened 2017-09-19
ENCD-3559 rob spitale

cricketsloan:
ENCD-3559 reding Dianes PR from master with test data


detrout:
Apparently I need to figure out how to run the encoded tests.

PR #1987, opened 2017-09-19
encode: alias prefix needs to be lower case.

detrout:
During testing at UCI we discovered that the example submit_file
submission failed because the encode prefix was upper case.


hitz:
Replaced by PR #1986

PR #1988, opened 2017-09-19
ENCD-3504 Upgrade Basic Genetic Modification and Lock Down Construct and RNAi

estherc:
The latest set of changes contains the following:
- Updates to the TALE upgrade to preserve both GM.notes, TALE.notes and TALE.talen_platform in the upgraded GM.notes field
- Fixed the calculated property biosample.applied_modifications to pull in modifications linked from either biosample or donor into a single property
- Embedding temporarily added to biosample.applied_modifications, genetic_modifications and model_organism_donor_modifications for the UI
- UI changes to display biosample.applied_modifications on biosample and experiment pages
- Experiment pages no longer display a GM table, instead GMs if any were applied are listed in the replicate table.
- Small updates to the manual migration data are still needed and will be added later.


src/encoded/static/components/experiment.js
```diff
@@ -741,6 +730,31 @@ const replicateTableColumns = {
         },
     },
 
+    genetic_modification: {
+        title: 'Modifications',
+        display: (condensedReplicate) => {
+            const replicate = condensedReplicate[0];
+            const gms = replicate.library && replicate.library.biosample && replicate.library.biosample.applied_modifications;
```
hitz: not that this is going to effect the request, but man JS is clunky here.  So is python I guess.

PR #1991, opened 2017-09-26
Allow adding/editing child objects when the linkFrom is specified as an abstract type

davisagli:
* When loading a child object to edit, use its `form` frame which includes the `@type` so the frontend knows what schema to use
* Make the client side schema validator aware of subtypes
* On the add form, show a dropdown menu to choose which subtype to add
* When serializing a new child for the server, include its `@type` so the server knows what to add
* Properly set links from the child object when the link property is an array

Note: This depends on https://github.com/ENCODE-DCC/snovault/pull/47

I've been testing by adding/editing quality metrics while editing a file. This needs QA on a demo instance but I'm opening this first to see if there are failures on travis...

@forresttanaka should review the styles on the add button; the dropdown version is not green like the single-type version.


forresttanaka:
I just pushed up a change so that the Add dropdown button on the form has the same green styling as the other buttons on the forms.

forresttanaka:
I looked over the Javascript changes and have no issues with what I saw. I did try running it and within an experiment form I opened a file, and opened that file's QC metrics object, and it all worked, while master just did nothing when I opened the QC metrics object.

PR #1996, opened 2017-09-28
ENCD-3489 dbxref rewrite, MGI.D external_id, and new dbxref Jest tests

forresttanaka:
The dbxref display mechanism was mostly code to handle special cases, and the cases didn’t have to be very special to require extra code.

So along with adding MGI.D to the list of dbxrefs/external_ids we handle, I rewrote all of dbxrefs to make maintenance easier. It now uses a kind of lame-o markdown language, where you have the dbxref URL with `{0}` wherever the specific dbxref value goes. To handle the special cases, the page that includes a dbxrefs display now passes the page’s context to <DbxrefList>. The dbxref code then has the opportunity to change the URL depending on what kind of object it gets displayed on, through a preprocessor mechanism.

I also added a post-processing mechanism so that some dbxrefs can do further value substitutions. The components that call <DbxrefList> now never need to pass special flags because the pre- and post-processor mechanisms use the passed context object to figure out the special cases.

I also finally added Jest tests for dbxrefs that tests all the different prefixes to make sure their URLs get built property.

I also commented the hell out of it with future new people and summer interns in mind.


hitz:
Can we get started on the domain name and github for LOML?

PR #2003, opened 2017-10-06
Encd 3562 gm audits

gabdank:
Adjusting the audits in biosample/experiment/replicate to work with GM instead of constructs.

Please merge it after 3538


gabdank:
@hitz conflicts resolved

PR #2007, opened 2017-10-12
ENCD-3595-missing-derived-from-audit-refactor

keenangraham:
This changes the logic for the missing derived_from audit. Before the raw files in the derived_from field of BAMs were only counted if they were FASTQs with the same status as the BAM _or_ if they were not deleted, replaced, revoked. This led to undesirable triggering of audit in some cases. For example:

- BAMs that came from CSFASTA/CSQUAL/FASTA instead of FASTQ.
- Deleted BAMs that came from revoked FASTQs.

Now raw files (still excluding tars) are counted if they have a higher or equal status-level value compared to BAM, based on STATUS_LEVEL dict in audit/items.py:
screen shot 2017-10-03 at 7 07 29 pm

Have also added tests of the audit to tests/test_audit_file.py.

### DIFF
--- a/src/encoded/audit/file.py
+++ b/src/encoded/audit/file.py
@@ -36,6 +36,10 @@ def audit_file_processed_derived_from(value, system):
                 and f['output_type'] == 'reads'
                 and f['output_category'] == 'raw data')):
 
+            # Audit shouldn't trigger if status isn't registered in STATUS_LEVEL dict.
+            if f['status'] not in STATUS_LEVEL or value['status'] not in STATUS_LEVEL:
+                return
+
             if STATUS_LEVEL[f['status']] >= STATUS_LEVEL[value['status']]:
                 fastq_bam_counter += 1
 
--- a/src/encoded/audit/file.py
+++ b/src/encoded/audit/file.py
@@ -3,6 +3,8 @@ from snovault import (
     audit_checker,
 )
 
+from .item import STATUS_LEVEL
+
 
 def audit_file_processed_derived_from(value, system):
     if value['output_category'] in ['raw data',
@@ -34,8 +36,7 @@ def audit_file_processed_derived_from(value, system):
                 and f['output_type'] == 'reads'
                 and f['output_category'] == 'raw data')):
 
-            if f['status'] not in ['deleted', 'replaced', 'revoked'] or \
-               f['status'] == value['status']:
+            if STATUS_LEVEL[f['status']] >= STATUS_LEVEL[value['status']]:
                 fastq_bam_counter += 1
 
             if f['dataset'] != value['dataset'].get('@id'):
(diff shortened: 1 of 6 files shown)


hitz:
this won't be merged until we sort out v62.

hitz:
OK I added 3 people to "review" request, I don't expect all 3 of you 2 review it.   But 2 might be good.

keenangraham:
Right, good point. I'm just wondering how this should be dealt with. Returning if file['status'] is not in STATUS_LEVEL (i.e. letting it silently fail) doesn't seem ideal. I could catch and raise an error but wasn't there some no-no about doing this within the audit framework?

keenangraham:
Third option is to just let any unknown status count.

keenangraham:
After talking to Idan have done two things to solve this problem.

First since missing derived_from audit relies on STATUS_LEVEL dict it won't function entirely properly if the file schema has statuses that aren't in STATUS_LEVEL. Instead of letting this fail silently or returning a default value the only way I see of maintaining this longterm is having a test that all statuses in file schema are in STATUS_LEVEL dict:

screen shot 2017-10-13 at 8 51 44 pm

This is in tests/test_audit_file.py (because it seems like a test of the audit to me). It imports STATUS_LEVEL from audit/item.py and uses load_schema from snovault.schema_utils (the alternative is to make a testapp.get request to /profiles/file.json). This is nice because you'll have a local test alerting you that STATUS_LEVEL needs to be updated in order for missing derived_from to function properly.

Second in order to prevent the audit from crashing if someone ignores this test failure I've put a default return value of -1 if the status is not found in STATUS_LEVEL. This means the audit will trigger in relation to any status in STATUS_LEVEL (because all of the values are higher). Returning a median value of 50 in this case seems confusing, because I want the failure to be apparent. I've also added a test to make sure this works.

gabdank:
@keenangraham I like the idea of test making sure all the file statuses are in item statuses dictionary. However, as you probably noticed the item statuses dictionary contains many other statuses (originating from other than file objects). I am not sure this has to go into this ticket, or maybe have a separate ticket, but if you are addressing the consistency of statuses - may be you should do it across the board and not only for file? 

As for the "-1" value - I disagree. I would rather not trigger anything - than trigger something irrelevant/erroneous. May be it is OK - because it is INTERNAL_STATUS audit, but if it was any publicly visible audit - I would strongly object reporting errors that doesn't make sense, only because the STATUS was not up to date in item.py

keenangraham:
Latest changes:
- Audit returns nothing if status is unknown.
- Test now in test_audit_item that checks STATUS_LEVEL dict against all statuses in schema.
- Added a number of statuses to STATUS_LEVEL dict and removed unused statuses from Publication object per ENCD-3627.

The test that pulls all schemas to compare to STATUS_LEVEL dict breaks the general test pattern a bit. Curious if it looks okay to @bek or @hitz or if they can think of better way. The general goal is to tie STATUS_LEVEL dict to any future schema changes, as missing derived_from audit won't work correctly otherwise.

hitz:
In the future don't mix null linting changes with code changes.

PR #2008, opened 2017-10-12
ENCD-3489 Restore RefSeq dbxref

forresttanaka:
Accidentally removed this dbxref prefix for v62rc3. Maybe a good way to demonstrate that new dbxrefs in this new system are very easy to add, as long as they’re easy URLs at least.


hitz:
should be named ENCD-3623

PR #2009, opened 2017-10-13
ENCD-3624 Avoid breaking page layouts on save (v62rc3 hotfix)

davisagli:


davisagli:
I guess this should get merged both to v62rc3 and to master?

PR #2015, opened 2017-10-17
updated "out of date analysis" to work properly with DNase

gabdank:
The audit will consider tech reps for DNase, instead of bio.reps that is considers for ChIP


tdreszer:
Idan and I had a run through of what this code does.  I don't think I would have comprehended it otherwise. No bugs that I can detect, but we discussed how it might be simplified.

hitz:
@gabdank there are conflicts now.

gabdank:
@hitz now it has no conflicts, but I want Jason to check it again, so I am closing this PR until I get green light from Jason

PR #2029, opened 2017-11-01
ENCD-3581 Donor UI cleanup

forresttanaka:
In addition to modifying the schema to include requested facets, I had to start embedding `lab` in the abstract Donor type so that we can facet on each donor’s lab. The derived donor types inherit this embedding.

Updated the display components to more cleanly inherit from a parent component because they mostly need the same display, but with minor variations. So I register donor-species-specific wrapper functions with the encoded view registry, and each of them calls a common <Donor> component.


src/encoded/static/components/donor.js
```diff
@@ -3,15 +3,18 @@ import PropTypes from 'prop-types';
 import _ from 'underscore';
 import url from 'url';
 import { Panel, PanelBody, PanelHeading } from '../libs/bootstrap/panel';
+import { auditDecor } from './audit';
 import { ExperimentTable } from './dataset';
 import { DbxrefList } from './dbxref';
 import { DocumentsPanel, DocumentsSubpanels } from './doc';
+import { FetchedData, Param } from './fetched';
```
caseylitton: Did you mean to leave this in?  FetchedData, Param are not used in the file.

src/encoded/types/donor.py
```diff
@@ -29,7 +29,8 @@ class Donor(Item):
         'documents',
         'documents.award',
         'documents.lab',
-        'documents.submitted_by'
+        'documents.submitted_by',
+        'lab'
```
caseylitton: Not a big deal, but when we're expanding lists like this we may as well add a trailing comma.  It results in a cleaner looking diff.  Don't resubmit a PR for it, just fyi for the future.
```
    ...
    'documents.submitted_by',
    'lab',
]
```

forresttanaka: Yeah, I’ve seen inconsistent usage of trailing commas in Python, so I’ve been equally inconsistent as a reaction. I’d like to see trailing commas for Python lists as a standard for us.

src/encoded/static/components/donor.js
forresttanaka: No! I need to remove that.

PR #2032, opened 2017-11-02
ENCD-3655 Move file assemblies calculation into ResultTable

davisagli:
(so it also works from object picker)

### DIFF
--- a/src/encoded/static/components/search.js
+++ b/src/encoded/static/components/search.js
@@ -1169,9 +1169,19 @@ export class ResultTable extends React.Component {
     constructor(props) {
         super(props);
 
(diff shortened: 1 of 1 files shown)


src/encoded/static/components/search.js
```diff
@@ -1169,9 +1169,19 @@ export class ResultTable extends React.Component {
     constructor(props) {
         super(props);
 
+        // Make an array of all assemblies found in all files in the search results.
+        let assemblies = [];
+        const results = this.props.context['@graph'];
```
forresttanaka: Can also just use `props.context['@graph']` since `props` is passed in as a parameter, though I know we use `this.props` for this situation all over the place, and that does work fine.

PR #2034, opened 2017-11-02
Encd 3647 schema docs

hitz:
just a doc edit.


estherc:
The content looks good. I could be OCD and make some changes to the formatting but ehhh...

If there is one thing to add it would be that you can check cloud-init-output.log soon (~within 30mins) after launching a demo (i.e. don't have to wait for indexing to finish) and you can immediately test any upgrade fixes by making them on the demo and re-running batchupgrade to see if it fixes the errors.

hitz:
@estherc this is updated, @caseylitton we can just merge it whenever.

PR #2039, opened 2017-11-07
Encd 3625 missing vis button

tdreszer:
Ready.  This branch also includes ENCD-3659.


src/encoded/visualization.py
```diff
@@ -2241,39 +2243,57 @@ def browsers_available(assemblies, status, files, types, item_type=None):
...
-            browsers.add('quickview')
-
-    if status not in VISIBLE_DATASET_STATUSES:
-        if status not in QUICKVIEW_STATUSES_BLOCKED:
-            return ["quickview"]
-        return []
+        vis_blob = None
+        if request is not None and accession is not None and status == 'released':
+            vis_blob = get_from_es(request, accession + "_" + assembly)  # use of find_or_make_acc_composite() will recurse!
+        if 'ucsc' not in browsers and 'ucsc_assembly' in mapped_assembly.keys():
```
caseylitton: If we're going to check if the value is in the browser set then we should just make browser a list.  The benefit of using a set it that sets do not allow duplicate values.  So this check is done again when browser.add() is called.

src/encoded/visualization.py
```diff
@@ -2241,39 +2243,57 @@ def browsers_available(assemblies, status, files, types, item_type=None):
...
+            if vis_blob is not None or files is None:  # If not checking files, be lenient
+                browsers.add('ucsc')
+            elif files is not None:
+                for file in files:
```
caseylitton: Two things.

1: 'file' is a keyword so we shouldn't use it.
2. We only need to run this for loop once.  See function rewrite below.

src/encoded/visualization.py
```diff
@@ -2241,39 +2243,57 @@ def browsers_available(assemblies, status, files, types, item_type=None):
...
+                    if file.get('assembly','') == assembly and file['status'] == 'released' \
+                        and file.get('file_format','') in ['bigBed', 'bigWig']:
+                        browsers.add('ucsc')
+                        break
 
+        if 'ensembl' not in browsers and 'ensembl_host' in mapped_assembly.keys():
```
caseylitton: Again we don't want to check if a value is duplicated in a set.

src/encoded/visualization.py
```diff
@@ -2241,39 +2243,57 @@ def browsers_available(assemblies, status, files, types, item_type=None):
...
+                for file in files:
+                    if file.get('assembly','') == assembly and file['status'] == 'released' \
+                        and file.get('file_format','') in ['bigBed', 'bigWig']:
+                        browsers.add('ensembl')
+                        break
+        if 'quickview' not in browsers and 'quickview' in mapped_assembly.keys():
+            # NOTE: quickview may not have vis_blob as 'in progress' files can also be displayed
+            #       Ideally we would also look at files' statuses and formats.  However, the (calculated)files
+            #       property only contains 'released' files so it doesn't really help for quickview!
+            if vis_blob is not None or (status not in QUICKVIEW_STATUSES_BLOCKED):
```
caseylitton: ```status not in QUICKVIEW_STATUSES_BLOCKED``` is not dependent on the main for loop.  So calculate this outside of the loop.

src/encoded/visualization.py
```diff
@@ -2241,39 +2243,57 @@ def browsers_available(assemblies, status, files, types, item_type=None):
...
+                browsers.add('quickview')
+        if len(browsers) == 3:  # No use continuing
```
caseylitton: I'm not a fan of this.  I made an explicit check of the values in my example function.

caseylitton:
I rewrote the function in a different perspective to isolate the various cases.  I also didn't like the duplicated code for the three browser types.  I think it's cleaner this way but that's just my perspective.  It does have examples of how to clean up the for loop so we're not running the same code repeatedly.    

```
def browsers_available(status, assemblies, types, item_type=None, files=None, accession=None, request=None):
    '''Returns list of browsers based upon vis_blobs or else files list.'''
    # NOTES:
    # When called by visualize calculated property, vis_blob should be in vis_cache, but if not files are used.
    #        When called by visindexer, neither vis_cache nor files are used (could be called 'browsers_might_work').
    # quickview may not have vis_blob as 'in progress' files can
    #       also be displayed. Ideally we would also look at files'
    #       statuses and formats.  However, the (calculated)files
    #       property only contains 'released' files so it doesn't
    #       really help for quickview!
    if "Dataset" not in types:
        return []
    if item_type is None:
        visualizabe_types = set(VISIBLE_DATASET_TYPES)
        if visualizabe_types.isdisjoint(types):
            return []
    elif item_type not in VISIBLE_DATASET_TYPES_LC:
            return []
    browsers = set()
    break_state = {'ensembl_host', 'quickview', 'ucsc_assembly'}
    file_assemblies = None
    check_vis_blob = False
    if request is None or accession is None or status != 'released':
        check_vis_blob = True
    quickview_blocked = False
    if status in QUICKVIEW_STATUSES_BLOCKED:
        quickview_blocked = True
    for assembly in assemblies:
        mapped_assembly = _ASSEMBLY_MAPPER_FULL.get(assembly)
        if not mapped_assembly:
            continue
        if not quickview_blocked:
            if 'quickview' in mapped_assembly.keys():
                browsers.add('quickview')
        if files is None:
            if 'ucsc_assembly' in mapped_assembly.keys():
                browsers.add('ucsc')
            if 'ensembl_host' in mapped_assembly.keys():
                browsers.add('ensembl')
        elif check_vis_blob:
            # use of find_or_make_acc_composite() will recurse!
            vis_blob = get_from_es(request, accession + "_" + assembly)
            if vis_blob is not None:
                if 'ucsc_assembly' in mapped_assembly.keys():
                    browsers.add('ucsc')
                if 'ensembl_host' in mapped_assembly.keys():
                    browsers.add('ensembl')
                if 'quickview' in mapped_assembly.keys():
                        browsers.add('quickview')
        elif files is not None:
            # Build the file assemblies once
            if file_assemblies is None:
                file_assemblies = []
                allowed_formats = ['bigBed', 'bigWig']
                if files:
                    file_assemblies = [
                        f['assembly'] for f in files
                        if f.get('assembly')
                        and f['status'] == 'released'
                        and file.get('file_format') in allowed_formats
                    ]
            if assembly in file_assemblies:
                if 'ucsc_assembly' in mapped_assembly.keys():
                    browsers.add('ucsc')
                if 'ensembl_host' in mapped_assembly.keys():
                    browsers.add('ensembl')
        if browsers == break_state:
            break
    return list(browsers)
```

src/encoded/visualization.py
tdreszer: 1. Fair point.
2. Not sure colapsing the files for loop saves anything because there is also a for assembly for loop.  I consciously duplicated the files for loop for readability of the different cases.

src/encoded/visualization.py
tdreszer: Fair point.  I was trying to keep the 3 browser cases separate for readability.

src/encoded/visualization.py
```diff
@@ -2241,39 +2243,57 @@ def browsers_available(assemblies, status, files, types, item_type=None):
...
     if 'accession' not in obj:
         return False
     if assembly is not None:
         assemblies = [ assembly ]
     else:
         assemblies = obj.get('assembly',[])
-    browsers = browsers_available(assemblies, obj.get('status','none'), obj.get('files'), obj.get('@type',[]))
+    files = None
+    if check_files:
+        files = obj.get('files',[])
```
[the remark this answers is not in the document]
tdreszer: This is not the same, as files=None means don't check but files=[] means check (for 'ucsc' and 'ensembl' but not for 'quickview').  So if files=[] the available browsers might be QuickView, while files=None could return all 3.

src/encoded/visualization.py
```diff
@@ -2440,7 +2460,7 @@ def generate_batch_hubs(context, request):
             if g_text == '':
                 log.debug('Requesting %s.%s#%s NO ASSEMBLY !!!' % (page,suffix,cmd))
                 g_text = json.dumps(results,indent=4)
-                assemblies = [result['assemblies'] for result in results['@graph']]
+                assemblies = [result.get('assemblies',[]) for result in results['@graph']]
```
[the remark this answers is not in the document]
tdreszer: Not sure that this is a more straightforward option.  I think this best change is actually to rip this code out as it only feeds a log.debug message that is never even seen.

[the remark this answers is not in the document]
tdreszer: Comment on your example rewrite below.  This would need further debugging, I can see at least one error, but that is understandable as I can't expect that you had a chance to run it.  While I can see the value of attempting to eliminate duplicate code and moving certain conditional statements outside loops so they are more efficient, I do not agree that your version represents an improvement for a couple of reasons.

1) You replace logic inefficiency with storage inefficiency.  Declaring variables to hold conditional results is not free.  In the case of quickview_blocked, for example, there are currently no more than 2 assemblies per experiment, so you replaced 2 conditional statements with 2 conditional statements.
2) You obscure the meaning of code when trying to make it more efficient.  While far from crystal clear, I think there was more simplicity in the original code due to the separation of the three types: ucsc, ensembl and quickview.  This code will certainly have to be touched again, for instance if a 4th option were supported, and your version will make that a little harder.
3) Your method requires going though all files, which might be a lengthy list, where as the original method could shorten that search.  Of course the original method might actually go through the files list twice, stopping on the first hit each time. 

Anyway, it would have been nice to do this face to face.  Do not misunderstand me, I do appreciate ALL your suggestions, even if some of them might generate debate.

tdreszer:
Checked in changes inspired by Casey's review.  I reproduced Kathrine's tests listed in the ticket, but she might be asked to take a look again.

PR #2047, opened 2017-11-15
ENCD-1972 Add status facet for analysis-step

forresttanaka:
Added a facet for lab in the analysis_step.json schema.

### DIFF
--- a/src/encoded/schemas/analysis_step.json
+++ b/src/encoded/schemas/analysis_step.json
@@ -319,5 +319,10 @@
             }
         }
     },
+    "facets": {
+        "status": {
+            "title": "Status"
+        }
+    },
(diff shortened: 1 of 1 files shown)


caseylitton:
Same here, we can't make a second commit for ENCD-1972 without some explaination.  Was this something missing from the original?  We'll want to add this information to the commit message....like 'Amend ENCD-1972...' if it's something missing.  Or if it reverts a part of a commit then, 'Partial Revert ENCD-1972....'

forresttanaka:
Cricket made a comment in the ENCD-1972 ticket that wasn't clear to me that she wanted this facet added or not, so I skipped it. After rc1, she made it clear she actually did want a facet here, so this PR adds that facet.

PR #2048, opened 2017-11-15
ENCD-3581 Restore Lab facet

forresttanaka:
Schema change to bring back the lab facet.


caseylitton:
This looks like a reversion of a change from ENCD-1972.  Why is it tagged as ENCD-3581?  

Adding 'ENCD-3581 Restore Lab facet'  as a commit message is going to be confusing.  I'll want to fix the name when I squash but I'm not sure how these two tickets are related.

PR #2050, opened 2017-11-16
ENCD-3691 Restore select button (for object picker) and audits to file listing view

davisagli:


davisagli:
I think "there is no button to view the error messages for individual accessions in the browse" in the ticket refers to the audits.

forresttanaka:
Oh, OK. I wasn’t sure what that meant.

PR #2052, opened 2017-11-20
ENCD-3656 include ssh keys in deploy

caseylitton:
Change to merge with v64rc4

This PR
- adds the deployer's ssh key to the cloud-config through deploy.py string format templating
- fixes pylint issues in deploy script
- removes key signing from all cloud configs
- download an authorized_keys file from an s3 bucket that should contain everyone's ssh key.

This will move the deployer's ssh key into authorized_keys2 as a backup but this will only work with dsa and rsa keys.  The s3 authorized_keys is written to ~/.ssh/authorized_keys.

### DIFF
--- a/src/encoded/commands/deploy.py
+++ b/src/encoded/commands/deploy.py
@@ -4,8 +4,9 @@ import re
 import subprocess
 import sys
 import datetime
-import base64
-import json
+
+from base64 import b64encode
+from os.path import expanduser
 
 
 BDM = [
@@ -26,10 +27,12 @@ BDM = [
         'NoDevice': "",
     },
 ]
+
+
 class spot_client(object):
     def __init__(self):
         self._spotClient = None
-    
+
     @property
     def spotClient(self):
         return self._spotClient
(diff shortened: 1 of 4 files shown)


src/encoded/commands/deploy.py
```diff
@@ -38,40 +41,47 @@ def spotClient(self):
     def spotClient(self, value):
         self._spotClient = value
 
+
 def get_spot_id(instance, client):
     SpotInstanceRequestId = instance['SpotInstanceRequests'][0]['SpotInstanceRequestId']
-    #print(list(instance.values()))
-    #print("SpotInstanceRequestId: %s" % SpotInstanceRequestId)
+    # print(list(instance.values()))
```
hitz: we probably could just delete these comments.

src/encoded/commands/deploy.py
```diff
@@ -38,40 +41,47 @@ def spotClient(self):
...
     return SpotInstanceRequestId
 
+
 def get_spot_code(instance, client, spot_id):
-    request = client.describe_spot_instance_requests(SpotInstanceRequestIds=[get_spot_id(instance, client)])
-    #print(list(request.values()))
+    request = client.describe_spot_instance_requests(
+        SpotInstanceRequestIds=[get_spot_id(instance, client)]
+    )
+    # print(list(request.values()))
```
hitz: remove

src/encoded/commands/deploy.py
```diff
@@ -38,40 +41,47 @@ def spotClient(self):
...
-    #print("\n Code Status: %s" % code_status)  
-    #for key, value in request.items():
-     #  if key == 'SpotInstanceRequests':
-      #      for item in value:
-       #         for i in item:
-        #            if i == 'Status':
-         #               for j in item[i]:
-          #                  if j == 'Code':
-           #                     code_status = item[i][j]
+    # print("\n Code Status: %s" % code_status)
```
hitz: remove all these

src/encoded/commands/deploy.py
```diff
@@ -131,11 +146,10 @@ def spot_instance_price_check(client, instance_type):
                     'us-west-2b',
                     'us-west-2c'
                 ],
-
-                'Name': 'product-description',
-                'Values': [
-                    'Linux/UNIX (Amazon VPC)'
-                ]
+                # 'Name': 'product-description',
```
hitz: remove

src/encoded/commands/deploy.py
```diff
@@ -200,37 +217,69 @@ def create_ec2_instances(client, image_id, count, instance_type, security_groups
...
+            ssh_pub_key = f.readline().strip()
+            return ssh_pub_key
+
+
+def get_user_data(commit, config_file, data_insert):
+    cmd_list = ['git', 'show', commit + config_file]
+    config_template = subprocess.check_output(cmd_list).decode('utf-8')
+    ssh_pub_key = read_ssh_key()
+    if not ssh_pub_key:
+        # We can block deploy or add empty line??
```
hitz: There is no reason in principle why someone should not be able to deploy a (demo) instance without a key.
So, I think it should just maybe print a warning:  "WARNING:  User is not authorized with ssh access to new instance because they have no ssh key"

src/encoded/commands/deploy.py
```diff
@@ -200,37 +204,80 @@ def create_ec2_instances(client, image_id, count, instance_type, security_groups
...
+    ssh_pub_key = read_ssh_key()
+    if not ssh_pub_key:
+        print(
+            "WARNING: User is not authorized with ssh access to "
+            "new instance because they have no ssh key"
+        )
+    data_insert['LOCAL_SSH_KEY'] = ssh_pub_key
+    # aws s3 authorized_keys folder
+    auth_base = 's3://encoded-conf-prod/ssh-keys'
+    auth_type = 'prod'
```
Bek: It would be safer if default value for auth_type was 'dev', and flipped to 'prod' only when specified otherwise. This can be achieved by setting default to 'dev' for '--profile-name' at argparse level. Although we don't follow the convention of always setting default argument values, in this case it is a neat opportunity.

caseylitton: I thought about this too, but 'prod' was chosen in the end for the auth_type default because it is the most limiting.  This way ssh access defaults to limited admin keys instead of all keys.              

I have a refactor deploy script branch on the heals of this one.  Maybe we should revisit this in there?  I'm already rearranging the args in there so it's a better place to fix the defaults.

PR #2054, opened 2017-11-21
ENCD-3707-scrub-proposed-preliminary

keenangraham:
Remove a few more mentions of proposed and preliminary statuses where they are no longer needed. You'll still find them mentioned in changelogs and upgrades but this is expected.

### DIFF
--- a/src/encoded/audit/experiment.py
+++ b/src/encoded/audit/experiment.py
@@ -2135,7 +2135,7 @@ def audit_experiment_documents(value, system, excluded_types):
     '''
     Experiments should have documents.  Protocol documents or some sort of document.
     '''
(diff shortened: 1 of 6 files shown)


caseylitton:
What about 'proposed' in these files?
-status-proposed in ./src/encoded/static/scss/encoded/modules/_characterizations.scss
-QUICKVIEW_STATUSES_BLOCKED in ./src/encoded/visualization.py
-proposed_count in ./src/encoded/schemas/award.json

forresttanaka:
I reverted the `proposed_count` change, but left the scss the way it is, as more things in there exist that aren’t used anymore, and ENCD-3629 will clean all of that out.

PR #2057, opened 2017-11-28
holding for 3704: ENCD-3705 limit collection

forresttanaka:
The key to this ticket is in the FacetChart component which now splits the facet terms into individual and “other” sections, if the number of terms exceeds the MAX_CHART_ITEMS constant. I store the values and labels as properties of the FacetChart component so that they’re accessible from the callbacks that actually draw the charts. Note that I _could_ store them as React component state too, but changing React state triggers a redraw, which would be unnecessary because these component properties are needed to draw the charts in the first place.

Changing MAX_CHART_ITEMS alone and rebuilding changes the number of items displayed before “other.”

Added two nested functions to the componentDidUpdate() method to generate query strings because both the legend-rendering callback and the doughnut-rendering click callback need to generate the same query strings under the same circumstances.

### DIFF
--- a/src/encoded/static/components/collection.js
+++ b/src/encoded/static/components/collection.js
@@ -5,7 +5,6 @@ import * as globals from './globals';
 import DataColors from './datacolors';
 
 
-
 const MAX_FACET_CHARTS = 3; // Maximum number of facet charts to display.
 const MAX_CHART_ITEMS = 9; // Maximum number of items to display in a chart; combine rest into "Other."
 
@@ -22,6 +21,17 @@ class FacetChart extends React.Component {
     }
 
     componentDidUpdate() {
+        // Utility function to generate a query string component using the given field and label.
+        function generateQuery(field, label) {
+            return `${field}=${encodeURIComponent(label).replace(/%20/g, '+')}`;
+        }
+
+        // Same as generateQuery, but generates a query string with multiple elements. The facet
+        // terms to use as a source of the data gets passed in the `terms` parameter.
+        function generateMultipleQuery(field, terms) {
+            return terms.reduce((combined, term) => `${combined}&${generateQuery(field, term.key)}`, '');
+        }
+
         // This method might be called because we're drawing the chart for the first time (in
         // that case this is an update because we already rendered an empty canvas before the
         // chart.js module was loaded) or because we have new data to render into an existing
@@ -77,7 +87,15 @@ class FacetChart extends React.Component {
                             const text = [];
                             text.push('<ul>');
                             for (let i = 0; i < chartData.length; i += 1) {
-                                const searchUri = `${baseSearchUri}&${facet.field}=${encodeURIComponent(chartLabels[i]).replace(/%20/g, '+')}`;
+                                let queryString = '';
+                                if (this.otherTerms.length && this.otherTermsIndex === i) {
+                                    // Handle a click in the "other" term. We need to generate a
+                                    // query string that combines all the "other" terms.
+                                    queryString = generateMultipleQuery(facet.field, this.otherTerms);
+                                } else {
+                                    queryString = generateQuery(facet.field, chartLabels[i]);
+                                }
+                                const searchUri = `${baseSearchUri}&${queryString}`;
                                 if (chartData[i]) {
                                     text.push(`<li><a href="${searchUri}">`);
                                     text.push(`<i class="icon icon-circle chart-legend-chip" aria-hidden="true" style="color:${chartColors[i]}"></i>`);
@@ -89,12 +107,21 @@ class FacetChart extends React.Component {
                             return text.join('');
                         },
                         onClick: (e) => {
+                            let queryString = '';
+
                             // React to clicks on pie sections
                             const activePoints = this.chartInstance.getElementAtEvent(e);
                             if (activePoints[0]) {
                                 const clickedElementIndex = activePoints[0]._index;
                                 const chartLabels = this.chartInstance.data.labels;
-                                const searchUri = `${baseSearchUri}&${facet.field}=${encodeURIComponent(chartLabels[clickedElementIndex]).replace(/%20/g, '+')}`;
+                                if (this.otherTerms.length && this.otherTermsIndex === clickedElementIndex) {
+                                    // The click was in the "other" term. We need to generate a
+                                    // query string that combines all the "other" terms.
+                                    queryString = generateMultipleQuery(facet.field, this.otherTerms);
+                                } else {
+                                    queryString = generateQuery(facet.field, chartLabels[clickedElementIndex]);
+                                }
+                                const searchUri = `${baseSearchUri}&${queryString}`;
                                 this.context.navigate(searchUri);
                             }
                         },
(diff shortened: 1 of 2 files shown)


src/encoded/static/components/collection.js
```diff
@@ -76,7 +87,17 @@ class FacetChart extends React.Component {
                             const text = [];
                             text.push('<ul>');
                             for (let i = 0; i < chartData.length; i += 1) {
-                                const searchUri = `${baseSearchUri}&${facet.field}=${encodeURIComponent(chartLabels[i]).replace(/%20/g, '+')}`;
+                                let queryString = '';
+                                if (this.otherTerms.length && this.otherTermsIndex === i) {
+                                    // Handle a click in the "other" term. We need to generate a
+                                    // query string that combines all the "other" terms.
+                                    queryString = generateMultipleQuery(facet.field, this.otherTerms);
```
caseylitton: Is this dependent on the loop?  It looks like 
```
queryString = generateMultipleQuery(facet.field, this.otherTerms);
```
could be called once when the chart is created since otherTerms is set in render.

src/encoded/static/components/collection.js
```diff
@@ -88,12 +109,23 @@ class FacetChart extends React.Component {
...
                             // React to clicks on pie sections
                             const activePoints = this.chartInstance.getElementAtEvent(e);
                             if (activePoints[0]) {
                                 const clickedElementIndex = activePoints[0]._index;
                                 const chartLabels = this.chartInstance.data.labels;
-                                const searchUri = `${baseSearchUri}&${facet.field}=${encodeURIComponent(chartLabels[clickedElementIndex]).replace(/%20/g, '+')}`;
+                                if (this.otherTerms.length && this.otherTermsIndex === clickedElementIndex) {
+                                    // The click was in the "other" term. We need to generate a
+                                    // query string that combines all the "other" terms.
+                                    queryString = generateMultipleQuery(facet.field, this.otherTerms);
```
caseylitton: Same as above, can this be called once when the chart is created?

src/encoded/static/components/collection.js
```diff
@@ -109,22 +141,58 @@ class FacetChart extends React.Component {
...
+
+            // Keep track of the "other" terms and its location (stored in this.otherTermsIndex) in
+            // the values/labels arrays so that the callbacks to handle clicks in the legend and
+            // the doughnut slices can handle the "other" term.
+            this.otherTerms = otherTerms;
+            this.otherTermsIndex = this.labels.length - 1;
+        } else {
+            // Remember we don't have enough terms to need an "other" term.
+            this.otherTerms = [];
+            this.otherTermsIndex = 0;
```
caseylitton: Same as above, I know what's going on but it's a confusing code block since 'otherTermsIndex' is not an index for 'otherTerms'.  I'm not telling you to change your variable name but when I first saw this I thought, 'Why are we setting the index variable to 0 when the array is empty?'

src/encoded/static/components/collection.js
```diff
@@ -109,22 +141,58 @@ class FacetChart extends React.Component {
...
+        }
+
         // Check whether we have usable values in one array or the other we just collected (we'll
-        // just use `this;values` here) to see if we need to render a chart or not.
+        // just use `this.values` here) to see if we need to render a chart or not.
```
caseylitton: Unnecessary changes!

src/encoded/static/components/collection.js
```diff
@@ -109,22 +141,58 @@ class FacetChart extends React.Component {
...
+        // as another data item to the array of values.
+        if (otherTerms.length) {
+            this.values.push(otherTerms.reduce((sum, term) => sum + term.doc_count, 0));
+            this.labels.push('other');
+
+            // Keep track of the "other" terms and its location (stored in this.otherTermsIndex) in
+            // the values/labels arrays so that the callbacks to handle clicks in the legend and
+            // the doughnut slices can handle the "other" term.
+            this.otherTerms = otherTerms;
+            this.otherTermsIndex = this.labels.length - 1;
```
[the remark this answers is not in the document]
hitz: maybe indexOfOther?  Just for future reference.

forresttanaka:
Discussed with Ben, and I’m certain your optimizations are correct (I haven’t looked in detail at it, but it feels like you’re probably correct), and Ben suggests I just look at that again for ENCD-3704 so that we avoid another QA cycle for this one. Ben added a comment to ENCD-3704 to remind me about it. This ticket exists to prepare for ENCD-3704.

On the one-character comment change, that’s based on our summer interns being led astray by typos in comments because none of the interns we’ve had were strong at Javascript, so they thought anything in a comment was gospel and would spend time researching what something like `this;values` meant. I’d really like to correct these when found around code I’m modifying. I don’t know if we’ll get summer interns again, but we’ve had them two years running.

caseylitton:
Okay, let me know which needs review and when to merge one or the other.   

I like marking the unnecessary changes, not to revert them, just to make sure they we're all on the same page.  I didn't know we had summer interns...reviewing their code should be fun.

forresttanaka:
We’re looking at merging this one, then once ENCD-3704 gets scheduled and I start work on it, to revisit your optimizations since I’ll be working in roughly the same code (certainly the same file). Really, this code isn’t very important right now in that very few people would care if it exists or not, much less optimized or not. But it’ll become more important once ENCD-3704 becomes real.

src/encoded/static/components/collection.js
```diff
@@ -109,22 +141,58 @@ class FacetChart extends React.Component {
...
+        // If we have more facet terms than we allow to be displayed (determined through
+        // MAX_CHART_ITEMS), then just get MAX_CHART_ITEMS to display as usual, and collect the
+        // rest to display as "other."
+        if (facet.terms.length > MAX_CHART_ITEMS) {
+            // We have more facet terms than we display individually. First get the terms to
+            // display as individual items.
+            terms = facet.terms.slice(0, MAX_CHART_ITEMS);
+
+            // Now collect the terms to display as "other."
+            otherTerms = facet.terms.slice(MAX_CHART_ITEMS);
```
davisagli: If there are 10 items we'll end up with an "other" wedge that only represents a single item, which seems non-ideal. So maybe slice at MAX_CHART_ITEMS - 1 (which basically reinterprets MAX_CHART_ITEMS as the max *including* the "other" item).

src/encoded/static/components/collection.js
```diff
@@ -109,22 +141,58 @@ class FacetChart extends React.Component {
...
 
         // Extract the arrays of labels from the facet keys, and the arrays of corresponding counts
         // from the facet doc_counts. Only use non-zero facet terms in the charts. IF we have no
         // usable data, both these arrays have no entries. componentDidMount assumes these arrays
         // have been populated.
         this.values = [];
         this.labels = [];
-        facet.terms.forEach((term) => {
+        terms.forEach((term) => {
             if (term.doc_count) {
```
davisagli: Hmm, so if there are terms with no doc_count we can end up with a chart that has fewer than MAX_CHART_ITEMS wedges but still has an "other"...

forresttanaka:
Thanks @davisagli. That’s good stuff. @caseylitton at this stage, as I don’t think it’s worth spending time on this ticket as it’s just some prep work for ENCD-3704, I recommend closing this, setting ENCD-3705 to in progress, and schedule for the same release as ENCD-3704 if @hitz agrees.

PR #2065, opened 2017-12-02
ENCD-3699 Remove mixed run type audit from DNase

gabdank:
1. excluding DNase from mixed run type audit
2. adding a test for that exclusion

### DIFF
--- a/src/encoded/audit/experiment.py
+++ b/src/encoded/audit/experiment.py
@@ -1785,7 +1785,9 @@ def audit_experiment_consistent_sequencing_runs(value, system, files_structure):
     if not value.get('replicates'):
         return
 
-    if value.get('assay_term_name') not in [
+    assay_term_name = value.get('assay_term_name')
+
+    if assay_term_name not in [
             'ChIP-seq',
             'DNase-seq',
             'genetic modification followed by DNase-seq']:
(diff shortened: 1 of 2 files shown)


src/encoded/audit/experiment.py
```diff
@@ -1785,7 +1785,9 @@ def audit_experiment_consistent_sequencing_runs(value, system, files_structure):
     if not value.get('replicates'):
         return
 
-    if value.get('assay_term_name') not in [
+    assay_term_name = value.get('assay_term_name')
+
```
caseylitton: Remove empty line

src/encoded/audit/experiment.py
```diff
@@ -1864,29 +1858,40 @@ def audit_experiment_consistent_sequencing_runs(value, system, files_structure):
...
-                             'in experiment {} '.format(value['@id']) + \
-                             'has endedness {} '.format(i_pairs) + \
-                             ' that differ from replicate {},'.format(keys[index_j]) + \
-                             ' which has {}.'.format(j_pairs)
-                    yield AuditFailure('mixed run types',
-                                       detail, level='WARNING')
-
-    return
+    # run type consistency is relevant only for ChIP-seq
+    if assay_term_name == 'ChIP-seq':
```
caseylitton: We shouldn't need to check this.  replicate_pairing_statuses will be empty if assay_term_name is not 'ChIP-seq'.  If we don't need this check then I believe we don't need to rearrange the code blocks.  

I suggest putting the code back to it's original state with the three line changes at the top of the file. 

Rearranging the code blocks adds a bunch of work to what could be a three line change.  If we keep the rearrangement then we'll need to fix the pep8 errors, string formatting, and there is probably a better way to write those nested for loops.  When working in preexisting files make as few changes as possible because it is easier and safer.  Lines of code that are changed need to be reviewed and fixed if needed.

PR #2067, opened 2017-12-05
ENCD-3721 Allow 'in vitro sample' to have biosample_term_name

estherc:
Changes include:

- Adding a dependency for NTR id with biosample_type = cell-free sample and making biosample_term_id required
- Bumping up the schema_version number to 15 and updating the changelog
- Adding an upgrade to convert all existing experiments using in vitro sample to cell-free sample, including adding biosample_term_id = NTR:0000471 and biosample_term_name = none
- Adding an upgrade test to verify the upgrade
- Updating fixtures and tests in other places within encoded to reflect this change


caseylitton:
Code change looks good, tests already passed prior to rebase.

PR #2070, opened 2017-12-06
[ES5] ENCD-3748 Return dataset facet back to Publications

Bek:


src/encoded/search.py
```diff
@@ -582,7 +582,7 @@ def format_facets(es_results, facets, used_filters, schemas, total, principals):
         if agg_name not in aggregations:
             continue
         terms = aggregations[agg_name][agg_name]['buckets']
-        if not any(bucket['doc_count'] > 0 for bucket in terms if 'doc_count' in bucket):
```
caseylitton: Since we're no longer using 'terms' in the condition we should move
```
terms = aggregations[agg_name][agg_name]['buckets']
```
down to above this condition block.
```
if facet_type == 'exists':
```
To group the code a little better.

src/encoded/search.py
```diff
@@ -582,7 +582,7 @@ def format_facets(es_results, facets, used_filters, schemas, total, principals):
...
+        if not aggregations[agg_name]['doc_count'] > 0:
```
caseylitton: We're using 
```
aggregations[agg_name]['doc_count']
```
twice so maybe set it to a variable
```
total = aggregations[agg_name]['doc_count']
```
Up to you.  I just find this way easier to read since we're adding it to a dict in the end.

PR #2073, opened 2017-12-06
[ES5] ENCD-3746 Fix pop error

tdreszer:
Third PR is a charm?  This can be patched to es5-master and vis_cache reindexed.  That should fix the trackhubs failures.


caseylitton:
Original PR notes from Tim:
```
Both v64rc2 and es5 demos could be patched with small fix. Without it, the secondary indexer is failing to fill vis_cache and large trackhubs are failing. I have tested locally and on a demo (though it was an impure demo with region_indexer v2 changes).
```

PR #2079, opened 2017-12-11
ENCD-3661 Render raw derived_from files in file graphs

forresttanaka:
### Jest tests

The vast bulk of the files changed involve the Jest tests in `src/encoded/static/components/__tests__/*.js`. graph-test.js tests now have to specify `selectedAssembly` and `selectedAnnotation` when calling `assembleGraph` because that can now be critical in determining whether nodes get rendered in the graph or not. Some changes also involve changing the fastq.js Jest test data from an object to an array of objects now that I added one fastq, and changing the references to this test data to reflect this. I also added a new Jest test to test a simple case specific to this branch’s changes.

### Local test data

I changed some files in local test data (/src/encoded/tests/data/inserts) to add a test case affected by this branch in experiment .

### filegallery.js changes

The key changes for this ticket are in the `assembleGraph` function in filegallery.js, as its job is to take a dataset’s files and make a Javascript object that describes the entire graph. Functions in graph.js then convert this object to a chart.js graph. As the existing algorithm only takes processed files (usually described as `matchingFiles` in the code, though I’ve corrupted the meaning a little bit for this ticket — more on that in a bit) and the files they directly derive from to add to the graph, this ticket changes the algorithm to travel up the `derived_from` chains of every processed file using a new recursive function, `collectDerivedFroms`. Its job is to generate a list of all parent `derived_from` files that eventually lead to each processed file.

One of the goals of this ticket was to have as little modification to existing graphing code as possible, as Ben has an idea on the table that might cause a major rewrite of graphing in 2018. So I kept the existing `allDerivedFroms` variable, but use the object generated by `collectDerivedFroms` to generate its contents. `allDerivedFroms` has a key for every file that exists in another file’s `derived_from` in the same dataset. I used to simply collect derived-from files by going through each processed file’s `derived_from` fields, but now I have to use the output of `collectDerivedFroms` to do this.

So I have a nested loop that turns the list from `collectDerivedFroms` on its head to generate `allDerivedFroms`, and that’s described in the comments.

Another major change is that `matchingFiles` no longer only holds files matching the current assembly and annotation (i.e. processed files). It now also holds all derived_from files because I now have to render those potentially with their own derived-from files and steps, so I use the same process to render raw files (which make up the bulk, though not all, of `derived_from` files) as I do processed files. Some of the large code deletions involve `derived_from` file processing because they’re now already included in `matchingFiles`.

Many of the large changes later in the file involve moving existing code around to fit the new model.

### Object key iterations

The new `derived_from` chain-traversing code does a _lot_ of iterating through objects’ keys. I’ve historically been using `Object.keys(obj).forEach` to traverse these as the AirBnB style guide doesn’t allow for…in loops, but new performance tests I did showed `Object.keys(obj).forEach` consistently slowest in current versions of Safari (11.0.1), Chrome (62.0.3202.94), and Firefox (57.0.1).

https://jsperf.com/object-keys-vs-for-in-with-closure/258

“Object.keys for loop” showed best or nearly best performance in current versions of Safari, Chrome, and Firefox, so I think this method is the way to go going forward for code like this that’s so dependent on traversing single-level objects. For less critical code, the clarity and scoping of `Object.keys(obj).forEach` might be preferable.


caseylitton:
Sorry, this got a little long...

Please rebase 'ENCD-3661-raw-derived' on origin/master and then force push back up to the remote.  Git tower has a way to rebase, https://www.git-tower.com/help/mac/branches-and-tags/merge-rebase.  If you're using pull then there are ways to configure pull so that it rebases instead of merges.  

Take a look at the commit history in command line.  Make a copy of your local feature branch.  Do a 'git log --pretty=format:"%h%x09%an%x09%x09%s"' .  You'll see that your feature commits are intermixed with origin/master commits.  Do a 'git rebase origin/master' and then the same git log.  Now all of your feature commits are sitting on top of 'ENCD-3710 Remove unused code from audits of Experiment (#2059)'.  Which is the current head of master.  This is want we want.   New code on top of master. 

If you were to git diff your original feature with the rebased feature it would be identical except for the histories.  Given that rebase is no more work than merge and ends up with a clean, readable commit history, unlike merging, please start using rebase exclusively to sync with master.

forresttanaka:
Closing in favor of a copy of this branch that has been rebased to master.

PR #2087, opened 2017-12-13
ENCD 3749: Update README for ES5 and dependency requirements

kakaly:
This PR contains

- ES5 instructions 

- Minor install dependency requirements

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,8 +38,6 @@ brew link --force postgresql@9.3
 brew install --force node@6
 brew link node@6 --force
 brew cask install java8
-brew tap garrow/homebrew-elasticsearch17
-brew install elasticsearch@1.7
 brew install elasticsearch@5.6
 pip3 install typing
 ```
--- a/README.md
+++ b/README.md
@@ -27,7 +27,57 @@ These are the primary software versions used in production, and you should be ab
   - `brew doctor`
 
 
-### **2. Python**  
+### **2. (Mac) Install or update other dependencies:**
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -27,7 +27,57 @@ These are the primary software versions used in production, and you should be ab
...
+brew install libevent libmagic libxml2 libxslt openssl graphviz nginx
+brew install freetype libjpeg libtiff littlecms webp chromedriver
+brew tap petere/postgresql
+brew install postgresql@9.3
+brew link --force postgresql@9.3
+brew install --force node@6
+brew link node@6 --force
+brew cask install java8
+brew tap garrow/homebrew-elasticsearch17
+brew install elasticsearch@1.7
```
caseylitton: Do we need this line and the one above it?  They are es 1.7

README.md
```diff
@@ -27,7 +27,57 @@ These are the primary software versions used in production, and you should be ab
   - `brew doctor`
 
 
-### **2. Python**  
+### **2. (Mac) Install or update other dependencies:**
```
caseylitton: Why did we change the order of the python and mac install?  Is that required?

kakaly: Actually, I had to do the mac installs before I even installed the python ones since they had some dependencies. Thats why I interchanged them

README.md
kakaly: I agree. I will remove the es1.7 instructions.

PR #2088, opened 2017-12-14
ENCD-3745 Add platform to file and exp facets

gabdank:
Adding platfrom embedding in file schema
Adding back the platform facet on experiment and file search pages


gabdank:
@hitz I think @keenangraham timed this demo against other demo and found no major differences in loading search pages for files.

PR #2091, opened 2017-12-15
ENCD-3743 Fix read depth audits for DNase and ChIP

gabdank:
1. DNase insufficient read depth - should be called now "low read depth"
2. TF and narrow peaks ChIP NOT_COMPLIANT controls that were called: "control low read depth" will be called "control insufficient read depth"


src/encoded/tests/test_audit_experiment.py
```diff
@@ -2185,6 +2185,8 @@ def test_audit_experiment_dnase_low_read_length(testapp,
                                                 'date_released': '2016-01-01',
                                                 'assay_term_name': 'DNase-seq'})
     res = testapp.get(base_experiment['@id'] + '@@index-data')
+    for error in collect_audit_errors(res):
+        print (error)
```
hitz: should we get rid of this print statement in test code?  Or maybe it gets swallowed in stdout anyway...

gabdank: removed the print statement (sorry, I forgot to do it)

PR #2094, opened 2017-12-19
ENCD-3579, ENCD-3473 - Batch download isues

kakaly:
Issue: Line 310 and 318 throws a stack trace when 'files' key is not present in the dict, 'exp'
Fix: Add an 'if block' and filter out the entries that do not have 'files' key
AWS Instance: http://ec2-52-36-22-200.us-west-2.compute.amazonaws.com/ has the fix

### DIFF
--- a/src/encoded/batch_download.py
+++ b/src/encoded/batch_download.py
@@ -307,7 +307,7 @@ def batch_download(context, request):
     files = [metadata_link]
     if 'files.file_type' in param_list:
         for exp in results['@graph']:
-            for f in exp.get('files',[]):
+            for f in exp.get('files', []):
                 if f['file_type'] in param_list['files.file_type']:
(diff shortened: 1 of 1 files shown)


src/encoded/batch_download.py
```diff
@@ -307,19 +307,21 @@ def batch_download(context, request):
     files = [metadata_link]
     if 'files.file_type' in param_list:
         for exp in results['@graph']:
-            for f in exp['files']:
-                if f['file_type'] in param_list['files.file_type']:
+            if 'files' in exp:
```
hitz: I think for f in exp.get('files',[]) better than the if ... for.

kakaly: Sounds good. I made the changes.

PR #2100, opened 2017-12-21
ENCD-3546 Create 'save and add' button to add forms

davisagli:


davisagli:
@forresttanaka I fixed the first one; for the second, do we have a style to make a button appear like a link?

forresttanaka:
It would have to be styled to fit that dropdown, I’m pretty sure.

I _think_ a viable alternative is to keep it an <a> but just remove the href. AirBnB is fine with that. It technically becomes an HTML5 placeholder link, but with a click handler, it seems like it should work.

PR #2105, opened 2018-01-12
ENCD-2567 New report page column selector modal

forresttanaka:
I used the existing Modal module used in several other places of the portal. Mostly this involved changing the existing ColumnSelector component to bring up this modal instead of the drop-down menu. I use the Modal in self-managed mode rather than plug-and-play mode so that its set of selected columns can be updated on mount — in plug-and-play mode, the modal doesn’t get mounted and unmounted as it gets displayed and dismissed, leaving its state with the selected columns potentially behind the reality coming from the URL query string.

I removed the toggleColumn method from the Report component because the columns no longer update as you select which columns you want to include or not. Instead a new setColumnState method gets called when the Modal gets dismissed, with the new set of selected columns set in there.

I also modified the Modal module itself to add custom CSS classes to the footer, as well as a custom title to the default Submit button, needed because the new report modal uses the default Submit button but renames it “Select.”

### DIFF
--- a/src/encoded/static/components/report.js
+++ b/src/encoded/static/components/report.js
@@ -400,9 +400,10 @@ class ColumnSelector extends React.Component {
             // Toggle the `visible` state corresponding to the column whose checkbox was toggled.
             // Then set that as the new React state which causes a redraw of the modal with all the
             // checkboxes in the correct state.
-            const oldColumns = Object.assign({}, prevState.columns);
-            oldColumns[columnPath].visible = !oldColumns[columnPath].visible;
-            return { columns: oldColumns };
+            const columns = Object.assign({}, prevState.columns);
+            columns[columnPath] = Object.assign({}, columns[columnPath]);
+            columns[columnPath].visible = !columns[columnPath].visible;
+            return { columns };
         });
     }
 
@@ -416,11 +417,12 @@ class ColumnSelector extends React.Component {
         // Called when the "Select all" button is clicked.
         this.setState((prevState) => {
             // For every column in this.state.columns, set its `visible` property to true.
-            const oldColumns = Object.assign({}, prevState.columns);
-            Object.keys(oldColumns).forEach((columnPath) => {
-                oldColumns[columnPath].visible = true;
+            const columns = Object.assign({}, prevState.columns);
+            Object.keys(columns).forEach((columnPath) => {
+                columns[columnPath] = Object.assign({}, columns[columnPath]);
+                columns[columnPath].visible = true;
             });
-            return { columns: oldColumns };
+            return { columns };
         });
     }
 
@@ -428,15 +430,16 @@ class ColumnSelector extends React.Component {
         // Called when the "Select (first) only" button is clicked.
         this.setState((prevState) => {
             // Set all columns to invisible first.
-            const oldColumns = Object.assign({}, prevState.columns);
-            const columnPaths = Object.keys(oldColumns);
+            const columns = Object.assign({}, prevState.columns);
+            const columnPaths = Object.keys(columns);
             columnPaths.forEach((columnPath) => {
-                oldColumns[columnPath].visible = false;
+                columns[columnPath] = Object.assign({}, columns[columnPath]);
+                columns[columnPath].visible = false;
             });
 
             // Now set the column for the first key to true so that only it's selected.
-            oldColumns[columnPaths[0]].visible = true;
-            return { columns: oldColumns };
+            columns[columnPaths[0]].visible = true;
+            return { columns };
         });
     }
 
@@ -575,7 +578,7 @@ class Report extends React.Component {
         // `newColumns` has the user's chosen columns from the modal, while `columns` has the
         // columns selected by the query string.
         const parsedUrl = url.parse(this.context.location_href, true);
-        parsedUrl.query.field = Object.keys(newColumns).filter(columnPath => newColumns[columnPath].visible);;
+        parsedUrl.query.field = Object.keys(newColumns).filter(columnPath => newColumns[columnPath].visible);
         delete parsedUrl.search;
         this.context.navigate(url.format(parsedUrl));
     }
--- a/src/encoded/static/components/report.js
+++ b/src/encoded/static/components/report.js
@@ -296,8 +296,10 @@ class ColumnSelectorControls extends React.Component {
         this.handleSortChange = this.handleSortChange.bind(this);
     }
 
-    // Called when the user changes the sorting option.
     handleSortChange(e) {
+        // Called when the user changes the sorting option. Sets a component state so that the
+        // controlled <select> component renders properly. It then calls the parent's callback to
+        // react to the new sorting option.
         this.setState({ selectedSort: e.target.value });
         this.props.handleSortChange(e.target.value);
     }
@@ -368,7 +370,11 @@ class ColumnSelector extends React.Component {
         }
 
         // Either a checkbox is being turned on, or it's being turned off and another checkbox is
-        // still checked. Change the component state to reflect the new checkbox states.
+        // still checked. Change the component state to reflect the new checkbox states. Presumably
+        // if the setState callback returned no properties setState becomes a null op, so the above
+        // test could be done inside the setState callback. The React docs don't say what happens
+        // if you return no properties (https://reactjs.org/docs/react-component.html#setstate) so
+        // I avoided doing this.
         this.setState((prevState) => {
             // Toggle the `visible` state corresponding to the column whose checkbox was toggled.
             // Then set that as the new React state which causes a redraw of the modal with all the
@@ -396,7 +402,7 @@ class ColumnSelector extends React.Component {
     }
 
     handleSelectOne() {
-        // Called when the "Select first only" button is clicked.
+        // Called when the "Select (first) only" button is clicked.
         this.setState((prevState) => {
             // Set all columns to invisible first.
             const columnPaths = Object.keys(prevState.columns);
--- a/src/encoded/static/components/report.js
+++ b/src/encoded/static/components/report.js
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import _ from 'underscore';
 import url from 'url';
+import { Modal, ModalHeader, ModalBody, ModalFooter } from '../libs/bootstrap/modal';
 import { svgIcon } from '../libs/svg-icons';
 import { FetchedData, Param } from './fetched';
 import * as globals from './globals';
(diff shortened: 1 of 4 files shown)


src/encoded/static/components/report.js
```diff
@@ -411,19 +552,14 @@ class Report extends React.Component {
...
-        const queryFields = parsedUrl.query.field ? (typeof parsedUrl.query.field === 'object' ? parsedUrl.query.field : [parsedUrl.query.field]) : undefined;
-        const columns = columnChoices(schema, queryFields);
 
-        const fields = [];
-        _.mapObject(columns, (column, path) => {
-            if (path === toggledPath ? !column.visible : column.visible) {
-                fields.push(path);
-            }
-        });
+        const fields = Object.keys(newColumns).filter(columnPath => newColumns[columnPath].visible);
```
caseylitton: Do we need the fields variable since we're now just putting it directly into parsedUrl.query.field?

forresttanaka: We could. Does this need to change? It’s zero cost as far as Javascript is concerned.

src/encoded/static/libs/bootstrap/modal.js
```diff
@@ -207,7 +207,7 @@ export class ModalFooter extends React.Component {
         // given function. Note: if you pass `null` in the submitBtn property, this component
         // thinks that's a function because of an old Javascript characteristic.
         if (submitBtn) {
-            submitBtnComponent = (typeof submitBtn === 'object') ? submitBtn : <button className="btn btn-info" onClick={this.submitModal}>Submit</button>;
+            submitBtnComponent = (typeof submitBtn === 'object') ? submitBtn : <button className="btn btn-info" onClick={this.submitModal}>{submitTitle || 'Submit'}</button>;
```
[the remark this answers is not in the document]
forresttanaka: That’s a good idea. Is it worth changing? I’m not asking rhetorically; I just don‘t know if it’s worth going through another QA cycle.

src/encoded/static/libs/bootstrap/modal.js
```diff
@@ -233,15 +233,14 @@ export class ModalFooter extends React.Component {
         }
 
         return (
-            <div className="modal-footer">
+            <div className={`modal-footer${addCss ? ` ${addCss}` : ''}`}>
```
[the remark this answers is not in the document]
forresttanaka: If `addCss` is an empty string, then it’s false and the final string should just be "modal-footer." If it’s not an empty string because someone passed it in, then it has to be added with a preceding space. If we didn’t need the space it could just be a conditionless ${addCss}. We could also just add a space to "modal-footer ", but then when `addCss` is empty, then you’d see `class="modal-footer "` with an unneeded space after it.

src/encoded/static/libs/bootstrap/modal.js
```diff
@@ -252,6 +251,8 @@ ModalFooter.propTypes = {
         PropTypes.object, // Submit button is a React component; just render it
         PropTypes.func, // Function to call when default-rendered Submit button clicked
     ]),
+    submitTitle: PropTypes.string, // Title for default submit button
+    addCss: PropTypes.string, // CSS classes to add to modal header
```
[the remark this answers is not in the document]
forresttanaka: I hate making the assumption of what someone would and would not expect, and just try to always add a comment to each property so that anyone reading it would know what it is without needing context. But yes, it should be footer. Should I change that? I once corrected a clearly incorrect comment and you said, “Unneeded change!”

src/encoded/static/components/report.js
caseylitton: No it's not vital so long as we keep it mind in the future.

src/encoded/static/libs/bootstrap/modal.js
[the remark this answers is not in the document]
caseylitton: No it's not vital so long as we keep it mind in the future.

src/encoded/static/libs/bootstrap/modal.js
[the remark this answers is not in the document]
caseylitton: Ah okay, I see the space.

src/encoded/static/libs/bootstrap/modal.js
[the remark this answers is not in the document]
caseylitton: Yes please do change it since we needed to edit the line of code for the feature.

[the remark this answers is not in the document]
forresttanaka: OK got it. Thanks!

src/encoded/static/components/report.js
```diff
@@ -282,53 +283,190 @@ ColumnHeader.propTypes = {
...
+
+// Displays a modal dialog with every possible column for the type of object being displayed.
+// This lets you choose which columns you want to appear in the report.
+class ColumnSelector extends React.Component {
+    constructor(props) {
+        super(props);
+
+        // Set the initial React states.
+        this.state = {
+            columns: props.columns, // Make (effectively) a local mutatable copy of the columns
```
davisagli: This doesn't make a mutable copy; it just adds a reference to the object that came in via props. We don't want to mutate an object that's controlled by another component, so we need to be careful to construct a new object whenever we make changes to this.state.columns.

src/encoded/static/components/report.js
```diff
@@ -282,53 +283,190 @@ ColumnHeader.propTypes = {
...
+        // still checked. Change the component state to reflect the new checkbox states. Presumably
+        // if the setState callback returned no properties setState becomes a null op, so the above
+        // test could be done inside the setState callback. The React docs don't say what happens
+        // if you return no properties (https://reactjs.org/docs/react-component.html#setstate) so
+        // I avoided doing this.
+        this.setState((prevState) => {
+            // Toggle the `visible` state corresponding to the column whose checkbox was toggled.
+            // Then set that as the new React state which causes a redraw of the modal with all the
+            // checkboxes in the correct state.
+            prevState.columns[columnPath].visible = !prevState.columns[columnPath].visible;
```
davisagli: https://reactjs.org/docs/react-component.html#setstate says:

> prevState is a reference to the previous state. It should not be directly mutated. Instead, changes should be represented by building a new object based on the input from prevState and props.

which means we should not be directly mutating prevState.columns

src/encoded/static/components/report.js
```diff
@@ -282,53 +283,190 @@ ColumnHeader.propTypes = {
...
+        // sets a new state for all checked report columns.
+        this.props.setColumnState(this.state.columns);
+    }
+
+    handleSelectAll() {
+        // Called when the "Select all" button is clicked.
+        this.setState((prevState) => {
+            // For every column in this.state.columns, set its `visible` property to true.
+            Object.keys(prevState.columns).forEach((columnPath) => {
+                prevState.columns[columnPath].visible = true;
```
davisagli: As above, avoid mutating prevState

src/encoded/static/components/report.js
```diff
@@ -282,53 +283,190 @@ ColumnHeader.propTypes = {
...
+        });
+    }
+
+    handleSelectOne() {
+        // Called when the "Select (first) only" button is clicked.
+        this.setState((prevState) => {
+            // Set all columns to invisible first.
+            const columnPaths = Object.keys(prevState.columns);
+            columnPaths.forEach((columnPath) => {
+                prevState.columns[columnPath].visible = false;
```
davisagli: As above, avoid mutating prevState

src/encoded/static/components/report.js
```diff
@@ -282,53 +283,190 @@ ColumnHeader.propTypes = {
...
+            });
+
+            // Now set the column for the first key to true so that only it's selected.
+            prevState.columns[columnPaths[0]].visible = true;
```
davisagli: As above, avoid mutating prevState

src/encoded/static/components/report.js
```diff
@@ -282,53 +283,190 @@ ColumnHeader.propTypes = {
...
+
+        // Get the column paths, sorting them by the corresponding column title if the user asked
+        // for that.
+        let columnPaths = Object.keys(columns);
+        if (this.state.sortOption === 'alpha') {
+            columnPaths = columnPaths.sort((aKey, bKey) => {
+                const aTitle = columns[aKey].title.toLowerCase();
+                const bTitle = columns[bKey].title.toLowerCase();
+                return (aTitle < bTitle ? -1 : (bTitle < aTitle ? 1 : 0));
+            });
```
davisagli: It would be best to define this sort comparison function at module scope, to avoid constructing a closure every time render is called.

src/encoded/static/components/report.js
```diff
@@ -282,53 +283,190 @@ ColumnHeader.propTypes = {
...
+                        {columnPaths.map((columnPath) => {
+                            const column = columns[columnPath];
+                            return <ColumnItem key={columnPath} columnPath={columnPath} column={column} toggleColumn={this.toggleColumn} />;
+                        })}
+                    </div>
+                </ModalBody>
+                <ModalFooter
+                    closeModal={this.props.closeSelector}
+                    submitBtn={this.submitHandler}
+                    submitTitle="Select"
```
davisagli: Since the buttons at the top of the modal also say "Select", I think this one would be clearer with a different title; maybe "Update Report"

src/encoded/static/components/report.js
```diff
@@ -411,19 +552,14 @@ class Report extends React.Component {
         this.context.navigate(url.format(parsedUrl));
     }
 
-    toggleColumn(toggledPath) {
+    setColumnState(newColumns) {
+        // Gets called when the user clicks the Select button in the ColumnSelector modal.
+        // `newColumns` has the same format as `columns` returned from `columnChoices`, but
+        // `newColumn s`has the user's chosen columns from the modal, while `columns` has the
```
davisagli: typo

src/encoded/static/components/report.js
forresttanaka: Is there a reference to this? I mean, the initial state is definitely a reference to the original props object, but setState _seems to_ use a copy, as I see no changes to the parent columns object when changes are made to this.state.columns in this component. The React docs don’t quite touch the topic, only saying to be careful about using props to set initial state, and I think this modal case falls under the OK-to-do case since it’s a modal with its own version of columns. I could see the argument that it’s just risky even if not documented.

src/encoded/static/components/report.js
forresttanaka: Yes, I missed that. I’ll change all those.

src/encoded/static/components/report.js
forresttanaka: Yeah, I agree.

src/encoded/static/components/report.js
davisagli: It may work in practice if React is giving us a copy, but since the React docs say not to do it, I would not assume that it will always continue to work.

davisagli: As long as you're going to change the code below that mutates things, it should be okay.

forresttanaka: React docs don’t say not to do it, but their example uses a simple value instead of an object, so mutating props wasn’t their concern. Their “Beware” note doesn’t apply to this modal, since if the props change, the modal would be unmounted. https://reactjs.org/docs/react-component.html#constructor

forresttanaka:
Yeah, that was dumb. For a while I was thinking it was an object of simple values, then was looking at it after and saw it wasn't. Was thinking of just using JSON.stringify as many like to do for deep copies. There’s also a way with underscore, but not sure it’s really better. Looking at options and will update.

davisagli:
I would guess using JSON is slower. It will be nice when we can use object spread in literals which was just approved for ES2018:
    const columns = {
        ...prevState.columns,
        columnPath: {
            ...prevState.columns[columnPath],
            visible: !prevState.columns[columnPath].visible,
        }
    };

(see http://2ality.com/2016/10/rest-spread-properties.html for more)

forresttanaka:
The JSON method isn’t fast. immutability-helper is nice, though people are saying anecdotally it’s not fast either (jsperf seems to be down). Might be best for future usage if this comes up a lot. Leaning towards just using the way you outlined for now. The underscore method is painful looking.

forresttanaka:
@davisagli Look OK you think?

forresttanaka:
Jason just reviewed a demo with all these changes and it worked fine from the user’s view.

PR #2106, opened 2018-01-16
Update travis java set defalt to java9

caseylitton:
testing through travis only


caseylitton:
Deployment is under a update,

PR #2113, opened 2018-01-22
ENCD-3803 Allow fastq_sig to accept 3 in SRR headers

gabdank:


estherc:
Regex looks fine but the comment for the fastq_signature property wasn't updated to reflect that read_end can now be (1/2/3) rather than just (1/2). This minor change should probably also be documented in the changelog as a minor change since schema version 12.

gabdank:
@estherc I have updated the changelog and the comment in file schema for fastq signature.

estherc:
OK, when this is merged with @thehushkat 's change in https://github.com/ENCODE-DCC/encoded/blob/ENCD-3341-file-output-types/src/encoded/schemas/changelogs/file.md, we should make sure they're listed under the same heading. I have started to note non-schema version bumping changes the same way in the changelogs on other branches.

PR #2115, opened 2018-01-25
ENCD-1934 Format schema displays

forresttanaka:
This branch combines my code with work from a summer intern. Two kinds of existing REST results get involved with this:

1. /profiles/ which returns JSON that only has keys for each object type, as well as the corresponding schemas as values — no `@type`.
1. /profiles/schemaname.json that contains the specific schema plus an `@type` of JSONSchema.

The changes to item.js in this branch handle the first. `<FallBack>` displays a generic JSON HTML page for any object with a non-registered `@type`, or for objects without an `@type`. The change here is redirecting to the `<AllSchemasPage>` component in schema.js if the path happens to be “/profiles/.” I verified generic JSON still gets displayed otherwise, such as displaying a GeneticModificationCharacterization object. `<AllSchemasPage>` doesn’t rely on other encoded components because it just generates a list of links to individual schema pages, as well as buttons to add a new object of each type.

The `<SchemaPage>` component in schema.js handles the second kind of REST result by registering itself to be called when an object with `@type` JSONSchema arrives. The bulk of this ticket’s changes lie in that component and its child components.

`<SchemaPage>` displays tabs using the existing TabPanel functions that lets the user select between the formatted schema through `<DisplayObject>` or the raw schema similar to how it does in production today using `<DisplayRawObject>`.

The complication with `<DisplayRawObject>` over the JSON display used in production now is I use the “brace” library to do some colorizing of the JSON instead of just displaying all-black text. `createJsonDisplay` handles this using the brace API which draws the JSON into a `<div>` with the HTML id that I pass to it. Because brace is a module the page loads on demand, I use Javascript promises so that I can track the brace instance.

`<DisplayRawObject>` takes the same schema data as `<DisplayRawObject>` but loops through each key (not included in `excludedTerms`) and uses `<TermDisplay>` to display the properties beneath them either as a simple string or as a JSON display depending on the type of each property.

For string values, booleans, or numbers, I use the `simpleTypeDisplay` map to display them depending on their type and simply display them directly. For arrays, it loops through their values and uses the `simpleTypeDisplay` for any strings, booleans, or numbers in the array. For objects, it uses the same brace library `<DisplayRawObjects>` uses, but only passes it the property being displayed within the schema instead of the whole schema. This happens in `<SchemaTermItemDisplay>` which also hides or show the JSON by reacting to clicks in the disclosure button.

The `<ChangeLog>` component that displays a markdown-formatted changelog for the schema being viewed has not changed from before. All other changes in this branch are for new styles for this page, in .scss files, including two new .scss files for _schema.scss and _item.scss.


src/encoded/static/components/item.js
```diff
@@ -19,11 +22,15 @@ const Fallback = (props, reactContext) => {
                 </div>
             </header>
             {typeof context.description === 'string' ? <p className="description">{context.description}</p> : null}
-            <section className="view-detail panel">
-                <div className="container">
-                    <pre>{JSON.stringify(context, null, 4)}</pre>
-                </div>
-            </section>
+            {path === '/profiles/' ?
```
caseylitton: Is this the preferred way to break a conditional?

forresttanaka: If you mean the line break, it’s the way we established quite a while ago because it seemed most easy formatting-wise when the true condition is long. AirBnB doesn’t have anything to say about it. React doesn’t recommend anything, but their examples are similar but they add some parentheses. I _think_ older React documentation didn’t use parens.

https://reactjs.org/docs/conditional-rendering.html#inline-if-else-with-conditional-operator

src/encoded/static/components/item.js
```diff
@@ -6,11 +6,14 @@ import { FetchedData, Param } from './fetched';
...
 import * as globals from './globals';
 import { AlternateAccession } from './objectutils';
+import AllSchemasPage from './schema';
 
 
 const Fallback = (props, reactContext) => {
     const context = props.context;
-    const title = typeof context.title === 'string' ? context.title : url.parse(reactContext.location_href).path;
+    const path = url.parse(reactContext.location_href).path;
+    const title = typeof context.title === 'string' ? context.title : (path === '/profiles/' ? 'Schemas' : path);
```
davisagli: I'm not a big fan of adding special cases to the `Fallback` component like this; it feels like sidestepping the view lookup system, and it puts code for a very specific case into a component that is meant to be very generic. Is there a reason we can't add a `@type` to the response from `/profiles/`? If so, maybe it would be better to make globals.contentViews.fallback check a mapping of path to `@type` before using this `Fallback` component?

hitz: yeah I don't like this either, should be fixed on the backend.

forresttanaka: Would this be a snovault change?

hitz: yeah I think so.

forresttanaka: OK. I’ll look into that.

forresttanaka: develop/snovault/src/snovault/schema_views.py:

```
@view_config(route_name='schemas', request_method='GET',
             decorator=etag_app_version_effective_principals)
def schemas(context, request):
    types = request.registry[TYPES]
    schemas = {}
    for type_info in types.by_item_type.values():
        name = type_info.name
        schemas[name] = _annotated_schema(type_info, request)

    schemas['_subtypes'] = subtypes = {}
    schemas['@type'] = ['JSONSchemas'];
    for name, type_info in types.abstract.items():
        subtypes[name] = type_info.subtypes

    return schemas
```

Results in:

```
...
        "changelog": "/profiles/changelogs/antibody_characterization.md",
        "@type": [
            "JSONSchema"
        ]
    },
    "@type": [
        "JSONSchemas"
    ],
    "TreatmentConcentrationSeries": {
        "title": "Treatment concentration series",
        "description": "Schema for submitting metadata for a treatment concentration series.",
        "id": "/profiles/treatment_concentration_series.json",
...
```

forresttanaka:
Jason reviewed a demo with the change for /profiles/ to have an `@type` property with ['JSONSchemas'] in it, and AllSchemasPage now gets called from the view registry when this `@type` appears. I returned item.js back to what it is in master.

This branch has a change to buildout.cfg to get the snovault branch ENCD-1934-schemas-at-type which just has the one change to add the `@type` property. I verified the demo has this version of snovault.

src/encoded/static/scss/style.scss
```diff
@@ -103,13 +103,15 @@ $brand-info: #4F689E;
...
+        "encoded/modules/schema",
         "encoded/modules/tooltip",
         "encoded/modules/badge",
         "encoded/modules/news",
         "encoded/modules/visualize",
         "encoded/modules/home",
         "encoded/modules/blocks",
-        "encoded/modules/layout-editor";
+        "encoded/modules/layout-editor",
+        "encoded/modules/item";
```
caseylitton: Again, if we reverted changes to item, is this impacted?  If this needs to be in here should it be in the schemas feature.

src/encoded/static/components/app.js
```diff
@@ -51,6 +51,8 @@ const portal = {
                 { id: 'pipelines', title: 'Pipelines', url: '/pipelines/' },
                 { id: 'datause', title: 'Release policy', url: '/about/data-use-policy/' },
                 { id: 'dataaccess', title: 'Data access', url: '/about/data-access/' },
+                { id: 'sep-mm-2' },
```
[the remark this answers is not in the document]
forresttanaka: This generates a gray-line separator in the Materials & Methods menu for the new Schemas item.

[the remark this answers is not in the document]
forresttanaka: Any object in the array with only an `id` and nothing else generates a gray-line separator. The ID itself is the unique React key.

src/encoded/static/components/item.js
```diff
@@ -1,7 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import url from 'url';
-import { collapseIcon } from '../libs/svg-icons';
```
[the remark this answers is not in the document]
forresttanaka: This line isn’t in master either, so why’s it showing as deleted?

src/encoded/static/scss/encoded/modules/_item.scss
```diff
@@ -0,0 +1,60 @@
+.no-indent {
```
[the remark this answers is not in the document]
forresttanaka: Oh, this change was from Chloe (summer intern) and I didn’t notice it was there. I removed the code that uses this. It doesn't have anything to do with items.js — just with some components she had added there that I removed. This should be backed out.

src/encoded/static/scss/style.scss
forresttanaka: Looks like Chloe added _items.scss. It didn’t exist before. So that line should be deleted.

src/encoded/tests/test_audit_item.py
```diff
@@ -76,6 +76,10 @@ def test_audit_item_status_level_dict_contains_all_statuses_in_schema(testapp):
     status_level_keys = STATUS_LEVEL.keys()
     schemas = testapp.get('/profiles/').json
     for title, schema in schemas.items():
+        # Ignore @type schemas property.
```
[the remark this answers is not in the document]
forresttanaka: Yes, that's a better test you have.

src/encoded/static/scss/encoded/modules/_panels.scss
```diff
@@ -478,6 +478,7 @@ $nav-tab-border-style: 1px solid $nav-tab-border-color;
 $nav-tab-border-radius: 4px;
 
 .tab-nav {
+    position: relative;
```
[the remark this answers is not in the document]
forresttanaka: This CSS only affects absolutely positioned items in a tabbed panel tab bar, and we only have one case of this: the one I added in this ticket. Anything not absolutely positioned within a .tab-nav would be unaffected by this.

src/encoded/tests/test_audit_item.py
```diff
@@ -76,6 +76,10 @@ def test_audit_item_status_level_dict_contains_all_statuses_in_schema(testapp):
...
+        if type(schema) is list:
```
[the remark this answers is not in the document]
forresttanaka: Oh, OK. I had seen some people using isinstance on and some using type, but I didn’t see a discussion about inheritance.

forresttanaka:
So I’ll back out those CSS changes in a few minutes. Did you want it to go through another round of QA @caseylitton?

forresttanaka:
…and I’ll change that test line to the test you have.

src/encoded/static/components/item.js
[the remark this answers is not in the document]
caseylitton: Oh, I found this while going through the commits.  It's not in the final diff and would not have been checked into master.

PR #2122, opened 2018-01-30
ENCD-3819 Add biosample preservation method

estherc:
* Added property **preservation_method** to biosample schema with enums _cryopreservation_ and _flash-freezing_
* Updated biosample changelog with non-schema incrementing change


estherc:
Pushed with comments added now to PMI, PMI_units and preservation_method properties in biosample.json

PR #2125, opened 2018-02-02
ENCD-3757 Adjust histone broad peak read depth audit

gabdank:
(1) the minimal threshold of 3M was moved to be 5M for everything except H3K9me3 where it was moved to be 100.
(2) the maximal insufficient threshold for broad marks was lowered from 40M to 35M
(3) the low read depth control audit that cold not be called (>20M but <45) moved into INTERNAL_ACTION


cricketsloan:
This data change is the one we requested.  I am also hoping that the code is now structured in a way that the numbers can be adjusted more easily.  I expect the argument for 25M not 35M will come up.

PR #2127, opened 2018-02-07
ENCD-3823,3864,3868 Filter out restricted files from batch download, refactor, and unit tests

kakaly:
The fix here is basically looking for 'restricted' key in the files and checking if its set to 'true'. If these two conditions are met, then skip adding this file to files.txt which eventually gets downloaded as part of batch download.

Demo URL: ec2-54-191-169-118.us-west-2.compute.amazonaws.com

### DIFF
--- a/src/encoded/batch_download.py
+++ b/src/encoded/batch_download.py
@@ -291,7 +291,6 @@ def metadata_tsv(context, request):
         content_disposition='attachment;filename="%s"' % 'metadata.tsv'
     )
 
-
 @view_config(route_name='batch_download', request_method='GET')
 def batch_download(context, request):
     # adding extra params to get required columns
@@ -305,32 +304,30 @@ def batch_download(context, request):
         search_params=request.matchdict['search_params']
     )
     files = [metadata_link]
-    if 'files.file_type' in param_list:
-        for exp in results['@graph']:
-            for f in exp.get('files', []):
-                if f['file_type'] in param_list['files.file_type']:
-                    if 'restricted' in f and f['restricted'] == 'true':
-                        continue
-                    else:
-                        files.append('{host_url}{href}'.format(
-                            host_url=request.host_url,
-                            href=f['href']
-                        ))
-    else:
-        for exp in results['@graph']:
-            for f in exp.get('files', []):
-                if 'restricted' in f and f['restricted'] == 'true':
-                    continue
-                else:
-                    files.append('{host_url}{href}'.format(
-                        host_url=request.host_url,
-                        href=f['href']
-                    ))
+    exp_files_gen = filter_restricted_files(results)
+    for exp_file in exp_files_gen:
+        if exp_file['file_type'] in param_list.get('files.file_type', []):
+            files.append(
+                '{host_url}{href}'.format(
+                    host_url=request.host_url,
+                    href=exp_file['href'],
+                )
+            )
     return Response(
         content_type='text/plain',
         body='\n'.join(files),
         content_disposition='attachment; filename="%s"' % 'files.txt'
     )
+
+
+def filter_restricted_files(results):
+    files_gen = (
+        exp_file
+        for exp in results['@graph']
+        for exp_file in exp.get('files', [])
+        if not exp_file.get('restricted') == True
+    )
+    return files_gen
 
 
 def lookup_column_value(value, path):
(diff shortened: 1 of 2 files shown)


hitz:
FYI each demo machine should also have a URL that is named by it's branch.demo.encodedcc.org

src/encoded/batch_download.py
```diff
@@ -309,17 +309,23 @@ def batch_download(context, request):
...
+                        files.append('{host_url}{href}'.format(
+                            host_url=request.host_url,
+                            href=f['href']
+                        ))
+    else:
+        for exp in results['@graph']:
+            for f in exp.get('files', []):
+                if 'restricted' in f and f['restricted'] == 'true':
+                    continue
+                else:
```
caseylitton: We could replace the entire if/else block with a generator expression that gets all the experiment files and filters them based on the restriction.  Then loop through the generator, do the file_type check, and create the url.
```
    exp_files_gen = (
        exp_file
        for exp in results['@graph']
        for exp_file in exp.get('files', [])
        if not exp_file.get('restricted') == 'true'
    )
    files = []
    for exp_file in exp_files_gen:
        if exp_file['file_type'] in param_list.get('files.file_type', []):
            files.append(
                '{host_url}{href}'.format(
                    host_url=request.host_url,
                    href=exp_file['href'],
                )
            )
```
Or generate the whole thing at once
```
    files = [
        '{host_url}{href}'.format(
            host_url=request.host_url,
            href=exp_file['href'],
        )
        for exp in results['@graph']
        for exp_file in exp.get('files', [])
        if (
            not exp_file.get('restricted') == 'true' and
            exp_file['file_type'] in param_list.get('files.file_type', [])
        )
    ]
files.append
    ('{host_url}/metadata/{search_params}/metadata.tsv'.format(
        host_url=request.host_url,
        search_params=request.matchdict['search_params']
    )
)
```
I think this second one is too much.  It may be hard to update.    The first one breaks it up into meaning parts.  First we do global filters that effect all files, like 'restriction' but we could add more in the future.  Then in the generator loop we can add more specific filters like the files_type.

Also, using .get on dicts can help make conditionals.

hitz:
I just wanted to chime in and say that the code in this "module" is wretched and I agree that it should be refactored (even heavily) in cases like these.  But I would suggest writing/modifying the unittests in test_batch_download.py.

kakaly:
Updated the review with unit tests

kakaly:
Demo: https://encd-3823-batchdownload-restrcited-e502543c9-karthik.demo.encodedcc.org/

kakaly:
ENCFF136ZBD -> this accession has 'restricted' set to true. I QAed it manually. When you download Roadmap files from production and look for 'ENCFF136ZBD' in files.txt, you will be able to find it. The files.txt from my demo URL will not have it.

src/encoded/batch_download.py
```diff
@@ -305,28 +304,40 @@ def batch_download(context, request):
...
     )
     files = [metadata_link]
-    if 'files.file_type' in param_list:
-        for exp in results['@graph']:
-            for f in exp.get('files', []):
-                if f['file_type'] in param_list['files.file_type']:
-                    files.append('{host_url}{href}'.format(
+    exp_files_gen = filter_restricted_files(results)
+    for exp_file in exp_files_gen:
+        if 'files.file_type' in param_list:
```
caseylitton: This files type check is a filter too, it's just conditional.  Add this to a function also.  Then, in the end we're just building the urls from a generator.  Splitting this up into functions will make testing easier.

caseylitton: I was playing around with this and I think we're going about it wrong.  We should make the exp file generator and then run it through our conditions.  This way it's easy to add more filters.
```
    type_check = None
    if 'files.file_type' in param_list:
        type_check = param_list.get('files.file_type', [])
    exp_files_gen = (
        exp_file
        for exp in results['@graph']
        for exp_file in exp['files']
    )
    files = []
    for exp_file in exp_files_gen:
        if exp_file.get('restricted'):
            continue
        elif type_check is not None and exp_file['file_type'] not in type_check:
            continue
        files.append(
            '{host_url}{href}'.format(
                host_url=request.host_url,
                href=exp_file['href'],
            )
        )
    return Response(
        body='\n'.join(files),
        content_disposition='attachment; filename="%s"' % 'files.txt',
        content_type='text/plain',
    )
```

kakaly:
Closing this one. The comments have been addressed in Reference #2176

PR #2132, opened 2018-02-12
ENCD-3341 Add new output_types to file json

thehushkat:
Added output_types to accommodate new files coming it from the Graveley Lab.


estherc:
Changes to file.json and file.md look fine to me.

PR #2133, opened 2018-02-13
ENCD-3835 Fix inconsistent platforms audit HiSeq2K 2.5K

gabdank:
replacing the use of platform names by platform IDs in the audit code
it resolves previously erroneous detection of HiSeq2K and 2.5K as inconsistent platforms


estherc:
Logic looks right and ontology term IDs match the term names they are supposed to replace.

PR #2144, opened 2018-02-22
Encd 3602 region indexer v2

tdreszer:
This PR depends upon snovault changes that are currently in PR "SNO-09-10 Indexer updates and reverts".   When that has been merged into snovault and a tag made, then this PR will need buildout.cfg changed to reference the new sno tag.


caseylitton:
Replaced with a rebased version: https://github.com/ENCODE-DCC/encoded/pull/2147

PR #2145, opened 2018-02-23
ENCD-3821 Show biosample characterization doc links

forresttanaka:
* Moved the <Documents> component out of antibody.js into doc.js (and renamed it <CharacterizationDocuments> so that biosample can now use the same component to display the detail panel on characterizations, as they now both display the same thing.
* Changed the <CharacterizationDetail> local variable name from `doc` to `characterization` because — since I’m now dealing with characterization documents (and characterizations are a form of document that happen to include documents), it was just getting confusing without the name change.


src/encoded/static/components/biosample.js
```diff
@@ -706,43 +706,56 @@ CharacterizationCaption.propTypes = {
...
-    const doc = props.doc;
+    const characterization = props.doc;
     const keyClass = `document__detail${props.detailOpen ? ' active' : ''}`;
-    const excerpt = doc.description && doc.description.length > EXCERPT_LENGTH;
+    const excerpt = characterization.caption && characterization.caption.length > EXCERPT_LENGTH;
+
+    // See if we need a list of documents or not. Documents without attachments don't get
+    // displayed.
+    const docs = characterization.documents && characterization.documents.length ?
+        characterization.documents.filter(doc => !!(doc.attachment && doc.attachment.href && doc.attachment.download))
```
davisagli: `!!expr` can be written as `Boolean(expr)` which I find more readable.

Here's the relevant eslint rule if we want to check for this sort of thing: https://eslint.org/docs/rules/no-implicit-coercion

src/encoded/static/components/doc.js
```diff
@@ -447,3 +447,34 @@ globals.documentViews.caption.register(QCAttachmentCaption, 'QualityMetric');
 
 // Register document preview rendering components
 globals.documentViews.preview.register(QCAttachmentPreview, 'QualityMetric');
+
+
+// Display a list of document links within a characterization.
+export const CharacterizationDocuments = (props) => {
+    const docs = props.docs.filter(doc => !!doc);
```
davisagli: ditto

src/encoded/static/components/doc.js
```diff
@@ -447,3 +447,34 @@ globals.documentViews.caption.register(QCAttachmentCaption, 'QualityMetric');
...
+        <dd>
+            {docs.map((doc) => {
+                if (doc && doc.attachment) {
+                    const attachmentHref = url.resolve(doc['@id'], doc.attachment.href);
+                    const docName = (doc.aliases && doc.aliases.length) ? doc.aliases[0] :
+                        ((doc.attachment && doc.attachment.download) ? doc.attachment.download : '');
+                    return (
+                        <div className="multi-dd dl-link" key={doc['@id']}>
+                            <i className="icon icon-download" />&nbsp;
+                            <a data-bypass="true" href={attachmentHref} download={doc.attachment.download}>
```
davisagli: I know you just moved this code so this is an unrelated side note, but is there a reason we don't always skip js-based navigation for links with a `download` attribute, so that we could avoid also using `data-bypass`?

forresttanaka: Do you mean modifying the navigate method to make an exception for download links?

davisagli: I'm suggesting that be created as a new ticket. Similar to ENCD-1813 which did so based on `@@download` in the URL.

PR #2151, opened 2018-02-26
ENCD-3879 Pin Alembic version

keenangraham:


versions.cfg
```diff
@@ -256,3 +256,8 @@ elasticsearch = 5.4.0
 
 # Added by buildout at 2018-01-17 17:26:37.230945
 selenium = 3.8.1
+
+# Required by:
+# encoded==66.0
+alembic = 0.9.8
+python-editor = 1.0.3
```
caseylitton: What is the python-editor for?

davisagli: It's a dependency of `alembic`. It looks like a small utility to help with opening the user's preferred text editor from Python.

forresttanaka:
As far as usage, I did a clean buildout in this branch and had no modified files afterwards, so it looks good.

versions.cfg
keenangraham: Dependency of Alembic.

PR #2154, opened 2018-02-28
ENCD-3841 Add collapsing sections on schema pages

forresttanaka:
Most of the change involves the new `<DisplayObjectSection>` component that replaces a loop inside `<DisplayObject>`. Now that each displayed section needs to keep its own state for being open or closed, it needed its own component to track that, and react to clicks in the new button.

I also added aria properties so that people using screen readers can better tell what they’re doing when opening and closing these sections. I should have added those to the open/closed state of the individual properties originally too.

### DIFF
--- a/src/encoded/static/components/schema.js
+++ b/src/encoded/static/components/schema.js
@@ -33,6 +33,9 @@ const excludedTerms = [
     'boost_values',
     'changelog',
     '@type',
+    'output_type_output_category',
+    'file_format_file_extension',
+    'sort_by',
 ];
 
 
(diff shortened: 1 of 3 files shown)


src/encoded/static/components/schema.js
```diff
@@ -33,6 +33,9 @@ const excludedTerms = [
     'boost_values',
     'changelog',
     '@type',
+    'output_type_output_category',
+    'file_format_file_extension',
```
davisagli: Why are these last two excluded?

forresttanaka: Because those are just static mappings not configured by submitters, and as they’re both root level in the schema, they caused the schema pages to be huge when I first put in these schema pages. With this ticket they would show up collapsed so they could appear without impacting the page too much, but I’m still not clear they offer any useful information to submitters though.

PR #2168, opened 2018-03-07
ENCD-3840 Add link to linkFrom linkTo schema properties

forresttanaka:
This PR relies on SNO-26-json-extension.

The most key new function is `collectLinks` which digs into schema properties looking for linkTo/linkFrom properties at any embedded levels and making arrays of all those properties it could find of each kind. These then get displayed as links.

The other part (relying on SNO-26) involves doing a GET request on the new /profiles-map/ object through the <FetchedData>/<Param> components, which then gets used to convert whatever properties `collectLinks` found to the corresponding schema paths.

### DIFF
--- a/src/encoded/static/components/schema.js
+++ b/src/encoded/static/components/schema.js
@@ -134,8 +134,8 @@ function collectLinks(prop) {
             // Simple linkTo or linkFrom property; adds its value to the array we're collecting.
             linksLists[key].push(key === 'linkFrom' ? prop[key].split('.')[0] : prop[key]);
         } else if (typeof prop[key] === 'object' && !Array.isArray(prop[key])) {
-            // The schema property key is not one of `nonlinkProps` and has a value that's an object.
-            // Look for linkTo properties within this object.
+            // The schema property is an object. Look for linkTo/LinkFrom properties within this
+            // object.
             const subLinksLists = collectLinks(prop[key]);
             linksLists.linkTo = linksLists.linkTo.concat(subLinksLists.linkTo);
             linksLists.linkFrom = linksLists.linkFrom.concat(subLinksLists.linkFrom);
--- a/src/encoded/static/components/schema.js
+++ b/src/encoded/static/components/schema.js
@@ -36,6 +36,7 @@ const excludedTerms = [
     'output_type_output_category',
     'file_format_file_extension',
     'sort_by',
+    'anyOf',
 ];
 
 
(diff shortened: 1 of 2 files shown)


src/encoded/static/components/schema.js
```diff
@@ -89,32 +94,143 @@ function createJsonDisplay(elementId, term) {
...
 }
 
 
+/**
+ * Convert the schema id property value to a page URI.
+ *
+ * @param {string} schemaId - value of id property within a schema
+ */
+function schemaIdToPage(schemaId) {
+    return schemaId.slice(0, -5);
```
davisagli: What's getting sliced off here?

src/encoded/static/components/schema.js
```diff
@@ -89,32 +94,143 @@ function createJsonDisplay(elementId, term) {
...
+function collectLinks(prop) {
+    const propKeys = Object.keys(prop);
+    const linksLists = { linkTo: [], linkFrom: [] };
+
+    propKeys.forEach((key) => {
+        if (key === 'linkTo' || key === 'linkFrom') {
+            // Simple linkTo or linkFrom property; adds its value to the array we're collecting.
+            linksLists[key].push(key === 'linkFrom' ? prop[key].split('.')[0] : prop[key]);
+        } else if (typeof prop[key] === 'object' && !Array.isArray(prop[key])) {
+            // The schema property key is not one of `nonlinkProps` and has a value that's an object.
```
davisagli: I don't see `nonLinkProps` anywhere except in this comment.

src/encoded/static/components/schema.js
forresttanaka: The .json extension.

src/encoded/static/components/schema.js
forresttanaka: Yeah, nonLinkProps used to be there but I changed it.

forresttanaka:
I looked into whether we could insert links into a brace editor, but there seems to be no way. I asked early on if having the links below was acceptable or if we should do something with the JSON instead, and didn’t hear much, and this seemed acceptable to those who tried it. This should probably be a @hitz and/or @jahilton question.

forresttanaka:
Also, in very early talks (I think at the Sunday meeting before the consortium meeting) when Noam suggested better schema pages instead of just the raw JSON we had, there was talk about maybe pop-up sub schemas or maybe graphs of schemas — something we can approach once people have more solid ideas about the best way to go moving forward.

PR #2174, opened 2018-03-12
ENCD-3915 Remove chromedriver lock from Travis

forresttanaka:
I assigned this ticket to a sprint and release, but I don’t really know how those should be set for this.


caseylitton:
@forresttanaka Does this fix the current testing issues without causing more?  We can get this in dev and add it to the release.

forresttanaka:
Doesn’t cause more problems as far as I know. It reverts .travis.yml to the way it was before ENCD-3869, as that change was temporary until Saucelabs got back in synch on their chromedriver and Chrome versions. The Travis tests for this branch have passed.

forresttanaka:
The [BDD test](https://travis-ci.org/ENCODE-DCC/encoded/jobs/352573392) with this change looks clean to me.

PR #2176, opened 2018-03-14
ENCD [3823, 3864, 3868] - filter restricted files, batch_download refactor, add unit tests

kakaly:
This PR addresses the comments from the previous PR for the same tickets:

ENCD 3823 - Filter out restricted files
ENCD 3864 - Refactor batch_download methods
ENCD 3868 - Add unit tests for batch_download helpers(partially implemented)

Demo: https://encd-3823-batchdownload-restrcited-016edb3f0-karthik.demo.encodedcc.org

Reference #2127

### DIFF
--- a/src/encoded/tests/test_batch_download.py
+++ b/src/encoded/tests/test_batch_download.py
@@ -1,8 +1,55 @@
 # Use workbook fixture from BDD tests (including elasticsearch)
-from .features.conftest import app_settings, app, workbook
+import json
+import pytest
+
+from encoded.tests.features.conftest import app
+from encoded.tests.features.conftest import app_settings
+from encoded.tests.features.conftest import workbook
+from encoded.batch_download import lookup_column_value
+from encoded.batch_download import restricted_files_present
+from encoded.batch_download import file_type_param_list
 
 
-def test_report_download(testapp, workbook):
+param_list_1 = {'files.file_type': 'fastq'}
+param_list_2 = {'files.title': 'ENCFF222JUK'}
+exp_file_1 = {'file_type': 'fastq',
+              'restricted': True}
+exp_file_2 = {'file_type': 'bam',
+              'restricted': False}
+exp_file_3 = {'file_type': 'gz'}
+
+
+@pytest.fixture
+def lookup_column_value_item():
+    item = {
+        'assay_term_name': 'ISO-seq',
+        'lab': {'title': 'John Stamatoyannopoulos, UW'},
+        'accession': 'ENCSR751ISO',
+        'assay_title': 'ISO-seq',
+        'award': {'project': 'Roadmap'},
+        'status': 'released',
+        '@id': '/experiments/ENCSR751ISO/',
+        '@type': ['Experiment', 'Dataset', 'Item'],
+        'biosample_term_name': 'midbrain'
+    }
+    return item
+
+
+@pytest.fixture
+def lookup_column_value_validate():
+    valid = {
+        'assay_term_name': 'ISO-seq',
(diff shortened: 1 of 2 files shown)


src/encoded/tests/test_batch_download.py
```diff
@@ -27,3 +63,30 @@ def test_report_download(testapp, workbook):
     #     b'', b'', b'', b'', b'', b'', b''
     # ]
     assert len(lines) == 47
+
+
+def test_batch_download_files_txt(testapp, workbook):
```
caseylitton: We can do a little more checking here without too much effort.  All results will have more than zero lines since the metadata_link is appended to everything.   

-assert that the first link is the metadata link.  We know that string exactly.
-assert there are more urls than just metadata link
-assert that the file urls are correct in someway...

This last one, you could do a regex to make sure all the urls fit our pattern, host_url files ENCFF... @@download ENCFF.ext.   Kinda overkill but certainly possible.  We should at least verify ONE file string format, like it starts with the correct host_url, and it contains 'files' and '@@download'.

src/encoded/tests/test_batch_download.py
```diff
@@ -27,3 +63,30 @@ def test_report_download(testapp, workbook):
...
+    results = testapp.get('/batch_download/type%3DExperiment')
+    assert results.headers['Content-Type'] == 'text/plain; charset=UTF-8'
+    assert results.headers['Content-Disposition'] == 'attachment; filename="files.txt"'
+    lines = results.body.splitlines()
+    assert len(lines) > 0
+
+
+def test_batch_download_restricted_files_present(testapp, workbook):
```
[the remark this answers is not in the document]
kakaly: Sure. Added tests for them.

src/encoded/tests/test_batch_download.py
kakaly: Thanks for this. I have addressed these validations in the code

davisagli:
Looks like this broke the build: https://travis-ci.org/ENCODE-DCC/encoded/jobs/359404421

```
src/encoded/tests/test_batch_download.py::test_batch_download_lookup_column_value FAILED
___________________ test_batch_download_lookup_column_value ____________________
lookup_column_value_item = {'@id': '/experiments/ENCSR751ISO/', '@type': ['Experiment', 'Dataset', 'Item'], 'accession': 'ENCSR751ISO', 'assay_term_name': 'ISO-seq', ...}
lookup_column_value_validate = {'@id': '/experiments/ENCSR751ISO/', '@type': ['Experiment', 'Dataset', 'Item'], 'assay_term_name': 'ISO-seq', 'audit': '', ...}
    def test_batch_download_lookup_column_value(lookup_column_value_item, lookup_column_value_validate):
        for path in lookup_column_value_validate.keys():
>           assert lookup_column_value_validate[path] == lookup_column_value(lookup_column_value_item, path)
E           assert ['Experiment', 'Dataset', 'Item'] == 'Experiment,Dataset,Item'
E            +  where 'Experiment,Dataset,Item' = lookup_column_value({'@id': '/experiments/ENCSR751ISO/', '@type': ['Experiment', 'Dataset', 'Item'], 'accession': 'ENCSR751ISO', 'assay_term_name': 'ISO-seq', ...}, '@type')
src/encoded/tests/test_batch_download.py:117: AssertionError
```

kakaly:
I ll take a look

kakaly:
Okay. lookup_column_value() returns a string. The validation data returned by the fixture for validating has a list should be a concatenated string of the objects contained in the list. But, the fixture returns a list for validation. I fixed it locally. Will open a ticket and submit a PR. Thanks @davisagli for finding this.

PR #2179, opened 2018-03-16
ENCD-3916 Fix for broken metadata queries

kakaly:
- Issue: metadata_tsv route fails when experiment_json object doesn't have the 'files' key

- Fix: Default 'files' to an empty list when there is no 'files' key

### DIFF
--- a/src/encoded/batch_download.py
+++ b/src/encoded/batch_download.py
@@ -241,7 +241,7 @@ def metadata_tsv(context, request):
     results = request.embed(path, as_user=True)
     rows = []
     for experiment_json in results['@graph']:
-        if experiment_json['files']:
+        if experiment_json.get('files', []):
             exp_data_row = []
             for column in header:
                 if not _tsv_mapping[column][0].startswith('files'):


kakaly:
Demo: https://encd-3916-broken-metadata-queries-153e03eb1-karthik.demo.encodedcc.org

src/encoded/batch_download.py
```diff
@@ -241,7 +241,7 @@ def metadata_tsv(context, request):
     results = request.embed(path, as_user=True)
     rows = []
     for experiment_json in results['@graph']:
-        if experiment_json['files']:
+        if experiment_json.get('files', []):
```
caseylitton: It's not broken, but we don't need the default empty list.  
```
if experiment_json.get('files')
```
 will be false if for many situations.  https://docs.python.org/2.4/lib/truth.html .  We probably want this condition to fail in all of those scenarios.  experiment_json.get('files', []) and experiment_json.get('files') give us the same result.  'if []' is False and, since None is the default of get, we end up with "if None" in the second case.

kakaly: Cool. Will refactor this one with my other tickets that have changes on this same method

PR #2180, opened 2018-03-16
ENCD-3933 Allow ES address to be set on deploy

caseylitton:
2018 03 22:  Booted up fine, waiting on initial indexing

1. bin/deploy -b kcl/encd-3933-template-es-address -n es-backend3-data --cluster-name es-backend3-cluster --instance-type m4.xlarge --elasticsearch yes --cluster-size 5
2. Waited for cluster/master to come up
3. bin/deploy -b kcl/encd-3933-template-es-address -n es-frontend2 --instance-type c4.8xlarge -i 172.31.18.178 -p 9201 -x

Notes:
1. builds 5 cluster machines and master cluster node.
3. builds the "frontend"


caseylitton:
@hitz Yes, I've been thinking about that since the first time I saw all the cloud-init yamls.  I like ansible because I have experience with it, but we should look into other automation tools before making a decision.

davisagli:
This seems okay. But I wonder if we should be looking up settings like this from environment variables rather than from the ini files so that we don't need to generate ini files.

FWIW, I like ansible too (but haven't yet used it in the context of programmatically deployed EC2 instances).

caseylitton:
replaced with combine PR https://github.com/ENCODE-DCC/encoded/pull/2228

PR #2181, opened 2018-03-16
ENCD-3886 Update deploy machine instance types

hitz:
Changed default ami in deploy scripts to use c5 and m5 series.  Way faster (and cheaper!)

### DIFF
--- a/checkfiles/deploy.py
+++ b/checkfiles/deploy.py
@@ -109,7 +109,7 @@ def main():
     parser.add_argument(
         '-n', '--name', type=hostname, help="Instance name")
     parser.add_argument(
-        '--image-id', default='ami-4b37d42b',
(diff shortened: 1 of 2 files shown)


caseylitton:
If we have a limited number of these machines then we shouldn't make them the default.  We should reserve them for admin testing and such things.

stuartmiyasato:
Just my $0.02, but I think that's the wrong approach.  Since I have not personally seen a case where Amazon has denied a limit increase request, I think we should estimate the maximum number of instances we need and make a new request.  Amazon may not grant the entire request, but they will normally bump up the value by some amount.  The fact that we get more resources at a cheaper price makes me think we should phase out the c4/m4 series.  (And in fact, that would be reasonable justification to Amazon for requesting a limit increase for the c5/m5 series.)

PR #2184, opened 2018-03-20
ENCD-3451 New summary page

forresttanaka:
* Removed the project and biosample chart colors out of the App context and into globals because they’re static and made more sense in globals than carried around in React, especially now that we have replicate chart colors in addition to the existing ones
* Tried to use awards.js charting functions and components the summer interns wrote as much as possible, which meant modifying them to fit a more general case. These charting functions should really be moved to a new file since they’re no longer specific to awards, but we’ll worry about that in another ticket as the charting code needs a lot of refactoring to reduce duplication.
* I included some embryonic code in statuslabels.js so that any front-end code can get consistent statuses for each kind of object and at different login levels, only defining the statuses for Experiment for now, as the only client is the summary page. After I wrote this, I found out Keenan was reworking statuses, so I think this has promise, but I’ll need to see what comes out of his efforts before really continuing with this in ENCD-3629.
* One back-end change to handle the /summary/ route. Only ?type=Experiment gives results. Most of this is just a simpler variant of the matrix code, as it doesn’t retrieve the `hits` data — only facet data.

### DIFF
--- a/src/encoded/search.py
+++ b/src/encoded/search.py
@@ -26,6 +26,7 @@ def includeme(config):
     config.add_route('matrix', '/matrix{slash:/?}')
     config.add_route('news', '/news/')
     config.add_route('audit', '/audit/')
+    config.add_route('summary', '/summary{slash:/?}')
     config.scan(__name__)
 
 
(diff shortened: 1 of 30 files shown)


forresttanaka:
Going to close this and PR later; this takes a lot of experimenting with real data, so the pushes are really stacking up. I’ll squash before the later PR.

PR #2185, opened 2018-03-21
ENCD-3918 | ENCD-3927 | ENCD-3929 move helper methods, fix pylint issues, add tests

kakaly:
**ENCD 3918 Refactor search at a high level**

- Adds helpers/helper.py module

- Removes helper methods from search.py

- Resolves dependencies

- Fixes lint errors


kakaly:
Depends on https://github.com/ENCODE-DCC/snovault/pull/91

src/encoded/batch_download.py
```diff
@@ -9,8 +9,8 @@
     parse_qs,
     urlencode,
 )
-from .search import iter_search_results
-from .search import list_visible_columns_for_schemas
+from encoded.search import iter_search_results
+from snowflakes.helpers.helper import list_visible_columns_for_schemas
```
hitz: I don't like the dependency on snowflakes literally.   Maybe we should move them into snovault/helpers?

kakaly: Sure. I can do that.

kakaly:
I moved helpers from snowflakes to snovault

caseylitton:
@kakaly I only looked at the pylint commit.  I mainly see line length fixes.

Don't use slashes to fix long lines.  There are many ways to fix lines that are too long and it depends on the code.  We'll go through them as they come up.  There isn't one right one but the slashes are old style.

I get 64 after I set my line length to 120.    There are a handful that we probably don't want to fix here, but many look fixable.  Let me know if these are not showing up.

caseylitton:
Also, can you edit the Pr to show that there are multiple encd tickets

kakaly:
@caseylitton - I changed the name of the PR. Interesting, I had fixed all the line length issues after running the linter. I ll check again. Thanks for letting me know.

kakaly:
@caseylitton - I ran pylint on src/encoded/search.py. I do not see any issues for line lengths. Not sure why.

caseylitton:
@kakaly Sorry, maybe that wasn't clear.  

I only see line length fixes in the pylint commit, but there are many more things to fix.  When I run pylint, I see 60+  issues that are not line length issues.  We can't fix them all but we'll want to fix some of them. 

Additionally, revert the line length edits you made.  Only fix the ones that are over 120 chars and don't use slashes to do it.

hitz:
Hey @kakaly is this going to get fixed soon?

kakaly:
@hitz - Yes. I will update this PR along with the tests I wrote for the helpers today.

src/encoded/tests/test_search.py
```diff
@@ -753,3 +888,115 @@ def test_format_facets_adds_pseudo_facet_for_extra_filters():
...
         'total': 42,
     }]
+
+
+def test_sort_query(sort_query_item):
+
+    sorted_query = sort_query(sort_query_item)
+    assert isinstance(sorted_query, OrderedDict) is True
+    for key in sort_query_item:
+        assert sort_query_item[key] == sorted_query[key]
```
caseylitton: Did we update this test or is it a copy/paste of an original?

This doesn't test if the keys are now sorted or if the sub dicts are ordered dicts.  The sort_query function does two things.    One it rebuilds the input dict with sorted keys.  Second it calls itself for all sub dicts.  So we should test for these two things.

Also don't use "real" data like sort_query_item() for tests that do not require real data.  It makes everything way more complex than it needs to be.  sort_query doesn't care what kind of dict it gets.

For example, 
```
test_dict = {
    '1': 1,
    '3': 3,
    '2': 2,
}
....
assert list(result_dict.keys()) == sorted(list(test_dict.keys()))
```  
Of course this is a partial example because it doesn't test the recursive property of the function.  The point is to make tests as simple as possible.  I wish there were better examples to follow in our code base but that's one reason we're updating them. 

 I didn't look into the other tests, but try updating them accordingly.  Some of them look data dependent but others don't.

src/encoded/tests/test_search.py
```diff
@@ -753,3 +888,115 @@ def test_format_facets_adds_pseudo_facet_for_extra_filters():
         ],
         'total': 42,
     }]
+
+
+def test_sort_query(sort_query_item):
+
```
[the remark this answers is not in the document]
kakaly: Done. Actually the existing tests had them, so I added it to the new ones. Removed it from everywhere now.

src/encoded/tests/test_search.py
```diff
@@ -753,3 +888,115 @@ def test_format_facets_adds_pseudo_facet_for_extra_filters():
...
+    args = build_aggregation_item[0]
+    validate = build_aggregation_item[1]
+    for index, arg in enumerate(args):
+        actual = validate[index]
+        agg_name, agg = build_aggregation(arg[0], arg[1])
+        assert agg == actual
+
+
+def test_normalize_query(normalize_query_item, registry):
+
```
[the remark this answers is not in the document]
kakaly: Done

src/encoded/tests/test_search.py
```diff
@@ -1,11 +1,172 @@
 # Use workbook fixture from BDD tests (including elasticsearch)
-from .features.conftest import app_settings, app, workbook
-from webob.multidict import MultiDict
 import pytest
+from collections import OrderedDict
+from webob.multidict import MultiDict
+from snovault import TYPES
+from encoded.tests.features.conftest import app_settings, app, workbook
+from snovault.helpers.helper import (
```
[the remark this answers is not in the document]
kakaly: Like we discussed, I am keeping these with encoded for now.

src/encoded/tests/test_search.py
```diff
@@ -753,3 +888,115 @@ def test_format_facets_adds_pseudo_facet_for_extra_filters():
         ],
         'total': 42,
     }]
+
+
+def test_sort_query(sort_query_item):
+
+    sorted_query = sort_query(sort_query_item)
+    assert isinstance(sorted_query, OrderedDict) is True
```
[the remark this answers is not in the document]
kakaly: Done

src/encoded/tests/test_search.py
kakaly: Sure. I updated this test with simple inputs and I am validating the output of the function by hardcoding valid OrderedDicts to test the sorting and recursive property of this function. I looked at other tests too. Those are having hardcoded conditions which cannot be tested without partial real data. For example, 

build_terms_filter() has 
if field.endswith('!'):
        field = field[:-1]
        if not field.startswith('audit'):

So, in order to get into these if blocks, I have a partial real data (like 'audit!'). Let me know if this sounds ok.

src/encoded/tests/test_search.py
```diff
@@ -1,12 +1,170 @@
...
+    return item
+
+
+@pytest.fixture
+def get_pagination_item():
+    requests = {FakeRequest((('from', '0'), ('limit', '25'))): ('0', 25),
+                FakeRequest((('from', '10'), ('limit', '50'))): ('10', 50),
+                FakeRequest((('from', '0'),)): ('0', 25),
+                FakeRequest((('limit', '50'),)): (0, 50),
+                FakeRequest((('key', 'value'),)): (0, 25)}
```
caseylitton: A possible output from get_pagination is 'size = None' when size is 'all' or ''.  That is not being covered in these tests.

src/encoded/tests/test_search.py
```diff
@@ -1,12 +1,170 @@
...
+    return requests
+
+
+@pytest.fixture
+def prepare_search_term_item():
+    requests = {FakeRequest((('type', 'Page'),
```
caseylitton: This first test doesn't need all these extra parameters.  It just needs to not contain 'searchTerm'.   So no args are just as good as these four.  Also why not group the ones that return early with '*' together.  May want to add a third one where the FakeRequest has searchTerm = '*'.


Also, without good reason let's not return a dictionary with key values pairs when we can be consistent and return lists of sets with the test object and test results.

src/encoded/tests/test_search.py
```diff
@@ -1,12 +1,170 @@
...
+                FakeRequest((('searchTerm', '@type:field'),)): 'embedded.@type:field',
+                FakeRequest((('searchTerm', 'type:s'),)): 'type\:s',
+                FakeRequest((('searchTerm', ' '),)): '*',
+                FakeRequest((('searchTerm', '@type'),)): '@type'}
+    return requests
+
+
+@pytest.fixture
+def set_sort_order_item():
+    requests = {FakeRequest((('type', 'Experiment'), ('sort', 'assay_title'))): ['*', True],
```
caseylitton: Again, no reason to use the dictionary keys...but why did the value structure change to a lists in this case?  We used sets in the other fixtures, why change the pattern?

In this case is looks like you'll want a third item in the test for clarity.  
```
return [
    (
        FakeRequest, searchTerm, boolean_expected_result
    ),
]
```

src/encoded/tests/test_search.py
```diff
@@ -753,3 +882,115 @@ def test_format_facets_adds_pseudo_facet_for_extra_filters():
...
+                             result)
+        assert res == validate
+
+
+def test_get_search_fields(registry):
+    # pylint: disable=redefined-outer-name, unused-argument
+    request = FakeRequest((('key', 'value'),))
+    request.registry = registry
+    doc_types = [['Experiment']]
+    for doc_type in doc_types:
```
caseylitton: We don't need a for loop.  We only have the one doc_types = ['Experiment'] .

src/encoded/tests/test_search.py
```diff
@@ -1,12 +1,170 @@
...
+             OrderedDict([('d', OrderedDict([('a', 'val'),
+                                             ('b', 'val')]))]))]
+    return item
+
+
+@pytest.fixture
+def get_pagination_item():
+    requests = {FakeRequest((('from', '0'), ('limit', '25'))): ('0', 25),
+                FakeRequest((('from', '10'), ('limit', '50'))): ('10', 50),
+                FakeRequest((('from', '0'),)): ('0', 25),
```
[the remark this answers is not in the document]
kakaly: For whatever reason, if I have just one pair of values, I need to have these in-trailing commas. Otherwise, it throws an error like this:

self = <[ValueError("too many values to unpack (expected 2)") raised in repr()] MultiDict object at 0x10b2070f0>, key = 'from'

    def __getitem__(self, key):
>       for k, v in reversed(self._items):
E       ValueError: need more than 1 value to unpack

eggs/WebOb-1.6.0-py3.4.egg/webob/multidict.py:96: ValueError

It seems like its a property of MultiDict. Not sure.

kakaly:
@hitz - I have fixed the tests and updated the PR. It is now passing locally for me. And, I have created a demo for this change. 
https://encd-3918-move-helper-methods-40eebb11f-karthik.demo.encodedcc.org/

zoldello:
This changes in this (2185) is subsumed by 2319- https://github.com/ENCODE-DCC/encoded/pull/2319/files. Therefore, this 2185 is closed in favor of 2319.

PR #2187, opened 2018-03-22
ENCD-3898 adding an audit flagging "tagging" GM without characterization

gabdank:
adding an audot that flags tagging genetic modifications without characterizations, and propagates the audit to the biosample and to the experiment level


src/encoded/audit/genetic_modification.py
```diff
@@ -0,0 +1,31 @@
+from snovault import (
+    AuditFailure,
+    audit_checker,
+)
+
+# flag genetic modifications with purpose "tagging" that lack genetic modification characterization.
+def audit_tagging_modification(value, system):
+    if value['purpose'] == 'tagging' and\
```
caseylitton: Use get for dicts unless we've already determined the key exists.  Also put multi line conditionals in () instead of using the /

src/encoded/audit/genetic_modification.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+function_dispatcher = {
+    'audit_tag': audit_tagging_modification,
+}
+
+@audit_checker('GeneticModification',
+               frame=['characterizations'])
+def audit_modification(value, system):
+    for function_name in function_dispatcher.keys():
```
caseylitton: We don't actually use the key so use values() instead of keys(). 

The third choice it items() that returns key and value

src/encoded/audit/genetic_modification.py
```diff
@@ -0,0 +1,31 @@
...
+        for failure in function_dispatcher[function_name](value, system):
```
caseylitton: Also, could you try this?  'yield from' should produce the same result without the 'for failure in ...etc'
```
for audit_function in function_dispatcher.values():
    yield from audit_function(value, system)
```

src/encoded/audit/genetic_modification.py
gabdank: purpose is a required property in the schema
but i have updated

src/encoded/audit/genetic_modification.py
gabdank: fixed

src/encoded/audit/genetic_modification.py
gabdank: done

PR #2188, opened 2018-03-22
ENCD-3739 rewriting chip control read depth audit

gabdank:
Re-writing the logic of chip-seq control read depth audit

### DIFF
--- a/src/encoded/audit/experiment.py
+++ b/src/encoded/audit/experiment.py
@@ -57,162 +57,180 @@ def audit_experiment_chipseq_control_read_depth(value, system, files_structure):
         target_name = value['target']['name']
         target_investigated_as = value['target']['investigated_as']
         if target_name not in ['Control-human', 'Control-mouse']:
-            for alignment_file in files_structure.get('alignments').values():
-                # initially was for file award
-                if not alignment_file.get('award') or \
-                    alignment_file.get('award')['rfa'] not in [
-                        'ENCODE3',
-                        'ENCODE4',
-                        'ENCODE2-Mouse',
-                        'ENCODE2',
-                        'ENCODE',
-                        'Roadmap']:
-                    continue
-                if alignment_file.get('lab') not in ['/labs/encode-processing-pipeline/']:
-                    continue
-                derived_from_files = list(
-                    get_derived_from_files_set([alignment_file], files_structure, 'fastq', True))
-                if not derived_from_files:
-                    continue
-                control_bam = get_control_bam(
-                    alignment_file,
-                    'ChIP-seq read mapping',
-                    derived_from_files,
-                    files_structure)
-                if control_bam is not False:
-                    control_depth = get_chip_seq_bam_read_depth(control_bam)
-                    control_target = get_target_name(derived_from_files)
-                    if control_depth is not False and control_target is not False:
-                        yield from check_control_read_depth_standards(
-                            control_bam,
-                            control_depth,
-                            control_target,
-                            True,
-                            target_name,
-                            target_investigated_as)
-    return
+            controls = value.get('possible_controls')
+            if controls:
+                controls_files_structures = {}
+                control_objects = {}
+                for control_experiment in controls:
+                    control_objects[control_experiment.get('@id')] = control_experiment
+                    controls_files_structures[control_experiment.get('@id')] = create_files_mapping(
+                        control_experiment.get('original_files'),
+                        files_structure.get('excluded_types'))
+                for peaks_file in files_structure.get('peaks_files').values():
+                    if not peaks_file.get('award') or \
+                        peaks_file.get('award')['rfa'] not in [
+                                'ENCODE3',
+                                'ENCODE4',
+                                'ENCODE2-Mouse',
+                                'ENCODE2',
+                                'ENCODE',
+                                'Roadmap']:
+                        continue
+                    if peaks_file.get('lab') not in ['/labs/encode-processing-pipeline/']:
+                        continue
+                    bio_reps = peaks_file.get('biological_replicates')
+                    if not bio_reps or len(bio_reps) > 1:
+                        continue
+
+                    derived_from_files = list(get_derived_from_files_set([peaks_file],
+                                                                         files_structure,
+                                                                         'bam',
+                                                                         True))
+                    derived_from_external_bams = [
+                        derived_from for derived_from in
+                        derived_from_files
+                        if (derived_from.get('dataset') != value.get('@id')
+                            and
+                            derived_from.get('dataset') in controls_files_structures
+                            and
+                            check_pipeline('ChIP-seq read mapping',
+                                           derived_from.get('@id'),
+                                           controls_files_structures[derived_from.get('dataset')]))]
+                    control_bam_details = []
+                    cumulative_read_depth = 0
+                    for bam_file in derived_from_external_bams:
 
 
-def check_control_read_depth_standards(value,
+                        control_depth = get_chip_seq_bam_read_depth(bam_file)
+                        control_target = get_control_target_name(
+                            bam_file.get('dataset'),
+                            control_objects)
+                            
+                        if control_target not in ['Control-human', 'Control-mouse']:
+                            detail = (
+                                'Control {} file {} ' +
+                                'has a target {} that is neither ' +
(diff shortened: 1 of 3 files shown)


src/encoded/audit/experiment.py
```diff
@@ -57,162 +57,177 @@ def audit_experiment_chipseq_control_read_depth(value, system, files_structure):
...
+            controls = value.get('possible_controls')
+            if controls:
+                controls_files_structures = {}
+                control_objects = {}
+                for control_experiment in controls:
+                    control_objects[control_experiment.get('@id')] = control_experiment
+                    controls_files_structures[control_experiment.get('@id')] = create_files_mapping(
+                        control_experiment.get('original_files'),
+                        files_structure.get('excluded_types'))
+                for peaks_file in files_structure.get('peaks_files').values():
```
caseylitton: All of the peaks_file conditionals under
```
for peaks_file in files_structure.get('peaks_files').values():
```
are filters for files_structure.get('peaks_files').values().  Lines 69 - 83 can be put in a generator expression much like 'derived_from_external_bams' below.  Although, 'derived_from_external_bams' should be a generator expression and not a list comprehension.  In general, if you have a list or set that you need to consume once, then you can make it a generator.

```
peaks_files_gen = (
    peaks_file
    for peak_file in files_structure.get('peaks_files').values()
    ...condition
    ...condition
)
for peaks_file in peaks_files:
    derived_from_files = etc...
    derived_from_external_bams = (
        derived_from
        derived_from in derived_from_files
        ...condition
        ...condition
    )
```

src/encoded/audit/experiment.py
```diff
@@ -57,162 +57,177 @@ def audit_experiment_chipseq_control_read_depth(value, system, files_structure):
...
+                    if not peaks_file.get('award') or \
+                        peaks_file.get('award')['rfa'] not in [
```
caseylitton: put the conditional list outside the loop.  Like where we build the controls objects.

src/encoded/audit/experiment.py
```diff
@@ -57,162 +57,177 @@ def audit_experiment_chipseq_control_read_depth(value, system, files_structure):
...
+                                'ENCODE',
+                                'Roadmap']:
+                        continue
+                    if peaks_file.get('lab') not in ['/labs/encode-processing-pipeline/']:
+                        continue
+                    bio_reps = peaks_file.get('biological_replicates')
+                    if not bio_reps or len(bio_reps) > 1:
+                        continue
+
+                    derived_from_files = list(get_derived_from_files_set([peaks_file],
```
caseylitton: 'get_derived_from_files_set '  returns a list when the last argument is true so we don't need to create a new list from the return value.

src/encoded/audit/experiment.py
```diff
@@ -57,162 +57,177 @@ def audit_experiment_chipseq_control_read_depth(value, system, files_structure):
...
+                        target_name,
+                        target_investigated_as)
+
+
+def get_control_target_name(control_id, control_objects):
+    control = control_objects.get(control_id)
+    if control and 'target' in control and \
+       'name' in control['target']:
+        return control['target']['name']
+    return False
```
caseylitton: We don't need a return statement at all.  It will return None which works as False most the time

src/encoded/audit/experiment.py
```diff
@@ -57,162 +57,177 @@ def audit_experiment_chipseq_control_read_depth(value, system, files_structure):
...
+                        peaks_file.get('assembly'),
+                        cumulative_read_depth,
+                        control_bam_details,
+                        target_name,
+                        target_investigated_as)
+
+
+def get_control_target_name(control_id, control_objects):
+    control = control_objects.get(control_id)
+    if control and 'target' in control and \
```
caseylitton: Much like sting concatenation,   we don't need slashes for new lines, just put the whole thing in parentheses.

src/encoded/audit/experiment.py
```diff
@@ -57,162 +57,177 @@ def audit_experiment_chipseq_control_read_depth(value, system, files_structure):
...
+    return False
+
+
+def check_pipeline(pipeline_title, control_file_id, file_structure):
+    control_file = file_structure.get('alignments')[control_file_id]
+    if ('analysis_step_version' in control_file and
+            'analysis_step' in control_file.get('analysis_step_version')):
+        pipelines = control_file.get('analysis_step_version').get('analysis_step').get('pipelines')
+        return pipeline_title in get_pipeline_titles(pipelines)
+    return False
```
caseylitton: Again we don't need a return false because the default is return None, which works in our case.  In general, functions should be boolean, where they return True or False, or they should return something or nothing.

src/encoded/audit/experiment.py
```diff
@@ -57,162 +57,177 @@ def audit_experiment_chipseq_control_read_depth(value, system, files_structure):
...
+
+def generate_control_bam_details_string(control_bam_details):
+    to_return = ''
+    for (file_id, depth, exp_id) in control_bam_details:
+        to_return += 'file {} from control experiment {} has {} usable fragments;'.format(
+            file_id,
+            exp_id,
+            depth
+        )
+    return to_return[:-1]
```
caseylitton: Just to clear, We're removing the last semi-colon from the return string?

src/encoded/audit/experiment.py
```diff
@@ -57,162 +57,177 @@ def audit_experiment_chipseq_control_read_depth(value, system, files_structure):
...
+
+    if not control_bam_details:
+        detail = ('The peaks file {} produced by ENCODE uniformly processing '
+                  'ChIP-seq pipeline has no control alignments specified.').format(
+                      peaks_file_id
+                  )
+        yield AuditFailure('missing control alignments', detail, level='ERROR')
+        return
+    control_details = generate_control_bam_details_string(control_bam_details)
+    prefix = ('Control alignment files ({}) '
```
caseylitton: add this in an else block after the 'if assembly'

src/encoded/audit/experiment.py
```diff
@@ -57,162 +57,177 @@ def audit_experiment_chipseq_control_read_depth(value, system, files_structure):
...
-                        yield from check_control_read_depth_standards(
-                            control_bam,
-                            control_depth,
-                            control_target,
-                            True,
-                            target_name,
-                            target_investigated_as)
-    return
+            controls = value.get('possible_controls')
+            if controls:
```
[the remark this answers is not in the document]
gabdank: done

src/encoded/audit/experiment.py
caseylitton: require re-demo

src/encoded/audit/experiment.py
caseylitton: require re-demo

src/encoded/audit/experiment.py
caseylitton: require re-demo

src/encoded/audit/experiment.py
caseylitton: require re-demo

src/encoded/audit/experiment.py
caseylitton: require re-demo

src/encoded/audit/experiment.py
caseylitton: require re-demo

src/encoded/audit/experiment.py
gabdank: done

src/encoded/audit/experiment.py
gabdank: made the 'get_derived_from_files_set' to return list even if the last argument if False
removed the unnecessary conversion in the code

src/encoded/audit/experiment.py
gabdank: done

src/encoded/audit/experiment.py
gabdank: done

src/encoded/audit/experiment.py
gabdank: yep

src/encoded/audit/experiment.py
gabdank: done

PR #2190, opened 2018-03-22
ENCD-3938 Adding new Series type "AggregateSeries"

gabdank:
Adding aggregate series to support HiC mega-maps


forresttanaka:
My ENCD-3938 PR will conflict directly with the app.js changes here, so the trailing comma will definitely get fixed once I resolve this conflict.

PR #2193, opened 2018-03-23
ENCD-3935 adding new biosample type "single cell"

gabdank:
new type was added to biosample, it required also chnaging the order of biosample summary - so it was changed too, and tests were updated

### DIFF
--- a/src/encoded/static/components/app.js
+++ b/src/encoded/static/components/app.js
@@ -90,6 +90,8 @@ const biosampleTypeList = [
     'stem cell',
     'in vitro differentiated cells',
     'induced pluripotent stem cell line',
+    "single cell",
+    "cell-free sample"
 ];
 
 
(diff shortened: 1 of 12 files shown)


src/encoded/static/components/app.js
```diff
@@ -90,6 +90,8 @@ const biosampleTypeList = [
     'stem cell',
     'in vitro differentiated cells',
     'induced pluripotent stem cell line',
+    "single cell",
+    "cell-free sample"
```
forresttanaka: To fit AirBnB style guides, this last item should have a trailing comma. I have changes to app.js coming up, so I’d recommend not worrying about this here and just fixing in my branch since this will show up in red once I rebase.

forresttanaka: My ENCD-3938 PR will conflict directly with the app.js changes here, so the trailing comma will definitely get fixed once I resolve this conflict.

PR #2200, opened 2018-03-27
ENCD-3451 Summary page

forresttanaka:
* Tried to use awards.js charting functions and components the summer interns wrote as much as possible, which meant modifying them to fit a more general case. These charting functions should really be moved to a new file since they’re no longer specific to awards, but we’ll worry about that in another ticket as the charting code needs a lot of refactoring to reduce duplication now that we have a better understanding of how we use charts.
* I included some embryonic code in statuslabels.js so that any front-end code can get consistent statuses for each kind of object and at different login levels, only defining the statuses for Experiment for now, as only the summary page uses this right now. After I wrote this part, I found out Keenan was reworking statuses, so I think this has promise, but I’ll need to see what comes out of his efforts before really continuing with this in ENCD-3629.
* One back-end change to handle the /summary/ route. Only ?type=Experiment gives results. Most of this is just a simpler variant of the matrix code, as it doesn’t retrieve the `hits` data — only facet data. I also added a couple facets I needed to the `matrix` object in experiment.py, as well as a new property to define the summary search.
* Removed the project and biosample chart colors out of the App context and into globals because they’re static and made more sense in globals than carried around in React, especially now that we have replicate chart colors in addition to the existing ones.
* Many of the file changes are only to import StatusLabel with object destructuring because that component is no longer the only/default exported component in statuslabel.js.
* Some of the changes to the scss files are to make printing the summary charts more usable, so that pieces don’t get split across pages.


caseylitton:
Be sure to add David to review your PR.  And let Karthik know since he's doing a big search update.

src/encoded/static/components/award.js
```diff
@@ -1648,22 +1667,103 @@ MilestonesTable.propTypes = {
...
 
+
+const ExperimentDateRenderer = (props) => {
+    const { experiments, awards } = props;
+
+    if (experiments) {
+        return <ExperimentDate experiments={experiments.facets} awards={awards} />;
+    }
+    return null;
+};
```
davisagli: Why does this component exist instead of just folding the logic into ExperimentDate?

Also the parameters don't seem right. You're now passing facets rather than experiments to the `experiments` parameter (it's been changed to expect that, but the name should be updated) and both the code that calls this and ExperimentDate use `award` (singular) instead of `awards`

src/encoded/static/components/award.js
```diff
@@ -1690,97 +1790,60 @@ const ExperimentDate = (props) => {
...
+        let accumulator = 0;
         // Make the data cumulative
-        const accumulatedData = dataset.map((term) => {
-            accumulator += term;
-            cumulativedataset = accumulator;
-            return cumulativedataset;
+        const accumulatedData = deduplicated.map((term) => {
+            accumulator += term.doc_count;
+            return accumulator;
         });
```
davisagli: You could do `accumulatedData = deduplicated.reduce((accumulator, term) => accumulator + term.doc_count)`, but the current way is probably more readable.

src/encoded/static/components/summary.js
```diff
@@ -0,0 +1,464 @@
...
+    linkUri: PropTypes.string.isRequired, // URI of base link for each bar to link to
+};
+
+SummaryStatusChart.contextTypes = {
+    navigate: PropTypes.func,
+};
+
+
+// Render the horizontal facets.
+class SummaryHorzFacets extends React.Component {
```
davisagli: Avoid abbreviations in class and variable names.

src/encoded/static/components/summary.js
```diff
@@ -0,0 +1,464 @@
...
+// property with this summary page. Might be good for a refactor later to share common code.
+class SummaryData extends React.Component {
+    constructor() {
+        super();
+        this.mediaQueryInfo = null;
+    }
+
+    componentDidMount() {
+        if (window.matchMedia) {
+            this.mediaQueryInfo = window.matchMedia('print');
```
davisagli: Thing I learned!

src/encoded/static/components/award.js
forresttanaka: I _love_ reduce(), but accumulatedData is an array with the same number of elements as deduplicated, and completely based on deduplicated’s contents. It’s used to display the y-axis values over time in the cumulative chart.

davisagli: Ah, righto. I was reading too fast.

src/encoded/static/components/award.js
forresttanaka: I really can’t explain. Given that I didn’t add even a one-line comment, I’m thinking I was trying something and didn’t back it out when I decided against it.

src/encoded/types/experiment.py
```diff
@@ -440,11 +442,13 @@ def replication_type(self, request, replicates=None, assay_term_name=None):
                 'assay_slims',
                 'target.investigated_as',
                 'month_released',
+                'date_submitted',
                 'files.file_type',
             ],
             'group_by': 'assay_title',
             'label': 'Assay',
         },
+        'summary_grouping': ['replication_type', 'status'],
```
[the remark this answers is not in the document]
forresttanaka: I should have thought of that. It’s now a separate object, and I reverted the `matrix` facet list to what it was before.

src/encoded/static/components/summary.js
davisagli: Just my own personal rule of thumb. It probably wasn't very ambiguous in this case, but generally I don't think saving a few bytes is worth making someone reading the code later (including future me) have to guess about what the abbreviation stands for.

PR #2201, opened 2018-03-27
ENCD-3944 Add organ slim shims & update ontology.json

jahilton:
Forced organ slims to some biosamples. Did standard update of
ontology.json & associated documentation. ENCD-3944


caseylitton:
PR name needs ENCD-####

PR #2204, opened 2018-03-28
[FIX] ENCD-3868 Failing test_batch_download_lookup_column_value

kakaly:
**Issue:** lookup_column_value() returns a string. The validation data returned by the validation fixture for validating a list should be a concatenated string of the objects contained in the list. But, the validation fixture returns a list for validation. 

**Fix:** Change the validation fixture to return a concatenated string of the objects contained in the list for validating a list.

### DIFF
--- a/src/encoded/tests/test_batch_download.py
+++ b/src/encoded/tests/test_batch_download.py
@@ -44,7 +44,7 @@ def lookup_column_value_validate():
         'award.project': 'Roadmap',
         '@id': '/experiments/ENCSR751ISO/',
         'level.name': '',
-        '@type': ['Experiment', 'Dataset', 'Item']
+        '@type': 'Experiment,Dataset,Item'
     }
     return valid
 


src/encoded/tests/test_batch_download.py
```diff
@@ -44,7 +44,7 @@ def lookup_column_value_validate():
         'award.project': 'Roadmap',
         '@id': '/experiments/ENCSR751ISO/',
         'level.name': '',
-        '@type': ['Experiment', 'Dataset', 'Item']
+        '@type': 'Experiment,Dataset,Item'
```
hitz: that looks totally bizarre to me.

kakaly: Yeah. Actually, I wrote this fixture. I think when I ran it, I dint validate it for a list. So, it passed locally. Then later I added list to the validation data and dint mind running it.

kakaly:
Reference #2205 .

PR #2207, opened 2018-03-29
[2ND] ENCD-3934 Update to original 3934 commit

caseylitton:
Replaces https://github.com/ENCODE-DCC/encoded/pull/2203

### DIFF
--- a/src/encoded/audit/library.py
+++ b/src/encoded/audit/library.py
@@ -1,0 +1,21 @@
+from snovault import (
+    AuditFailure,
+    audit_checker,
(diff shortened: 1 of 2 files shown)


forresttanaka:
@gabdank Right now in dev, running a local instance fails to index library ENCLB001SCL because its plate_location is “5:3” which doesn’t match the library schema which has the regex "^([ABCDEFGH]([1-9]|1[0-2]))$" The local instance still starts fine, but it has that error in the console output and is missing that library of course.

PR #2210, opened 2018-04-02
REG-9 Fix vis button for region search

tdreszer:
Simple change to fix the visualize button on region-search.  Forrest should review code.


src/encoded/static/components/region_search.js
```diff
@@ -189,6 +198,10 @@ class AdvSearch extends React.Component {
         const id = url.parse(this.context.location_href, true);
         const region = id.query.region || '';
 
+        if (this.state.genome === '') {
+            this.state.genome = context.assembly ? context.assembly : regionGenomes[0].value;
```
forresttanaka: React’s `this.state` is immutable through any mechanism besides React’s `setState()`. Did `setState()` not work here? Calling `setState()` within `render()` is really tricky because that can cause infinite render loops, but I think it should work here because of the `if` statement that gates it.

src/encoded/static/components/region_search.js
```diff
@@ -199,7 +199,7 @@ class AdvSearch extends React.Component {
         const region = id.query.region || '';
 
         if (this.state.genome === '') {
-            this.state.genome = context.assembly ? context.assembly : regionGenomes[0].value;
+            this.setState({ genome: context.assembly ? context.assembly : regionGenomes[0].value });
```
forresttanaka: Sorry; I should have noticed this aspect too from the first time. Do you have the eslint extension in your code editor? I’m not sure it would warn you for this specific thing or not, but it runs afoul of [rule 15.7](https://github.com/airbnb/javascript#comparison--unneeded-ternary).

`context.assembly || regionGenomes[0].value` would satisfy AirBnB.

PR #2214, opened 2018-04-06
ENCD-3961 eslint / AirBnB npm package updates

forresttanaka:
For code that doesn’t comply with this version of AirBnB/eslint, I generally had a choice of:

1. Rewrite code to follow the rule
2. Disable the rule in the eslintrc permanently
3. Disable the rule in the eslintrc with the goal of enabling the rule later
4. Disable the rule in the specific section of code

I only took option 1 where the risk was very small, like changing indentation. Some were borderline, like converting from string refs to function refs.

I took option 2 where keeping the rule seems untenable going forward, like one React component per file, or converting to two spaces per indent.

I took option 3 where the number of violations made the thought of putting eslint-disable comments in every relevant place seemed too much, or if I wasn’t sure if rewriting the code now or in the future is a good idea. So those rules in .eslintrc.js are marked with a `// FF` comment after them as a reminder that new code should follow the rule even though the rule is disabled.

I took option 4 where changing the code seemed to have too much risk of accidentally changing the functionality, but I wanted to keep the rule going forward. So if you change this very code, you’ll have a reminder right above it to fix this if you choose, and any new code should fit the rule. In some cases it might look like it makes more sense to disable a certain lint check on an entire file instead of over individual sections, but I felt the disabling individual sections would help us remember the rule if we added a new element to that file.

I changed the .eslintrc file from a JSON file to a Javascript file, as it allows commenting without running afoul of JSON linters. I tried to use ES6 module syntax in .eslintrc.js, but I can’t get linting to work without commonJS syntax.

The largest danger I think with this change is the requirement for either a .isRequired or a default property value, as this applied all over our code base. In most cases this is straightforward, but there are some where the component gets passed `undefined` but no check is done to make sure it exists before using it, yet some circumstance seems to avoid a problem. I decided to leave these as much as possible and file future tickets if bugs (which would already be happening without this branch) are found.

QA would just mean trying everything, which is likely a waste of QA’s focused time. I’ll describe what I think is a good path for this ticket in Jira.


davisagli:
npm-shrinkwrap.json lost a LOT of lines...did you forget to pass the `--dev` flag to `npm shrinkwrap` to include dev dependencies?

forresttanaka:
> did you forget to pass the --dev flag to npm shrinkwrap to include dev dependencies

It appears so. I’ll update that and push.

PR #2219, opened 2018-04-13
ENCD-3973 Redo how GM characterization documents get displayed

forresttanaka:
The basic cause of this problem is that I had expected DocumentRender to be called even if the GM/characterization document search 404ed, but it doesn’t.  So if the GM had no documents of its own, and none of the GM characterizations had documents, then no GM characterizations would appear. If even one document existed, then they would all appear. All the local test data either had no characterizations or some with documents, so it always checked out.

To fix this, instead of using `<FetchedData>` to retrieve the documents, I use `requestSearch` which does a search GET request for document objects that’s promise based, then sets a couple of states with the results of the GET request which causes the document panels to update with the results along with the embedded characterizations.

### DIFF
--- a/src/encoded/static/components/genetic_modification.js
+++ b/src/encoded/static/components/genetic_modification.js
@@ -319,7 +319,7 @@ Attribution.propTypes = {
 // Render a panel containing GM document panels and GM characterization panels associated with a
 // GM object.
 const DocumentsRenderer = ({ characterizations, modificationDocuments, characterizationDocuments }) => {
-    // Don't need to check for characterizationDocuments.length because these only get displyed
+    // Don't need to check for characterizationDocuments.length because these only get displayed
     // within characterization panels.
     if (characterizations.length || modificationDocuments.length) {
         // Make a mapping of characterization document @ids to characterization documents
@@ -334,7 +334,7 @@ const DocumentsRenderer = ({ characterizations, modificationDocuments, character
         // make a copy of each and modify that instead.
         const characterizationsCopy = characterizations.map((characterization) => {
             const copy = Object.assign({}, characterization);
-            copy.documents = _.compact(characterization.documents.map(documentAtId => characterizationDocumentMap[documentAtId]));
+            copy.documents = characterization.documents ? _.compact(characterization.documents.map(documentAtId => characterizationDocumentMap[documentAtId])) : [];
             return copy;
         });
 
@@ -357,6 +357,25 @@ DocumentsRenderer.propTypes = {
 };
 
 
+/**
+ * Collect an array of all characterization document @ids.
+ *
+ * @param {array} characterizations - Array of characterizations from GM context object.
+ * @return {array} - Array of all characterization document @ids from all characterizations in the GM context object.
+ */
+const collectCharacterizationAtIds = (characterizations) => {
+    const characterizationDocumentAtIds = [];
+    if (characterizations && characterizations.length) {
+        characterizations.forEach((characterization) => {
+            if (characterization.documents && characterization.documents.length) {
+                characterizationDocumentAtIds.push(...characterization.documents);
+            }
+        });
+    }
+    return characterizationDocumentAtIds;
+};
+
+
 // Render the entire GeneticModification page. This is called by the back end as a result of an
 // attempt to render an object with an @type of GeneticModification.
 class GeneticModificationComponent extends React.Component {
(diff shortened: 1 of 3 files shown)


forresttanaka:
Need to fix one corner case I just noticed.

src/encoded/static/components/genetic_modification.js
```diff
@@ -316,190 +316,229 @@ Attribution.propTypes = {
...
+                        } else {
+                            modificationDocuments.push(document);
+                        }
+                    });
+                }
+
+                // If the resulting documents are different from what we have in state, then
+                // update the state.
+                if (this.geneticModificationComponentChanged(modificationDocuments, characterizationDocuments)) {
+                    this.setState({ modificationDocuments, characterizationDocuments });
```
davisagli: It feels odd to do it this way when React already has `shouldComponentUpdate` that is called when the state changes.

src/encoded/static/components/genetic_modification.js
```diff
@@ -316,190 +316,229 @@ Attribution.propTypes = {
...
+        this.requestDocuments = this.requestDocuments.bind(this);
+        this.geneticModificationComponentChanged = this.geneticModificationComponentChanged.bind(this);
+    }
+
+    componentDidMount() {
+        this.requestDocuments();
+    }
+
+    componentDidUpdate() {
+        this.requestDocuments();
```
davisagli: I haven't tried it, but it looks like this would result in 2 requests? A change to props would trigger this, which would fetch the query and update the state, which would trigger this again. No?

src/encoded/static/components/genetic_modification.js
forresttanaka: Was wondering whether to add a comment on that, and I guess I should have. The timing of shouldComponentUpdate isn’t quite right for this case, and solving that would have been, in my view, much less clear than just saying, “If the results are the same as what we had in state, then that’s it for re-rendering until something else changes.” I checked other resources to see if there was some clever way, but I mostly saw people just adding an if condition on the setState instead of using shouldComponentUpdate. It’s great when props are changing, but in this case we don’t know what’s changing until the GET request comes back.

src/encoded/static/components/genetic_modification.js
forresttanaka: Yes, absolutely, on initial page load. I saw this as not a significant issue. Should I put more time into avoiding that? As I said, I’m fixing one corner case right now.

src/encoded/static/components/genetic_modification.js
```diff
@@ -316,190 +316,229 @@ Attribution.propTypes = {
...
-    }
 
-    // Now filter out any characterization documents out of modDocs.
-    const modDocs = modCharDocs.filter(doc => props.charDocs.indexOf(doc['@id']) === -1);
+        // Need to replace the `documents` array of @ids in each characterization with an array of
+        // corresponding document objects. We can't touch the GM object's characterizations so we
+        // make a copy of each and modify that instead.
+        const characterizationsCopy = characterizations.map((characterization) => {
+            const copy = Object.assign({}, characterization);
+            copy.documents = _.compact(characterization.documents.map(documentAtId => characterizationDocumentMap[documentAtId]));
```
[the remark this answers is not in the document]
forresttanaka: It doesn’t _seem_ to be, but I agree I should check for that. I do elsewhere.

hitz:
the 404 point is an interesting one.  As I recall it drove Karen nuts to have the pages' log filled with 404s as "expected empty set".  

@forresttanaka this actually confuses me now:   On this page: (currently): https://www.encodeproject.org/experiments/ENCSR442FIY/

There is a 404 from this query: https://www.encodeproject.org/search/?type=Experiment&possible_controls.accession=ENCSR442FIY

But this API call does NOT return a 404 (try format=json).  So you must be 404'ing it using react's .fetch()?

src/encoded/static/components/genetic_modification.js
davisagli: It seems to me that we only want to reissue the request if the props changed, not if the state changed. I'm not sure I understand the update scenario though -- what would cause the props to change in a way that requires reissuing the request?

forresttanaka:
That search 404s for me.

screen shot 2018-04-13 at 10 19 30

forresttanaka:
Empty searches have always 404ed for us, so if that needs to change, another back-end ticket needs to be filed.

hitz:
Yeah, search.py expressly sets empty->404 according to vintage 2014 doctrine: https://webmasters.googleblog.com/2014/02/faceted-navigation-best-and-5-of-worst.html

forresttanaka:
For anyone examining the network logs, the last change I made Friday only reduces the number of initial GET requests from two to one if you’re logged out on initial page load. If you’re logged in this change makes no difference — you still get two GET requests because all pages render at least twice if you’re logged in on initial page load: once without any user information and with logged-out data, and once with user information and all data once the session-properties GET request completes.

PR #2221, opened 2018-04-17
ENCD-3979 Fix zc buildout for pypi update

davisagli:
* Use currently recommended approach to bootstrapping (pip install zc.buildout and then run buildout bootstrap). Note that I removed the version pins. That probably feels like a bad idea at first glance, but my rationale is that the point of the version pins was to ensure consistent builds, and that hasn't worked out in practice. It seems to be more important to have recent versions that are compatible with the PyPI servers. Thanks to the new approach, we no longer need bootstrap.py or ez_setup.py.
* Update allow-hosts to use the domains that PyPI switched to yesterday


forresttanaka:
I won’t be approving this simply because someone else should.

Has the build-from-clean process changed with this branch?

davisagli:
> Has the build-from-clean process changed with this branch?

See the updates in the README. It would be great if someone could try this out.

forresttanaka:
It built fine! But dev-servers had a problem immediately:

```
Home-iMac:encoded ftanaka$ bin/dev-servers development.ini --app-name app --clear --init --load
Traceback (most recent call last):
  File "bin/dev-servers", line 78, in <module>
    import snovault.dev_servers
  File "/Users/ftanaka/Sites/encoded/develop/snovault/src/snovault/__init__.py", line 33, in <module>
    from .schema_utils import load_schema  # noqa
  File "/Users/ftanaka/Sites/encoded/develop/snovault/src/snovault/schema_utils.py", line 261
    yield from orig_uniqueItems(validator, uI, instance, schema)
             ^
SyntaxError: invalid syntax
```

davisagli:
@forresttanaka what version of python did you use? `yield from` was introduced in 3.3; we should be using 3.4.

caseylitton:
I built locally and a demo.  Only issue I had was that I needed to delete old pip versions of setuptools and or zc packages.    Merging code now.

PR #2222, opened 2018-04-17
Manually Install Java jdk-8u172

caseylitton:
#### Built a demo with #2221 and the java update.
* i-0e4f03731f291061d.instance.encodedcc.org
* ssh java-testing.instance.encodedcc.org
* https://java-testing.demo.encodedcc.org


caseylitton:
webppa updated to 8u172 before we merged.

PR #2224, opened 2018-04-19
ENCD-3950 adding dependency to experiment to require date_submitted to submitted experiments

gabdank:


src/encoded/schemas/experiment.json
```diff
@@ -143,11 +143,19 @@
                         }
                     }
                 },
+                {
```
keenangraham: You are putting dependency on experiment but upgrading all of the children of dataset? I don't think it will be enforced on anything but experiment.

src/encoded/tests/test_schema_experiment.py
```diff
@@ -107,3 +109,14 @@ def test_experiment_valid_statuses(status, testapp, experiment):
...
+
+
+def test_submission_date_dependency(testapp, experiment_no_error):
+    expt = testapp.post_json('/experiment', experiment_no_error).json['@graph'][0]
+    res = testapp.patch_json(expt['@id'], {
+        'status': 'submitted'}, expect_errors=True)
+    assert res.status_code == 422
+    res = testapp.patch_json(expt['@id'], {
+        'status': 'submitted',
+        'date_submitted': '2000-10-10'}, expect_errors=True)
```
keenangraham: Why do we expect_errors here?

keenangraham: Also I think you can use this idiom instead of the assert res.status_code: 
`testapp.patch_json(expt['@id'], {'status': 'submitted'}, status=422)`
`testapp.patch_json(expt['@id'], { 'status': 'submitted', 'date_submitted': '2000-10-10'}, status=200)`

src/encoded/schemas/experiment.json
gabdank: you are right
fixing

gabdank:
@keenangraham I have updated the branch according to your review comments

src/encoded/schemas/changelogs/reference.md
```diff
@@ -1,5 +1,9 @@
 ## Changelog for reference.json
 
+### Schema version 16
```
keenangraham: I don't see dependency for this one?

src/encoded/schemas/changelogs/project.md
```diff
@@ -1,5 +1,9 @@
 ## Changelog for project.json
 
+### Schema version 15
```
keenangraham: I don't see dependency for this one?

src/encoded/schemas/experiment.json
```diff
@@ -143,11 +143,19 @@
...
+                            "enum": ["submitted"]
+                        }
+                    }
+                },
                 {
                     "not": {
                         "properties": {
                             "status": {
-                                "enum": ["released", "revoked"]
+                                "enum": ["released", "revoked", "submitted"]
```
keenangraham: What happens when someone goes from in progress -> released, skipping submitted? This seems to allow released files with no date_submitted.

src/encoded/upgrade/dataset.py
```diff
@@ -364,3 +364,21 @@ def dataset_16_17(value, system):
 def dataset_17_18(value, system):
     if value['status'] == "started":
         value['status'] = "in progress"
+
+
+@upgrade_step('experiment', '18', '19')
```
keenangraham: I suppose you can't inherit dependency from parent dataset. Copying this into all of the children objects and keeping them updated seems like a pain.

src/encoded/schemas/changelogs/project.md
gabdank: missed that one

src/encoded/schemas/changelogs/reference.md
gabdank: missed that one

src/encoded/schemas/experiment.json
gabdank: the dependency we add now will allow your scenario. We have to backfill date submitted to many released datasets, before we ad dependency that required date_submitted to released datasets as well.

src/encoded/tests/test_types_series.py
```diff
@@ -8,7 +8,8 @@ def base_reference_epigenome(testapp, lab, award):
         'lab': lab['uuid'],
         'status': 'submitted'
     }
-    return testapp.post_json('/reference-epigenomes', item, status=201).json['@graph'][0]
+    print (testapp.post_json('/reference_epigenome', item))
```
keenangraham: This print looks like it should be removed.

src/encoded/tests/test_schema_experiment.py
keenangraham: Change name to test_experiment_blah_blah. Also I think you reverted the changes that cleaned this up. i.e. `testapp.patch_json(expt['@id'], {'status': 'submitted'}, status=422)`.

src/encoded/upgrade/dataset.py
```diff
@@ -364,3 +364,10 @@ def dataset_16_17(value, system):
...
+def dataset_18_19(value, system):
```
keenangraham: I know I didn't do this for the last upgrade I wrote but we could clean this up to use all single quotes instead of a mix of single and double quotes.

gabdank: fixed in the upgrade I have added

src/encoded/tests/test_types_series.py
gabdank: removed

gabdank:
@keenangraham updated the branch and pushed

PR #2225, opened 2018-04-19
ENCD-3216 Hide antibody characterizations via user access level

forresttanaka:
The core of this change is to filter AntibodyLot.characterizations (embedded on antibody pages) by status based on your current logged-in and admin states, and I extend my past half-hearted attempt at status revision which now seems necessary to provide in some form in ENCD-3629.

You can still see unreleased antibody characterizations in the JSON, but for now the team feels this is acceptable until and unless we figure out something better than a front-end-only change.

Where this Jest test will go eventually, I don’t know. Probably it’ll just be a kind of random sample of types, or just these two.

### DIFF
--- a/src/encoded/static/components/antibody.js
+++ b/src/encoded/static/components/antibody.js
@@ -12,14 +12,17 @@ import { DbxrefList } from './dbxref';
 import { DocumentsPanel, Document, DocumentPreview, CharacterizationDocuments } from './doc';
 import { RelatedItems } from './item';
 import { AlternateAccession } from './objectutils';
-import { StatusLabel } from './statuslabel';
+import { StatusLabel, getObjectStatuses, sessionToAccessLevel } from './statuslabel';
 
 
 const LotComponent = (props, reactContext) => {
     const context = props.context;
 
-    // Sort characterization arrays
-    const characterizations = _(context.characterizations).sortBy(characterization => (
+    // Sort characterization arrays, filtering for the current logged-in and administrative status.
+    const accessLevel = sessionToAccessLevel(reactContext.session, reactContext.session_properties);
+    const viewableStatuses = getObjectStatuses('AntibodyCharacterization', accessLevel);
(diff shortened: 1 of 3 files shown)


src/encoded/static/components/antibody.js
```diff
@@ -12,14 +12,17 @@ import { DbxrefList } from './dbxref';
...
 
 const LotComponent = (props, reactContext) => {
     const context = props.context;
 
-    // Sort characterization arrays
-    const characterizations = _(context.characterizations).sortBy(characterization => (
+    // Sort characterization arrays, filtering for the current logged-in and administrative status.
+    const accessLevel = sessionToAccessLevel(reactContext.session, reactContext.session_properties);
+    const viewableStatuses = getObjectStatuses('AntibodyCharacterization', accessLevel);
+    let characterizations = context.characterizations.filter(characterization => viewableStatuses.indexOf(characterization.status) !== -1);
```
davisagli: Since we're using `babel-polyfill`, I think you should be able to reliably use `Array.includes`. However it's probably worth confirming that it works in IE11.

src/encoded/static/components/statuslabel.js
```diff
@@ -8,47 +8,79 @@ import * as globals from './globals';
...
+/**
+ * Maps the current session information from the <App> React context to an access level from the
+ * objectStatusLevels array.
+ *
+ * @param {object} session - From encoded session context
+ * @param {object} sessionProperties - From encoded session_properties context
+ */
+export const sessionToAccessLevel = (session, sessionProperties) => {
+    const loggedIn = !!(session && session['auth.userid']);
+    const administrativeUser = loggedIn && !!(sessionProperties && sessionProperties.admin);
```
davisagli: Use `Boolean(expr)` instead of `!!expr` for readability. (Turn on https://eslint.org/docs/rules/no-implicit-coercion if we want to enforce this.)

PR #2227, opened 2018-04-20
ENCD-3980 Organize ini files and vars

caseylitton:
I diff'd the derived ini file results with origin/dev and they were the same less the changes in line placement.  And I deploy'd a single machine without issue.

Not meant to be a final version but this should be more consistent.


caseylitton:
replaced with combine PR https://github.com/ENCODE-DCC/encoded/pull/2228

PR #2231, opened 2018-04-24
[2nd] ENCD-3980 Fix base ini ontology path

caseylitton:


ini-templates/base-template.ini
```diff
@@ -6,7 +6,7 @@ download_proxy = https://download.encodeproject.org/
 elasticsearch.server = ${elasticsearch.server}
 file_upload_bucket = encoded-files-dev
 file_upload_profile_name = ${file_upload_profile_name}
-ontology_path = ${aws_ip_ranges_path}
```
hitz: This is the correct change, but maybe the mapping for aws_ip_ranges_path got lost?  Looks OK, still in base.ini.

caseylitton: It was a copy paste error.

PR #2240, opened 2018-05-02
ENCD-3981 remove histone modification

jahilton:


src/encoded/tests/test_upgrade_target.py
```diff
@@ -70,3 +81,10 @@ def test_target_upgrade_status_5_6(upgrader, target_5):
         'target', target_5, current_version='5', target_version='6')
     assert value['schema_version'] == '6'
     assert value['status'] == 'current'
+
+
+def test_target_upgrade_status_6_7(upgrader, target_6):
```
keenangraham: You could rename this function to something like `test_target_upgrade_remove_histone_modification_6_7`.

src/encoded/tests/test_upgrade_target.py
```diff
@@ -38,6 +38,17 @@ def target_5(target):
     return item
 
 
+@pytest.fixture
+def target_6(target):
+    item = target.copy()
+    item.update({
+        'schema_version': '5',
```
keenangraham: I think this should be schema_version six.

src/encoded/tests/test_upgrade_target.py
```diff
@@ -38,6 +38,17 @@ def target_5(target):
...
+        'status': 'proposed',
```
keenangraham: Proposed no longer exists.

src/encoded/tests/test_upgrade_target.py
jahilton: renamed the function & pushed change

src/encoded/tests/test_upgrade_target.py
jahilton: schema_version:6 and pushed change

src/encoded/tests/test_upgrade_target.py
jahilton: status:current and pushed change

PR #2241, opened 2018-05-03
ENCD-4011 Fix wrong vars in buildout candidate config file

caseylitton:


candidate.cfg
```diff
@@ -5,5 +5,4 @@ extends = buildout.cfg
 accession_factory = encoded.server_defaults.enc_accession
 file_upload_bucket = encode-files
 blob_bucket = encoded-blobs
-region_search_instance = region-search.production.encodedcc.org:9200
-indexer_processes = 16
\ No newline at end of file
+region_search_instance = region-search.production.encodedcc.org:9201
```
hitz: shouldn't this be the local ES cluster IP?

caseylitton: Yea, I was wondering about that...I was going to ask @tdreszer but he's out all week.

caseylitton: I looked into this.  region_search_instance is not in any *-template.ini so it's never making it to and *.ini file.  Meaning it doesn't look like it's currently used.

PR #2243, opened 2018-05-07
ENCD-4016 adding community to users

gabdank:
adding community to all users viewing_groups


keenangraham:
Could you add upgrade test?

PR #2246, opened 2018-05-08
ENCD-3629 Refactor and redesign status displays.

forresttanaka:
The primary goal of this branch is to simplify the code to display statuses, with the secondary goal of updating their appearance so they don’t look so much like buttons (which prompted me to change the button appearance to a white background long ago). The statuslabel.js file has changed name to status.js, and the `<Status>` component is its default export, along with a couple utility functions if needed.

`fileStatusList` had been used to display file statuses specifically. This has been removed because `<Status>` now covers all object types.

Because some classes and status contents have changed, some of the Jest tests had to change. I also added new Jest tests specifically for the `<Status>` component and utility functions.

Object statuses no longer display in the header, so audits now appear in the header by themselves and we no longer have a need for the .status-line CSS class. Status now displays as a definition-list item. Many of the .js file changes relate to this.

The unique antibody status panel has become a proper panel and uses the new `<Status>` component to display what used to be colored circles. This applies to AntibodyLot search results as well.

Facet term components have become more complicated over time, so since I’m now using the `<Status>` component within Status facet terms, I redid the HTML and CSS using flexbox instead of floats and inlines. I tested this successfully in IE11 and MS Edge as well.

I removed `globals.statusClass` as it hasn’t served much purpose up to now, and now serves none with the new `<Status>` component.

The vast majority of the change to `AntibodyComponent` in search.js was to convert it to a stateless component to satisfy AirBnB. It otherwise has similar changes to other listing components.

One criticism I could see for much of this concept is the required synching of the statuses between the schemas and `objectStatuses` in status.js. I could imagine a separate GET request for a special `/statuses/` endpoint that gets this information as needed, or the automatic inclusion of this information in the context of every page. I didn’t see either of these as practical, though I could see the argument for the first as it’s only needed rarely, for getObjectStatuses. The `<Status>` component doesn’t need this.

summary.js has an odd mechanism in `SummaryStatusChart` caused by the first render being without login information even if you’re logged in, followed by another couple renders as login information arrives, but chart.js hasn’t been loaded before the final render happens. Then when chart.js loads and the callback gets called, it has all the data it needs, but for the logged-out case. But by that time, we’ve already gotten new data passed to componentDidMount which had no chart to update with new data.

I use an object data property that signals when we got new data but chart.js wasn’t loaded. Once chart.js loads and the initial chart gets drawn, it checks this property and forces a React update with the logged-in data if it’s true. If anyone has suggestions for a better mechanism than this, please let me know.


src/encoded/static/components/status.js
```diff
@@ -0,0 +1,352 @@
+import React from 'react';
```
hitz: this has nothing to do with the JS code per se - but aren't we duplicating this login in python @keenangraham ?  I feel like this should be configure in the schemas or some other shared config file.

src/encoded/static/scss/encoded/modules/_status.scss
```diff
@@ -0,0 +1,237 @@
+// All status indicator colors. As statuses change in the schemas, should see if corresponding
+// changes are needed here.
+$status-colors: (
```
hitz: And now we have to define colors for each status text string?   I guess it's not trivial to set this in SASS programattically.

forresttanaka: Yes, which is what we were doing before this change as well, though in a very inconsistent way (sometimes in a loop, sometimes statically). Now all statuses are mapped to corresponding colors in a loop.

forresttanaka:
I updated the first comment with new notes. Since this has already been approved, Phil can you just indicate whether you approve or not? I also personally recommend [this](https://blog.github.com/2015-01-21-how-to-write-the-perfect-pull-request/) to avoid some problems we’ve had in the past. We’ve also sometimes seen feedback where it wasn’t clear whether the approval was waiting on that feedback to get addressed, or was just a note that didn’t affect the approval.

zoldello:
@forresttanaka: Here is my review: 
Code review of ENCD-3629


Major (May be serious)

- src/encoded/static/components/__tests__/experiment-test.js
	- Why did you remove a unit test in this file? 

- src/encoded/static/components/__tests__/status-test.js
    - Why are you testing multiple features in the same test? My understanding is that a test should have only one expect statement. 
	- Line 50, why don’t you test for 0 (less than 1) and 100 (large number)? There is a chance the code will receive a number greater than 1 and it should handle it gracefully
	- I am not sure of the point of shallow() ad using that for test. Since shallow seems external to this file, I assume some other file should be responsible for unit test. This is akin to the jQuery test rather than us unit testing jQuery.
	- Line 253, comments should never tell developers what they can read themselves.
	- What is the point of line 259? It does not contribute anything it seems.
	- Line 258 seems janky. If business rule change, this code may be hard to witch-hunt. However, I cannot think of a better way to implement what it does. 
	- Line 290. I think using ‘i’ as a variable is great in for-loops. But I recommend something more descriptive
here, like index or counter. To illustrate, line 301 is a bit confusing since readers will not get a sense of what ‘i’ is without remembering or looking at the context (people do not always do what they ought to.)
	- Lines 504 and 510- These comments seem unnecessary. We can assume readers are a bit familiar with react 101.
	- Line 515 seems to be misplaced. It is better to place it near tooltip.
	- Line 540: Maybe you should format the json to improve readability.
	- Line 582: No need of parenthesis for single-parameter arrow function 
	- Line 606- Code is self-explanatory. 


- src/encoded/static/components/award.js
	-  Line 2174- Why is a css class name “pi”? I doubt if it refers to the number. 
	- Link 2179- Do we suppose to return null or <span></span>/“”  for empty html tags?


- src/encoded/static/components/dataset.js
    - Line 105: I think a div does to suppose to be an immediate child of a dl-tag. Even though we are not using IE7 that strongly dislikes that, it may still be incorrect HTML syntax technically
    - Line 299- Valid HTML (dl has div as an immediate child)?
    - Line 469- Valid HTML (dl has div as an immediate child)?
    - Line 626- Valid HTML (dl has div as an immediate child)?
    - Line 1019- Why are you using “size” attribute? Why not place it in a css class?
    - Line 1062, 1119, 1206- Why are you using “size” attribute? Why not place it in a css class?
    - Line 1303-- Valid HTML (dl has div as an immediate child)?

- src/encoded/static/components/donor.js
    - Line 40 (numerous times)- Valid HTML (dl has div as an immediate child)?  
    - Line 505: Why is there “component” in class name
    - Comments in Lines 658 - 665 seem unneccessary  


- src/encoded/static/components/experiment.js
    - Line 466- Why have a h4 wrapped by a div?



- src/encoded/static/components/file.js
    - Line 42- Size attribute? If this is meant for a property, it may be confused with the old HTML size attribute
    - Line 695- Why have an attribute called “css”?

- src/encoded/static/components/search.js
    - Line 651- I think it is incorrect HTML to have a div as a child of an anchor tag
    - Line 653- Use <em> over css?



- src/encoded/static/scss/encoded/_base.scss
    - Line 138 - 158: Are the numbers in the code- “Magic numbers”?



Minor (just my personal taste that may have little relevance)
	- Why is there an ‘@’ in variable names like  “@type” in line 37 of: 
	- Should be have the word “target” in test
	- In each unit test, should a space separate unit test components of: arrange, act, assert; so developers can know what part is test component?
	- My understand is that JavaScript unit test should not test DOM, it should only test JavaScript. That was true with knockout.js. However, since react.js has the virtual DOM and HTML-like code embedded, it may not confirm to this rule. I am not sure. 
	- I thought there should be only one class in every file..?

forresttanaka:
> Why did you remove a unit test in this file?

Because it tested a status line that no longer exists with this branch. No status lines exist anymore.

> Why are you testing multiple features in the same test? My understanding is that a test should have only one expect statement.

Do you have a reference for this? Jest examples have multiple expect statements. In any case, that's not part of this ticket.
https://facebook.github.io/jest/docs/en/tutorial-react.html

> Line 50, why don’t you test for 0 (less than 1) and 100 (large number)? There is a chance the code will receive a number greater than 1 and it should handle it gracefully

If it received a number greater than one or less than, then this test would fail as it should; there should be exactly one. If there were multiple or no .status--in-progress elements, that would be incorrect and this test would flag that.

> I am not sure of the point of shallow() ad using that for test

`shallow` renders the component so it can be tested. It’s part of enzyme. Without `shallow` no React component can be tested.
https://github.com/airbnb/enzyme/blob/master/docs/guides/jest.md

> Line 253, comments should never tell developers what they can read themselves.

There are only 105 lines in this file. Are you referring to another file? And that’s _purely_ opinion — AirBnB says nothing about comment length or content.

> Why is a css class name “pi”? I doubt if it refers to the number

“PI” is well-known around here and every scientific organization as Principal Investigator. Even the schema uses this term. Mike Cherry is one. context.pi comes directly from our database — that’s what it’s called there. You can ask questions like that separately.

> Link 2179- Do we suppose to return null or /“” for empty html tags?

null:
https://reactjs.org/docs/conditional-rendering.html#preventing-component-from-rendering

>  I think a div does to suppose to be an immediate child of a dl-tag.

This is incorrect HTML we decided years ago to allow, as all browsers we supported even back then do fine with this. It was the only way we knew of to achieve what we wanted with word wrap and responsiveness. If you have a better solution that uses correct HTML, please file a ticket. That’s not part of this ticket.

> Why are you using “size” attribute? Why not place it in a css class

That‘s a good option too, but I felt the size was such a common thing to want to specify, it deserved its own property, and limit the “css” property to one-off cases. Plus you get a slight bit of error checking from React dev builds — the property only allows “small,” “standard,” and “large” (or nothing) or you get an error. You can’t check that with CSS classes.

> Why is there “component” in class name

That’s a personal standard I chose for components that get used with HoCs (the audit HOC in this case). The final decorated component has the same name without “Component.” I don’t require using “Component” for anyone else, and AirBnB doesn’t specify anything for this case.

> Comments in Lines 658 - 665 seem unnecessary

They’re _extremely_ necessary to me. Revisiting code and wondering what properties are for is not productive. I’ll often remember, but I often won’t, and I don’t want to spend any time trying to predict which ones I’ll remember and which ones I won’t. This is a matter of opinion, and I generally don’t ding people for having too many comments or too few. In one code review I did, I told them I just couldn't understand the code because it was complex with zero comments. I think that’s the only time, because I simply couldn’t complete the code review without them.

> Why have a h4 wrapped by a div?

The div has CSS classes. That `<h4>` isn’t on its own line — it’s part of a panel header, and in some cases it might not be an `<h4>` but something else entirely that needs to be in the same position, or the <h4> might need to share with something else in the same section. For that, I just always use the same div for the same place on every page that has this. That’s not part of this ticket though.

> Size attribute? If this is meant for a property, it may be confused with the old HTML size attribute

That’s true. Maybe statusSize or badgeSize.

> Why have an attribute called “css”?

It’s to add CSS classes to the wrapper.

> I think it is incorrect HTML to have a div as a child of an anchor tag

It is incorrect HTML. I don’t remember if I felt there was no choice or what. But that wasn’t modified for this ticket.

> Use over css

I don’t understand. Also that wasn’t modified for this ticket.

> Are the numbers in the code- “Magic numbers”?

Yes, but very well-known ones. I remember using those numbers in a college project.:
https://en.wikipedia.org/wiki/Luma_(video)

forresttanaka:
> Why is there an ‘@’ in variable names like “@type” 

That’s part of the JSON-LD style.

> Should be have the word “target” in test

I’m not sure what you’re referring to.

> In each unit test, should a space separate unit test components of: arrange, act, assert; so developers can know what part is test component

We haven’t established a standard for that. If you can find a reference to something like that, that would be great.

> My understand is that JavaScript unit test should not test DOM, it should only test JavaScript

Jest and enzyme are designed to test React components by drawing into a virtual DOM specific to Jest testing, using jsdom. You’ll see both Javascript tests that don‘t use React or enzyme along with React component tests in our Jest tests.

> I thought there should be only one class in every file

That _is_ part of AirBnB, and I disabled it just because it would require so much rewrite of our code base for little gain that I could see. If you’d like to see that happen and can make a case for it that convinces Ben, I’d be fine with it, though I personally don’t want to do that work.

The other thing about this rule that I worried about is that it might encourage people to make big React components instead of breaking them up, to avoid making new files and importing what you need. I’m not sure that’s true, but I could see that happening.

forresttanaka:
Also, if in your own new code you want to put all your React components each in their own files, go right ahead. For me, anything not specified in AirBnB is programmer’s discretion. I don’t like control freaks, and I try desperately not to be one myself.

zoldello:
We can talk about the points like what I mean by target  and the
arrange/act/assert tomorrow if you like. You got me on the control-freak
thing. I sometimes tend to follow rules but sometimes do not stop to think
of utility.

hitz:
@zoldello Did @forresttanaka answer your questions?  Maybe we could take it off line if you have further comment.

zoldello:
@hitz Yes, Forrest and I spoke about my review offline.

forresttanaka:
I’ll work on the conflicts once I get some unrelated things cleared off.

PR #2255, opened 2018-05-18
ENCD-4034 SVG tooltip and minor JS clean up

zoldello:
Remove title tag-children from svg tags. Did some code cleanup on svg-icons.js to make it better comply with eslint. Added preliminary .editconfig file.


src/encoded/static/libs/svg-icons.js
```diff
@@ -40,3 +38,5 @@ var collapseIcon = module.exports.collapseIcon = function(collapsed, addClasses)
         </svg>
     );
 };
+
+module.exports.svgIcon = collapseIcon;
```
forresttanaka: Thanks for moving this file forward! We’re moving everything to ES6, so should do that for this file too. So `import React from 'react'` and `export const…` instead of `module.exports.…` would be great. The AirBnB eslint extension I think requires this, though I’m not sure. AirBnB itself does: https://github.com/airbnb/javascript#modules

.editconfig
```diff
@@ -0,0 +1,11 @@
...
+
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+#indent_style = space
+#indent_size = 2
+trim_trailing_whitespace = true
+insert_final_newline = true
```
forresttanaka: I’m not sure what editor this file is for? I’m guessing it doesn’t affect anyone it’s not relevant to.

forresttanaka:
Both Travis CI tests failed. The Pyramid one was because the `components/__tests__/antibody-test.js` Jest test failed because it was looking for something in svgIcons, so that’ll need some adjustment. Not sure why the browser test failed. I couldn’t find some text for some reason.

.editconfig
zoldello: Thanks for the good catches. I will address them. The .editconfig file tells text editors what settings to apply like how many spaces are in a tab. This helps make different people using different text editors comply to a team-agreed text-editing standard. We can discuss later on if it is needed (I assumed it is) and what needs to be in it if it is kept.

caseylitton: @zoldello In your home folder you can make a global git ignore file for stuff like this. 

https://gist.github.com/subfuzion/db7f57fff2fb6998a16c

We don't want to check in personal configuration to the main repository.  In the global git ignore you should just add '.editconfig' so git won't check it in.

zoldello:
@caseylitton .editConfig's purpose is to store text editor configuration information; different text editors (sublime, visual studio, atom, etc) can read it and then know how to format code (http://editorconfig.org/). It helps with code-style uniformity (less of a problem with python, but big in JavaScript). I can remove the .editConfig file if you desire. I should of had discussed it with the group before putting it in.

forresttanaka:
I was about to suggest leaving it out of this commit, and then have Jason/Idan/Ben put it on the agenda for @zoldello to present it at a group or project meeting. I doubt that file hurts anything, but I'd like to see it in action with Sublime and VSCode.

PR #2256, opened 2018-05-21
ENCD-3911 Added a break above document box

zoldello:


forresttanaka:
What might also work _and_ take advantage of the display:block of the <p> and <ItemPanel> is just to move <ItemPanel> inside the <col-sm-12> div along with the description line. Since they’re both blocks, they shouldn’t overlay each other as long as they’re in the same "col-sm-12" block.

Would be interesting to test on something that's _not_ a document, but I’m not sure if any other object types use item.js. It would have to be an object whose `@type` gets registered with `globals.panelViews.register` but not `globals.contentViews.register`.

zoldello:
@forresttanaka I terminated the old QA test server and created a new one. The url is: https://encd-3911-fix-box-overflow-f0195d572-phil.demo.encodedcc.org

zoldello:
@forresttanaka That was a good catch.

PR #2257, opened 2018-05-22
ENCD-3364  fixed text and link button misalignment

zoldello:
This is a tricky one. I took out the flex changes made by a prior developer. I could not see a strong reason why it was  needed in the first place and no other code seemed to use it.

### DIFF
--- a/src/encoded/static/scss/encoded/modules/_visualize.scss
+++ b/src/encoded/static/scss/encoded/modules/_visualize.scss
@@ -50,11 +50,14 @@
         line-height: 30px;
         font-weight: bold;
         font-size: 1.1rem;
+        width: 100px;
+        word-wrap: break-word;
     }
 
     @at-root #{&}__assembly-title {
         @extend .browser-selector__assembly;
         font-size: 1.3rem;
+
     }
 
     @at-root #{&}__browsers {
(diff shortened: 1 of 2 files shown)


forresttanaka:
Does this branch mix ENCD-3883 and ENCD-3364?

zoldello:
@forresttanaka I am normally careful when changing branches locally and checking out branches. I added a commit to undo the ENCD-384 commit (biosample.js file involving PMI.) Now, only one file should be modified.

zoldello:
@forresttanaka: I will correct it so that the buttons are flushed.

src/encoded/static/scss/encoded/modules/_visualize.scss
```diff
@@ -46,20 +46,21 @@
...
         font-weight: bold;
         font-size: 1.1rem;
+        width: 100px;
+        word-wrap: break-word;
     }
 
     @at-root #{&}__assembly-title {
         @extend .browser-selector__assembly;
         font-size: 1.3rem;
+
```
caseylitton: This looks unnecessary too, right?  For now let's air on the side of "best practice" and make as little changes as possible.

zoldello:
@caseylitton: Removed space added unnecessarily.

PR #2261, opened 2018-05-24
ENCD-3917-fix matrix text overlap

zoldello:
I could not reproduce this locally and it seems from the description that this may be a hard one for the reporter to duplicate too. So I fixed the issue with the F12 tool and ported over the change. I will try in the test environment. The change is minor and can cause little harm.

### DIFF
--- a/src/encoded/static/scss/encoded/modules/_facet.scss
+++ b/src/encoded/static/scss/encoded/modules/_facet.scss
@@ -92,7 +92,7 @@
 .facet-list li {
     padding: 3px 10px;
     margin-bottom: 1px;
-    margin-right: 15px;
+    margin-right: 5px;
     border-radius: 5px;
     position: relative;
     font-size: 0.9rem;
@@ -274,11 +274,11 @@ div.meta-status {
     display: inline;
     list-style-type: none;
     border: none;
-    margin-right: 5px;  
+    margin-right: 5px;
 }
 
 .data-row ul li a{
-    display: inline;    
+    display: inline;
 }
 
 .data-row img {
@@ -327,7 +327,7 @@ div.meta-status {
     }
 }
 
- 
+
 .main-panel {
     position: relative;
     z-index: 0;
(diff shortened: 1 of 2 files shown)


src/encoded/static/scss/encoded/modules/_facet.scss
```diff
@@ -274,11 +274,11 @@ div.meta-status {
     display: inline;
     list-style-type: none;
     border: none;
-    margin-right: 5px;
```
caseylitton: tldr;
Best practice, don't commit code that didn't need to be changed.  Fixing the syntax issues with our code base is a great thing to do but not at the expense of our engineering process. 


How did this line get changed?  Did you do it manually or was it automated?  

In either case these changes are "unnecessary" considering the ENCD ticket.  I know this may seem silly, it's just some spaces, but there are other things to think about.

One is that we're making a change to our repository and not documenting it.  We don't want to sneak anything into our release no matter how small.  In the end, being able to look at the commit history and trust we see all that we've done has a higher priority then making unnecessary changes.  

Another is that we're adding extra work in the middle of the sprint.  Possibly a lot of extra work that provides little benefit.  In this PR for example we've changed 4 lines of code but only needed to change one.  Now there is four times as much code to review and test.  Consider how this scales. Imagine you need to review dozens of PRs and hundreds of lines of code where 75% of them are not related to the feature.  It wastes a lot of time looking at code that didn't need to be changed.  And if it needs to be changed then it's worth the effort to make a ticket.

That said, having the syntax in the repository cleaned up is beneficial.  We as a team don't want to be crazy about adhering to rules like these, but it is important to  understand the possible consequences of violating them.  Also, if we're going to do it then let's do it all the way.  Make a ticket to fix the syntax in the file.  Then fix all the syntax issues in the file.  For example, on line 280 there should be a space between the curly bracket and the 'a', right?  Why would we fix one and not the other?

zoldello:
@caseylitton I took out the space at line 330 I believe I added unnecessarily. For text time, I will ensure my text editor does not auto-add spaces, like it did in some other lines.

zoldello:
I am requesting that this PR be abandoned because the issue it addressed was fixed by ENCD-3629 (https://github.com/ENCODE-DCC/encoded/commit/e55a52cd785dce576a1694328326e45603ed6a63#diff-cf1b4a48e0291374cf93ea97bf583641)

PR #2270, opened 2018-05-30
added hypen for long words

zoldello:
Firefox and other browsers have slight differences. For example, Firefox adds an extra dash if the word breaks on a dash while Chrome does not. But essentially, all evergreen browsers and IE11 should be the same.

### DIFF
--- a/src/encoded/static/scss/encoded/modules/_key-value-display.scss
+++ b/src/encoded/static/scss/encoded/modules/_key-value-display.scss
@@ -23,6 +23,9 @@ dl.key-value, dl.key-value-doc {
         width: auto;
         word-wrap: break-word;
         word-break: break-word;
+        -webkit-hyphens: auto;
+        -ms-hyphens: auto;
+        hyphens: auto;
     }
 
     dt {


hitz:
This PR has no ticket in the title!  I guess it's ENCD-3727?

src/encoded/static/scss/encoded/modules/_key-value-display.scss
```diff
@@ -23,6 +23,9 @@ dl.key-value, dl.key-value-doc {
         width: auto;
         word-wrap: break-word;
         word-break: break-word;
+        -webkit-hyphens: auto;
+        -ms-hyphens: auto;
+        hyphens: auto;
```
forresttanaka: Do we need the browser prefixes? Seems like all relevant browsers support the `hyphens` property. https://caniuse.com/#search=hyphen

forresttanaka: Not really requesting changes per se; just _if_ the browser prefixes aren’t needed, seems like you could delete them.

zoldello: Good point. I used caniuse when working on this problem. In order to support Safari, "-webkit-hypens" is need. IE 11 supports hypens only with "-ms-hypens".  Mozilla, Edge and Chrome edge versions work with the standard "hypens". I got this information on caniuse by hovering over each box with a gray dash with a yellow background (far right on certain boxes)  reading what non-standard means is needed to enable support hypens.

PR #2271, opened 2018-05-30
ENCD-3306-fix-line-in-table

zoldello:


zoldello:
@hitz Good observation. The CSS class: pair-bottom is still used. That is why I left it. The name did not seem fitting for non-paired items so I made group-paired class. This can be re-addressed in a future refactor if this will appear to be confusing.

PR #2272, opened 2018-05-31
ENCD-3695-upgrade-to-boto3

keenangraham:
Uses boto3 for file creation/download and moto for testing. Relaxed some versions in versions.cfg so moto would install. Will have to pin down again


src/encoded/tests/test_file_download.py
```diff
@@ -0,0 +1,79 @@
...
+    }
+    return item
+
+
+@mock_sts
+def test_external_creds():
+    from encoded.types.file import external_creds
+    creds = external_creds('mock_bucket', 'mock_object', 'mock_name')
+    assert 'upload_credentials' in creds
+    assert creds['bucket'] == 'mock_bucket'
```
hitz: why only check 'bucket' property here?

keenangraham: No reason. Can add more.

keenangraham: Added.

PR #2273, opened 2018-05-31
ENCD-4048-update-publication-data-raw-file-table

zoldello:
Remove isogenic/anisogenic column and a supporting class dedicated for it


zoldello:
remove remnants of SOME biological_replicates still being used when they ought not to (note, not all.)

PR #2276, opened 2018-06-04
ENCD-4057-properly-format-data-on-biosample-page

zoldello:


src/encoded/static/libs/measurement.js
```diff
@@ -0,0 +1,28 @@
...
+            return '0';
+        }
+
+        const measurement = `${magnitudeAsFloat} ${unit}`.trim();
+
+        return magnitudeAsFloat === 1
+          ? measurement
+          : `${measurement}s`;
+    }
+}
```
forresttanaka: What’s the advantage of a class with a single static method instead of a function without a class?

src/encoded/static/components/biosample.js
```diff
@@ -34,9 +35,10 @@ class BiosampleComponent extends React.Component {
         // Build the text of the synchronization string
         let synchText;
         if (context.synchronization) {
+            const synchronizationTime = Measurement.format(context.post_synchronization_time, context.post_synchronization_time_units);
             synchText = context.synchronization +
                 (context.post_synchronization_time ?
-                    ` + ${context.post_synchronization_time}${context.post_synchronization_time_units ? ` ${context.post_synchronization_time_units}` : ''}`
+                    `${synchronizationTime}`
```
forresttanaka: Don’t need the string template; just `{synchronizationTime}` would work.

src/encoded/static/libs/measurement.js
zoldello: Measurement seems like functionality we will expand on. That is why I made it a class. I made it a static because it is a utility, so does not depend on this. You got a valid point if you think it violates YAGNI and should be a function or utility-object with just a collection of function as these use less memory.

src/encoded/static/components/biosample.js
zoldello: Good catch.

src/encoded/static/libs/measurement.js
forresttanaka: I’m not going to gate approval on this; I just thought I would have difficulty explaining it to someone else. If you have a vision for how it might be expanded on in the future, I’ve no problem with it. I suppose even if it eventually got more static functions and no methods, there’s a name-space aspect to it (“This has to do with Measurement.”).

zoldello:
I can make this a function if you like. Still, do you mind if I stop by and
talk about this? Anyone can always add methods to it without breaking
functionality (and it is a common pattern). The namespace aspect is
probably the best thing about it I believe. Maybe I am understanding
something.

forresttanaka:
I’m WFH today, and most Tuesdays. It’s really up to you. If you want to change it, that’s fine. If you want to keep it, that’s fine too as far as I’m concerned. I just wasn’t sure of the reason, and I see you have reasons, so I’m happy with that. I’ll approve either way after the string template change.

zoldello:
Mind if I call you? I have no problem changing it to a function or object
literal (and your approach is better on resources). Still, I think it is a
good idea to share a bit of my thought-process because you saying it is not
clear is a sign maybe I am doing something wrong and do not know.... I
wanna fix that :)

zoldello:
Nevermind. I re-looked at other util functions and I "get" what you mean. I
will update the code.

zoldello:
Done.

src/encoded/static/components/biosample.js
```diff
@@ -34,9 +35,11 @@ class BiosampleComponent extends React.Component {
         // Build the text of the synchronization string
         let synchText;
         if (context.synchronization) {
+            const synchronizationTime = formatMeasurement(context.post_synchronization_time, context.post_synchronization_time_units);
+
             synchText = context.synchronization +
                 (context.post_synchronization_time ?
-                    ` + ${context.post_synchronization_time}${context.post_synchronization_time_units ? ` ${context.post_synchronization_time_units}` : ''}`
+                    synchronizationTime
```
forresttanaka: Doesn't this just print “synchronizationTime” or am I crazy? Doesn’t it have to be `{synchronizationTime}`? Or am I crazy?

src/encoded/static/libs/formatMeasurement.js
```diff
@@ -0,0 +1,26 @@
...
+    if (magnitudeAsFloat === 0) {
+        return '0';
+    }
+
+    const measurement = `${magnitudeAsFloat} ${unit || ''}`.trim();
+
+    return magnitudeAsFloat === 1
+          ? measurement
+          : `${measurement}s`;
+}
```
forresttanaka: AirBnB prefers the:

```
const formatMeasurement = function formatMeasurement() {…};
export default formatMeasurement;
```

…syntax (https://github.com/airbnb/javascript#functions). But two things about that:

1. The eslint editor plugin with AirBnB extension doesn’t call this out (probably their bug).
2. This conflicts with Redux which prefers:

```
const formatMeasurement = (magnitude, unit) => {
};
export default formatMeasurement;
```

I prefer the Redux preference lately, but I think for now this is all programmer’s discretion. I do see the reasoning for AirBnB’s opinion, and I might go back to that. I have no issue with inconsistencies with this, personally.

src/encoded/static/components/biosample.js
zoldello: It is a string not in JSX html, so it is not needed. However, I update it because the '+' was not being printed as in prod.

forresttanaka: Oh, you’re right. I _am_ crazy.

zoldello: I updated the formatMeasurement structure. The Airbnb method in my experience works if you need "this" function context to be the parent function. Otherwise arrow function is better. I went with Redux. I did not go that route initially because other utils used "function" rather than arrow.

src/encoded/static/libs/formatMeasurement.js
```diff
@@ -0,0 +1,28 @@
+
+/**
+* Takes a magnitude and unit (singular form) and returns the measurement
+* with the proper unit in singular or plural, depending on magnitude.
+* Invalid measurements are turned as- Unknown
+* @param {number} magnitude - Measured value
```
forresttanaka: “are returned as”?

src/encoded/static/libs/formatMeasurement.js
```diff
@@ -0,0 +1,28 @@
...
+    // parseFloat used because parseInt only keeps integer
+    const magnitudeAsFloat = parseFloat(magnitude);
+
+    if (Number.isNaN(magnitudeAsFloat)) {
+        return 'Unknown';
+    }
+
+    if (magnitudeAsFloat === 0) {
+        return '0';
+    }
```
forresttanaka: So a 0 magnitude should never show any units? I didn’t see that in the ticket and just wanted to be sure.

src/encoded/static/libs/formatMeasurement.js
```diff
@@ -0,0 +1,28 @@
...
+
+    const measurement = `${magnitudeAsFloat} ${unit || ''}`.trim();
+
```
forresttanaka: Seems odd to me to add a space deliberately, then trim it if not needed. Couldn’t it be ``${unit ? `${unit}` : ''}``? I doubt there’s any real performance difference, but it just seems odd.

forresttanaka:
Wait, why is it saying two of my comments are out of date, and the code really is different from what GitHub showed me minutes ago, yet there‘s no sign Phil pushed anything in the past few minutes. The comments still apply, but that’s really weird.

hitz:
@forresttanaka maybe you were looking at a stale view, he only pushed last night?

forresttanaka:
@hitz I don’t know. I only brought up the code review view maybe half an hour ago. That would imply GitHub can provide stale views.

zoldello:
@forresttanaka What is a sample of a stale comment you see? Do you need me to hold off updating the param-comment?

forresttanaka:
@zoldello I think the comments still apply though the code has changed.

zoldello:
@forresttanaka How many comments did you make? I see only one (to specify return value in comment.) If that is the only one, then I updated the code.

src/encoded/static/libs/formatMeasurement.js
```diff
@@ -0,0 +1,29 @@
...
+/**
+* Takes a magnitude and unit (singular form) and returns the measurement
+* with the proper unit in singular or plural, depending on magnitude.
+* Invalid measurements are turned as- Unknown
+* @param {string} magnitude - Measured value, range or unknown
```
forresttanaka: Actually, I was just not sure what “are turned as- Unknown” means.

src/encoded/static/libs/formatMeasurement.js
```diff
@@ -0,0 +1,29 @@
...
+        return 'Unknown';
+    }
+
+    // parseFloat used because parseInt only keeps integer
+    // remember, in JavaScript, 1 === 1.0 returns true
+    const magnitudeValue = parseFloat(magnitude);
+    const isMagnitudeNumeric = !Number.isNaN(magnitudeValue);
+    const isUnitSingular = !isMagnitudeNumeric || magnitudeValue === 1;
+
+    const measurement = `${magnitude} ${unit || ''}`.trim();
```
forresttanaka: Seems odd to me to add a space deliberately, then trim it if not needed. Couldn’t it be ${unit ? `${unit}` : ''}? I doubt there’s any real performance difference, but it just seems odd. I had one more outdated comment, but I see that one really doesn’t apply anymore.

zoldello: I added trim in the slim change there is a extra space before or after magnitude. The computational cost seemed low, so I tossed it in. I can take out trim if you insist. ${unit || '' and ${unit ? ` ${unit}` : ''} are equivalent. I selected the former because it is more brief. I can use what you have (${unit ? ` ${unit}` : ''}) if that is clearer.

PR #2277, opened 2018-06-05
ENCD-4055-removed-some-styling-on-raw-data

zoldello:


src/encoded/static/components/filegallery.js
```diff
@@ -720,17 +720,17 @@ class RawFileTable extends React.Component {
                                                     {groupFiles[0].replicate && groupFiles[0].replicate.library ? <span>{groupFiles[0].replicate.library.accession}</span> : <i>N/A</i>}
                                                 </td>
                                             : null}
-                                            <td className={groupClass}>
+                                            <td className={groupBottom}>
```
caseylitton: As discussed.   groupBottom and groupClass variable name is a refactor and not needed for feature pr.

PR #2281, opened 2018-06-05
ENCD-4064-release-endpoint

keenangraham:
Initial release/unrelease functionality implemented only for files, along with corresponding private/public view permission in s3. Submitting PR for preliminary code review not for merging.


src/encoded/tests/testing_views.py
```diff
@@ -9,6 +9,7 @@
 )
 from ..types.base import paths_filtered_by_status
 from snovault.attachment import ItemWithAttachment
+from encoded.types.base import Item as EncodedItem
```
hitz: Why did you rename this Class EncodedItem?  Do you also load Snovault.Item?

src/encoded/types/base.py
```diff
@@ -63,6 +67,11 @@ def _award_viewing_group(award_uuid, root):
     (Allow, 'group.submitter', ['add']),
 ]
 
+# New status must contain current status in list to be valid transition.
+STATUS_TRANSITION_TABLE = {
+    'released': ['in progress'],
```
hitz: So this is "TO": [ "FROM1", "FROM2"...]??

src/encoded/types/file.py
```diff
@@ -378,6 +382,59 @@ def create(cls, registry, uuid, properties, sheets=None):
...
+            external['key']
+        ).put(ACL='private')
+
+    def set_status(self, new_status, request, parent=True):
+        properties = self.upgrade_properties()
+        status = properties.get('status')
+        # Is valid transition?
+        if status not in STATUS_TRANSITION_TABLE[new_status]:
+            # Raise failure if this is primary object.
+            if parent:
```
hitz: So if parent=False there is no error even if the transition isn't allowed?  This is kind of confusing.  I suspect this is just an un-implemented feature.

hitz: or it raises and error otherwise silently does nothing?

src/encoded/types/file.py
```diff
@@ -378,6 +382,59 @@ def create(cls, registry, uuid, properties, sheets=None):
...
+            return external
+        else:
+            raise HTTPNotFound()
+
+    def set_public_s3(self):
+        external = self._get_external_sheet()
+        boto3.resource('s3').ObjectAcl(
+            external['bucket'],
+            external['key']
+        ).put(ACL='public-read')
```
hitz: where are 'public-read' and 'private' defined?  Are these base-level named ACLs in S3?

keenangraham: Yes, these are ACLs that Amazon has defined.

src/encoded/types/file.py
keenangraham: Unimplemented. The idea is that if you are trying to change status of primary object it should fail if it's not a valid a transition. However if we add code to consider child objects after considering the primary object (e.g. releasing files in an experiment after releasing the experiment) then it shouldn't necessarily fail if one of the child objects isn't a valid transition.

src/encoded/types/base.py
keenangraham: Yes, if you are trying to go to released you must be in progress.

src/encoded/tests/testing_views.py
keenangraham: Yeah, this is already loaded above for the other fixtures:
```
from snovault import (
    Item,
    calculated_property,
    collection,
)
```

Also we need to specifically use the encoded Item here to get it to work.

PR #2284, opened 2018-06-06
Kcl/remove data master node

caseylitton:
Testing removing master data node from es cluster


caseylitton:
replaced by ENCD-4072 Remove es master from deployment #2302

PR #2289, opened 2018-06-09
ENCD-4077 Cast deploy volmne size to int

caseylitton:
When deploying regulome release 1 Yunhai found a bug. We usually do not set the volume size on deploy. The volmne size is a string in the deploy.py but AWS only accepts int.


hitz:
Not sure why this branch isn't being analyzed by Codacy.

caseylitton:
Closed since this is built on regulome.   Replaced with 

ENCD-4077 Fix deploy volume size type #2304 

for encoded merge.

PR #2296, opened 2018-06-12
ENCD-4020 Allowing specification of Sets as controls

gabdank:


src/encoded/audit/experiment.py
```diff
@@ -2436,16 +2436,28 @@ def audit_experiment_control(value, system, excluded_types):
...
+                        control.get('biosample_term_name'),
+                        value['biosample_term_name']
+                      )
             yield AuditFailure('inconsistent control', detail, level='ERROR')
     return
 
 
+def is_matching_biosample_control(dataset, biosample_term_id):
+    if (dataset['@type'][0] == 'Experiment'):
+        return dataset.get('biosample_term_id') == value.get('biosample_term_id')
```
keenangraham: You are calling this function on line 2439: `is_matching_biosample_control(control, value.get('biosample_term_id'))` and inside the function returning `dataset.get('biosample_term_id') == value.get('biosample_term_id')` Am I missing something? Where is the value variable coming from in the function?

src/encoded/audit/experiment.py
```diff
@@ -2436,16 +2436,28 @@ def audit_experiment_control(value, system, excluded_types):
...
+                        control['@id'],
+                        control.get('biosample_term_name'),
+                        value['biosample_term_name']
+                      )
             yield AuditFailure('inconsistent control', detail, level='ERROR')
     return
 
 
+def is_matching_biosample_control(dataset, biosample_term_id):
+    if (dataset['@type'][0] == 'Experiment'):
```
keenangraham: Unnecessary round brackets.

src/encoded/audit/experiment.py
```diff
@@ -2436,16 +2436,28 @@ def audit_experiment_control(value, system, excluded_types):
...
+        return dataset.get('biosample_term_id') == value.get('biosample_term_id')
+    else:
+        for term in dataset.get('biosample_term_id'):
+            if term != biosample_term_id:
```
keenangraham: Something like 
```
if any([term != biosample_term_id for term in dataset.get('biosample_term_id')]):
    return False
```

src/encoded/audit/experiment.py
```diff
@@ -2534,11 +2547,23 @@ def audit_experiment_ChIP_control(value, system, files_structure):
             # The binding group agreed that ChIP-seqs all should have an input control.
             detail = 'Experiment {} is ChIP-seq and requires at least one input control, as agreed upon by the binding group. {} is not an input control'.format(
                 value['@id'],
-                control['@id'])
+                control_dataset['@id'])
             yield AuditFailure('missing input control', detail, level='NOT_COMPLIANT')
     return
 
 
+def is_control_dataset(dataset):
```
keenangraham: One suggestion is to add some unit tests for your added functions `is_matching_biosample_control` and `is_control_dataset`. That way I would get a better understanding of what these are supposed to do and there would be some concrete use cases that you know work as expected.

src/encoded/audit/experiment.py
```diff
@@ -2534,11 +2547,23 @@ def audit_experiment_ChIP_control(value, system, files_structure):
...
+    if 'target' not in dataset:
+        return False
+    if (dataset['@type'][0] == 'Experiment'):
```
keenangraham: Unnecessary round brackets.

keenangraham:
Also I'm assuming there are already passing tests for this particular audit?

src/encoded/audit/experiment.py
gabdank: fixed

gabdank:
@keenangraham please review the unit tests I have added

src/encoded/tests/test_audit_experiment.py
```diff
@@ -2915,3 +2924,43 @@ def test_audit_experiment_with_biosample_not_missing_nih_consent(testapp, experi
         ]
     )
 
+
+def test_is_matching_biosample_control(testapp, biosample, ctrl_experiment, treatment_time_series):
+    from encoded.audit.experiment import is_matching_biosample_control
+    exp = testapp.get(ctrl_experiment['@id'] + '@@index-data')
```
keenangraham: Is there a reason for using the `@@index-data` view instead of `@@embedded` view? Didn't realized `@@index-data` existed but neat!

PR #2297, opened 2018-06-13
ENCD-4048-rep-num-in-pub-n-exp

zoldello:


src/encoded/static/components/filegallery.js
```diff
@@ -388,6 +415,55 @@ FileTable.refTableColumns = {
...
+            <i className="icon icon-circle status-indicator__icon" />
+            <div className="status-indicator__label">{status}</div>
+        </div>
+    );
+};
+
+FileStatusLabel.propTypes = {
+    file: PropTypes.object.isRequired, // File whose status we're displaying
+};
+
```
forresttanaka: Where is FileStatusLabel used? I don’t see it called from anywhere. eslint also doesn’t give an error for me, so that’s kind of odd.

zoldello: @forresttanaka I am still digging into this. FileStatusLabel was removed and replaced with a Status controller. I am digging into this.

zoldello:
@forresttanaka I removed FileStatusLabel. You removed it in a PR for ENCD-3629 and may have been brought back during one of the many merge conflicts that afflicted this file.

zoldello:
This PR is closed. It is overridden by: https://github.com/ENCODE-DCC/encoded/pull/2311

PR #2307, opened 2018-06-19
ENCD-3997 Create Gene object

yunhailuo:
### Demo machine:
TBD

### Comments
I haven't put any gene on it. I'll work on the genes for current targets. I'm not sure when I should load genes, now or after code review?

### Developer notes
* ENCD-3997
  - Add gene.json to schemas; adopted from target.json
  - Add gene.py totypes;adopted from target.py
  - Add gene to the Order array in loadxl.py
  - Add sample data to test gene object; adopted from target.json
  - Add changelog for the gene object
* ENCD-3998
  - Embed and linkTo gene in target
  - rev link to and linkFrom targets in gene
  - Update target test data with gene embedded
  - Update target changelog

### DIFF
--- a/src/encoded/schemas/changelogs/gene.md
+++ b/src/encoded/schemas/changelogs/gene.md
@@ -1,0 +1,24 @@
+## Changelog for gene.json
+
+### Schema version 1
+
+* The Gene object is isolated from the Target object which will hold metadata from Target and replace some properties of Target. It is initially created by adopting the schema and type of target object. It has seven essential properties:
+  - geneid: NCBI Entrez GeneID.
+  - ncbi_entrez_status: the status of corresponding record in NCBI Entrez Gene database. One of "live", "secondary", "discontinued". For details, see https://www.ncbi.nlm.nih.gov/data_specs/schema/NCBI_Entrezgene.mod.xsd
+  - symbol: gene symbol approved by official nomenclature such as HGNC, MGI, FlyBase, WormBase.
+  - name: the full gene name provided by NCBI Entrez Gene database. Since all genes in ENCODE portal are approved by official nomenclature, this name will be preferably name approved by the official nomenclature.
+  - synonyms: alternative symbols/names referring to the gene.
+  - dbxrefs: external resources related to the gene. As for now, only the following databases are allowed:
+    * HGNC
+    * MGI
+    * FlyBase
+    * WormBase
+    * ENSEMBL
+    * MIM
+    * UniProtKB
+    * Vega
+    * miRBase
+    * IMGT/GENE-DB
+  - go_annotations: a list of gene ontology annotations of the gene, including all three aspects: Biological Process, Cellular Component, and Molecular Function. Each ontology annotation should include GO ID, ontology term name, annotation evidence code and the aspect of ontology (P: biological process, F: molecular function or C: cellular component).
+
+* Genes on the ENCODE portal are imported from and synced with NCBI Entrez Gene database. Only a subset of genes, which have Entrez GeneID as well as ID and symbol from official nomenclature, are used and maintained here. New genes to be added should comply with this requirement.
--- a/src/encoded/schemas/gene.json
+++ b/src/encoded/schemas/gene.json
@@ -1,0 +1,165 @@
+{
+    "title": "Gene",
+    "description": "Schema for submitting a gene.",
+    "id": "/profiles/gene.json",
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "type": "object",
+    "required": ["geneid", "symbol", "dbxrefs", "organism"],
+    "identifyingProperties": ["uuid", "geneid"],
+    "additionalProperties": false,
+    "mixinProperties": [
+        { "$ref": "mixins.json#/schema_version" },
+        { "$ref": "mixins.json#/uuid" },
+        { "$ref": "mixins.json#/notes" },
+        { "$ref": "mixins.json#/standard_status" }
+    ],
+    "properties": {
+        "schema_version": {
+            "default": "1"
+        },
+        "geneid": {
+            "title": "NCBI Entrez GeneID",
+            "description": "NCBI Entrez GeneID of official nomenclature approved gene.",
+            "comment": "An ENCODE gene (object) needs to have an Entrez GeneID and needs to be approved by (i.e. have an ID and gene symbol from) offical nomenclature such as HGNC, MGI, FlyBase, WormBase.",
+            "type": "string",
+            "permission": "import_items",
+            "uniqueKey": true
+        },
+        "ncbi_entrez_status": {
+          "title": "NCBI Entrez Gene Status",
+          "type": "string",
+          "default": "live",
+          "enum" : [
+              "live",
+              "secondary",
+              "discontinued"
+          ],
+          "permission": "import_items"
+        },
+        "symbol": {
+            "title": "Gene symbol",
+            "description": "Gene symbol approved by the official nomenclature.",
+            "type": "string",
+            "format": "gene_name",
+            "permission": "import_items"
+        },
+        "name": {
+            "title": "Gene name",
+            "description": "The full gene name preferably approved by the official nomenclature.",
+            "type": "string",
+            "format": "gene_name",
+            "permission": "import_items"
+        },
+        "synonyms": {
+            "title": "Synonyms",
+            "description": "Alternative symbols that have been used to refer to the gene.",
+            "comment": "The list of synonyms is maintained using information from NCBI Entrez. Synonyms not included in NCBI gene_info can be added but are not actively maintained.",
+            "type": "array",
+            "uniqueItems": true,
+            "items": {
+                "title": "Synonym",
+                "description": "An alternative symbol/name referring to the gene.",
+                "type":  "string",
+                "format": "gene_name"
+            },
+            "permission": "import_items"
+        },
+        "dbxrefs": {
+            "title": "External identifiers",
+            "description": "Unique identifiers from external resources.",
+            "comment": "This list should have at least an ID from the official nomenclature. The list of other dbxrefs is maintained using information from NCBI Entrez. Identifiers not included in NCBI gene_info can be added but are not actively maintained.",
+	          "type": "array",
+            "default": [],
+            "uniqueItems": true,
+            "items": {
+                "title": "External identifier",
+                "description": "A unique identifier from external resource (e.g. HGNC, MGI, FlyBase, WormBase, ENSEMBL, MIM, UniProtKB, Vega, miRBase).",
+                "comment": "Submit as database_name:id. See changelog for potential databases.",
+                "type":  "string",
+                "pattern": "^((HGNC:HGNC:\\d+)|(MGI:MGI:\\d+)|(FlyBase:FBgn\\d+)|(WormBase:WBGene\\d+)|(ENSEMBL:ENS[A-Z]*G\\d+)|(MIM:\\d+)|(UniProtKB:[0-9A-Z]+)|(Vega:OTT[A-Z]+G\\d+)|(miRBase:MI\\d+)|(IMGT/GENE-DB:[0-9A-Za-z/\\(\\)\\-]+))$"
+            },
+            "permission": "import_items"
+        },
+        "go_annotations": {
+            "title": "Gene Ontology annotations",
+            "descrption": "Gene ontology annotations of the gene, including all three aspects: Biological Process, Cellular Component, and Molecular Function.",
+            "comment": "Data imported from the Gene Ontology Consortium.",
+            "type": "array",
+            "uniqueItems": true,
+            "items": {
+                "title": "Gene Ontology annotation",
+                "description": "A gene ontology annotation of the gene function.",
+                "type": "object",
+                "required": ["go_id", "go_evidence_code", "go_aspect"],
+                "properties": {
+                    "go_id": {
+                        "title": "GO term identifier",
+                        "description": "GO ID of the annotation in the format of \"GO:xxxxxxx\"",
+                        "type": "string",
+                        "pattern": "^(GO:\\d+)$"
+                    },
+                    "go_name": {
+                        "title": "GO term name",
+                        "description": "GO term name of the annotation.",
+                        "type": "string"
+                    },
+                    "go_evidence_code": {
+                        "title": "GO evidence code",
+                        "description": "An evidence code indicating how this GO annotation is supported.",
+                        "type": "string",
+                        "pattern": "^([A-Z]{2,3})$"
+                    },
+                    "go_aspect": {
+                        "title": "GO annotation category",
+                        "description": "The namespace or ontology to which the GO annotation belongs.",
+                        "comment": "One of P (biological process), F (molecular function) or C (cellular component).",
+                        "type": "string",
+                        "enum": ["P", "F", "C"]
+                    }
+                }
+            },
+            "permission": "import_items"
+        },
+        "organism": {
+            "title": "Organism",
+            "description": "Organism of the gene.",
+            "comment": "See organism.json for available identifiers.",
+            "type": "string",
+            "linkTo": "Organism",
+            "permission": "import_items"
+        }
+    },
+    "facets": {
+        "organism.scientific_name": {
+            "title": "Organism"
+        }
+    },
+    "columns": {
+        "geneid": {
+            "title": "NCBI Entrez GeneID"
+        },
+        "symbol": {
+            "title": "Gene symbol"
+        },
+        "synonyms": {
+            "title": "Synonyms"
+        },
+        "dbxrefs": {
+            "title": "External resources"
+        },
+        "organism.scientific_name": {
+            "title": "Species"
+        }
+    },
+    "boost_values": {
+        "geneid": 1.0,
+        "symbol": 1.0,
+        "synonyms": 1.0,
+        "dbxrefs": 1.0,
+        "organism.name": 1.0,
+        "organism.scientific_name": 1.0,
+        "organism.taxon_id": 1.0
+    },
+    "changelog": "/profiles/changelogs/gene.md"
+
+}
--- a/src/encoded/schemas/gene.json
+++ b/src/encoded/schemas/gene.json
@@ -85,12 +85,13 @@
             "descrption": "Gene ontology annotations of the gene, including all three aspects: Biological Process, Cellular Component, and Molecular Function.",
             "comment": "Data imported from the Gene Ontology Consortium.",
             "type": "array",
+            "default": [],
             "uniqueItems": true,
             "items": {
                 "title": "Gene Ontology annotation",
                 "description": "A gene ontology annotation of the gene function.",
                 "type": "object",
-                "required": ["go_id", "go_evidence_code", "go_aspect"],
+                "required": ["go_id", "go_name", "go_evidence_code", "go_aspect"],
                 "properties": {
                     "go_id": {
                         "title": "GO term identifier",
--- a/src/encoded/schemas/gene.json
+++ b/src/encoded/schemas/gene.json
@@ -4,7 +4,7 @@
     "id": "/profiles/gene.json",
     "$schema": "http://json-schema.org/draft-04/schema#",
     "type": "object",
-    "required": ["geneid", "symbol", "dbxrefs", "organism"],
+    "required": ["geneid", "ncbi_entrez_status", "symbol", "dbxrefs", "organism"],
     "identifyingProperties": ["uuid", "geneid"],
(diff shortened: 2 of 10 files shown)


src/encoded/schemas/target.json
```diff
@@ -16,7 +16,7 @@
     ],
     "properties": {
         "schema_version": {
-            "default": "8"
+            "default": "9"
```
hitz: We need to remove organism from target and change all references of target.organism to target.gene.organism.

hitz: target.gene_name should be target.gene.symbol.       I am not sure how we reconcile target.label and target.gene -- but I guess this was always an issue (and one of the reasons we don't allow non-admins to submit)

src/encoded/schemas/target.json
```diff
@@ -99,12 +107,13 @@
         }
     },
     "boost_values": {
-        "gene_name": 1.0, 
-        "label": 1.0, 
-        "dbxref": 1.0, 
-        "aliases": 1.0, 
-        "organism.name": 1.0,  
-        "organism.scientific_name": 1.0, 
+        "gene.symbol": 1.0,
```
hitz: add "gene.symbol" to facets - this would let us group H3 modifications together.

src/encoded/tests/data/inserts/gene.json
```diff
@@ -0,0 +1,247 @@
...
+            "FlyBase:abrupt",
+            "FlyBase:FBgn0264442",
+            "FlyBase:ab-RB",
+            "FlyBase:CG43860-RD"
+        ],
+        "symbol": "abrupt",
+        "organism": "/organisms/dmelanogaster/",
+        "status": "released"
+    }
+]
```
hitz: I wonder if we should add some rna genes for testing, like mIR precursors etc.

src/encoded/schemas/gene.json
```diff
@@ -0,0 +1,73 @@
...
+    "additionalProperties": false,
+    "mixinProperties": [
+        { "$ref": "mixins.json#/schema_version" },
+        { "$ref": "mixins.json#/uuid" },
+        { "$ref": "mixins.json#/notes" },
+        { "$ref": "mixins.json#/standard_status" }
+    ],
+    "properties": {
+        "schema_version": {
+            "default": "1"
```
hitz: I guess we are going to add something like coordinates: { 'GRCh38': { chromosome: "X", start: "44949", end: "65990"}, "hg19": {P...} for version 2?

src/encoded/types/gene.py
```diff
@@ -0,0 +1,81 @@
...
+    find_root,
+)
+
+
+@collection(
+    name='genes',
+    unique_key='gene:name',
+    properties={
+        'title': 'Genes',
+        'description': 'Listing of ENCODE genes',
```
hitz: "ENCODE" genes may be not quite right...  Maybe this should just be "Listing of genes"

src/encoded/types/gene.py
```diff
@@ -0,0 +1,81 @@
...
+    @calculated_property(schema={
+        "title": "Title",
+        "type": "string",
+    })
+    def title(self, request, organism, symbol):
+        organism_props = request.embed(organism, '@@object')
+        return u'{} ({})'.format(symbol, organism_props['scientific_name'])
+
+    @calculated_property(schema={
+        "description": "Targets targeting this gene.",
```
hitz: I feel like this is a little jargon-y.  But Antibody targets isn't 100% accurate either, is it?

src/encoded/schemas/target.json
yunhailuo: As mentioned last week, this will be step 2 & 3 like this:
1. Get this gene schema and target schema in this or one release.
2. Post thousands of gene data/object with the following two copies: "target.gene_name" -> "gene.symbol" and "target.dbxref" -> "gene.dbxref"
3. Patch targets with gene embedding.
4. In a next release, replace "gene_name" and "dbxref" with "gene.symbol" and "gene.dbxref" target schema. Pop/delete "gene_name" and "dbxref" from target data/object using upgrade.

src/encoded/schemas/target.json
yunhailuo: Added in ENCD-3998-modification-1 commit.

src/encoded/tests/data/inserts/gene.json
yunhailuo: Added 29 non-coding genes in ENCD-3997-modification-1 commit. They cover different categories. It is helpful since I need to update schema (regex for dbxref) accordingly. Thanks.

src/encoded/schemas/gene.json
yunhailuo: I'd say v2 since I'm not sure how this will be used. Here is my current thought on schema:
```
"coordinates": {
    "title": "Coordinates",
    "description": "Genome positions for different assemblies of this gene.",
    "type": "array",
    "default": [],
    "uniqueItems": true,
    "items": {
        "title": "Coordinate",
        "description": "Genome positions for one assembly of this gene.",
        "type":  "object",
        "additionalProperties": false,
        "required":["assembly", "chromosome", "start", "end"],
        "properties": {
            "assembly": {
                "title": "Mapping assembly",
                "description": "The genome assembly to which modification coordinates relate. e.g. GRCh38.",
                "type": "string",
                "enum": [
                    "ce10",
                    "ce11",
                    "dm3",
                    "dm6",
                    "GRCh38",
                    "hg19",
                    "mm10",
                    "mm9"
                    ]
            },
            "chromosome": {
                "title": "Chromosome",
                "description": "The number (or letter) designation for the target chromosome. e.g. 1, 2L, or X",
                "type": "string"
            },
            "start": {
                "title": "Start",
                "description": "The starting coordinate.",
                "type": "integer"
            },
            "end": {
                "title": "End",
                "description": "The ending coordinate (possibly equal to the starting coordinate for one base.)",
                "type": "integer"
            }
        }
    }
}
```

src/encoded/types/gene.py
yunhailuo: Corrected in ENCD-3997-modification-1 commit.

src/encoded/types/gene.py
yunhailuo: Corrected as "List of associated targets." in ENCD-3997-modification-1 commit.

yunhailuo:
I also added a "synonyms" property in "ENCD-3997-modification-1" commit.
If the code is OK, I'll launch a demo later.

src/encoded/schemas/target.json
hitz: Agreed, I was wrong about the order.

src/encoded/schemas/changelogs/gene.md
```diff
@@ -0,0 +1,21 @@
...
+### Schema version 1
+
+* The Gene object is isolated from the Target object which will hold metadata from Target and replace some properties of Target. It is initially created by adopting the schema and type of target object. It has six essential properties:
+  - geneid: NCBI Entrez GeneID.
+  - symbol: gene symbol approved by official nomenclature such as HGNC, MGI, FlyBase, WormBase.
+  - synonyms: Alternative symbols/names referring to the gene.
+  - dbxrefs: external resources related to the gene. As for now, only the following databases are allowed:
+    * HGNC
+    * MGI
+    * FLYBASE
```
estherc: This should be FlyBase to match what's in the namespace schema

src/encoded/schemas/changelogs/gene.md
```diff
@@ -0,0 +1,21 @@
...
+    * WormBase
+    * Ensembl
```
estherc: This should be ENSEMBL to match what's in the namespace schema

src/encoded/schemas/changelogs/target.md
```diff
@@ -3,6 +3,10 @@
 ### Schema version 8
 
 * Move status property to standard_status mixin.
+* Added "targeted_genes" property for the list of genes targeted by assay or antibody associated with the target.
```
estherc: These changes don't require an upgrade as they're new properties, so perhaps they should be listed as minor changes since schema version 7... or 8.

src/encoded/schemas/gene.json
```diff
@@ -0,0 +1,107 @@
...
+            "comment": "This list should have at least an ID from the official nomenclature. The list of other dbxrefs is maintained using information from NCBI Entrez. Identifiers not included in NCBI gene_info can be added but are not actively maintained.",
+	          "type": "array",
+            "default": [],
+            "uniqueItems": true,
+            "items": {
+                "title": "External identifier",
+                "description": "A unique identifier from external resource (e.g. HGNC, MGI, FLYBASE, WormBase, Ensembl, MIM, RefSeq, UniProtKB, Vega, miRBase).",
+                "comment": "Submit as database_name:id. See changelog for potential databases.",
+                "type":  "string",
+                "pattern": "^((HGNC:HGNC:[0-9a-zA-Z/\\-]+)|(MGI:MGI:\\d+)|(FLYBASE:[0-9a-zA-Z\\-]+)|(WormBase:[0-9a-zA-Z\\.]+)|(Ensembl:ENS[A-Z]*G\\d+)|(MIM:\\d+)|(RefSeq:[A-Z]{2}_\\d+)|(UniProtKB:[0-9A-Z]+)|(Vega:OTT[A-Z]+G\\d+)|(miRBase:MI\\d+))$"
```
estherc: The pattern for FlyBase and Ensembl should be "FlyBase" and "ENSEMBL" to match what's in the namespace schema.

src/encoded/schemas/target.json
```diff
@@ -38,17 +38,39 @@
...
             "comment": "Submit only the identifier (e.g. HMFN0395 or 22809).",
             "type": "string",
             "format": "gene_name"
         },
+        "modification": {
+            "title": "Modification to the corresponding gene.",
+            "description": "A textual description of modification(s) made to the corresponding gene in this targets (e.g. epitope tag(s), post-translational modifications)",
+            "comment": "This property will be used to calculate the proper label of the target. See computed_label property for details.",
+            "type": "string",
+            "default": ""
```
estherc: We shouldn't do defaults or blank defaults if this is not a required property. I think we should think about making this import_items if It stays a string for now or make it a set of restricted enums matching the enums for GeneticModification.introduced_tags

src/encoded/schemas/target.json
```diff
@@ -38,17 +38,39 @@
...
+        },
         "label": {
-            "title": "Common name with modification",
-            "description": "Common name for the target including post-translational modifications, if any.",
-            "comment": "Submit the common name of the gene with modification (e.g. H3K4me3, eGFP-E2F1, or POLR2AphosphoS2).",
+            "title": "Preferred target label",
+            "description": "An easy to understand label preferred over  computed label.",
+            "comment": "Do not submit. DCC personnel is responsible for submission of this property. In general, the format is the common name of the gene with modification (e.g. H3K4me3, eGFP-E2F1, or POLR2AphosphoS2).",
```
estherc: This comment is not enforced by import_items. Not sure if it should be but the description doesn't match what will be enforced.

src/encoded/schemas/target.json
hitz: yes I think that is correct; we want it to default to null (no property exists), not "".

src/encoded/schemas/changelogs/target.md
yunhailuo: What is the correct way/format to list the changes as minor changes? Target is version 8 in production now.

src/encoded/schemas/target.json
hitz: it should be import-items.

yunhailuo: For this comment and the comment below, should the whole target object be admin only? I think @hitz mentioned nobody really post target objects.

estherc: Well, if you made this change then that effectively makes the whole target object admin only. We can try it and see if we get an avalanche of requests for new ones. I think it will be unlikely but we can always relax it later.

yunhailuo:
This PR is refactored so that it contains only one commit for ENCD-3997. There is new branch for ENCD-3998: [ENCD-3998-link-target-to-gene](https://github.com/ENCODE-DCC/encoded/tree/ENCD-3998-link-target-to-gene). The history has been rewrite and will cause conflicts. Sorry about that.

src/encoded/static/components/gene.js
```diff
@@ -0,0 +1,150 @@
...
+        } else if (context.organism.name === 'dmelanogaster' || context.organism.name === 'celegans') {
+            const organismPrefix = context.organism.name === 'dmelanogaster' ? 'FBgn' : 'WBGene';
+            const baseUrl = context.organism.name === 'dmelanogaster' ? globals.dbxrefPrefixMap.FlyBase : globals.dbxrefPrefixMap.WormBase;
+            geneRef = _.find(context.dbxref, ref => ref.indexOf(organismPrefix) !== -1);
+            if (geneRef) {
+                sep = geneRef.indexOf(':') + 1;
+                baseName = geneRef.substring(sep, geneRef.length);
+                geneLink = baseUrl + baseName;
+            }
+        }
```
forresttanaka: Unless I’m not understanding something, I think this whole if/else if/else if block can be eliminated by just using the `<DbxrefList>` component, and adding dependencies to the schema. The danger is that `dbxrefHref` isn’t complete; it doesn’t do any pre- or post-processing of the URL. They might not be needed in this case, but this seems like a dangerous way to go unless you really had a reason that `<DbxrefList>` wouldn’t work, in which case that should be commented here, otherwise everyone working on this code in the future will wonder why it duplicates so much work in dbxref.js. Also, the `geneLink` variable can end up undefined with a couple different paths through this block, and you’ll have an undefined `<a>` href later in this component if that happens.

src/encoded/static/components/gene.js
```diff
@@ -0,0 +1,150 @@
...
+                                <dt>Official gene name</dt>
+                                <dd>{context.name}</dd>
+                            </div>
+                        : null}
+
+                        {context.synonyms ?
+                          <div data-test="synonyms">
+                              <dt>Synonyms</dt>
+                              <dd><ul>
+                                  {context.synonyms.map((synonym, i) =>
```
forresttanaka: You can use `synonym` as the `<li>` key and remove `i` which would be preferable, though it’s not always possible to avoid using `i` as the key. If it’s possible two entries in the `synonyms` array could be the same, then you’d need to keep using `i`. The same thing applies to the `go_bp_ids` and `go_mf_id` arrays below.

src/encoded/static/components/search.js
```diff
@@ -508,6 +509,53 @@ const Target = auditDecor(TargetComponent);
...
+    auditDetail: PropTypes.func.isRequired, // Audit decorator function
+};
+
+GeneComponent.contextTypes = {
+    session: PropTypes.object, // Login information from <App>
+};
+
+const Gene = auditDecor(GeneComponent);
+
+globals.listingViews.register(Gene, 'Gene');
```
forresttanaka: We’re moving towards listingViews being in the modules they represent, so in this case `GeneComponent` would be in gene.js and called `ListingComponent` if you were to be consistent with other modules (you can look at genetic_modification.js for just one example). If you prefer to leave it here, then don’t worry about it, but moving listingViews to their own modules has been the way we’ve been going, just so that search.js doesn’t keep growing and possibly needing to import parts specific to certain object types.

Also, like above, you should remove the eslint-disable comments and make this a stateless component.

yunhailuo:
Thanks for your review and comments, @forresttanaka . They are very helpful. I hope my commits addressed them all. Please feel free to make more suggestions.

src/encoded/static/components/gene.js
```diff
@@ -0,0 +1,181 @@
...
+                </div>
+            </header>
+
+            <div className="panel">
+                <dl className={itemClass}>
+                    {/* TODO link to NCBI Entrez page? */}
+                    <div data-test="gendid">
+                        <dt>Entrez GeneID</dt>
+                        <dd>
+                            <a href={dbxrefHref('GeneID', context.geneid)}>{context.geneid}</a>
```
forresttanaka: All looks good! Except I’m not sure why you need `dbxrefHref` here instead of `<DbxrefList>` which would put in the links for you? `dbxrefHref` is only for very unusual cases (Targets needed something odd, and I don’t remember why). If the issue is that `<DbxrefList>` needs an array, you can pass `dbxrefs={[context.geneid]}`, or maybe you could make `<DbxrefUrl>` external and use that instead. We just never needed that before, but no problem having it external now.

yunhailuo: I don't have a good reason. I want to make the text in display just the ID while I thought I'll need to pass "GeneID:{context.geneid}" to DbxrefList which will make the display have a "GeneID" prefix. Is that right?

forresttanaka: That’s right.

src/encoded/static/components/gene.js
```diff
@@ -77,49 +77,13 @@ const Gene = (props) => {
...
-                              <ul>
-                                  {context.go_mf_ids.map(goMFID =>
-                                      <li key={goMFID}>
-                                          <span>{goMFID}</span>
-                                      </li>
-                                  )}
-                              </ul>
-                          </dd>
-                      </div>
+                    {context.go_annotations ?
```
forresttanaka: _Might_ want to make it `context.go_annotations && context.go_annotations.length > 0` here, but it depends if you can have an empty array here or not. If you can’t, then no reason to change this.

yunhailuo: Thanks.

src/encoded/types/gene.py
```diff
@@ -0,0 +1,30 @@
...
+    name_key = "geneid"
+    embedded = ['organism']
+
+    @calculated_property(schema={
+        "title": "Title",
+        "type": "string",
+    })
+    def title(self, request, organism, symbol):
+        organism_props = request.embed(organism, '@@object')
+        return u'{} ({})'.format(symbol, organism_props['scientific_name'])
```
caseylitton: Did you consider using organism_props.get('scientific_name')?  In this case the function would return 
```
'whatever-symbol (None)' 
```
when the key doesn't exist.  The current implementation would fail if for some reason the key wasn't there.  But that could be a desired result?

estherc: This should probably never happen as the loading script should not be loading any gene objects for organisms that aren't already registered. We should probably be quite clear in the schema/docs as to which organisms are supported.

However, organism.scientific_name isn't required and it probably should be. We should address that in a separate ticket.

src/encoded/schemas/gene.json
```diff
@@ -0,0 +1,165 @@
...
+            "title": "Gene Ontology annotations",
+            "descrption": "Gene ontology annotations of the gene, including all three aspects: Biological Process, Cellular Component, and Molecular Function.",
+            "comment": "Data imported from the Gene Ontology Consortium.",
+            "type": "array",
+            "uniqueItems": true,
+            "items": {
+                "title": "Gene Ontology annotation",
+                "description": "A gene ontology annotation of the gene function.",
+                "type": "object",
+                "required": ["go_id", "go_evidence_code", "go_aspect"],
```
estherc: Just curious - why require everything but the GO term name? Is there a scenario in which the loading script will have all but this property to load?

src/encoded/schemas/gene.json
```diff
@@ -0,0 +1,165 @@
...
+                "type":  "string",
+                "pattern": "^((HGNC:HGNC:\\d+)|(MGI:MGI:\\d+)|(FlyBase:FBgn\\d+)|(WormBase:WBGene\\d+)|(ENSEMBL:ENS[A-Z]*G\\d+)|(MIM:\\d+)|(UniProtKB:[0-9A-Z]+)|(Vega:OTT[A-Z]+G\\d+)|(miRBase:MI\\d+)|(IMGT/GENE-DB:[0-9A-Za-z/\\(\\)\\-]+))$"
+            },
+            "permission": "import_items"
+        },
+        "go_annotations": {
+            "title": "Gene Ontology annotations",
+            "descrption": "Gene ontology annotations of the gene, including all three aspects: Biological Process, Cellular Component, and Molecular Function.",
+            "comment": "Data imported from the Gene Ontology Consortium.",
+            "type": "array",
```
estherc: Maybe we also need a default empty [] value for this property as it doesn't seem to show up in objects that don't have any GO terms in this property (even in the inserts that have an empty list value). That way if people are operating on these objects to pull out GO annotations, they would not get key errors.

src/encoded/schemas/gene.json
yunhailuo: Sorry, that's my negligence. I will required it. I hesitated initially because I saw it's not mandatory in GAF files. It's obviously required and I just confirmed that in OBO format.

PR #2308, opened 2018-06-19
ENCD-4093 Fix initial region indexing failure

tdreszer:
Fixes the initial region indexing bug, which turned out to be a race condition in handing off the staged uuids list.  Tested this solution on a demo, (but had to artificially open the race condition window wide to catch it).


tdreszer:
This should be 72, and can get into regulome via a future rebase on dev.

hitz:
yes, it's a fix for the region-search indexer that is in production.

PR #2310, opened 2018-06-20
ENCD-4088 Remove extra code added in 3306

zoldello:


forresttanaka:
My one question just to be sure: the PR is titled “Remove extra code…” but all I see is code added in this PR. That’s not an accident though, right?

forresttanaka:
I hear you talking to Casey so I’ll ask here: what problem does this branch fix? I thought it was for moving the bold line to the proper place, but it seems to have to do with adding columns, or maybe adding them back? Which ticket describes the original problem this branch fixes?

zoldello:
@forresttanaka I updated the PR title. 
This ticket about starting over. We in a sense reverted the code to a pre-3306 state and then added just 3306 code. So it should not be polluted with 4048 code. 

3306 adds the bold line on not just the second row. 

4048 initial check-in removed all replicate columns (entitled: isogenic number or anisogenic number on the UI.) Part of the "pre-3306 revert" is to bring back that column.

src/encoded/static/components/filegallery.js
```diff
@@ -533,6 +572,9 @@ class RawSequencingTable extends React.Component {
 
                                     return (
                                         <tr key={file['@id']}>
+                                            {i === 0 ?
+                                              <td rowSpan={groupFiles.length} className={`${bottomClass} merge-right table-raw-merged table-raw-biorep`}>{groupFiles[0].biological_replicates[0]}</td>
+                                              : null}
```
forresttanaka: We don’t (and AirBnB doesn’t) have a standard for how to indent conditional expressions, but we do need to enforce four-space indents. I’m surprised eslint doesn’t mark this.

zoldello: Maybe an .editorconfig file will help with catching editor-specific things eslint misses. I will dig into that when I get bandwidth. I pushed the 4 indent. Thanks for keeping an eye on that.

zoldello:
@forresttanaka. The "groupClass" code at line 737 and above adds the bold line. Because it is similar to dev, it does not show in the diff.

PR #2311, opened 2018-06-21
Encd 4048 rep num 06 22

zoldello:


src/encoded/static/components/filegallery.js
```diff
@@ -169,12 +170,17 @@ export class FileTable extends React.Component {
                 return 'proc';
             });
 
+            // showReplicateNumber matches with show-functionality. It has to
+            // be NOT (!)-ed to match with hide-functionality
+            FileTable.procTableColumns.biological_replicates.hide = () => !showReplicateNumber;
```
forresttanaka: Would it be better to put this into the FileTable.procTableColumns definition (around line 317) instead of on its own here, or is there a reason this property can only be set if `datasetFiles.length` isn’t zero?

src/encoded/static/components/filegallery.js
```diff
@@ -572,6 +586,9 @@ class RawSequencingTable extends React.Component {
 
                                     return (
                                         <tr key={file['@id']}>
+                                            {showReplicateNumber && i === 0 ?
+                                              <td rowSpan={groupFiles.length} className={`${bottomClass} merge-right table-raw-merged table-raw-biorep`}>{groupFiles[0].biological_replicates[0]}</td>
+                                              : null}
```
forresttanaka: We’ll need four spaces indent here instead of two.

src/encoded/static/components/filegallery.js
```diff
@@ -610,7 +627,11 @@ class RawSequencingTable extends React.Component {
 
                                 return (
                                     <tr key={file['@id']} className={rowClasses.join(' ')}>
-                                        <td className="table-raw-biorep">{file.biological_replicates && file.biological_replicates.length ? file.biological_replicates.sort((a, b) => a - b).join(', ') : 'N/A'}</td>
+                                        {
+                                          showReplicateNumber ?
+                                          <td className="table-raw-biorep">{file.biological_replicates && file.biological_replicates.length ? file.biological_replicates.sort((a, b) => a - b).join(', ') : 'N/A'}</td> :
+                                          null
```
forresttanaka: Need four spaces indent here too. Also, while I don’t think every person needs to do the same ternary indenting style as other people, I think each person should generally be consistent with themselves. You have braces on their own line here, but on the same line as the first term in other places.

src/encoded/static/components/filegallery.js
```diff
@@ -708,7 +731,9 @@ class RawFileTable extends React.Component {
 
                         {!this.state.collapsed ?
                             <tr>
-                                <th>{replicationDisplay(meta.replicationType)}</th>
+                                {showReplicateNumber ?
+                                  <th>{replicationDisplay(meta.replicationType)}</th> :
+                                  null}
```
forresttanaka: Need four spaces indent here.

src/encoded/static/components/filegallery.js
```diff
@@ -742,6 +767,11 @@ class RawFileTable extends React.Component {
                                     // Prepare for run_type display
                                     return (
                                         <tr key={file['@id']}>
+                                            {showReplicateNumber && i === 0 ?
+                                              <td rowSpan={groupFiles.length} className={`${bottomClass} merge-right table-raw-merged table-raw-biorep`}>
+                                                  {groupFiles[0].biological_replicates.length ? <span>{groupFiles[0].biological_replicates[0]}</span> : <i>N/A</i>}
+                                              </td>
+                                            : null}
```
forresttanaka: And four spaces indent here.

src/encoded/static/components/filegallery.js
```diff
@@ -778,7 +808,9 @@ class RawFileTable extends React.Component {
 
                                 return (
                                     <tr key={file['@id']} className={rowClasses.join(' ')}>
-                                        <td className="table-raw-biorep">{(file.biological_replicates && file.biological_replicates.length) ? file.biological_replicates.sort((a, b) => a - b).join(', ') : 'N/A'}</td>
+                                        {showReplicateNumber ?
+                                          <td className="table-raw-biorep">{(file.biological_replicates && file.biological_replicates.length) ? file.biological_replicates.sort((a, b) => a - b).join(', ') : 'N/A'}</td> :
+                                          null}
```
forresttanaka: And four spaces here.

src/encoded/static/components/filegallery.js
zoldello: That makes sense but is hard. FileTable.procTableColumns is an object by itself and has no knowledge of showReplicateNumber. Using "this" (function context) to specific showReplicateNumber gets hairy because "this" changes depending on if you use function or arrow function. Passing in showReplicate in a parameter is hairy too because sorttable does not accept showReplicateNumber and I am trying to do as less surgery as possible. Does this make sense?

zoldello:
@caseylitton I mistakenly added a change after Forrest code review. I reverted it.

PR #2312, opened 2018-06-21
ENCD-4087-fixed-flybase-search

zoldello:
1. Flybase changed their search url. This change reflects this update
2. Dataset and some other search criteria were removed. See the jira ticket for details if interested.


zoldello:
@caseylitton Addressed. A unit test was failing and that was fixed.

PR #2315, opened 2018-06-25
ENCD-4089-Deactivated-rep-num-in-annot-table

zoldello:


src/encoded/static/components/dataset.js
```diff
@@ -225,7 +225,7 @@ class AnnotationComponent extends React.Component {
                 </Panel>
 
                 {/* Display the file widget with the facet, graph, and tables */}
-                <FileGallery context={context} encodevers={encodevers} />
+                <FileGallery context={context} encodevers={encodevers} showReplicateNumber={false}  />
```
caseylitton: I think there are two spaces after 
```
showReplicateNumber={false}
```
The ticket is already merge ready, but I guess we should fix it?

forresttanaka: Probably should be fixed. Looks like it causes an eslint error. @zoldello Do you have eslint configured in Atom in a way that uses the encoded .eslintrc file?

caseylitton: Sounds good.  I'll mark this in PR as review requested.  I don't think we need a new QA though, right?

src/encoded/static/components/dataset.js
```diff
@@ -225,7 +225,7 @@ class AnnotationComponent extends React.Component {
                 </Panel>
 
                 {/* Display the file widget with the facet, graph, and tables */}
```
caseylitton: Please fix eslint error with the double spaces in the other comment.  Check with Forrest, but we probably don't need a new demo/qa cycle.

src/encoded/static/components/dataset.js
caseylitton: oh, it showed up in codacy.  I should look at that more

PR #2325, opened 2018-07-06
ENCD-4105 Add DGGR dbxref processing

forresttanaka:


src/encoded/static/components/__tests__/dbxref-test.js
```diff
@@ -561,6 +561,25 @@ describe('Test individual dbxref types', () => {
...
+            const context = { '@type': ['FlyDonor'] };
+            const wrapper = mount(
+                <DbxrefList context={context} dbxrefs={['DGGR:109027', 'DGGR:101084']} />
+            );
+
+            dbxLinks = wrapper.find('a');
+        });
+
+        it('has the correct links', () => {
+            expect(dbxLinks.length).toBe(2);
```
zoldello: Testing for the length and testing for the correct values of links are two separate concerns. One succeeding those not mean the other one will. My recommendation would be to break them up to two unit test.

forresttanaka: Do you have a relevant citation to support that for React Jest tests?

I see Facebook tests for any number of things in one Jest test:
https://github.com/facebook/react/blob/f762b3abb1d951b366f3c8a2354aeedb7fbc32ff/packages/react/src/__tests__/ReactChildren-test.js#L813

Even Jest examples themselves do too:
https://github.com/facebook/jest/blob/master/examples/timer/__tests__/infinite_timer_game.test.js

Doesn't mean they’re right, but I think splitting up every function in this file needs some outside justification.

zoldello: TL;DR Answer: 
It is not big deal and seems to be a React-way of thinking about unit test. I will pass this.

Long Story Answer:
Consider these-

Look briefly at this question: https://stackoverflow.com/questions/2878717/multiple-asserts-in-a-unit-test

Look at answer: https://stackoverflow.com/a/2878912/178550

Note, Roy Osherove's "The Art of Unit Testing" is an older book that was groundbreaking in pushing unit testing. It is a much nicer read in my opinion than Ken Beck's pioneering and older TDD book which covered unit test. Anyway, assert is equivalent to Enzyme's "expect". 

Essentially, each unit test should test a concept- Ideally have only one assert (expect-test) or  a few that test a concept with a few expects. I should be able to look at a test run from Jest and each failed test should match one failed thing and not many. If it tells me many, I am getting into integration test territory. In this case, if the test fails, it could be that the length is wrong OR the content is wrong. Testing one thing or concept also is in support of Uncle Bob's (Robert Martin) single responsibility principle, which is part of his SOLID principles (https://en.wikipedia.org/wiki/SOLID). I do admit, following this idea may lead to having a lot more files and methods. I also do admit too this be a draconic principle fading away like React did  with having HTML in JavaScript (used to be an automatic code-review fail; not it is standard).

src/encoded/static/components/__tests__/dbxref-test.js
```diff
@@ -561,6 +561,25 @@ describe('Test individual dbxref types', () => {
         });
     });
 
+    describe('Test DGGR', () => {
```
caseylitton: I spent some time yesterday looking through all the tests in the file.  It would take some effort to explain and maybe you're already aware, but this new DGGR test only effectively tests if the new DGGR key exists in the mapping.  Everything else has already been tested almost 50 times.

forresttanaka: The goal is to make sure that each prefix/suffix combination generates the correct URL. The fact that DGGR currently uses a similar style mapping to others isn’t theoretically known from the point of view of the test writer. They only know what should go in and what should come out for each prefix. Some prefixes go through a more complicated process than others.

caseylitton: Sorry, this got long again!  These are the kind of things we could discuss if we had regular code review meetings.

I think the goal is great, but we are doing a lot of work to accomplish it.  My concern is that we've established a less than ideal testing pattern that doesn't cover all the code and seems more like busy work.  To be clear, I'm not suggesting to update all the tests right now because that would be out of the scope of the PR and probably isn't a high priority.  I took the time to look at them and believe there is some value to be gained from sharing my perspective.  Possibly that value is that javascript/react testing is vastly different than what I've experienced in python.

Consider that for every item in dbxrefPrefixMap we're calling DbxrefList once and then calling DbxrefUrl at least twice.  dbxrefPrefixMap has 30 items so DbxrefList gets tested 30 times when it just creates a list.  We should be able to fully test that function with 3 or 4 test cases.  DbxrefUrl is more complex, but does it require 60+ tests that grow at 2 times the number of keys in the map?  I would guess it needs about 15 to 20 test cases that grow at a fraction of the number of map keys because most the keys and values are very similar.  Our testing for DbxrefUrl essentially boils down to checking a map key
```
const urlProcessor = dbxrefPrefixMap[prefix];
```
and string replace function
```
let url = urlPattern.replace(/\{0\}/g, encodeURIComponent(value));
```
but there is actually a lot more going on in there that isn't tested.  If our goal is to check urls, I suspect we could check N urls in dbxrefPrefixMap with one function, not N functions.  Then we wouldn't need to add 20 lines of test code for every new item added to dbxrefPrefixMap. 
  
There are many goals of testing.  One could be that we don't need to add new tests everytime we update code.  It's possible that when a developer adds a new item to dbxrefPrefixMap, without writing any tests, they get feedback as to the mistakes that were made.  One of them could be that they didn't write a test to check the url.  Another could be that they didn't add a pattern, or that the pattern doesn't have a '{0}'. 

The important takeaways here are that our test pattern scales with dbxrefPrefixMap, retests some line of dozens of times with essentially the same input while not testing others at all, and that we will never be done writing tests for this file.  


   I have limited experience in JS/React but it looks like the Work to Value ratio of the test pattern in this file is pretty large and will just get larger over time.   Work being developer effort writing the test, review effort(imagine the tests give feedback to the dev instead of having to find them review), and then actual test run time. 

If you think this is reasonable in JS/React then it might be neat to assign the update ticket to me.  Then I could get some real world experience in our front end with a low priority ticket.

PR #2326, opened 2018-07-09
ENCD-4017 Fulfill ECP requests for home page additions.

forresttanaka:
Two new search fields feature on the home page in place of the descriptive text next to the ENCODE assay category selector image. One simply copies the functionality of the search field in the upper right of every page. The other uses the SCREEN website in two ways:

1. To perform a search of a given term
2. To suggest search terms given a partial search term.

The vast majority of the changes exist in home.js including the new forms and nascent code for a combobox — a combination text-input field an drop-down list. I filed ENCD-4121 to refactor this combobox to its own module. I also filed ENCD-4120 to refactor other components to use the tooltip.js module this branch introduced.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .aws/
 .coverage
 .gitconfig
+.vscode/
 /.installed.cfg
 /.mr.developer.cfg
 /.cache/
--- a/src/encoded/static/components/app.js
+++ b/src/encoded/static/components/app.js
@@ -34,8 +34,11 @@ const portal = {
             title: 'Encyclopedia',
             children: [
                 { id: 'aboutannotations', title: 'About', url: '/data/annotations/' },
+                { id: 'sep-mm-1' },
+                { id: 'annotationvisualize', title: 'Visualize (SCREEN)', url: 'http://screen.encodeproject.org/' },
                 { id: 'annotationmatrix', title: 'Matrix', url: '/matrix/?type=Annotation&encyclopedia_version=4' },
                 { id: 'annotationsearch', title: 'Search', url: '/search/?type=Annotation&encyclopedia_version=4' },
+                { id: 'annotationmethods', title: 'Methods', url: 'http://screen.encodeproject.org/index/about' },
             ],
         },
         {
@@ -821,6 +824,7 @@ class App extends React.Component {
         return request;
     }
 
+    /* eslint-disable class-methods-use-this */
     fallbackNavigate(href, fragment, options) {
         // Navigate using window.location
         if (options.replace) {
@@ -833,6 +837,7 @@ class App extends React.Component {
             }
         }
     }
+    /* eslint-enable class-methods-use-this */
 
     receiveContextResponse(data) {
         // title currently ignored by browsers
--- a/src/encoded/static/libs/bootstrap/tooltip.js
+++ b/src/encoded/static/libs/bootstrap/tooltip.js
@@ -1,0 +1,120 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+
+// This module lets you have a tooltip pop up when any component is hovered over or focused. The
+// general usage is:
+//
+// <Tooltip trigger={<Trigger />} tooltipId="html-id-unique-on-page" css="custom-css-classes">
+//     <TooltipContent />
+// </Tooltip>
+//
+// The component that the user can hover over to trigger the tooltip is given in the `trigger`
+// property and appears in the DOM whereever you place the <Tooltip> component. Put the contents
+// of the tooltip between the opening and closing <Tooltip> tags. This can simply be a string if
+// that's all you need, or a full-fledged component.
+
+export default class Tooltip extends React.Component {
+    constructor() {
+        super();
+        this.state = {
+            tooltipDisplayed: false,
+        };
+        this.timer = null;
+        this.hoveringOverTooltip = false;
+        this.handleTriggerMouseEnter = this.handleTriggerMouseEnter.bind(this);
+        this.handleTriggerMouseLeave = this.handleTriggerMouseLeave.bind(this);
+        this.handleTooltipMouseEnter = this.handleTooltipMouseEnter.bind(this);
+        this.handleTooltipMouseLeave = this.handleTooltipMouseLeave.bind(this);
+        this.handleTriggerFocus = this.handleTriggerFocus.bind(this);
+        this.handleTriggerBlur = this.handleTriggerBlur.bind(this);
+    }
+
+    handleTriggerMouseEnter() {
+        // Mouse entered the tooltip trigger element.
+        this.setState({ tooltipDisplayed: true });
+
+        // If we happen to have a running timer, clear it so we don't hide the tooltip while
+        // hovering over the trigger.
+        if (this.timer) {
+            clearTimeout(this.timer);
+            this.timer = null;
+        }
+    }
+
+    handleTriggerMouseLeave() {
+        // Start a timer that might hide the tooltip after a second passes. It won't hide the
+        // tooltip if they're now hovering over the tooltip itself.
+        this.timer = setTimeout(() => {
+            this.timer = null;
+            if (!this.hoveringOverTooltip) {
+                this.setState({ tooltipDisplayed: false });
+            }
+        }, 1000);
+    }
+
+    handleTriggerFocus() {
+        // Show tooltip when keyboard focus lands on tooltip trigger.
+        this.setState({ tooltipDisplayed: true });
+    }
+
+    handleTriggerBlur() {
+        // Hide tooltip when keyboard focus leaves tooltip trigger.
+        this.setState({ tooltipDisplayed: false });
+    }
+
+    handleTooltipMouseEnter() {
+        // Mouse started hovering over the tooltip itself. Prevent the tooltip from hiding even if
+        // the one-second trigger time expires.
+        this.hoveringOverTooltip = true;
+    }
+
+    handleTooltipMouseLeave() {
+        // Mouse stopped hovering over the tooltip itself. If the timer's not running, hide the
+        // tooltip, otherwise let the existing timer hide the tooltip when it expires.
+        this.hoveringOverTooltip = false;
+        if (!this.timer) {
+            this.setState({ tooltipDisplayed: false });
+        }
+    }
+
+    render() {
+        const { trigger, direction, tooltipId, css } = this.props;
+        const tooltipCss = `tooltip ${direction}`;
+        const wrapperCss = `tooltip-container${css ? ` ${css}` : ''}`;
+        return (
+            <div className={wrapperCss}>
+                {this.state.tooltipDisplayed ?
+                    <div className={tooltipCss} role="tooltip" id={tooltipId} onMouseEnter={this.handleTooltipMouseEnter} onMouseLeave={this.handleTooltipMouseLeave}>
+                        <div className="tooltip-arrow" />
+                        <div className="tooltip-inner">{this.props.children}</div>
+                    </div>
+                : null}
+                <button
+                    aria-describedby={this.state.tooltipDisplayed ? tooltipId : ''}
+                    onMouseEnter={this.handleTriggerMouseEnter}
+                    onMouseLeave={this.handleTriggerMouseLeave}
+                    onFocus={this.handleTriggerFocus}
+                    onBlur={this.handleTriggerBlur}
+                    className="tooltip-container__trigger"
+                >
+                    {trigger}
+                </button>
+            </div>
+        );
+    }
+}
+
+Tooltip.propTypes = {
+    trigger: PropTypes.element.isRequired, // Visible tooltip triggering component
+    tooltipId: PropTypes.string.isRequired, // HTML ID of tooltip <div>; unique within page
+    direction: PropTypes.oneOf(['left', 'top', 'bottom', 'right']), // Direction of bootstrap tooltip location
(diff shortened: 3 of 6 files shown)


src/encoded/static/libs/bootstrap/tooltip.js
```diff
@@ -0,0 +1,120 @@
...
+        // Mouse stopped hovering over the tooltip itself. If the timer's not running, hide the
+        // tooltip, otherwise let the existing timer hide the tooltip when it expires.
+        this.hoveringOverTooltip = false;
+        if (!this.timer) {
+            this.setState({ tooltipDisplayed: false });
+        }
+    }
+
+    render() {
+        const { trigger, direction, tooltipId, css } = this.props;
```
zoldello: Ah ha! You did not define "direction" in the comment on top of the class. Is "placement" like Twitter Bootstrap tooltip a better name?

src/encoded/static/libs/bootstrap/tooltip.js
```diff
@@ -0,0 +1,120 @@
...
+        const tooltipCss = `tooltip ${direction}`;
```
zoldello: Personal Preference Comment: Why is there no defensive programming to default to a direction if an unknown value is provided.

src/encoded/static/libs/bootstrap/tooltip.js
```diff
@@ -0,0 +1,120 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+
+// This module lets you have a tooltip pop up when any component is hovered over or focused. The
+// general usage is:
+//
+// <Tooltip trigger={<Trigger />} tooltipId="html-id-unique-on-page" css="custom-css-classes">
```
zoldello: Why not call "tooltipId" simply "id"?

Personal Preference Question: Instead of passing a component into another component, why not just pass a id of another element and read off the element or set the child of Tooltip as the target?

src/encoded/static/libs/bootstrap/tooltip.js
```diff
@@ -0,0 +1,120 @@
+import React from 'react';
```
zoldello: Personal Preference Question: Tooltip seems like a common problem. Did you find any viable react-components plugins with toooltip support. My wild guess is at least someone wrote a react-wrapper for Twitter's bootstrap tooltip.

src/encoded/static/libs/bootstrap/tooltip.js
```diff
@@ -0,0 +1,120 @@
...
+        this.timer = setTimeout(() => {
+            this.timer = null;
+            if (!this.hoveringOverTooltip) {
+                this.setState({ tooltipDisplayed: false });
+            }
+        }, 1000);
+    }
+
+    handleTriggerFocus() {
+        // Show tooltip when keyboard focus lands on tooltip trigger.
```
zoldello: Personal Preference Comment: This seems like a comment that describes the function. It may be advantageous to use JSDoc comment style and have this on top of the method

src/encoded/static/libs/bootstrap/tooltip.js
```diff
@@ -0,0 +1,120 @@
...
+Tooltip.propTypes = {
+    trigger: PropTypes.element.isRequired, // Visible tooltip triggering component
+    tooltipId: PropTypes.string.isRequired, // HTML ID of tooltip <div>; unique within page
+    direction: PropTypes.oneOf(['left', 'top', 'bottom', 'right']), // Direction of bootstrap tooltip location
+    css: PropTypes.string, // CSS classes to add to tooltip-container wrapper class
+    children: PropTypes.node,
+};
+
+Tooltip.defaultProps = {
+    direction: 'bottom',
```
zoldello: where are "bottom", "top", "left", "right" used?

src/encoded/static/components/app.js
```diff
@@ -34,8 +34,11 @@ const portal = {
             title: 'Encyclopedia',
             children: [
                 { id: 'aboutannotations', title: 'About', url: '/data/annotations/' },
+                { id: 'sep-mm-1' },
```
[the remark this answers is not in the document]
forresttanaka: This is a React key. This mechanism’s been around for a very long time and was never commented. Should I comment it? I don’t mind, but I’ve gotten enough criticism the past few months for commenting things in detail that I’m a bit reticent.

src/encoded/static/libs/bootstrap/tooltip.js
```diff
@@ -0,0 +1,120 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+
+// This module lets you have a tooltip pop up when any component is hovered over or focused. The
```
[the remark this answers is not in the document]
forresttanaka: I’m a fan of JSDoc, though I haven’t yet run their tool on our code base. I haven’t yet seen an _official_ way to use JSDoc with React components though, so I haven’t done it. Is there one?

src/encoded/static/libs/bootstrap/tooltip.js
```diff
@@ -0,0 +1,120 @@
...
+        this.handleTriggerBlur = this.handleTriggerBlur.bind(this);
+    }
+
+    handleTriggerMouseEnter() {
+        // Mouse entered the tooltip trigger element.
+        this.setState({ tooltipDisplayed: true });
+
+        // If we happen to have a running timer, clear it so we don't hide the tooltip while
+        // hovering over the trigger.
+        if (this.timer) {
```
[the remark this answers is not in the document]
forresttanaka: setTimeout cannot return 0: https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setInterval#Return_value

src/encoded/static/libs/bootstrap/tooltip.js
```diff
@@ -0,0 +1,120 @@
...
+            clearTimeout(this.timer);
+            this.timer = null;
+        }
+    }
+
+    handleTriggerMouseLeave() {
+        // Start a timer that might hide the tooltip after a second passes. It won't hide the
+        // tooltip if they're now hovering over the tooltip itself.
+        this.timer = setTimeout(() => {
+            this.timer = null;
```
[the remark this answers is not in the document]
forresttanaka: Because line 49 wouldn’t be executed until after a second passed. It’s part of the callback at timer expiry.

src/encoded/static/libs/bootstrap/tooltip.js
forresttanaka: Yeah, “placement” sounds better.

src/encoded/static/libs/bootstrap/tooltip.js
forresttanaka: defaultProps defines “bottom” as default. propTypes defines the accepted values.

src/encoded/static/libs/bootstrap/tooltip.js
forresttanaka: This is a React-recommended way to do this (https://reactjs.org/docs/components-and-props.html#composing-components). Works well (used it a lot in ClinGen) and is simple and straightforward. The child of the tooltip is already the content of the tooltip. We could add a `TooltipTrigger` component as a required child, sort of like `<Param>`, but then we’d have to check that exactly one exists, then render it explicitly. This way is much simpler, which is probably why the React team recommends it.

I don‘t typically call these kinds of properties `id` in case someone confuses it with an HTML id, or in case we need more than one.

src/encoded/static/libs/bootstrap/tooltip.js
forresttanaka: I’m sure, but we’ve run into _so_ many problems using other people’s React components with them not getting supported long term, and then not working with new versions of React so we always ended up having to write our own anyway. The entire React `<forms>` module used to be a third-party React module, but their support was so bad David eventually ripped it out and rewrote the whole thing himself.

src/encoded/static/libs/bootstrap/tooltip.js
forresttanaka: If you have something to describe official ways to support JSDoc in methods, I’d like to see it. Possibly not for this branch though as this is so high priority.

src/encoded/static/libs/bootstrap/tooltip.js
forresttanaka: I don’t understand the question.

src/encoded/static/libs/bootstrap/tooltip.js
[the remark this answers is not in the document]
zoldello: @forresttanaka Thanks. I looked at that doc but did not search for "zero".

src/encoded/static/libs/bootstrap/tooltip.js
zoldello: @forresttanaka What I mean is that a user can put in "this-should-not-work" as the value of direction. I doubt if you were expecting that. So, the code ought to set a value for cases where a user puts in something he/she/gender-neutral ought not put to put in.

src/encoded/static/libs/bootstrap/tooltip.js
zoldello: @forresttanaka This is not big deal. Still, out of curiosity, I will stop by tomorrow. I think what I think this means is not what it does mean.

src/encoded/static/libs/bootstrap/tooltip.js
zoldello: @forresttanaka I assume direction means placement/position of the tooltip. I assume assume it is changeable by users since you expose it as a props. If these are true, where are alternates to "bottom"-setting used in the code?

src/encoded/static/components/app.js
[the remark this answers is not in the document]
zoldello: I think commenting is an art that will never please everyone. I will look this up as I have not seen it before. Since you say it has been used for a long time, it is safe to assume everyone should know it.

src/encoded/static/libs/bootstrap/tooltip.js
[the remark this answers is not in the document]
zoldello: React components are just JavaScript, so JSDoc itself should work fine. If react-ness is needed, I am sure someone made a library that somehow makes JSDoc and react work together. I found one on a quick search online, albeit, it has little stars.

src/encoded/static/libs/bootstrap/tooltip.js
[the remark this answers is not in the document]
zoldello: Good point. The function- setTimeout is not called immediately.

src/encoded/static/libs/bootstrap/tooltip.js
zoldello: similar to how they are used on classes... http://usejsdoc.org/tags-function.html? Again, this no big deal and the least of our worries.

src/encoded/static/libs/bootstrap/tooltip.js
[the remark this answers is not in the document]
forresttanaka: Is this going to gate approval? We have never specified JSDoc as a requirement; it’s just something I’ve done for regular functions as an experiment. I personally would like to see official support for React before we even consider making that an encode requirement.

src/encoded/static/libs/bootstrap/tooltip.js
forresttanaka: If they put “this-should-not-work” in a dev build, they’ll get an error in the console because of PropTypes. defaultProps sets “bottom” as default. That’s the whole point of PropTypes and default: value and type checking.

src/encoded/static/libs/bootstrap/tooltip.js
forresttanaka: I really worry something like shopping cart will take _weeks_ to code review over things like this that in the end change nothing.

src/encoded/static/libs/bootstrap/tooltip.js
forresttanaka: They’re in tooltips.js PropTypes, which in dev builds also does live value checking. As you see in modals.js, I’m a fan of documenting everything. But the past few months I’ve been burned enough about commenting too much or commenting too little that I’ve just about given up.

zoldello:
The code look fine. My comments cosmetics or commentaries.

.gitignore
```diff
@@ -1,6 +1,7 @@
 .aws/
 .coverage
 .gitconfig
+.vscode/
```
caseylitton: What's this from?  Is it part of the JIRA ticket?

forresttanaka: It’s not part of this ticket. VSCode sometimes makes this directory, and I had accidentally committed it at one point while working on this branch before realizing .vscode/ had been created.

caseylitton: Are you removing it?  Or do you think we should keep it in the repo? 

I'm back and forth on it.  It might be beneficial to .gitignore standard personal development environment files, like .vscode, or in my case for vim, .swp files.  On the other hand an individual's development environment isn't really a part of the encoded project.  And it could get out of hand pretty quick.  There are alternate solutions too like adding it to the .gitignore_global in the home directory.

forresttanaka: I don’t really care. Some other people here use VSCode and might accidentally commit .vscode/ too, so I felt like this was a way to head off what I had done accidentally. If you prefer I used .gitignore_global, I’ll do that.

caseylitton: Yea okay, let's leave it in but in general push people to using .gitignore_global for "personal" type things.

forresttanaka: That sounds good. Probably should document that somewhere.

PR #2331, opened 2018-07-10
ENCD-4116-import-vis-defines

zoldello:


buildout.cfg
```diff
@@ -53,7 +53,7 @@ webtest = git https://github.com/Pylons/webtest.git
 WSGIProxy2 = git https://github.com/lrowe/WSGIProxy2.git
 zope.sqlalchemy = git https://github.com/zopefoundation/zope.sqlalchemy.git
 pytest-bdd = git https://github.com/lrowe/pytest-bdd.git branch=allow-any-step-order
-snovault = git https://github.com/ENCODE-DCC/snovault.git rev=1.0.11
+snovault = git https://github.com/ENCODE-DCC/snovault.git branch=snovault-refactor-low-lvl
```
hitz: This change shouldn't be merged right @caseylitton?

src/encoded/helpers/__init__.py
```diff
@@ -0,0 +1 @@
+# package
```
hitz: Do we need this?  It's innocuous enough...  I thought empty __init__ should be empty.

caseylitton: I think this is a copy/paste thing.  I don't add it to new __init__.py files but it exists in other ones.

buildout.cfg
caseylitton: I tell people to do it this way because for whatever reason it shows up as a merge conflict.  So it actually can't be merged as is.  We revert the commit after finished with QA and code review.

caseylitton:
replaced by https://github.com/ENCODE-DCC/encoded/pull/2422

PR #2338, opened 2018-07-17
Fix encd 4083

caseylitton:
initial fix for 4083.  still needs to be tested.  hopefully get into v73.


caseylitton:
Replaced by https://github.com/ENCODE-DCC/encoded/pull/2341

PR #2340, opened 2018-07-18
ENCD-4033-added-header-to-search-page

zoldello:


hitz:
Ha @forresttanaka I just made similar comments on the Jira ticket.

hitz:
This build is failing because it doesn't have @zoldello 's fix in it.

PR #2344, opened 2018-07-23
ENCD-4141 Add alias namespaces 

jahilton:


src/encoded/tests/data/inserts/document.json
```diff
@@ -10,7 +10,7 @@
     },
     {
         "aliases": [
-            "encode:HAoSMC_PC_8121902.2P_certificate_of_analysis"
+            "alex-visel:HAoSMC_PC_8121902.2P_certificate_of_analysis"
```
estherc: Just a small typo here - should be "axel-visel" and it's causing test_load_workbook to fail.

jahilton: Yes, think we caught it at the same time.
Thanks!

PR #2347, opened 2018-07-24
ENCD-4118: set minimum value to 0 for ch coordinates in GM

weiweizhong2018:


src/encoded/schemas/genetic_modification.json
```diff
@@ -449,12 +449,14 @@
                 "start": {
                     "title": "Start",
                     "description": "The starting coordinate.",
-                    "type": "integer"
+                    "type": "integer",
+                    "minimum": 0
                 },
                 "end": {
                     "title": "End",
                     "description": "The ending coordinate (possibly equal to the starting coordinate for one base.)",
```
estherc: The change looks good. I think only the description needs updating for the end property as in a 0-based system, to indicate 1 base would mean end = start+1. end = start is also possible for indicating a position where an insertion occurs, for example, but the description just needs to choose one of those examples, not inappropriately combine both.

PR #2348, opened 2018-07-24
ENCD-4134 add RIN number property to Library

weiweizhong2018:


src/encoded/schemas/library.json
```diff
@@ -272,6 +272,13 @@
             "type": "string",
             "pattern": "(^[0-9]+-[0-9]+$|^[<>][0-9]+$)"
         },
+        "rna_integrity_number":{
```
caseylitton: Need a space after the ":"
```
"rna_integrity_number": {
```

PR #2350, opened 2018-07-25
ENCD-4144 Fix failing impersonation test button size

zoldello:
This is tricky. Re-writing the HTML to use grid-system would require a lot of testing (I think is ideal) since this code is reused a lot and in many ways. Lowering accession-class (left hand side of table) would also be tricky. The safe-bet is rise the right-hand-side to be higher than left. I did a light search and I could not find a place where that would be a problem,


zoldello:
@forresttanaka Take 2.

zoldello:
@forresttanaka Made the code update and updated the demo

PR #2351, opened 2018-07-26
ENCD-4149 Allow bin test log arg for workbook loading

caseylitton:
This allows bin/test to be called with a --log arugment which will limit the logging output from loadxl.load_all when called from tests.features.conftest.workbook.

Create and troubleshooting tests with that use workbook is a little slow and painful.  This will limit the scroll back in a backwards compatible manner.

### DIFF
--- a/src/encoded/loadxl.py
+++ b/src/encoded/loadxl.py
@@ -7,8 +7,9 @@ import os.path
 
 text = type(u'')
 
+DEFAULT_LOG_LEVEL = logging.INFO
 logger = logging.getLogger('encoded')
-logger.setLevel(logging.INFO)  # doesn't work to shut off sqla INFO
+logger.setLevel(DEFAULT_LOG_LEVEL)  # doesn't work to shut off sqla INFO
 
 
 ORDER = [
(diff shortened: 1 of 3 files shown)


src/encoded/loadxl.py
```diff
@@ -7,8 +7,9 @@
 
 text = type(u'')
 
+DEFAULT_LOG_LEVEL = logging.INFO
 logger = logging.getLogger('encoded')
-logger.setLevel(logging.INFO)  # doesn't work to shut off sqla INFO
+logger.setLevel(DEFAULT_LOG_LEVEL)  # doesn't work to shut off sqla INFO
```
hitz: not sure this is necessary, we don't do it throughout the code.

src/encoded/loadxl.py
```diff
@@ -90,6 +91,18 @@
     'cross_correlation_plot'
 ]
 
+
+def _reset_log_level(log_level):
+    '''
+    Resets the default log level, usually for running bin tests
+    '''
+    if logger and log_level is not DEFAULT_LOG_LEVEL:
```
hitz: oooh that's why

PR #2352, opened 2018-07-26
ENCD-4146 Restore FileDownloadButton constructor

forresttanaka:
Added a constructor to `<FileDownloadButton>` so that `this` is defined properly in `onMouseEnter` and `onMouseLeave`. I don’t know how this ever worked before, but it definitely did for some reason.


zoldello:
@forresttanaka Out of curiosity; I take it that the mouse-functionality did not work before..? I do not see how it would without this change.

forresttanaka:
I distinctly remember testing this ages ago, but how it could have worked, I don’t know.

PR #2353, opened 2018-07-27
ENCD-4131-set-status-endpoint

keenangraham:
Implements a method for releasing/deleting/revoking/archiving objects and their subobjects.

### DIFF
--- a/src/encoded/types/file.py
+++ b/src/encoded/types/file.py
@@ -147,6 +147,8 @@ class File(Item):
         'step_run',
     ]
     set_status_down = []
+    public_s3_statuses = ['released', 'archived', 'revoked']
+    private_s3_statuses = ['uploading', 'in progress', 'replaced', 'deleted']
 
     @property
     def __name__(self):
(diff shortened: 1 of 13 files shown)


src/encoded/types/file.py
```diff
@@ -404,36 +408,26 @@ def set_private_s3(self):
...
+            new_status,
+            request,
+            parent=parent,
+        )
+        if not status_set:
+            return False
         # Change permission in S3.
         try:
-            if new_status == 'released':
+            if new_status in ['released', 'archived', 'revoked']:
```
caseylitton: These to lists in the conditionals would be better as class vars.

src/encoded/types/quality_metric.py
```diff
@@ -20,6 +20,10 @@
     })
 class QualityMetric(ItemWithAttachment, CalculatedAssayTermID, Item):
     base_types = ['QualityMetric'] + Item.base_types
+    set_status_up = [
+        'step_run'
```
caseylitton: I would have put commas after all the final list items by default.

PR #2356, opened 2018-08-01
ENCD-4157 Make search view with tests

caseylitton:
This has two commits, first I made some tests then just moved previous search view code into a pyramid view class. 

Other changes would be to change the other search calls in batch_download, report, etc to SearchView.  This seemed to work with any issue.

Philip has some test we could add but I wanted to get the PR out to review and create a demo.


caseylitton:
Good learning but we're going to try to merge original search prs.

PR #2365, opened 2018-08-07
ENCD-4166 RIN internal action audit

gabdank:
Adding an audit for missing RIN in RNA libraries

### DIFF
--- a/src/encoded/audit/experiment.py
+++ b/src/encoded/audit/experiment.py
@@ -2910,18 +2910,14 @@ def audit_RNA_library_RIN(value, system, excluded_types):
     '''
     An RNA library should have a RIN specified.
     '''
-
-    RNAs = ['RNA',
-            'polyadenylated mRNA',
-            'miRNA']
-
+    RNAs = ['RNA', 'polyadenylated mRNA', 'miRNA']
     for rep in value['replicates']:
         if (rep['status'] not in excluded_types and
            'library' in rep and rep['library']['status'] not in excluded_types and
            rep['library']['nucleic_acid_term_name'] in RNAs and
            'rna_integrity_number' not in rep['library']):
-            detail = ('Metadata of RNA library {} lacks specification of ' +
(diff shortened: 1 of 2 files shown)


src/encoded/audit/experiment.py
```diff
@@ -2906,6 +2906,25 @@ def audit_library_RNA_size_range(value, system, excluded_types):
...
+    RNAs = ['RNA',
+            'polyadenylated mRNA',
+            'miRNA']
+
+    for rep in value['replicates']:
+        if (rep['status'] not in excluded_types and
+           'library' in rep and rep['library']['status'] not in excluded_types and
+           rep['library']['nucleic_acid_term_name'] in RNAs and
+           'rna_integrity_number' not in rep['library']):
+            detail = ('Metadata of RNA library {} lacks specification of ' +
```
caseylitton: The + can be removed

src/encoded/audit/experiment.py
```diff
@@ -2906,6 +2906,25 @@ def audit_library_RNA_size_range(value, system, excluded_types):
     return
 
 
+def audit_RNA_library_RIN(value, system, excluded_types):
+    '''
+    An RNA library should have a RIN specified.
+    '''
+
```
caseylitton: Can we collapse this all onto one line?  
```
RNAs = ['RNA', 'polyadenylated mRNA', 'miRNA']
for rep in value['replicates']:
```
It's a pretty short list.  If there is a concern about needing to add elements to it in the future then
```
RNAs = [
    'RNA',
    'polyadenylated mRNA',
    'miRNA',
]
for rep in value['replicates']:
```
would be prefered.

src/encoded/audit/experiment.py
gabdank: fixed

src/encoded/audit/experiment.py
gabdank: fixed

PR #2372, opened 2018-08-16
Encd 3998 link target to gene

yunhailuo:
As mentioned in the ticket, there are three commits in case we are not confident on the part for removing dbxref and gene_name:
1. Just link gene and target
2. Change dbxref and gene_name in Target to calculated property. I feel it's safer than the next change.
3. Completely remove dbxref and gene_name. This leads to a lot of small changes scattered around. JS changes only happen here.

Passed all tests (locally and on Travis). And local app looks good. Waiting for gene post and demo.


src/encoded/static/components/__tests__/dbxref-test.js
```diff
@@ -29,10 +29,10 @@ describe('Test individual dbxref types', () => {
         let dbxLinks;
 
         beforeAll(() => {
-            const context = { '@type': ['Target'], gene_name: 'CXXC1' };
+            const context = { '@type': ['Gene'], symbol: 'CXXC1' };
             const wrapper = mount(
                 <DbxrefList
-                    dbxrefs={['HGNC:hCGBP', 'HGNC:ZCGPC1']}
+                    dbxrefs={['HGNC:HGNC:24343']}
```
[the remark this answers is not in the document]
estherc: Oh I see, I think you updated the regex for dbxrefs in Gene for both HGNC and MGI to include our namespace AND the namespace used by those resources. I think it's redundant, as in both cases, I think the namespaces exist to distinguish from the Entrez IDs that are also numeric and we already do that with our namespacing.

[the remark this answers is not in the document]
yunhailuo: That's how Entrez did, and how HGNC and MGI do by default in their link (check links in these search results: https://www.genenames.org/cgi-bin/search?search_type=all&search=adnp&submit=Submit). I can change and upgrade Gene if needed.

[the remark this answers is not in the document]
estherc: I understand, but I'm fairly certain they do that in order to distinguish it from the Entrez ID which is also all numerics. However, since we already namespace the Entrez ID and the HGNC and MGI IDs, it seems redundant and confusing to do it twice.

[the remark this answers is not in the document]
yunhailuo: OK. Will work on an upgrade for Gene.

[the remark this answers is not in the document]
hitz: yeah I agree double namespacing is wrong.  Even when good people do it.

src/encoded/tests/test_audit_experiment.py
```diff
@@ -3137,4 +3137,4 @@ def test_is_control_dataset(testapp, control_target, ctrl_experiment, publicatio
     assert is_control_dataset(series_embedded) == True
     file_set = testapp.get(publication_data['@id'] + '@@index-data')
     file_set_embedded = file_set.json['embedded']
-    assert is_control_dataset(file_set_embedded) == False
\ No newline at end of file
+    assert is_control_dataset(file_set_embedded) == False
```
caseylitton: This is an unfortunate but common unnecessary change.  Seems like some people's text editors add these while other's remove them.  

I get around this and many other similar issues by always using git add --patch.  It's weird to become accustom to, but a great tool that leads to more consistent coding practices.

src/encoded/tests/test_audit_antibody_lot.py
```diff
@@ -98,4 +98,3 @@ def test_audit_control_characterizations(testapp, antibody_lot, base_target1):
     errors = res.json['audit']
     print(errors)
     assert 'NOT_COMPLIANT' not in errors
-
```
caseylitton: Did you remove this empty line on purpose or did your editor do it automatically?

src/encoded/tests/test_audit_antibody_characterization.py
```diff
@@ -106,12 +106,12 @@ def base_antibody(award, lab, source, organism, target):
...
 
 @pytest.fixture
-def recombinant_target(testapp, organism):
+def recombinant_target(testapp, organism, gene):
     item = {
         'organism': organism['uuid'],
         'label': 'HA-ABCD',
         'investigated_as': ['transcription factor', 'recombinant protein'],
-        'gene_name': "ABCD"
+        'targeted_genes': [gene['uuid']]
```
caseylitton: This is another situation where it is good practice to add a comma at the end.

src/encoded/upgrade/target.py
```diff
@@ -157,3 +157,15 @@ def target_7_8(value, system):
     # Shouldn't be any of these.
     elif status == 'replaced':
         value['status'] = 'deleted'
+
+
+@upgrade_step('target', '8', '9')
+def target_8_9(value, system):
+    # https://encodedcc.atlassian.net/browse/ENCD-3998
+    value.pop('gene_name', '')
```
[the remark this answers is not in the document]
yunhailuo: I think not giving default will lead to `KeyError` when the `gene_name` doesn't exist:

https://docs.python.org/3.4/library/stdtypes.html#dict.pop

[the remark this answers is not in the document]
caseylitton: nice

src/encoded/schemas/target.json
```diff
@@ -82,6 +98,9 @@
         },
         "investigated_as": {
             "title": "Target of assay"
+        },
+        "modification": {
+            "title": "Modification in targets"
```
hitz: Should be "Protein Modification"

src/encoded/tests/data/inserts/target.json
```diff
@@ -1,12 +1,6 @@
 [
     {
-        "dbxref": [
-            "ENSEMBL:ENSG00000101126",
-            "GeneID:23394",
-            "HGNC:ADNP1",
-            "UniProtKB:Q9H2P0"
-        ],
-        "gene_name": "ADNP",
+        "genes": ["23394"],
```
hitz: So the HGNC numerical ID is a valid alias -> pointer to the gene... that might be a little confusing.

src/encoded/tests/data/inserts/target.json
```diff
@@ -377,71 +241,51 @@
...
+        "status": "released"
+    },
+    {
         "status": "released",
-        "dbxref": [
-            "UniProtKB:Q71DI3"
+        "uuid": "452c5ea6-5880-499d-b8bb-392790fd2b89",
+        "organism": "human",
+        "genes": [
+            "8335",
```
hitz: This is because there are two identical copies of H2A right?

src/encoded/static/components/antibody.js
```diff
@@ -67,18 +67,28 @@ const LotComponent = (props, reactContext) => {
...
         organismTerms.push(`targets.organism.scientific_name=${scientificName}`);
         organismTips.push(scientificName);
 
         // Add to the information on gene names from the targets
-        if (geneName && geneName !== 'unknown') {
-            geneComponents.push(<span key={key}>{i > 0 ? <span> + {geneName}</span> : <span>{geneName}</span>}</span>);
-            geneTerms.push(`targets.gene_name=${geneName}`);
-            geneTips.push(geneName);
+        if (geneName.length) {
+            geneName.map(
```
forresttanaka: .map() is really designed to convert one array to another, not as a generic looping structure. You can replace with .forEach() so the fact that no conversion happens is clear to others (though you’ll need to convert to an explicit return).

_But_ this also looks like it _might_ be an actual mapping operation. It _looks_ like you can keep .map() but rather than pushing geneComponents you can just return the span, and assign the result of .map() to geneComponents.

You’re probably also getting a Javascript warning in the console, if not an error, because you’re assigning the same `key` to multiple `<spans>`. If `symbol` is unique in each iteration, you can use `key={symbol}` instead to get rid of the error.

src/encoded/static/components/antibody.js
```diff
@@ -67,18 +67,28 @@ const LotComponent = (props, reactContext) => {
...
+                symbol => geneComponents.push(
+                    <span key={key}>{geneComponents.length > 0 ? <span> + {symbol}</span> : <span>{symbol}</span>}</span>
+                )
+            );
+            geneTerms.push(
+                ...geneName.map(
+                    symbol => `targets.genes.symbol=${symbol}`
+                )
+            );
```
forresttanaka: Similar thing here. I _think_ you can say geneTerms = geneName.map(…) if I’m understanding correctly.

src/encoded/static/components/antibody.js
```diff
@@ -67,18 +67,28 @@ const LotComponent = (props, reactContext) => {
...
+            geneTips.push(...geneName);
```
forresttanaka: You’re cloning the geneName array into geneTips? You’ll probably want to comment this as this is a highly unusual way to do that, or you can change to:

geneTips = geneName.slice(0); // Very common
geneTips = [...geneName]; // Much less common, but MDN-approved

src/encoded/static/components/target.js
```diff
@@ -13,28 +12,12 @@ class Target extends React.Component {
...
-                sep = geneRef.indexOf(':') + 1;
-                baseName = geneRef.substring(sep, geneRef.length);
-                geneLink = baseUrl + baseName;
-            }
-        }
+        const geneIDs = context.genes.map(
+            gene => `GeneID:${gene.geneid}`
+        );
+
+        const dbxrefs = [].concat(...context.genes.map(gene => gene.dbxrefs));
```
forresttanaka: Simpler to do:

```
const dbxrefs = [...context.genes.map(gene => gene.dbxrefs)]
```

or probably just:

```
const dbxrefs = context.genes.map(gene => gene.dbxrefs)
```

...because it doesn’t appear to matter if the array is cloned or not.

src/encoded/static/components/antibody.js
forresttanaka: Actually, I’m not clear you need to clone geneName. Can you just use it directly below instead of cloning to geneTips?

src/encoded/tests/data/inserts/target.json
yunhailuo: This is NCBI Entrez ID which is the key for gene object. It's confusing without namespace. Do you think add some explanation to the schema will help?

src/encoded/tests/data/inserts/target.json
yunhailuo: Yes, and there are actually more. Please check my comments in WRAN-911. In my current plan, WRAN-911 needs to get in first before we can merge this.

src/encoded/static/components/antibody.js
yunhailuo: Thank you very much for all these comments, especially the detailed explanations. They are very helpful and let me learn a lot!

src/encoded/static/components/__tests__/dbxref-test.js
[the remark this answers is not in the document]
gabdank: I believe double name-spacing was resolved

src/encoded/static/components/target.js
yunhailuo: gene.dbxrefs itself is an array. I'm trying to flat the array here. I think I need a flat array [here](https://github.com/ENCODE-DCC/encoded/pull/2372/files/73d38b88d200695c4563e350a956e480a67e7194#diff-1b76489ca7832bd8baec06af55cf04a7R63). Does this sound right/reasonable?

forresttanaka: You’re converting an array of gene objects into a corresponding array of gene dbxrefs?

forresttanaka: Oh wait, I have to think about this. You want a simple array of dbxref strings that combines all the arrays in all genes?

forresttanaka: I think that’s what you’re doing, right? In that case, I think this is fine but could you add a comment so it’s clear what this is for?

yunhailuo: Sure. Will add a comment.

To clarify, target.genes is an array of gene objects. For a gene object, gene.dbxrefs is an array of dbxrefs (strings). This PR will remove the dbxref property from target object. To display dbxrefs related to a target, using/extracting dbxrefs from its related/targeted genes. So yes, it will be a a simple array of dbxref strings that combines all the (gene.dbxrefs) arrays in all genes. Is this making sense now?

src/encoded/static/components/antibody.js
yunhailuo: Pushed a new commit. Could you help me check if the way of defining and assigning variables is good practice?

yunhailuo:
Thanks Forrest. Definitely need to correct it.

yunhailuo:
Replaced by #2416 

There are quite some re-organization and changes for a change of plan. Close this one

PR #2378, opened 2018-08-23
ENCD-3692-search-refactoring-v3

zoldello:

PR #2379, opened 2018-08-23
ENCD-3692-search-refactoring

zoldello:
replaces https://github.com/ENCODE-DCC/encoded/pull/2378


buildout.cfg
```diff
@@ -53,7 +53,8 @@ webtest = git https://github.com/Pylons/webtest.git
 WSGIProxy2 = git https://github.com/lrowe/WSGIProxy2.git
 zope.sqlalchemy = git https://github.com/zopefoundation/zope.sqlalchemy.git
 pytest-bdd = git https://github.com/lrowe/pytest-bdd.git branch=allow-any-step-order
-snovault = git https://github.com/ENCODE-DCC/snovault.git rev=1.0.16
+# snovault = git https://github.com/ENCODE-DCC/snovault.git rev=1.0.16
+snovault = git https://github.com/ENCODE-DCC/snovault.git branch=SNO-31-search-refactoring-v3
```
caseylitton: Reminder to revert before mering.

src/encoded/helpers/__init__.py
```diff
@@ -0,0 +1 @@
+# package
```
caseylitton: Remove all these # package things.  I know it's in other ones but they don't have a purpose.  

You can add a doc string at the top explaining the reason for the package.
```
"""
Container for code shared by multiple scripts or modules.
"""
```

src/encoded/helpers/helper.py
```diff
@@ -0,0 +1,118 @@
...
+        self.summary_report = {
+            'href': request.route_path('summary', slash='/') + search_base,
+            'title': 'View summary report',
+            'icon': 'summary',
+        }
+        self.summary_matrix = {
+            'href': request.route_path('matrix', slash='/') + search_base,
+            'title': 'View summary matrix',
+            'icon': 'th',
+        }
```
caseylitton: Can you look into how to get rid of the 'no newline' at the end of the file?  I think we want new lines.

src/encoded/viewconfigs/__init__.py
```diff
@@ -0,0 +1 @@
+# package
```
caseylitton: add doc string instead of package comments

src/encoded/viewconfigs/news.py
```diff
@@ -0,0 +1,48 @@
...
+        self.result['@id'] = '/news/' + self.search_base
+        self.result['@type'] = ['News']
+        self.result['notification'] = ''
+
+        # construct doc_types
+        self.construct_doc_types(['Page'])
+
+        # construct query
+        static_items = [('type', 'Page'), ('news', 'true'), ('status', 'released')]
+        query = self.construct_query(static_items, False)
```
caseylitton: construct_query doesn't exist in SearchView or BaseView.   Same with construct_doc_types.  It looks like these functions were removed from SearchView 27 days ago in snovault commit 'phil modifications to kathik refactor' on your 'SNO-31-search-refactoring-v3' branch.   Please go back and make sure inherited method in all the subviews exist in their parents somewhere.

Two things.  

1. Make a list of urls used to test the search refactor feature.  There should be a few for each of the 6 views, search, audit, summary, news, matrix, and report.  Why a few for each?  Each view has conditions, so a good list will hit each condition.  Each time we make an update to the PR we need to check this list first before code review.  It looks like news hasn't worked for a month.

2. Given that all the views eventually inherit from BaseView and all views have similar, required methods.  Good practice would be adding all the functions that must be overridden by a child class to the base class as
```
def construct_query(self, what_ever_args):
    '''
    Contruct Query for such and such
    - Must be overridden by child classes.
    '''
    raise NotImplementedError
```
for example.  This way, when someone creates a new view they know what is required.  Do some googling for NotImplementedError to see how people use it in practice.  It helps to define how parent classes should be implemented.  Add a couple NotImplementedError functions to the base view and then sync up with me to check it out.

src/encoded/viewconfigs/views.py
```diff
@@ -0,0 +1,109 @@
...
+        doc_types = [context.type_info.name]
+    else:
+        doc_types = request.params.getall('type')
+    
+    if '*' in doc_types:
+        doc_types = ['Item']
+    
+    return doc_types
+
+def iter_search_results(context, request):
```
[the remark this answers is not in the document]
zoldello: It was done this way to abstract away the need for batch_download.py of knowing about the decorator- view_config; albeit not a strong one. Still, do you want it removed?

src/encoded/viewconfigs/views.py
```diff
@@ -0,0 +1,109 @@
...
+@view_config(context=AbstractCollection, permission='list', request_method='GET', name='listing')
+def collection_view_listing_es(context, request):
+    # Switch to change summary page loading options
+    if request.datastore != 'elasticsearch':
+        return collection_view_listing_db(context, request)
+    return search(context, request)
+
+@view_config(route_name='search', request_method='GET', permission='search')
+def search(context, request, search_type=None, return_generator=False):
+    default_doc_types = [
```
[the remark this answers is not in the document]
zoldello: default_doc_types was made an argument because many other objects call SearchView and they have different values for default_doc_types. So, adding it as a class variable in SearchView may not meet the needs of all objects and introduce encode-specific settings into snovault.  Does this make sense to you?

src/encoded/viewconfigs/views.py
```diff
@@ -0,0 +1,109 @@
...
+        'Biosample',
+        'Dataset',
+        'GeneticModification',
+        'Page',
+        'Pipeline',
+        'Publication',
+        'Software',
+        'Target',
+    ]
+    search = SearchView(context, request, search_type, return_generator, default_doc_types)
```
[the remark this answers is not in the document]
zoldello: - I assume this is more of a style change, if true, I have done this and on anything I touch in this commit (although not previous ones). 
- I made a comment about why default_doc_types was needed in a previous comment.
- I do not get my TODO from this comment- "Please review the reset of the code changes in this PR for the same issue. Python allows keyword args to be passed in positionally, but we as an organization don't want it." I was trying to set default values; "positionally" came along.

src/encoded/viewconfigs/news.py
zoldello: I did some updates on news.py. Now, you see JSON but not the css-styling. I will address that in an upcoming commit. 

On the two things
1. Okay, I will write it up and place it in Jira; after the addressing the code-feedback

2. I do not fully get this. There is no method that must be overridden.

src/encoded/viewconfigs/views.py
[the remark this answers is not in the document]
caseylitton: Can you list the locations where this occurs?

Also, did you fix the default_data_types?

src/encoded/viewconfigs/views.py
```diff
@@ -0,0 +1,115 @@
...
+
+        if hasattr(ti.factory, 'summary_data'):
+            views.append(view_item.summary_report)
+    
+    return search.preprocess_view(views=views, search_result_actions=search_result_actions)
+
+
+@view_config(route_name='report', request_method='GET', permission='search')
+def report(context, request):
+
```
caseylitton: please remove all the empty space inside these view_config  functions.  I asked Karthik to do it a long time ago and I guess it slipped by.

src/encoded/viewconfigs/views.py
```diff
@@ -0,0 +1,115 @@
...
+        return_generator, 
+        default_doc_types
+    )
+    doc_types = _get_doc_types(context, request)
+    views = []
+    view_item = View_Item(search.request, search.search_base)
+    if len(doc_types) == 1:
+        ti = search.types[doc_types[0]]
+        views.append(view_item.tabular_report)
+
```
caseylitton: Same thing here, lets clean this code up by removing these empty lines.

src/encoded/tests/test_search.py
```diff
@@ -753,3 +888,115 @@ def test_format_facets_adds_pseudo_facet_for_extra_filters():
         ],
         'total': 42,
     }]
+
+
+def test_sort_query(sort_query_item):
+
```
caseylitton: I asked Karthik to remove these extra spaces a while ago.  Please remove them from the top of all these tests functions.

src/encoded/viewconfigs/views.py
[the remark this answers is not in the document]
zoldello: Thinking about it, you are right. My way probably more Uncle Bob-like, but adds a small bit of complexity with no significant benefit.

src/encoded/viewconfigs/news.py
```diff
@@ -0,0 +1,98 @@
...
+
+        # Insert the facet data into the results.
+        schemas = [self.types[doc_type].schema for doc_type in doc_types]
+        self.result['facets'] = self.format_facets(
+            es_results, 
+            facets, 
+            used_filters, 
+            schemas, total, 
+            self.principals)
+
```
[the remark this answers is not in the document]
zoldello: Let's talk about what linting/flake8/pep8 tool you use to catch these. None of my tools are.

src/encoded/viewconfigs/views.py
zoldello: What empty space

src/encoded/viewconfigs/views.py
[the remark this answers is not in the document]
zoldello: Phil- Note clear.

caseylitton:
replaced by https://github.com/ENCODE-DCC/encoded/pull/2422

PR #2389, opened 2018-08-30
ENCD-4153-alter-select-distinct-values

keenangraham:
More efficient assay_term_id calculated property for FileSets and use of skip_calculated=true. This is joint PR with Snovault ticket.

### DIFF
--- a/src/encoded/types/dataset.py
+++ b/src/encoded/types/dataset.py
@@ -29,7 +29,7 @@ import datetime
 
 
 def item_is_revoked(request, path):
-    return request.embed(path, '@@object').get('status') == 'revoked'
+    return request.embed(path, '@@object?skip_calculated=true').get('status') == 'revoked'
 
 
 def calculate_assembly(request, files_list, status):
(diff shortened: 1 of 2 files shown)


src/encoded/types/dataset.py
```diff
@@ -29,7 +29,7 @@
 
 
 def item_is_revoked(request, path):
-    return request.embed(path, '@@object').get('status') == 'revoked'
+    return request.embed(path, '@@object?skip_calculated=true').get('status') == 'revoked'
```
hitz: explicitly adding the parameter skip_calculated using ? doesn't seem 100% kosher.  Like I feel it would be cleaner if parameters could be passed to request.embed directly?

src/encoded/types/shared_calculated_properties.py
```diff
@@ -150,16 +150,19 @@ def assay_term_name(self, request, files):
...
         "title": "Assay term id",
         "type": "array",
         "items": {
             "type": 'string',
         },
     })
-    def assay_term_id(self, request, files):
-        return request.select_distinct_values(
-            'replicate.experiment.assay_term_id', *files)
+    def assay_term_id(self, request, assay_term_name):
```
hitz: is this just faster?

keenangraham:
Tests in snovault. Though it would be nice to have more unit tests for the calculated property functions themselves.

src/encoded/types/dataset.py
caseylitton: In python 3.5 you can use multiple sets of kwargs.  That is how I would set this up.  request embed would take a dict of **query_args and a dict of **options.  so skip_calc would be a query arg and the is_user would be an option.

keenangraham:
Removed snovault branch @caseylitton. Ready for rereview.

PR #2392, opened 2018-09-04
ENCD-4196 Reduce cloud init size

caseylitton:
https://encodedcc.atlassian.net/browse/ENCD-4196

This was becoming a nuisance.  Should have just removed comments and empty lines from all cloud-init files.

I did a quick test on the demo.  I'll be using this update all week building other clusters and demos for es worker stuff.

### DIFF
--- a/cloud-config-cluster.yml
+++ b/cloud-config-cluster.yml
@@ -1,3 +1,4 @@
+#cloud-config
 ssh_authorized_keys:
   - %(LOCAL_SSH_KEY)s
 apt_sources:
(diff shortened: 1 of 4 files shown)


cloud-config-cluster.yml
```diff
@@ -1,3 +1,4 @@
+#cloud-config
```
caseylitton: This is needed but was missing.  We don't really use this file any more.  It
s use case is building the old style cluster with es master on the front end machine.  Do we want to continue maintaining this deployment?

cloud-config-elasticsearch.yml
```diff
@@ -139,97 +123,60 @@ users:
...
     */5 * * * * nobody /opt/cloudwatchmon/bin/mon-put-instance-stats.py --mem-util --swap-util --disk-space-util --disk-path=/ --from-cron
-
 - path: /etc/elasticsearch/elasticsearch.yml
   content: |
     network.host: 0.0.0.0
     http.port: 9201
     transport.tcp.port: 9299
     node.master: %(ES_MASTER)s
     node.data: %(ES_DATA)s
-    # if 5 nodes then min is 3 since (5/2) + 1
```
caseylitton: I liked this comment, may want to keep?

cloud-config.yml
```diff
@@ -238,18 +224,6 @@ runcmd:
 - sudo mv /etc/mailname /etc/mailname.OLD
 - sudo echo "$PUBLIC_DNS_NAME" | sudo tee --append /etc/mailname
 - sudo service postfix restart
-# - sleep 10
```
hitz: this can definitely go, it's a terrible place for actually useful documentation

PR #2395, opened 2018-09-07
ENCD-4198 Clean up indexer init

caseylitton:
Goes with SNO-51 PR, This also make the vis and region indexer string names into variables

The encoded process did need an instance of an Indexer attached to the requests for multi processing indexing, but only because update_object used self.es.  But ELASTICSEARCH is attached to every request and update_object gets a request. 

Removing that dependency allows update_object_in_snapshot in mpindexer.py to call update_object as a staticmethod.

Lo and behold there is a bunch of failing tests...I'll see if those can be easily resolved.  If not then dropping PR.


caseylitton:
Not necessary before es worker

PR #2397, opened 2018-09-10
ENCD-4197 red audit for missing step_run

gabdank:


src/encoded/audit/file.py
```diff
@@ -5,6 +5,27 @@
 
 from .item import STATUS_LEVEL
 
+def check_award_condition(value, awards):
+    return value.get('award', None) and value.get('award')['rfa'] in awards
```
keenangraham: Perhaps `value.get('award', {}).get('rfa')` in case the award doesn't have an rfa for whatever reason.

src/encoded/audit/file.py
```diff
@@ -5,6 +5,27 @@
...
+                           'deleted',
+                           'revoked',
+                           'uploading',
+                           'content error',
+                           'upload failed']:
+        return
+    if value['output_category'] in ['raw data',
+                                    'reference']:
+        return
+    if check_award_condition(value.get('dataset'), ['ENCODE3', 'ENCODE4']):
```
keenangraham: Isn't `value.get('dataset')` here just the accession of the experiment? How would passing this into check_award_condition work?

src/encoded/audit/file.py
gabdank: fixed

src/encoded/audit/file.py
gabdank: 'dataset' is embedded in file audits (see line 356) , it is not just an accession

gabdank:
@keenangraham , I have fixed according to your comments.

PR #2401, opened 2018-09-11
ENCD-4200-fix-fb

zoldello:


zoldello:
@caseylitton: Done. Strangely, I did not get notification about your change request. I found this by looking at a different PR.

PR #2404, opened 2018-09-12
ENCD-4173 Add Paul to test user inserts

p-sud:
I added myself to user.json so I can log in to local demos.


p-sud:
@emro You are correct, 'ENCORE' viewing group doesn't appear anywhere else in user.json. I just copied and pasted the json from my profile on the ENCODE portal.

forresttanaka:
Might want to ask Jason or Idan about it. Looks like very recent profiles on production have been getting ENCORE, but older ones don’t. ENCORE is a relatively new project.

hitz:
I don't think ENCORE is used in the LOCAL test  database so it's not relevant.  Probably should be added for completeness.

PR #2408, opened 2018-09-13
ENCD-4212 Change encoded home dir on deploy

caseylitton:
Need to do on all cloud inits.  This is only for demo.   Also the ENCD-4196 is on here so only look at other commit.


caseylitton:
This did not end up working.  Not vital so closing PR.  Moving jira ticket back to todo.

PR #2413, opened 2018-09-19
ENCD-4221 Fix button styles

emro:


emro:
Did not submit correctly, submitted a new pull request instead.

PR #2416, opened 2018-09-24
ENCD-3998 Link target gene

yunhailuo:
Replacing #2372 

I've got a lot of re-organization and some changes in the code for a change in plan (keeping dbxref for now). Sorry about this but I'd appreciate it if I can get a new code review.

This PR also contain a little bit code for ENCD-4000, in the "modifications" property of "target.json".


src/encoded/static/components/antibody.js
```diff
@@ -62,23 +62,26 @@ const LotComponent = (props, reactContext) => {
...
         organismComponents.push(<span key={key}>{i > 0 ? <span> + <i>{scientificName}</i></span> : <i>{scientificName}</i>}</span>);
         organismTerms.push(`targets.organism.scientific_name=${scientificName}`);
         organismTips.push(scientificName);
 
         // Add to the information on gene names from the targets
-        if (geneName && geneName !== 'unknown') {
-            geneComponents.push(<span key={key}>{i > 0 ? <span> + {geneName}</span> : <span>{geneName}</span>}</span>);
-            geneTerms.push(`targets.gene_name=${geneName}`);
-            geneTips.push(geneName);
+        if (geneName.length) {
```
forresttanaka: Very minor, but the AirBnB style guide specifies numeric comparisons, even against 0, must be explicit https://github.com/airbnb/javascript#comparison--shortcuts

src/encoded/static/components/antibody.js
```diff
@@ -87,7 +90,7 @@ const LotComponent = (props, reactContext) => {
     const crumbs = [
         { id: 'Antibodies' },
         { id: organismComponents, query: organismQuery, tip: organismTips.join(' + ') },
-        { id: geneComponents.length ? geneComponents : null, query: geneQuery, tip: geneTips.join(' + ') },
+        { id: geneComponents.length ? geneComponents : null, query: geneQuery, tip: geneName.join(' + ') },
```
forresttanaka: comparison--shortcuts from AirBnB again.

src/encoded/static/components/dbxref.js
```diff
@@ -65,9 +65,9 @@ export const dbxrefPrefixMap = {
     HGNC: {
         pattern: 'http://www.genecards.org/cgi-bin/carddisp.pl?gene={0}',
         preprocessor: (context) => {
-            // For dbxrefs in targets, use target.gene_name instead of the dbxref value.
-            if (context['@type'][0] === 'Target' && context.gene_name) {
-                return { altValue: context.gene_name };
+            // For dbxrefs in targets, use the symbol of the first gene in target.genes instead of the dbxref value.
+            if (context['@type'][0] === 'Target' && context.genes) {
+                return { altValue: context.genes[0].symbol };
```
forresttanaka: If target.genes exists, are you guaranteed to have a first item (context.genes[0]) or could there be a case that this array is empty?

src/encoded/static/components/target.js
```diff
@@ -64,10 +46,12 @@ class Target extends React.Component {
                             <dd>{context.label}</dd>
                         </div>
 
-                        {context.gene_name && geneLink ?
+                        {geneIDs.length ?
```
forresttanaka: comparison--shortcuts rule.

src/encoded/static/components/antibody.js
yunhailuo: Thank you for pointing this out. Very good practice to learn.

src/encoded/static/components/dbxref.js
yunhailuo: Initially, I think this is OK because target.genes shouldn't be an empty array if the schema and dependency works properly. However, now I think you are absolutely right. I should never rely on that in the code. I'll add ` && context.genes.length > 0`.

yunhailuo: One more clarification for this question. If target.genes is indeed an empty array, there is a chance the link goes bad (because there is no good altValue). I think this is expected and (has to be) acceptable.

yunhailuo:
@gabdank You are talking about boost_values, are you? If yes, they are to replace whatever is there for target.gene_name.

yunhailuo:
I'll need another review one commit 29629827474558d839aa2ebd7267f6169403b4d0 I just pushed. It is to prevent target having genes from different organisms. @hitz @gabdank 

Thanks @keenangraham for discussing about the code!

yunhailuo:
Amino acid one-letter code will be `^[AC-IK-Y]$` based on https://www.uniprot.org/help/sequences

There are unspecified amino acid (Xaa; X), selenocysteine (Sec; U) and pyrrolysine (Pyl; O), which are special cases. I'm not sure if we should exclude them, @gabdank?

hitz:
I would keep the odd Amino acid codes.

gabdank:
@hitz why keep the odd Amino acid codes?

yunhailuo:
Not sure if it is strong enough to argue keeping them, they are rare but are standard code for proteins. Like [SELENON|https://www.uniprot.org/uniprot/Q9NZV5]. Pyrrolysine is encoded by stop codon UAG but not in human.

yunhailuo:
Sounds reasonable. Restrict to 20 amino acid code now, `^[ACDEFGHIKLMNPQRSTVWY]$`.

yunhailuo:
Tests have been added, @hitz @gabdank .

src/encoded/schemas/target.json
```diff
@@ -31,25 +45,88 @@
...
-            "format": "gene_name"
+        "genes": {
+            "title": "Targeted genes",
+            "description": "List of gene(s) targeted by corrsponding assay or antibody.",
+            "comment": "See gene.json for available identifiers and details.",
+            "type": "array",
+            "uniqueItems": true,
+            "items": {
+                "title": "Targeted Gene",
+                "description": "A gene targeted by corrsponding assay or antibody.",
```
gabdank: corrsponding typo, should be corresponding

yunhailuo:
The new commit fixes a typo and the Breadcrumbs of antibody page. In general, I made it similar to the dev and may have the same problem to be fixed in the new ticket. The key is "geneName" is now an array of strings rather than a single string. May I have another code review here, @forresttanaka ?

forresttanaka:
@yunhailuo I looked over your latest branch and I see no issues.

yunhailuo:
Thanks Forrest!

PR #2417, opened 2018-09-24
ENCD-4195 Ontology update sept18

jahilton:
ENCD-4195
Changing artery organ slim to arterial blood vessel
Forcing germinal matrix to slim to brain
Updating EFO owl file link
Typical ontology update
Add cell_slims property & facets to Experiments & Biosamples (ENCD-3555)


caseylitton:
closing since this is stalled on a backlog ticket.  The branch will still exists, and we can just reopen when needed.

PR #2420, opened 2018-09-27
REGULOME local region tests

yunhailuo:
I made this on top of v75.1 so that it's separated from regulome codes and hopefully easy to review.


src/encoded/types/file.py
```diff
@@ -507,6 +510,14 @@ def download(context, request):
             ExpiresIn=36*60*60
         )
     else:
+        if request.registry.settings.get('testing'):
```
caseylitton: I was just reimagining the s3/test conditional in the download function...Maybe this is a little more readable.
```
external = context.propsheets.get('external', {})
    if not external.get('service') == 's3':
        if asbool(request.registry.settings.get('testing')):
            body, headers = _get_test_file_response(filename)
            if body:
                return Response(body=body, headers=headers)
            else:
                raise HTTPNotFound(
                    detail='File download could not read body from file {}.'.format(
                        filename,
                    )
                )
        else:
            raise HTTPNotFound(
                detail='External service {} not expected'.format(external.get('service'))
            )
    conn = boto3.client('s3')
    location = conn.generate_presigned_url(
        ClientMethod='get_object',
        Params={
            'Bucket': external['bucket'],
            'Key': external['key'],
            'ResponseContentDisposition': 'attachment; filename=' + filename
        },
        ExpiresIn=36*60*60
    )
    if asbool(request.params.get('soft')):
```
Also I moved the check/open test file code into it's own function. 
```
def _get_test_file_response(filename):
    '''Look for and return test file response'''
    filedir = resource_filename('encoded', 'tests/data/documents/')
    filepath = os.path.join(filedir, filename)
    if os.path.exists(filepath):
        mime_type, _ = mimetypes.guess_type(filepath)
    try:
        with open(filepath, 'rb') as file_handler:
            body = file_handler.read()
            headers = {'Content-Type': mime_type}
            return body, headers
    except Exception:  # pylint: disable=broad-except
        return None, None
```
I think this second part, separating the test file code, is better because we could more easily write units tests for building a response body/header.  As opposed to the current implementation where we can only verify all of the download code through integration testing via a request object.

src/encoded/types/file.py
```diff
@@ -492,24 +495,32 @@ def download(context, request):
...
-                'Key': external['key'],
-                'ResponseContentDisposition': 'attachment; filename=' + filename
-            },
-            ExpiresIn=36*60*60
-        )
-    else:
-        raise HTTPNotFound(
-            detail='External service {} not expected'.format(external.get('service'))
-        )
+    if not external.get('service') == 's3':
```
keenangraham: @caseylitton But now you are trying to use boto3 even if the service isn't s3?

yunhailuo: I think `if not external.get('service') == 's3'`, this download method will either return a local file or raise `HTTPNotFound` or other exceptions. So when it tries to use boto3, it should have confimed `external.get('service') == 's3'`. Do I understand the code correctly? And should we make it more explicit/clear?

src/encoded/types/file.py
yunhailuo: Good suggestions! I adopted them. Thank you.

src/encoded/types/file.py
keenangraham: Ah, was missing the HTTPNotFound before. Thanks, makes sense.

yunhailuo: No problem. Thank you for taking a look here.

yunhailuo:
Replaced by #2427

PR #2425, opened 2018-10-03
ENCD-4185-s3-public-url

keenangraham:
Exposes google_transfer and s3_uri calculated properties on released/archived/revoked files.

### DIFF
--- a/src/encoded/types/file.py
+++ b/src/encoded/types/file.py
@@ -1,4 +1,5 @@
 from botocore.exceptions import ClientError
+from botocore.config import Config
 from snovault import (
     AfterModified,
     BeforeModified,
@@ -27,6 +28,7 @@ from urllib.parse import (
 )
 import base64
 import boto3
+import botocore
 import datetime
 import logging
 import json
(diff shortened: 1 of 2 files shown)


src/encoded/types/file.py
```diff
@@ -332,6 +341,56 @@ def file_type(self, file_format, file_format_type=None):
...
+            "comment": "Do not submit. Value is calculated from file metadata.",
+            "type": "string",
+            "notSubmittable": True, 
+        }
+    )
+    def s3_uri(self):
+        try:
+            external = self._get_external_sheet()
+        except HTTPNotFound:
+            return ''
```
keenangraham: I do wish there was a better option here than just returning an empty string if the file doesn't have an s3 external sheet (there are a few, for whatever reason). Returning None doesn't seem right because I think that would show up in the JSON.

keenangraham: `if you don't return at all - would not that mean that the property will not be there?`

All Python functions return None if there isn't an explicit return value. Not sure if the schema is smart enough to remove this field from JSON though, or use None as the value.

`would not that be the desired result for files without s3 external sheet?`
My thinking was that it would be more important to be consistent in having that field in all released/archived/revoked files for people who are expecting that field to be there, even if the value is blank.

`how are restricted files handled?`

This is a good point. Restricted files probably shouldn't calculated these properties, and actually the set_status endpoint for files should probably not try to set the ACL to public when these are released (separate ticket).

keenangraham:
```
file_size is a duplication of file_size property
md5sum_base64 is derived from md5sum
url field has a different meaning for Biosample (and other objects),
file object has href property, maybe we should be consistent with that?
```

The idea is that the base64 md5sum isn't very straightforward to calculate, and calculating the proper URL and base64 md5sum for anything like 50,000 files for a Google URL list transfer TSV is slow/a lot of work. I think we should make it easy for our users to do this. While file size is duplicated keeping these three fields together means you can extract just the google_transfer field without looking anywhere else.

The href is a different download endpoint and I don't think we are ready to get rid of it. Since it is a subfield (google_transfer.url) I don't think it will conflict with biosample.

src/encoded/types/file.py
gabdank: The schema is "smart " enough to not put None in calculated values

gabdank:
I was not thinking about a conflict between biosample.url and file.url - I was thinking about consistency in schema naming of different values. I don't like the fact URL means a different thing in different places.

keenangraham:
Updated according to @hitz comments in ticket.

PR #2431, opened 2018-10-11
ENCD-9999 add inital and reindex

caseylitton:
CI test failed for no reason, I restarted them.


caseylitton:
For skillz lab 2018

PR #2433, opened 2018-10-12
ENCD-4254 Raise file validation error audit level in exp

p-sud:


src/encoded/audit/experiment.py
```diff
@@ -376,7 +376,10 @@ def audit_experiment_with_uploading_files(value, system, files_structure):
                           'with the status {}.'.format(value['@id'],
                                                        file_object['@id'],
                                                        file_object['status']))
-                yield AuditFailure(category, detail, level='WARNING')
+                if category == 'file validation error':
```
caseylitton: If we add a level variable to the category conditional this can be a little simpler.
```
            category = None
            level = None
            if file_object['status'] in ['upload failed', 'content error']:
                category = 'file validation error'
                level = 'ERROR'
            elif file_object['status'] == 'uploading':
                category = 'file in uploading state'
                level = 'WARNING'
            if category and level:
                detail = ('Experiment {} contains a file {} '
                          'with the status {}.'.format(value['@id'],
                                                       file_object['@id'],
                                                       file_object['status']))
                yield AuditFailure(category, detail, level=level)
```
Checking 'and level' is just for "safety".   I don't know if sending level=None to AuditFailure would crash.  Up to you with that update though.

PR #2437, opened 2018-10-12
ENCD-4142 update assembly 2

weiweizhong2018:


src/encoded/types/shared_calculated_properties.py
```diff
@@ -284,7 +284,17 @@ class CalculatedVisualize:
     def visualize(self, request, hub, accession, assembly, status, files):
         hub_url = urljoin(request.resource_url(request.root), hub)
         viz = {}
-        for assembly_name in assembly:
+        vis_assembly = set()
+        viewable_file_formats = ['bigWig', 'bigBed']
+        viewable_file_status = ['released','in progress']
```
caseylitton: This is a pep8 issue.  Need a space after commas.  
```
viewable_file_status = ['released', 'in progress']
```
Do you have linting running in your texteditor?  It's tough to use since there are so many pep8 issues preexisting in our files, but we can go over how it can be helpful.

Also, since we're now using this list in two different locations, we should add it at the top of the file as a constant.
```
VIEWABLE_FILE_STATUS = ['released','in progress']
```

src/encoded/types/shared_calculated_properties.py
```diff
@@ -284,7 +284,17 @@ class CalculatedVisualize:
...
+
```
caseylitton: This is a python thing, but in general don't add empty space in python functions.   It's a visual thing.  Two spaces between functions, no spaces in functions.   If it's a class, then one space in between functions.

src/encoded/types/shared_calculated_properties.py
```diff
@@ -284,7 +284,17 @@ class CalculatedVisualize:
...
+        for properties in files:
```
caseylitton: This can be updated to a set comprehension expression.  When to do this?  When creating a list, set, or dict for one time use using ADD conditions

Here we're filtering the files based on some simple AND conditions, putting them a set, and then consuming that set when creating the viz dict.   
```
vis_assembly = {
    properties['assembly']
    for properties in files
    if properties.get('file_format') in viewable_file_formats
    if properties.get('status') in viewable_file_status
    if 'assembly' in properties
}
```
Python's built in syntax for this common programming pattern can run faster with less memory.

I also add dict.get(key).  This is safer than dict[key] because if the key doesn't exist the get method return None.  And there is nothing wrong with checking 'if None in list'.  So long as None is not in the list!

src/encoded/types/shared_calculated_properties.py
forresttanaka: https://www.python.org/dev/peps/pep-0008/#blank-lines

caseylitton: Thanks! @forresttanaka  It's good to have references.  I've never actually read this one.

@weiweizhong2018 Don't try to follow PEP8 word for word.  It is just a set of good team develop practices.  The point is to provide general guidelines to improve readability through code consistency.  Of course in order to accomplish that goal for a wide variety of teams PEP8 will need to be a little ambiguous.  Every instance of a team will need to use PEP8 as a guide to setup their own specific standards.

More details below, if needed.  Don't feel obligated to read it.  :)

For example,
```
Use blank lines in functions, sparingly, to indicate logical sections.
```
PEP8 needs to be vague to allow for the needs of a wide variety of teams.  But we are a specific team with individual team needs.  Senior wranglers, associate wranglers, junior and senior developers will all have a different take on words like 'sparingly' and 'logical sections'.  Given our team, the easiest application of this rule is to say, 'No spaces in functions'.  It's a simple, unambiguous rule that leads to a less chaotic code base.

In PRs, when I see spaces before **return statements**, **for loops**, **if blocks** or after function declarations, I wonder if it was on purpose, by accident, or due to a misunderstanding in how python uses indentation to structure code.  A lot of people learned to program in Java or C++ where spaces were needed for code readability.  Python is a little different and doesn't usually require spaces around common elements because indents are required instead.     

Please put this on the lower end of your coding priorities however.  The most important thing to remember is that someone will be reading and updating your code in the future.  And not to make unnecessary changes.  In the long run this will make our lives easier.

PR #2438, opened 2018-10-13
SNO-47 ENCD-4211 Dump indexer data to json with flags

caseylitton:


caseylitton:
This has been folded into ENCD-4198-cleanup-some-es-files-indexer-init and SNO-57-51-cleanup-indexer

PR #2440, opened 2018-10-16
ENCD-4198 Cleanup some es files indexer init

caseylitton:
Associated with SNO-51 and SNO-57 PRs.  This has ENCD-4209 but that should be merged in v76.  this is going in v77


caseylitton:
This has been replaced , keeping branh though

PR #2442, opened 2018-10-17
ENCD-4205 biosample type

yunhailuo:


yunhailuo:
I don't understand why "BROWSER=Chrome" test fails. I restarted it three times. And the specific one failed (test_nodata) all three times passes locally.

hitz:
What happens when you run the test locally?  It looks like it's failing to set up the server at all.   Actually looks like it's broken since yesterday @caseylitton

yunhailuo:
All Travis tests are in good shape after Saucelabs back up.

caseylitton:
@hitz Yea, It looks fine now, It must not of have the JAVA fix earlier.

src/encoded/static/components/biosample_type.js
```diff
@@ -0,0 +1,166 @@
...
+    auditDetail: PropTypes.func.isRequired, // Audit decorator function
+};
+
+ListingComponent.contextTypes = {
+    session: PropTypes.object, // Login information from <App>
+};
+
+const Listing = auditDecor(ListingComponent);
+
+globals.listingViews.register(Listing, 'BiosampleType');
```
forresttanaka: Need a last carriage return.

src/encoded/static/components/biosample_type.js
```diff
@@ -0,0 +1,166 @@
...
+    ];
+
+    // Get a list of reference links, if any
+    const references = pubReferenceList(context.references);
+
+    return (
+        <div className={itemClass}>
+            <header className="row">
+                <div className="col-sm-12">
+                    <Breadcrumbs root="/search/?type=Biosample" crumbs={crumbs} />
```
forresttanaka: `root` should be "/search/?type=BiosampleType" I think. I don’t think breadcrumbs will show any results with "Biosample" here.

yunhailuo: Right.

yunhailuo:
Rebased onto dev (ENCD-4195) and revived. The last commit 0391b48130834d7641f727ae2cc6c3d22e06e890 is the only new commit from the last code review https://github.com/ENCODE-DCC/encoded/pull/2442#pullrequestreview-166233689 . Other details are in the ticket.

src/encoded/types/biosample_type.py
```diff
@@ -0,0 +1,112 @@
...
+        return self.name()
+
+    @calculated_property(condition='term_ids', schema={
+        "title": "Organ slims",
+        "type": "array",
+        "items": {
+            "type": "string",
+        },
+    })
+    def organ_slims(self, registry, term_ids):
```
caseylitton: Did you consider putting this in a static function?

For example,
```
@staticmethod
def _get_ontology_slims(registry, term_ids, slim_key):
    slims = set()
    for term_id in term_ids:
        if term_id in registry['ontology']:
            slims.update(registry['ontology'][term_id][slim_key])
    return list(slims)

@calculated_property(condition='term_ids', schema={
    ...
})
def organ_slims(self, registry, term_ids):
    return self._get_ontology_slims(registry, term_ids, 'organs')

@calculated_property(condition='term_ids', schema={
    ...
})
def cell_slims(self, registry, term_ids):
    return self._get_ontology_slims(registry, term_ids, 'cells')

etc...
```
We're doing the same pattern 5 times with just a change of key value.

Then try set comprehension too
```
@staticmethod
def _get_ontology_slims(registry, term_ids, slim_key):
    slims = set(
        registry['ontology'][term_id][slim_key]
        for term_id in term_ids
        if term_id in registry['ontology']
    )
    return list(slims) 
```

PR #2443, opened 2018-10-17
ENCD-4266 Stub for SNO-60 report view test bug

zoldello:
This is the ENCODED branch needed for SNO-60


caseylitton:
just a stub for sno-60 no encoded changes.

PR #2444, opened 2018-10-17
made copyig matrix data a deep copy

zoldello:


zoldello:
Cancelled because it is not needed.

PR #2448, opened 2018-10-22
[2nd] ENCD-3159 Remove cart toggle from dataset results

forresttanaka:
Update for already-merged PR. With this update, the cart toggle button only appears with experiment search results. With RC1 they appear on the search results of all dataset-derived types (FileSet, PublicationData, etc). Carts only support experiments. The toggle button already doesn't appear on non-experiment dataset-derived object pages, so that part’s fine.

### DIFF
--- a/src/encoded/static/components/search.js
+++ b/src/encoded/static/components/search.js
@@ -364,7 +364,6 @@ globals.listingViews.register(Experiment, 'Experiment');
 
 
 const DatasetComponent = (props, reactContext) => {
-    const { activeCart } = props;
     const result = props.context;
     let biosampleTerm;
     let organism;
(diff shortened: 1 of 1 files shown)


src/encoded/static/components/search.js
```diff
@@ -452,11 +451,6 @@ const DatasetComponent = (props, reactContext) => {
                         <div><strong>Project: </strong>{result.award.project}</div>
                     </div>
                 </div>
-                {activeCart ?
-                    <div className="result-item__cart-control">
-                        <CartToggle element={result} />
```
emro: I think you can take out CartToggle at the top because this is the only place I see you including that component.

forresttanaka: That’s weird. I already removed CartToggle from the top, but for some reason this PR still shows it.

forresttanaka: Oh wait, that's not right. I’m being confused.

forresttanaka: OK, I’m straightened out. CartToggle is still used in ExperimentComponent in that file. I just removed it from DatasetComponent.

emro: oh, I see, that's right.

PR #2452, opened 2018-10-24
[2nd] ENCD-3998 debug target organism ID

yunhailuo:


src/encoded/types/target.py
```diff
@@ -36,13 +37,13 @@ def unique_keys(self, properties):
...
         "linkTo": "Organism"
     })
-    def organism(self, properties=None):
+    def organism(self, properties=None, return_uuid=False):
         if properties is None:
             properties = self.upgrade_properties()
+        root = find_root(self)
         if 'target_organism' in properties:
-            return properties['target_organism']
+            organism_uuid = properties['target_organism']
```
keenangraham: Do you know if the target_organism is a uuid here? Was assuming it would be @id since it is a specified property. Passing the @id to root.get_by_uuid would return blank result. We should add test for this.

yunhailuo: Two unit tests added for posting non-gene targets with target_organism being @id and uuid respectively. Both cases have the organism property calculated correctly.

yunhailuo:
Added a test on experiment audits to catch this bug as we talked about, @keenangraham .

PR #2454, opened 2018-10-24
REG-54 Get regulome peak details

yunhailuo:


caseylitton:
replaced by https://github.com/ENCODE-DCC/encoded/pull/2457

PR #2459, opened 2018-10-25
ENCD-4267 Remove files redirect from apache config

forresttanaka:
The redirect from /files/ to /search/?type=file caused the /files/#!add mechanism to fail. “files” is the only object type that gets redirected, and according to Bek he thinks Ben wanted this because our old /files/ page killed the system. As our new /files/ page with the charts doesn’t kill the system, we can go back to _not_ redirecting.


keenangraham:
Nothing should rely on this redirect.

PR #2462, opened 2018-10-26
ENCD-4280 Remove table margin

emro:


src/encoded/static/scss/encoded/modules/_facet.scss
```diff
@@ -79,6 +79,14 @@
...
 }
 
+.box.facets {
+    .horizontal {
+        .facet {
+            border-bottom: 0;
+        }
+    }
+}
+
```
forresttanaka: Could this border-bottom be instead added to just after line 22?

emro: Yes, thanks!

PR #2467, opened 2018-10-29
ENCD-4223 Make internal_tags badges clickable to search pages

forresttanaka:
* I replaced a lot of duplicated code with a new InternalTags component that’s in objectutils.js, as the HTML generation is a bit more complicated with the links.
* I feel like these new `<a>` should have title attributes to help explain to sighted users what these buttons will lead to, but the vast majority of a11y sites say to never use title attributes as they might get read by some screen readers, duplicating the alt attribute on the ``, so I didn’t use the title attribute. I’m thinking this is something yet to be worked out in the industry.
* Fixed a couple eslint errors in objectutils.js.
* Removed the internal_tags checks for `<FlyWomDonor>` and `<MouseDonor>` because only HumanDonor has `internal_tags` as a property.

### DIFF
--- a/src/encoded/static/components/donor.js
+++ b/src/encoded/static/components/donor.js
@@ -308,13 +308,6 @@ const MouseDonor = (props) => {
                             </div>
                         : null}
 
-                        {context.internal_tags && context.internal_tags.length > 0 ?
-                            <div className="tag-badges" data-test="tags">
-                                <dt>Tags</dt>
-                                <dd><InternalTags context={context} /></dd>
-                            </div>
-                        : null}
-
                         {context.submitter_comment ?
                             <div data-test="submittercomment">
                                 <dt>Submitter comment</dt>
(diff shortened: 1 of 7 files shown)


src/encoded/static/components/objectutils.js
```diff
@@ -634,3 +634,28 @@ AlternateAccession.propTypes = {
...
+        return <a href={tagSearchUrl} key={tag}><img src={`/static/img/tag-${tag}.png`} alt={`Search for all ${context['@type'][0]} with internal tag ${tag}`} /></a>;
+    });
+    return <span className={css}>{tagBadges}</span>;
+};
+
+InternalTags.propTypes = {
+    /** encode object being displayed */
+    context: PropTypes.object.isRequired,
+    /** Optional CSS class to assign to <span> surrounding all the badges */
+    css: PropTypes.string,
```
emro: It doesn't look like you're using "css" on any of the pages where you're including this component. You think it might be useful in the future, though?

forresttanaka: Almost certainly I used the `css` prop in the past. I figured it could be useful in the future.

src/encoded/static/components/donor.js
```diff
@@ -320,13 +308,6 @@ const MouseDonor = (props) => {
                             </div>
                         : null}
 
-                        {tagBadges ?
-                            <div className="tag-badges" data-test="tags">
-                                <dt>Tags</dt>
-                                <dd>{tagBadges}</dd>
-                            </div>
-                        : null}
-
```
emro: It looks like you're taking the internal tags off of the mouse donor and fly worm donor. We don't want those any more?

forresttanaka: They don’t exist in the mouse and fly worm donor schemas (any more? Not sure if they used to) — only the human schema.

PR #2468, opened 2018-10-29
ENCD-4282 Add PGP internal_tags badge

forresttanaka:
Just adding an image for an existing PGP internal_tags value.


forresttanaka:
As this only involves the addition of an image, assigning code review to @gabdank to judge whether the image is what he wants.

tag-pgp

hitz:
Looks OK to me, as long as the helix is right-handed.   Not 100% on the colors but seems fine.

forresttanaka:
Oh, should say the colors come from the Harvard PGP site that seems to be the highlighted PGP site. https://pgp.med.harvard.edu/

PR #2469, opened 2018-10-30
ENCD-4289 adding new tag and gene_id targeting to GM

gabdank:


caseylitton:
REplaced with https://github.com/ENCODE-DCC/encoded/pull/2491

PR #2472, opened 2018-10-30
ENCD-4278 Submitter-saved carts beta

keenangraham:
Permission modification to make cart object editable by owner but viewable by everyone.

### DIFF
--- a/src/encoded/static/components/cart/clear.js
+++ b/src/encoded/static/components/cart/clear.js
@@ -92,16 +92,17 @@ const mapStateToProps = (state, ownProps) => ({
     user: ownProps.sessionProperties,
 });
 const mapDispatchToProps = (dispatch, ownProps) => ({
-    onClearCartClick: elementAtIds => dispatch(removeMultipleFromCartAndSave(elementAtIds, ownProps.sessionProperties.user, ownProps.sessionProperties.admin, ownProps.fetch)),
+    onClearCartClick: elementAtIds => dispatch(removeMultipleFromCartAndSave(elementAtIds, ownProps.sessionProperties.user, !!(ownProps.session && ownProps.session['auth.userid']), ownProps.fetch)),
 });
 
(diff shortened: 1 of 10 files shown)


src/encoded/static/components/cart/clear.js
```diff
@@ -92,16 +92,17 @@ const mapStateToProps = (state, ownProps) => ({
     user: ownProps.sessionProperties,
 });
 const mapDispatchToProps = (dispatch, ownProps) => ({
-    onClearCartClick: elementAtIds => dispatch(removeMultipleFromCartAndSave(elementAtIds, ownProps.sessionProperties.user, ownProps.sessionProperties.admin, ownProps.fetch)),
+    onClearCartClick: elementAtIds => dispatch(removeMultipleFromCartAndSave(elementAtIds, ownProps.sessionProperties.user, !!(ownProps.session && ownProps.session['auth.userid']), ownProps.fetch)),
```
hitz: I don't understand why this particular replacement of admin->logged in is negated.

hitz: oh only admins are allowed to clear>?

forresttanaka:
The admin flag is in session_properties, but checking for logged-in state for anyone comes from session. Since none of this relies on being an admin anymore, I changed the object I used to decide whether people can do cart stuff or not.

src/encoded/static/components/cart/clear.js
forresttanaka: It’s negated negated, so positive. It’s just a trick to convert something that’s truthy to a real boolean. In this case it probably isn’t required, but I tend to like to have real booleans where booleans are expected — a style thing. I never require this of others, but tend to do that myself.

keenangraham:
Closing in favor of https://github.com/ENCODE-DCC/encoded/pull/2525

PR #2474, opened 2018-10-30
ENCD-4231 matched set controls

forresttanaka:
I moved ExperimentTable and its dependencies from dataset.js to typeutils.js, and I moved ControllingExperiments from experiment.js to typeutils.js. This is to avoid the possibility of having circular imports, and that’s the purpose of typeutils.js — to have React components and functions that are specific for encode object types but could get shared between related types. In this case, both Series and Experiments could both rely on the same ControllingExperiments component, which depends on ExperimentTable.

### DIFF
--- a/src/encoded/static/components/dataset.js
+++ b/src/encoded/static/components/dataset.js
@@ -56,6 +56,7 @@ class AnnotationComponent extends React.Component {
         const context = this.props.context;
         const itemClass = globals.itemClass(context, 'view-item');
         const adminUser = !!(this.context.session_properties && this.context.session_properties.admin);
+        const experimentsUrl = `/search/?type=Experiment&possible_controls.accession=${context.accession}`;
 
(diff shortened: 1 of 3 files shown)


src/encoded/static/components/dataset.js
```diff
@@ -56,6 +56,7 @@ class AnnotationComponent extends React.Component {
         const context = this.props.context;
         const itemClass = globals.itemClass(context, 'view-item');
         const adminUser = !!(this.context.session_properties && this.context.session_properties.admin);
+        const experimentsUrl = `/search/?type=Experiment&possible_controls.accession=${context.accession}`;
```
zoldello: It is not a big deal. It may be a good idea to move this to a class-wide variable. Not just to follow the D.R.Y principle, but to make it  easier to identify this url path in case it changes and it needs to be witch-hunted.

PR #2476, opened 2018-10-31
ENCD-3665-add-file-format-audit

p-sud:


p-sud:
I forgot that id() is a builtin, now fixed

src/encoded/audit/file.py
```diff
@@ -335,6 +335,27 @@ def audit_duplicate_quality_metrics(value, system):
...
+    output_category = value.get('output_category')
+    file_format = value.get('file_format')
+    file_id = value.get('@id')
+    raw_file_formats = ['rcc', 'idat', 'CEL', 'csfasta', 'fastq', 'csqual',
+                        'sra']
+    # These three formats can be either raw or processed, so ignore them
+    if file_format in ['tar', 'tsv', 'fasta']:
+        return
+    if file_format in raw_file_formats and output_category != 'raw data':
+        detail = 'File {} has file format {} '.format(file_id, file_format) + \
```
caseylitton: I realize this is done throughout our code base so it's a copy/paste job, but it's not good string concatenation.   Try to update these issues as you change the lines.  Using slash for line continuation is an old style, parens are current standards for multi line strings.  Also we only need one format call.
```
        detail = (
            'File {} has file format {} corresponding to raw data but '
            'has an output category of {}'.format(
                file_id,
                file_format,
                output_category,
            )
        )
``` 
Please fix here and on the other one below.

src/encoded/audit/file.py
```diff
@@ -335,6 +335,27 @@ def audit_duplicate_quality_metrics(value, system):
                 level='INTERNAL_ACTION'
             )
 
+def audit_file_output_category(value, system):
```
caseylitton: Two spaces between functions unless they are in a class.  Same with below the function.

src/encoded/audit/file.py
```diff
@@ -335,6 +335,27 @@ def audit_duplicate_quality_metrics(value, system):
...
+        detail = 'File {} has file format {} '.format(file_id, file_format) + \
+                 'corresponding to raw data but has an output category ' + \
+                 'of {}.'.format(output_category)
+        yield AuditFailure('inconsistent file format and output category',
+                           detail, level = 'ERROR')
+    elif file_format not in raw_file_formats and output_category == 'raw data':
+        detail = 'File {} has file format {} '.format(file_id, file_format) + \
+                 'corresponding to processed data, but has an output ' + \
+                 'category of raw data'
+        yield AuditFailure('inconsistent file format and output category',
```
caseylitton: Since we're using this string twice we should put it in a var above the conditionals.  Usually hard-code sentences are better in a variable no matter how many times they are used.
```
    raw_file_formats = ['rcc', 'idat', 'CEL', 'csfasta', 'fastq', 'csqual', 'sra']
    audit_msg = 'inconsistent file format and output category' 
    # These three formats can be either raw or processed, so ignore them
    if file_format in ['tar', 'tsv', 'fasta']:
...
...
        yield AuditFailure(audit_msg, detail, level='ERROR')
```
Also note I removed the spaces around level = Error.  Thats a pep8 thing.  No spaces in equal signs in 
 function arguments.

PR #2477, opened 2018-11-01
ENCD-4199 Ignore npm shrinkwrap

caseylitton:


.gitignore
```diff
@@ -15,6 +15,7 @@
 /eggs/
 /extends/
 /node_modules/
+/npm-shrinkwrap.json
```
zoldello: Good idea. It is recommended that this file be generated when needed. See- https://docs.npmjs.com/files/shrinkwrap.json

PR #2481, opened 2018-11-02
ENCD-4252-disabled-download-restricted-files-admin

zoldello:


zoldello:
@forresttanaka We spoke about the adminUser variable. Do you need more work done?

forresttanaka:
At the very least, `adminUser` needs to be removed form `<DownloadIcon>` because the resulting style is wrong, with the download icon not being clickable, but also not appearing grayed as it should.

And from that change, I think you’ll probably have eslint errors from unused variables and props for `adminUser`, so you’ll probably have to remove `adminUser` from various places because it won’t be used anymore.

zoldello:
@forresttanaka  Thanks for the observation. I looked at the line of interest without considering the periphery.

PR #2482, opened 2018-11-02
Rebased regulome dev v77

emro:
rebased and deleted npm-shrinkwrap


caseylitton:
move to fork and we have a new reg master PR

PR #2487, opened 2018-11-05
ENCD-4250 replace synthetic organism with synthetic tag in target

yunhailuo:
1. If target is investigated as "synthetic_tag", it can't have target_organism or genes specified and won't have the calculated property "organism".
2. If target have either target_organism or genes (mutually exclusive), it can't be investigated as "synthetic_tag" and it will have the calculated property "organism".
3. AntibodyLot.lot_reviews.organisms can be an empty list.
4. Experiment audit will consider "synthetic tag" as tag thus won't raise "not tagged antibody". The same change is applied to antibody_characterization audit.
5. There will be changes in the changelog. Some of them are related to ENCD-3998 which hasn't been documented correctly (schema version, etc)


src/encoded/static/components/antibody.js
```diff
@@ -272,14 +282,18 @@ const AntibodyStatus = (props) => {
 
         // Look at all organisms in current lot_review. They go under this lot_review's status
         const statusNode = statusTree[lotReview.status];
+        if (!lotReview.organisms.length) {
```
forresttanaka: For [AirBnb](https://github.com/airbnb/javascript#comparison--shortcuts) compliance, this should be `lotReview.organisms.length === 0`.

yunhailuo: Updated. Thank you.

PR #2488, opened 2018-11-05
ENCD-3877-annotation-download-btn

zoldello:


src/encoded/helpers/helper.py
```diff
@@ -79,19 +79,18 @@ def search_result_actions(request, doc_types, es_results, position=None):
...
 
-    # generate batch download URL for experiments
-    # TODO we could enable them for Datasets as well here, but not sure how well it will work
+    # generate batch download URL for experiments and annotation
     # batch download disabled for region-search results
-    if '/region-search/' not in request.url:
-        #if (doc_types == ['Experiment'] or doc_types == ['Annotation']) and any(
-        if (doc_types == ['Experiment']) and any(
+    # TODO: Can potentially be enabled for dataset...
+    if '/region-search/' not in request.url \
```
caseylitton: Slashes are old style, just wrap it in a set of parens
```
if (
        '/region-search/' not in request.url and
        doc_types in (['Experiment'], ['Annotation']) and
        any(
            bucket['doc_count'] > 0
            for bucket in aggregations['files-file_type']['files-file_type']['buckets']
        )
    ):
    pass
```


and maybe...at the top of the file add a var
```
BATCH_DOWNLOAD_DOC_TYPES = [
    ['Experiment'],
    ['Annotation'],
]
```
so then 
```
if (
        '/region-search/' not in request.url and
        doc_types in BATCH_DOWNLOAD_DOC_TYPES and
        any(
            bucket['doc_count'] > 0
            for bucket in aggregations['files-file_type']['files-file_type']['buckets']
        )
    ):
    pass
```
This would allow us to turn on/off batch_download per doc type by editing a var.  Which could be put in a config ini as opposed to the application code.  

Just a thought, it does change the scope of the JIRA ticket and could be better in the long run

zoldello: Done

caseylitton:
This is stalled on something,  Closing to remove from PR list

PR #2489, opened 2018-11-05
ENCD-4195 ontology update sept18

jahilton:
ENCD-4195
Summary of changes in this branch
* change the organ slim from 'artery' to 'arterial blood vessel'
* new cell_slim
* remove 'stem cell' and 'induced pluripotent stem cell line' biosample_types
* general ontology update


src/encoded/upgrade/biosample.py
```diff
@@ -330,4 +330,23 @@ def biosample_19_20(value, system):
...
+    return
+
+
+@upgrade_step('biosample', '21', '22')
+def biosample_21_22(value, system):
+    # https://encodedcc.atlassian.net/browse/ENCD-3555
+    if value.get('biosample_type') == 'induced pluripotent stem cell line':
+        value['biosample_type'] = 'cell line'
+    if value.get('biosample_type') == 'stem cell':
+        if value.get('biosample_term_name') in ['MSiPS', 'E14TG2a.4', 'UCSF-4', 'HUES9', 'HUES8', 'HUES66', 'HUES65',
```
caseylitton: Why not add the both large lists to shared.py and then import in the three upgrade files like
```
from .shared import (
    ENCODE2_AWARDS, 
    REFERENCES_UUID,  
    BIOSAMPLE_TERM_NAMES,
    BIOSAMLE_TERM_DESCS,
)
```
They looked the exact same to me?

PR #2492, opened 2018-11-08
ENCD-4033 Search page type-specific header (Req SNO-66)

forresttanaka:
* The overriding change is I now do a GET on a new endpoint from SNO-66 that gives us a mapping of `@type` for all possible objects to their corresponding schema “title” property so we can display that for each type=something on search, matrix, and summary pages.
* App.componentDidMount now does a conditionless fetch, so the server-render-test now supplies a mocked fetch just so <App> doesn’t crash on that test.
* None of the search-related pages had an `<h1>` element, so since I was making all the search titles consistent, I changed all of them to use `<h1>` as well.

Requires SNO-66-profiles-titles.


src/encoded/viewconfigs/matrix.py
```diff
@@ -44,10 +44,7 @@ def __init__(self, context, request):
     @staticmethod
     def _set_result_title(type_info):
         '''Helper function for class and child classes'''
-        if type_info.name == 'Annotation':
-            title = 'Encyclopedia'
-        else:
-            title = type_info.name + ' Matrix'
+        title = type_info.name + ' matrix'
```
zoldello: This is a method children classes can use. It may be  good idea to sanitize (like check for falsey) and check for None (null)...

zoldello: This may be a question for Jason/Idan. ENCD-4244 enables users to view page: audit/type=Annotation. Should that page have "matrix" in its title (see auditview.py)

src/encoded/static/scss/encoded/modules/_summary.scss
```diff
@@ -10,8 +10,7 @@
         margin-bottom: 30px;
 
         h1 {
-            margin: 0 0 5px;
-            font-size: 1.7rem;
+            @include facet-title;
```
zoldello: Great idea.

src/encoded/static/components/objectutils.js
```diff
@@ -634,3 +634,43 @@ AlternateAccession.propTypes = {
...
+    // Determine the search page doc_type title to display at the top of the facet list.
+    let facetTitle = '';
+    const docTypes = searchResults.filters.length > 0 ? searchResults.filters.filter(searchFilter => searchFilter.field === 'type') : [];
+    if (docTypes.length === 1) {
+        facetTitle = reactContext.profilesTitles[docTypes[0].term];
+    }
+
+    // If one object type was requested, render it *into* the given wrapper component if one was
+    // given.
+    if (facetTitle) {
```
zoldello: Is this a good alternative, whereby you check if you can do something, if you cannot, do not try and return, otherwise try:

if (!facetTitle) {
 return null;
}
￼
const facetTitleComponent = <span>{facetTitle}</span>;
￼return wrapper ? wrapper(facetTitleComponent) : facetTitleComponent;

forresttanaka: Back when I was in college, it was generally bad form to return from the middle of a function. Things have changed, but I generally prefer to return only at the end if there’s no functional difference.

PR #2493, opened 2018-11-09
ENCD-4279 Biosample preservation

weiweizhong2018:


jahilton:
I think this should get a new test or be incorporated into one of the existing biosample_summary tests.

src/encoded/types/experiment.py
```diff
@@ -223,6 +223,7 @@ def biosample_summary(self,
                                 biosampleObject.get('post_treatment_time'),
                                 biosampleObject.get('post_treatment_time_units'),
                                 treatment_objects_list,
+                                None,
```
jahilton: Can you explain the purpose of the addition of this line?

gabdank: this line is to eliminate the specification of preservation method in the experiment biosample summary. It was added per my request, trying to keep the experiment level summaries terse.

weiweizhong2018: The generate_summary_dictionary method in biosample is changed to have preservation_method as an input parameter. Here in the experiment class, this method is called to display biosample summary in the experiment summary. Therefore, we need to supply a value for the preservation_method parameter here. Idan said that he does not want preservation_method to be displayed in biosample.summary of the experiment, so it is set to 0. I am going to change it to two lines of code: preservation_method=None and then pass preservation_method as a parameter.

weiweizhong2018: I added preservation_method in test_biosample_summary.

jahilton: Thank you for clarifying - makes sense (especially with the code update)

PR #2504, opened 2018-11-15
ENCD-4330 Stop using _all for indexing

caseylitton:


src/encoded/viewconfigs/matrix.py
```diff
@@ -61,7 +62,7 @@ def _construct_query(self):
             del query['query']['query_string']
         else:
             query['query']['query_string']['fields'].extend(
-                ['_all', '*.uuid', '*.md5sum', '*.submitted_file_name']
+                [RESOURCES_INDEX, '*.uuid', '*.md5sum', '*.submitted_file_name']
```
keenangraham: These should be field names not indices names?

keenangraham:
I think the confusing thing here is that `_all` is being used differently depending on if it is Python ES API or ES mapping? Right @hitz?

src/encoded/viewconfigs/matrix.py
hitz: no it's chr indexes that has to be skipped.

keenangraham: Yeah but these are `fields` not indices.

caseylitton: reverting this change.

caseylitton:
Encoded does not need any changes associated with SNO-68.  Keeping branch to set snovault change.

PR #2509, opened 2018-11-19
ENCD-4338 Allow repo to be specified on deploy

caseylitton:


src/encoded/commands/deploy.py
```diff
@@ -558,6 +561,8 @@ def hostname(value):
         help="Deploy to production AWS")
     parser.add_argument('--availability-zone', default='us-west-2a',
         help="Set EC2 availabilty zone")
+    parser.add_argument('--git-repo', default='https://github.com/ENCODE-DCC/encoded.git',
+            help="Git repo to checkout branches")
```
hitz: Maybe should show example with http://*.git

PR #2510, opened 2018-11-19
ENCD-4250 synthetic tag again

yunhailuo:
This PR is a retry on #2487 which is pulled out of v78 by #2501 . The first commit is to revert #2501 . The revert should be clean, and if so there should be no need to review those changes since it has been reviewed.

The second commit fixes three bugs mentioned during v78rc1 QA by Idan:
1. "synthetic tag" fails to exclude target_organism (ENCD-4250)
2. "modifications" depends on "genes" (ENCD-4250)
3. "synthetic tag" objects have the calculated title property (ENCD-4295)
It would need code review and QA.

For the third bug specifically, I removed `condition='organism'` on the title property since "synthetic tag" objects won't have organism. All targets seem like still have titles. This might due to the fact that "organism" has `define=True` and thus is calculated before title. I'm not quite sure so I'd like to have @keenangraham to take a look here.


src/encoded/schemas/changelogs/target.md
```diff
@@ -1,11 +1,16 @@
...
+### Schema version 9
 
-* Move status property to standard_status mixin.
 * Added "genes" property for the list of genes targeted by assay or antibody associated with the target.
+* Removed the gene_name property.
 * The "organism" property is now a calculated property. The organism of a target is either specified directly through the new "target_organism" property or defined indirectly by its target genes. The "target_organism" property and the "genes" property are mutually exclusive.
 * Added "modification" property to represent modification(s) made to the corresponding wild-type gene product.
+* Added "synthetic tag" and dependency to address synthetic targets (FLAG) properly.
+
+### Schema version 8
```
keenangraham: From the diff it looks like some of these properties were added to version 8 at some point. FYI you can add new changes that didn't require upgrade using `Minor changes` header to keep them separate from the changes that did require a version bump, e.g. https://github.com/ENCODE-DCC/encoded/blob/2445f38faeb454fc12fa5091c85139cce928c613/src/encoded/schemas/changelogs/analysis_step_version.md#minor-changes-since-schema-version-4

src/encoded/tests/test_audit_replicate.py
```diff
@@ -65,8 +65,8 @@ def test_audit_inconsistent_modification_tag(
         library_1):
     testapp.patch_json(experiment['@id'], {'assay_term_name': 'ChIP-seq',
                                            'target': target_H3K27ac['@id']})
-    testapp.patch_json(target['@id'], {'investigated_as': ['recombinant protein'],
-                                       'label': 'FLAG'})
+    testapp.put_json(target['@id'], {'investigated_as': ['synthetic tag'],
```
keenangraham: Do you really mean to `put` here?

src/encoded/tests/test_schema_target.py
```diff
@@ -187,3 +195,26 @@ def test_target_organism_uuid(testapp, human):
...
+def test_synthetic_tag_target(testapp, target_synthetic_tag, human, ctcf):
+    flag_target = testapp.post_json(
+        '/target', target_synthetic_tag
+    ).json['@graph'][0]
+    assert flag_target['title'] == 'FLAG (Synthetic tag)'
+
+    syn_and_org = {'target_organism': human['uuid']}
+    syn_and_org.update(target_synthetic_tag)
+    res = testapp.post_json('/target', syn_and_org, expect_errors=True)
+    assert res.json['status'] == 'error'
```
keenangraham: Use more specific error if possible. If just validation error you can say something like `testapp.post_json('/target', syn_and_org, status=422)` instead and get rid of assert statement.

src/encoded/tests/test_schema_target.py
```diff
@@ -187,3 +195,26 @@ def test_target_organism_uuid(testapp, human):
...
+    syn_and_org['investigated_as'] = ['synthetic tag', 'tag']
```
keenangraham: This just tests if synthetic tag still triggers dependency issue if part of a group?

src/encoded/tests/test_schema_target.py
```diff
@@ -187,3 +195,26 @@ def test_target_organism_uuid(testapp, human):
...
+    res = testapp.post_json('/target', syn_and_org, expect_errors=True)
+    assert res.json['status'] == 'error'
+
+    syn_and_gene = {'genes': [ctcf['uuid']]}
+    syn_and_gene.update(target_synthetic_tag)
+    res = testapp.post_json('/target', syn_and_gene, expect_errors=True)
+    assert res.json['status'] == 'error'
+    syn_and_gene['investigated_as'] = ['synthetic tag', 'tag']
+    res = testapp.post_json('/target', syn_and_gene, expect_errors=True)
+    assert res.json['status'] == 'error'
```
keenangraham: Could also add test of posting target_organism and genes together, which are mutually exclusive?

src/encoded/types/antibody_lot.py
```diff
@@ -165,9 +165,10 @@ def lot_reviews(characterizations, targets, request):
         if 'histone' in target['investigated_as']:
             is_histone = True
 
-        organism = target['organism']
+        organism = target.get('organism')
```
keenangraham: What if this returns None and gets store in `target_organisms[organism] = target['@id']` below?

src/encoded/types/antibody_lot.py
```diff
@@ -165,9 +165,10 @@ def lot_reviews(characterizations, targets, request):
...
         target_organisms = {'all': []}
-        target_organisms['all'].append(organism)
+        if organism:
+            target_organisms['all'].append(organism)
```
keenangraham: Why are we even using `all`?

src/encoded/types/target.py
```diff
@@ -80,13 +86,19 @@ def __name__(self):
...
+            source = organism.upgrade_properties()['name']
+        return u'{}-{}'.format(properties['label'], source)
 
     def __resource_url__(self, request, info):
         request._linked_uuids.add(str(self.uuid))
         # Record organism uuid in linked_uuids so linking objects record
         # the rename dependency.
-        request._linked_uuids.add(str(self.organism(return_uuid=True)))
+        organism_uuid = self.organism(return_uuid=True)
+        if organism_uuid:
```
keenangraham: I would probably use `if not organism_uuid` in the two checks above instead of `if organism_uuid is None` in order to make it consistent with this. I don't think you'll run into any problems by not checking specifically for None.

src/encoded/types/target.py
```diff
@@ -80,13 +86,19 @@ def __name__(self):
 
     def _name(self, properties):
         root = find_root(self)
-        organism = root.get_by_uuid(self.organism(properties=properties, return_uuid=True))
-        organism_props = organism.upgrade_properties()
-        return u'{}-{}'.format(properties['label'], organism_props['name'])
+        organism_uuid = self.organism(properties=properties, return_uuid=True)
+        if organism_uuid is None:
+            source = properties['investigated_as'][0].replace(' ',  '_')
```
keenangraham: Do we always want first item, and is this deterministic ordering? I guess investigated_as is required, so this won't be empty at least.

src/encoded/types/target.py
```diff
@@ -65,13 +67,17 @@ def organism(self, properties=None, return_uuid=False):
...
         "title": "Title",
         "type": "string",
     })
-    def title(self, request, label, organism):
-        organism_props = request.embed(organism, '@@object')
-        return u'{} ({})'.format(label, organism_props['scientific_name'])
+    def title(self, request, label, organism, investigated_as):
+        if organism is None:
+            # ENCD-4250 investigated_as must be just ['synthetic tag']
+            source = investigated_as[0].capitalize()
```
keenangraham: What's preventing investigated as being patched as e.g. `["transcription factor", "synthetic tag"]`?

src/encoded/types/target.py
```diff
@@ -54,6 +54,8 @@ def organism(self, properties=None, return_uuid=False):
                 )
                 raise ValidationFailure('body', ['genes'], msg)
             organism_uuid = next(iter(organism_uuids))
+        else:
+            return None
```
keenangraham: This doesn't show up in JSON object as `organism: None` right?

src/encoded/tests/test_schema_target.py
yunhailuo: Great suggestion! If just for successful posting, I can also get rid of `assert res.json['status'] == 'success'`, right?

src/encoded/audit/replicate.py
```diff
@@ -60,8 +60,8 @@ def audit_inconsistent_modifications_tag(value, system):
         if len(tags_names) > 0:
             antibody_targets = get_ab_targets(value)
             for ab_target in antibody_targets:
-                if 'recombinant protein' in ab_target['investigated_as'] or \
-                   'tag' in ab_target['investigated_as']:
+                if any(i in ab_target['investigated_as']
```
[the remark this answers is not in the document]
yunhailuo: Change back to the old pattern to avoid confusions and minimize changes

src/encoded/schemas/changelogs/target.md
yunhailuo: It may not make sense here but the previous change (ENCD-3998) should have a version bump (we do have `target_8_9`) but I forgot to update the changelog.

src/encoded/tests/test_audit_replicate.py
yunhailuo: Yes. It needs to get rid of the `target_organism` property in target.

src/encoded/tests/test_schema_target.py
yunhailuo: Correct.

src/encoded/tests/test_schema_target.py
yunhailuo: Test added.

src/encoded/types/antibody_lot.py
yunhailuo: Actually, I'd like to have a `None` here so that I know it has a synthetic target. But it triggers an error during indexing on local launch. These codes here (including the one below) work but don't really make sense and would be addressed in ENCD-4299. My plan is to have this in before ENCD-4299.

src/encoded/types/antibody_lot.py
yunhailuo: As mentioned above, it doesn't make sense and will be addressed in ENCD-4299.

src/encoded/types/target.py
yunhailuo: Updated.

src/encoded/types/target.py
yunhailuo: The way "investigated_as" work is if it has "synthetic tag", it cannot have other investigated_as and it won't have target_organism or genes thus no organism calculated. In another word, if a target doesn't have organism calculated, it should be `properties['investigated_as'] == ['synthetic tag']`.

src/encoded/types/target.py
yunhailuo: As mentioned above, a target should not have `["transcription factor", "synthetic tag"]`. A test is added for this patch.

src/encoded/types/target.py
yunhailuo: Correct. Won't have the "organism" calculated property at all.

PR #2514, opened 2018-11-19
ENCD-4272-no-add-treatments

keenangraham:
Remove lab from treatments to prevent submitters from adding/editing.

### DIFF
--- a/src/encoded/schemas/treatment.json
+++ b/src/encoded/schemas/treatment.json
@@ -28,13 +28,6 @@
         "schema_version": {
             "default": "10"
         },
-        "lab": {
-            "title": "Lab",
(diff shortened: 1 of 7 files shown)


src/encoded/schemas/treatment.json
```diff
@@ -26,14 +26,7 @@
...
         "schema_version": {
-            "default": "10"
-        },
-        "lab": {
-            "title": "Lab",
-            "description": "Lab associated with the submission.",
-            "comment": "See lab.json for list of available identifiers.",
-            "type": "string",
-            "linkTo": "Lab"
+            "default": "11"
```
hitz: Oh this is hard to read with gitdiff.   It's correct though.

PR #2520, opened 2018-11-21
ENCD-4334 add status facet

zmyers23:
Added the status facet to Experiment objects in Matrix view and AntibodyLot objects in Search view.


zmyers23:
Removed commit adding the status facet to AntibodyLot search page.

PR #2524, opened 2018-11-26
ENCD-4306-remove-improper-styling

zoldello:


src/encoded/static/scss/bootstrap/_forms.scss
```diff
@@ -124,7 +124,6 @@ output {
   border: 1px solid $input-border;
   border-radius: $input-border-radius;
   @include box-shadow(inset 0 1px 1px rgba(0,0,0,.075));
-  @include transition(border-color ease-in-out .15s, box-shadow ease-in-out .15s);
```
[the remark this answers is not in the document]
zoldello: Great catch. I am used to bootstrap being used as a single huge file rather than using component-files; so this slipped past me. I did an approach-reset. I just reset transition to default values only on the input text box of interest and not the entire site.

zoldello:
@forresttanaka I made a change to this, so it needs to be re-reviewed.

src/encoded/static/scss/encoded/modules/_forms.scss
```diff
@@ -197,4 +198,4 @@ input[type="text"], input[type="password"], input[type="date"], input[type="date
 	border: none;
 	background: transparent;
 	color: #428bca;
-}
+}
```
forresttanaka: Lost a CR?

zoldello: Fixed.

PR #2528, opened 2018-11-27
ENCD-4312 genetic modification audit

p-sud:


src/encoded/audit/genetic_modification.py
```diff
@@ -0,0 +1,34 @@
...
+                missing_reagents = True
+        elif method == 'RNAi':
+            if not value.get('rnai_sequences'):
+                missing_reagents = True
+        elif method == 'TALEN':
+            missing_reagents = True
+        elif not value.get('documents'):
+            missing_reagents = True
+
+    if missing_reagents:
```
caseylitton: This conditional block can be indented to be under the "if not value.get('reagents'):" since missing reagents will always be False. 

If we're doing that then "missing_reagents = False" can be brought under "if not value.get('reagents'):" too.

src/encoded/audit/genetic_modification.py
```diff
@@ -0,0 +1,34 @@
...
+        detail = 'Genetic modification {} of method {} is missing reagents'.format(
+            value['@id'],
+            method)
```
caseylitton: When expanding lines drop the last paren (or bracket or whatever) to the next line.
```
'Genetic modification {} of method {} is missing reagents'.format(
    value['@id'],
    method,
)
```
I think it is more common to find code like the above then all the other arbitrary ways people might expand lines that get long.

src/encoded/audit/genetic_modification.py
```diff
@@ -0,0 +1,34 @@
...
+            if not value.get('guide_rna_sequences'):
+                missing_reagents = True
+        elif method == 'RNAi':
+            if not value.get('rnai_sequences'):
+                missing_reagents = True
+        elif method == 'TALEN':
+            missing_reagents = True
+        elif not value.get('documents'):
+            missing_reagents = True
+
```
caseylitton: No need for spaces like this in python.  The required indents create "natural" visual cues, so we don't need spaces in general.

src/encoded/audit/genetic_modification.py
```diff
@@ -0,0 +1,34 @@
...
+def audit_genetic_modification_reagents(value, system):
+    '''
+    Genetic modifications with missing reagents will be flagged by this audit.
+    Modifications with acceptable substitutes for reagents will not be flagged,
+    depending on their method. See the method dependecy in file.json for more
+    detail.
+    '''
+    missing_reagents = False
+
+    if not value.get('reagents'):
```
caseylitton: See if you can combine all of the conditionals in 'any'
```
if not value.get('reagents'):
    method = value['method']
    if any([
            method == 'CRISPR' and not value.get('guide_rna_sequences'),
            method == 'RNAi' and not value.get('rnai_sequences'),
            method == 'TALEN',
            not value.get('documents'),
        ]):
        detail = 'Genetic modification {} of method {} is missing reagents'.format(
            value['@id'],
            method,
        )
        yield AuditFailure('missing genetic modification reagents', detail, level='ERROR')
```
Another python built in is 'all'.  Which would require all the conditions to be true.  'any' will return true after finding the first true condition.  Don't trust the code above, I didn't test it.

p-sud: It would be nicer if I could combine the conditionals into an 'any', but checking for 'documents' should only be performed for methods other than 'CRISPR', 'TALEN', and 'RNAi'. The logic above would return True if the 'method' is CRISPR and 'value' had 'guide_rna_sequences' but not 'documents', which is not the desired behavior.

src/encoded/audit/genetic_modification.py
p-sud: That makes sense, I'll make those two changes.

src/encoded/audit/genetic_modification.py
caseylitton: nice

PR #2530, opened 2018-11-28
ENCD-4273 Inconsistent antibody handling

forresttanaka:


src/encoded/static/scss/encoded/modules/_breadcrumbs.scss
```diff
@@ -4,9 +4,8 @@
     border-bottom: 1px solid #c0c0c0;
 
     > li {
-        text-transform: uppercase;
         font-weight: bold;
-        font-size: 12px;
+        font-size: 1rem;
```
yunhailuo: Don't really know how to review this file. Looks fine and may not need a review. But just to be sure, this is good, right?

forresttanaka: With lower-case breadcrumbs, the font seemed a bit small. Also I’ve been converting from px to rem units, so this was a good time to convert that one.

PR #2532, opened 2018-11-28
ENCD-3508 add requirement of zygosity for TALENs

p-sud:


p-sud:
@jahilton sure thing

p-sud:
Code changes from this ticket are being moved into ENCD-4312

PR #2535, opened 2018-11-30
ENCD-4323 Promote released objects

emro:
Change links in navigation to display released objects


src/encoded/static/components/antibody.js
```diff
@@ -105,11 +105,13 @@ const LotComponent = (props, reactContext) => {
         { id: geneComponents.length > 0 ? geneComponents : null, query: geneQuery, tip: geneTips.join(' + ') },
     ];
 
+    const crumbsReleased = (context.status === "released");
```
forresttanaka: This will apply to almost every breadcrumb change: strings (but not HTML attribute values) have to use single quotes: https://github.com/airbnb/javascript#strings

src/encoded/static/components/navigation.js
```diff
@@ -301,13 +301,17 @@ export const Breadcrumbs = (props) => {
                     accretingTip += crumb.tip ? (accretingTip.length ? ' and ' : '') + crumb.tip : '';
                 }
 
+                // Append released tag to link and link title if the object is released itself
+                const releasedLink = (props.crumbsReleased) ? '&status=released' : '';
```
forresttanaka: Have to add `crumbsReleased` to propTypes: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/prop-types.md

PR #2537, opened 2018-12-04
ENCD-4014 library restriction enzyme audit

weiweizhong2018:


src/encoded/audit/experiment.py
```diff
@@ -48,6 +48,34 @@
...
+    if value['status'] in ['deleted', 'replaced', 'revoked']:
+        return
+    if 'replicates' not in value:
+        return
+    frag_methods = set()
+    using_restriction_enzyme = 0
+    for rep in value['replicates']:
+        if rep.get('status') not in excluded_types and 'library' in rep:
+            lib = rep['library']
+            if lib.get('status') not in excluded_types and \
```
caseylitton: Slashes are old style.  Just use ()
```
if (lib.get('status') not in excluded_types and
    'fragmentation_method' in lib):
```
and for strings too
```
 detail = (
    'Experiment {} '.format(value['@id'])
    'contains libraries with inconsistant restriction enzymes {} '.format(
        frag_methods
    )
)
```

src/encoded/audit/experiment.py
```diff
@@ -48,6 +48,34 @@
...
+                    'fragmentation_method' in lib:
+                if 'restriction' in lib['fragmentation_method']:
+                    using_restriction_enzyme = 1
+                frag_methods.add(lib['fragmentation_method'])
+    if len(frag_methods) > 1 and using_restriction_enzyme > 0:
+        detail = 'Experiment {} '.format(value['@id']) + \
```
caseylitton: Don't mix string functions.   I know this pattern is all over our code base so try to fix it when updating.

The detail can be written with one string format function 
```
detail = 'Experiment {} contains libraries with inconsistent restriction enzymes {}'.format(
    value['@id'],
    frag_methods
)
```
or 
```
detail = (
    'Experiment {} contains libraries with '
    'inconsistent restriction enzymes {}'.format(
        value['@id'],
        frag_methods,
    )
)
```

src/encoded/audit/experiment.py
```diff
@@ -3676,6 +3704,7 @@ def check_award_condition(experiment, awards):
     'audit_library_biosample': audit_experiment_library_biosample,
     'audit_target': audit_experiment_target,
     'audit_mixed_libraries': audit_experiment_mixed_libraries,
+    'audit_HiC_restriction_enzyme_in_libaries': audit_HiC_restriction_enzyme_in_libaries,
```
caseylitton: don't forget to update this when making function name lowercase

src/encoded/audit/experiment.py
```diff
@@ -48,6 +48,34 @@
...
+    if value['assay_term_name'] != 'HiC':
+        return
+    if value['status'] in ['deleted', 'replaced', 'revoked']:
+        return
+    if 'replicates' not in value:
+        return
+    frag_methods = set()
+    using_restriction_enzyme = 0
+    for rep in value['replicates']:
+        if rep.get('status') not in excluded_types and 'library' in rep:
```
caseylitton: A couple things,

'lib' is not a good variable name.  Common names should be avoided and 3 letters is the bare minimum.  Maybe 'rep_lib' if I wanted it to be terse.  But really I would have done 'rep' as ' value_rep' and 'lib' as 'value_rep_lib' because then it's clear where the variable comes from in the nested code block.  Readability is a high priority for code.  I don't care if you change this here but just FYI, Optimize for readability.

Also, I would have written this more explicitly.  
```
rep_lib = rep.get('library')
rep_status = rep.get('status')
if rep_lib and rep_status and rep_status not in excluded_types:
   ...etc
```
The current code leaves some questions.   Do we want these situations to pass
- if rep['library'] happens to be None or zero or empty list or empty dict? 
- If None is in excluded_types and rep['status'] is None or 'status' is not in rep?

We may want these cases to succeed but I wouldn't do it loosely.   

The pattern of putting the output of dict.get('key') into a var and then saying 'if var' handles the cases where the key doesn't exists or the value for the key is set to (None, False, 0, or empty object).  It's better to code to this standard.

PR #2538, opened 2018-12-04
ENCD-4353-refactor-query-and-matrix-from-baseview

zoldello:


caseylitton:
does not adhere to JIRA ticket.

PR #2539, opened 2018-12-04
ENCD-4353-Refactor-out-matrix-and-result

zoldello:
Only functional code changes.

### DIFF
--- a/src/encoded/viewconfigs/matrix.py
+++ b/src/encoded/viewconfigs/matrix.py
@@ -34,9 +34,6 @@ class MatrixView(BaseView):  #pylint: disable=too-few-public-methods
     ]
     def __init__(self, context, request):
         super(MatrixView, self).__init__(context, request)
-        if self._view_name == 'matrix':  # hack
-            self._result['matrix'] = ''
-            self._matrix = ''
         self._view_item = View_Item(request, self._search_base)
         self._facets = []
         self._schema = None
@@ -47,7 +44,7 @@ class MatrixView(BaseView):  #pylint: disable=too-few-public-methods
         title = type_info.name + ' matrix'
         return title
 
-    def _construct_query(self):
+    def _construct_query(self, result_filters, matrix_x_y):
         '''Helper method for preprocessing view'''
         search_fields, _ = get_search_fields(self._request, self._doc_types)
         query = get_filtered_query(
@@ -63,7 +60,7 @@ class MatrixView(BaseView):  #pylint: disable=too-few-public-methods
             query['query']['query_string']['fields'].extend(
                 ['_all', '*.uuid', '*.md5sum', '*.submitted_file_name']
             )
-        used_filters = self._set_query_aggs(query)
+        used_filters = self._set_query_aggs(query, result_filters, matrix_x_y)
         return query, used_filters
 
     def _construct_result_views(self, type_info):
@@ -76,10 +73,11 @@ class MatrixView(BaseView):  #pylint: disable=too-few-public-methods
             views.append(self._view_item.summary_report)
         return views
 
-    def _construct_xygroupings(self, query, filters, negative_filters):
+    # pylint: disable=no-self-use
+    def _construct_xygroupings(self, query, filters, negative_filters, matrix_x_y):
         '''Helper Method for constructing query'''
-        x_grouping = self._matrix['x']['group_by']
-        y_groupings = self._matrix['y']['group_by']
+        x_grouping = matrix_x_y['x']['group_by']
+        y_groupings = matrix_x_y['y']['group_by']
         x_agg = {
             "terms": {
                 "field": 'embedded.' + x_grouping,
@@ -128,14 +126,14 @@ class MatrixView(BaseView):  #pylint: disable=too-few-public-methods
         if msg:
             raise HTTPBadRequest(explanation=msg)
 
-    def _set_query_aggs(self, query):
+    def _set_query_aggs(self, query, result_filters, matrix_x_y):
         '''Helper Method for constructing query'''
         query_filters = query['post_filter'].pop('bool')
         filter_collector = {'post_filter': {'bool': query_filters}}
         used_filters = set_filters(
             self._request,
             filter_collector,
-            self._result,
+            result_filters,
             filter_exclusion=self._filter_exclusion
         )
         filters = filter_collector['post_filter']['bool']['must']
@@ -144,8 +142,8 @@ class MatrixView(BaseView):  #pylint: disable=too-few-public-methods
             (field, facet)
             for field, facet in self._schema['facets'].items()
             if (
-                field in self._matrix['x']['facets'] or
-                field in self._matrix['y']['facets']
+                field in matrix_x_y['x']['facets'] or
+                field in matrix_x_y['y']['facets']
                 )
         ]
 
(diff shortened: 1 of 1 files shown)


src/encoded/viewconfigs/matrix.py
```diff
@@ -202,50 +196,60 @@ def preprocess_view(self):
...
+            'filters': [],
+            '@id': matrix_route + self._search_base,
+            '@type': ['Matrix'],
+            'notification': '',
+            'title': self._set_result_title(type_info),
+            'matrix': matrix,
+            'view': self._construct_result_views(type_info),
+        }
+        search_query, used_filters = self._construct_query(search_result, matrix)
+        es_results = self._elastic_search.search(
```
zoldello: You mentioned this in the ticket:

Second Note: Move see how much of the results creation could be moved below the es query search line. https://github.com/ENCODE-DCC/encoded/blob/dev/src/encoded/viewconfigs/matrix.py#L216

There is a Texas standoff situation where es_results needs search_result and search_result needs es_results; breaking this standoff could be tricky and not necessarily pretty. I recommend we leave it.

caseylitton: Because of 'construct_query'?  I suggested looking into to why construct_query needs search_results and how that dependency was impacted by ENCD-4345/SNO-69.  I believe we've already fixed this.

But good job.  This is exactly what I wanted you to find per the Second Note.  A important step in cleaning up the back end matrix code is identifying unnecessary dependencies so we can fix them.  I think this is the line of code that lead us to correcting the set_filters via ENCD4345/SNO-69.

zoldello: I might can move the es_results above search_result. However, I think it would be better if I do that in a different ticket. The changes are tangential to ENCD-4353 (this ticket). 

How important is that to you? Two different tickets perform refactoring on matrix.py and adding a third ticket to the mix is doable if it is important.

caseylitton: The only reason we're handing in 'search_result' to '_construct_query' is because 'set_filters' may add a 'filters' key with a list of terms dicts.  Let's make this dependency more clear by not passing in the final search_results dict.  Just pass in a dict with a 'filters' key.  

This will allow us to move the construction of the final search results down lower in the function.

src/encoded/viewconfigs/matrix.py
```diff
@@ -202,50 +196,60 @@ def preprocess_view(self):
...
+            index=self._es_index
+        )
         aggregations = es_results['aggregations']
         total = aggregations['matrix']['doc_count']
-        self._result['matrix']['doc_count'] = total
-        self._result['matrix']['max_cell_doc_count'] = 0
-        self._result['facets'] = self._format_facets(
+        search_result['matrix']['doc_count'] = total
+        search_result['matrix']['max_cell_doc_count'] = 0
+        search_result['facets'] = self._format_facets(
```
caseylitton: Don't add the facets to the search_results since it won't exist at this point due to other change.

src/encoded/viewconfigs/matrix.py
```diff
@@ -202,50 +196,60 @@ def preprocess_view(self):
...
-            y_groupings + [x_grouping]
+            y_groupings + [x_grouping],
+            matrix
         )
-        self._result['matrix']['y'][y_groupings[0]] = aggregations['matrix'][y_groupings[0]]
-        self._result['matrix']['x'].update(aggregations['matrix']['x'])
-        self._result.update(
+        search_result['matrix']['y'][y_groupings[0]] = aggregations['matrix'][y_groupings[0]]
+        search_result['matrix']['x'].update(aggregations['matrix']['x'])
+        search_result.update(
```
caseylitton: Also don't update search result since it doesn't exist yet.   Just store the actions returned from 'search_result_actions' in a var.

src/encoded/viewconfigs/matrix.py
```diff
@@ -202,50 +196,60 @@ def preprocess_view(self):
...
             search_result_actions(
                 self._request,
                 self._doc_types,
                 es_results
             )
         )
-        self._result['total'] = es_results['hits']['total']
-        if self._result['total']:
-            self._result['notification'] = 'Success'
+        search_result['total'] = es_results['hits']['total']
```
caseylitton: Now that we've collected everything we need to make search_results, build it here with remaining keys.

src/encoded/viewconfigs/matrix.py
```diff
@@ -128,14 +125,14 @@ def _validate_items(self, type_info):
         if msg:
             raise HTTPBadRequest(explanation=msg)
 
-    def _set_query_aggs(self, query):
+    def _set_query_aggs(self, query, search_result, matrix):
```
caseylitton: Let try removing the matrix dependency here too.  Put this in a separate commit than the search_results change in the other comments.  

_set_query_aggs only needs a single list for 
```
field in matrix['x']['facets'] or
                field in matrix['y']['facets']
```
and these two things 
```
        x_grouping = matrix['x']['group_by']
        y_groupings = matrix['y']['group_by']
```
for _construct_xygroupings

src/encoded/viewconfigs/matrix.py
zoldello: Let me know if this issue still exist. I refactored but I could not see a reason to move this.

src/encoded/viewconfigs/matrix.py
zoldello: Let me know if the principle behind this still exist. I could not see a reason to move it after the refactor

src/encoded/viewconfigs/matrix.py
zoldello: I forgot move this to a separate commit. Is ther separate commit important to you?

src/encoded/viewconfigs/matrix.py
```diff
@@ -202,50 +197,68 @@ def preprocess_view(self):
...
-        es_results = self._elastic_search.search(body=query, index=self._es_index)
+        matrix['search_base'] = search_route + self._search_base
+        matrix['clear_matrix'] = matrix_route + '?type=' + self._doc_types[0]
+        matrix_x_y = {
+            'x': matrix['x'],
+            'y': matrix['y'],
+        }
+        search_query, used_filters = self._construct_query(
+            result_filters,
+            matrix_x_y
```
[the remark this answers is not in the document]
zoldello: Yes, you are correct and I normally try to keep name consistent. I used the name "matrix_x_y" in process_view to avoid a name collision in this method with "matrix". However, for methods called by process_view directly or indirectly, like "construct_query", they properly should use "matrix". This also applies to other use of "matrix" in other classes.Do you prefer I used "matrix_x_y" anyway?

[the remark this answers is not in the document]
caseylitton: Which other classes use these MatrixView functions?

src/encoded/viewconfigs/matrix.py
```diff
@@ -76,10 +73,11 @@ def _construct_result_views(self, type_info):
             views.append(self._view_item.summary_report)
         return views
 
-    def _construct_xygroupings(self, query, filters, negative_filters):
+    # pylint: disable=no-self-use
```
zoldello: @caseylitton This is not part of the requirements. However, I added it to silence Codacy (this method out to be a static function as "self" is never used). It is something that needs attention in the future.

caseylitton: Let remove this, the codacy issue it not a problem.  If we do want to fix it the remove self from the arguments and make add the staticmethod decorator.

zoldello: I removed the comment to disable pylint warning. Going forward, I will not rebase while a PR is undergoing review.

PR #2549, opened 2018-12-13
SNO-73-ENCD-4456 Add uuid queue module

caseylitton:
only failed test admin user bdd test

### DIFF
--- a/buildout.cfg
+++ b/buildout.cfg
@@ -36,6 +36,14 @@ auto-checkout = snovault
 ini-templates = ${buildout:directory}/ini-templates
 es-ip = localhost
 es-port = 9201
+primary_indexer_queue_type = Simple
+primary_indexer_queue_server = true
+primary_indexer_queue_worker = true
+primary_indexer_queue_worker_processes = 16
+primary_indexer_queue_worker_chunk_size = 1024
+primary_indexer_queue_worker_batch_size = 5000
+primary_indexer_queue_worker_get_size = 2000000
+
 
 [sources]
 behave = git https://github.com/behave/behave.git
@@ -53,7 +61,7 @@ webtest = git https://github.com/Pylons/webtest.git
 WSGIProxy2 = git https://github.com/lrowe/WSGIProxy2.git
 zope.sqlalchemy = git https://github.com/zopefoundation/zope.sqlalchemy.git
 pytest-bdd = git https://github.com/lrowe/pytest-bdd.git branch=allow-any-step-order
-snovault = git https://github.com/ENCODE-DCC/snovault.git rev=1.0.25
+snovault = git https://github.com/ENCODE-DCC/snovault.git branch=SNO-73-Add-base-uuid-queue
 
 [versions]
 # Hand set versions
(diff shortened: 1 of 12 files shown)


demo.cfg
```diff
@@ -1,2 +1,3 @@
 [buildout]
 extends = buildout.cfg
+primary_indexer_queue_type = REDIS_LIST
```
caseylitton: This is just for testing, it should work on demos but the default should be 'simple'

candidate.cfg
```diff
@@ -1,5 +1,6 @@
 [buildout]
 extends = buildout.cfg
+primary_indexer_queue_type = REDIS_LIST
```
caseylitton: We should use REDIS_LIST for now, but there are faster options for read/write like REDIS_PIPE_EXEC queue type

buildout.cfg
```diff
@@ -53,7 +59,8 @@ webtest = git https://github.com/Pylons/webtest.git
 WSGIProxy2 = git https://github.com/lrowe/WSGIProxy2.git
 zope.sqlalchemy = git https://github.com/zopefoundation/zope.sqlalchemy.git
 pytest-bdd = git https://github.com/lrowe/pytest-bdd.git branch=allow-any-step-order
-snovault = git https://github.com/ENCODE-DCC/snovault.git rev=1.0.24
+# snovault = git https://github.com/ENCODE-DCC/snovault.git rev=1.0.24
+snovault = git https://github.com/ENCODE-DCC/snovault.git branch=SNO-73-Add-base-uuid-queue
```
caseylitton: Flag for removal

rc.cfg
```diff
@@ -1,2 +1,3 @@
 [buildout]
 extends = buildout.cfg
+primary_indexer_queue_type = BASE_QUEUE_TYPE
```
caseylitton: RC can user REDIS_LIST too, I must have been testing the base queue type.

src/encoded/region_indexer.py
```diff
@@ -79,7 +79,9 @@ def includeme(config):
     config.scan(__name__)
     config.add_route('_regionindexer_state', '/_regionindexer_state')
     registry = config.registry
-    registry['region'+INDEXER] = RegionIndexer(registry)
+    is_region_indexer = registry.settings.get('regionindexer')
```
caseylitton: There are a number of processes running, this change only create a region indexer for the region indexer process.

src/encoded/vis_indexer.py
```diff
@@ -46,7 +47,9 @@ def includeme(config):
     config.add_route('_visindexer_state', '/_visindexer_state')
     config.scan(__name__)
     registry = config.registry
-    registry['vis'+INDEXER] = VisIndexer(registry)
+    is_vis_indexer = registry.settings.get('visindexer')
```
caseylitton: There are a number of processes running, this change only create a vis indexer for the vis indexer process.

src/encoded/region_indexer.py
```diff
@@ -388,6 +390,18 @@ def get_from_es(request, comp_id):
         '''Returns composite json blob from elastic-search, or None if not found.'''
         return None
 
+    def update_objects(self, request, uuids, force):
```
caseylitton: Update objects was move to region and vis indexer child classes because snovault indexer class has a different update objects now.

src/encoded/vis_indexer.py
```diff
@@ -245,12 +248,27 @@ def all_visualizable_uuids(registry):
...
+        self.es = registry[ELASTIC_SEARCH]
+        self.esstorage = registry[STORAGE]
+        self.index = registry.settings['snovault.elasticsearch.index']
         self.state = VisIndexerState(self.es, self.index)  # WARNING, race condition is avoided because there is only one worker
 
     def get_from_es(request, comp_id):
         '''Returns composite json blob from elastic-search, or None if not found.'''
         return None
 
+    def update_objects(self, request, uuids, xmin):
```
caseylitton: Update objects was move to region and vis indexer child classes because snovault indexer class has a different update objects now.

buildout.cfg
```diff
@@ -36,6 +36,14 @@ auto-checkout = snovault
 ini-templates = ${buildout:directory}/ini-templates
 es-ip = localhost
 es-port = 9201
+primary_indexer_queue_type = Simple
+primary_indexer_queue_server = true
+primary_indexer_queue_worker = true
+primary_indexer_queue_worker_processes = 16
+primary_indexer_queue_worker_chunk_size = 1024
+primary_indexer_queue_worker_batch_size = 5000
```
caseylitton: Simple queue should have a batch_size of 2000000 or greater than the number of objects.

PR #2552, opened 2018-12-21
ENCD-4359 Fix recursive is_part_of audit function

p-sud:


src/encoded/audit/biosample.py
```diff
@@ -178,15 +178,27 @@ def audit_biosample_part_of_consistency(value, system):
...
-        return False
-    if part_of_term_id in ontology[term_id]['part_of']:
-        return True
-    else:
-        parents = []
-        for x in ontology[term_id]['part_of']:
-            parents.append(is_part_of(x, part_of_term_id, ontology))
-        return any(parents)
+    """
+    Given the term_ids for a child and parent biosample pair as obtained from the
```
caseylitton: May be good to have examples of the arguments in the doc string.  Also, I'd like to know more about them to verify the functionality.

src/encoded/audit/biosample.py
```diff
@@ -178,15 +178,27 @@ def audit_biosample_part_of_consistency(value, system):
 # utility functions
 
 def is_part_of(term_id, part_of_term_id, ontology):
```
caseylitton: Is there a test for this?  If not, where we would put it?  Would be good to add one.

p-sud:
Example usage, auditing ENCBS651ADZ:

is_part_of('CL:0000598', 'UBERON:0002037', ontology)

src/encoded/audit/biosample.py
p-sud: There's no test for the is_part_of helper function, although there are some for the audit itself. I don't know how to access the ontology from the test app, though

gabdank: I think @p-sud you can mock-up ontology for the test purposes

p-sud:
Added an ontology fixture and test for is_part_of

src/encoded/tests/test_audit_biosample.py
```diff
@@ -229,3 +272,9 @@ def test_audit_biosample_part_of_consistency_ontology_part_of(testapp,
         errors_list.extend(errors[error_type])
     assert all(error['category'] != 'inconsistent biosample_term_id' for error in errors_list)
 
+
+def test_is_part_of(ontology):
+    from encoded.audit.biosample import is_part_of
+    assert is_part_of('UBERON:0002469', 'UBERON:0001007', ontology)
```
caseylitton: This doesn't cover all the cases.  I'll explain in person

PR #2554, opened 2019-01-08
ENCD-4325 Juicebox browser

forresttanaka:
This branch has both front end and back end changes, with the bulk being in the front end.

The back end changes do change the API for datasets in that the `visualize` property used to contain a dict with assemblies as keys, and each assembly was a dict with the browser names as keys and the corresponding URL as the value.

This changes in this branch so that each assembly is now an array of browser names because the URL processing now gets handled in the front end because the URLs can change depending on user selections.

### DIFF
--- a/src/encoded/static/components/filegallery.js
+++ b/src/encoded/static/components/filegallery.js
@@ -9,11 +9,12 @@ import { auditDecor, auditsDisplayed, AuditIcon } from './audit';
 import { FetchedData, Param } from './fetched';
 import * as globals from './globals';
 import { Graph, JsonGraph, GraphException } from './graph';
-import { requestFiles, DownloadableAccession, BrowserSelector } from './objectutils';
+import { requestFiles, DownloadableAccession } from './objectutils';
 import { qcIdToDisplay } from './quality_metric';
 import { softwareVersionList } from './software';
 import { SortTablePanel, SortTable } from './sorttable';
 import Status from './status';
+import { visOpenBrowser, visFilterBrowserFiles, visFileSelectable, visSortBrowsers } from './vis_defines';
 
 
 const MINIMUM_COALESCE_COUNT = 5; // Minimum number of files in a coalescing group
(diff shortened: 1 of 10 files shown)


src/encoded/vis_defines.py
```diff
@@ -1518,22 +1537,34 @@ def browsers_available(
         if not vis_blob and file_assemblies is None and files is not None:
             file_assemblies = visualizable_assemblies(assemblies, files)
         if ('ucsc' not in browsers
-                and 'ucsc_assembly' in mapped_assembly.keys()):
+                and 'ucsc_assembly' in mapped_assembly.keys()
+                and file_types is not None
```
caseylitton: Can we add this above the conditional block so we don't have to repeat it.
```
if file_types is None:
    continue
if ('ucsc' not in browsers
    ...
```

src/encoded/static/components/vis_defines.js
```diff
@@ -0,0 +1,203 @@
...
+    case 'Quick View': {
+        const fileQueries = files.map(file => `accession=${globals.atIdToAccession(file['@id'])}`);
+        href = `/search/?type=File&assembly=${assembly}&dataset=${dataset['@id']}&${fileQueries.join('&')}#browser`;
+        break;
+    }
+    case 'hic': {
+        const parsedUrl = url.parse(datasetUrl);
+        delete parsedUrl.path;
+        delete parsedUrl.search;
+        delete parsedUrl.query;
```
emro: Why do we need these deletions?

src/encoded/static/components/filegallery.js
```diff
@@ -122,29 +125,30 @@ export class FileTable extends React.Component {
...
+            filterOptions,
             filePanelHeader,
             encodevers,
-            selectedFilterValue,
-            filterOptions,
             showFileCount,
+            browserOptions,
             session,
             adminUser,
+            selectedFilterValue,
```
emro: It was a little confusing how these were rearranged, but I see that just the one parameter was added.

PR #2555, opened 2019-01-13
ENCD-4396 Citation note

forresttanaka:
Made a mod to the modals.js library because that was causing the `<body>` to shift horizontally when displaying a modal, yet seemed to have no real function, and in fact was code that wasn’t even executed for most modals.


caseylitton:
I restarted bdd tests.

src/encoded/schemas/user.json
```diff
@@ -94,7 +94,8 @@
         }
     },
     "boost_values": {
-        "title": 1.0
+        "title": 1.0,
+        "viewing_groups": 1.0
```
emro: What does this change do?

src/encoded/static/scss/encoded/_base.scss
```diff
@@ -65,8 +65,23 @@ body { background: $mainThemeColor; font-size: 1rem; }
...
 
-#slot-application, #application {
-	height: 100%;
+#slot-application {
+	height: 100vh;
+}
+
+#application {
+    min-height: 100%;
+
```
emro: I don't think that "height: 100vh;" and "min-height: 100%" are necessary. Am I missing a case when they're important?

forresttanaka: Without either, the footer ends up below the content instead of at the bottom of the window on short pages. Just about every page I saw on this had a different way, many that didn’t work. This way worked on all browsers we support. This is an example of a short page without the min-height:100% enabled.

no-100

src/encoded/schemas/user.json
forresttanaka: That was likely a left-over from a test. Removed

PR #2556, opened 2019-01-14
ENCD-4393-add-gene-alignments

khzlin:


caseylitton:
Can you fix the pr name syntax?  Capital ENCD-####.

Also there is one failing bdd test.  I'm not sure if it is related to the feature update.  Maybe rerun the bdd tests.  And look into the tests associated with your code change, if they exist.

src/encoded/schemas/analysis_step.json
```diff
@@ -112,6 +112,7 @@
                     "filtered regions",
```
caseylitton: flagging for tests/pr name

khzlin: the one failing bad test passed.

PR #2567, opened 2019-01-31
ENCD-4251 Combine reorganize technical replicates

yunhailuo:


src/encoded/types/file.py
```diff
@@ -269,6 +271,36 @@ def technical_replicates(self, request, registry, root, replicate=None):
...
+        "type": "array",
+        "items": {
+            "title": "Library",
+            "description": "The nucleic acid library sequenced.",
+            "comment": "See library.json for available identifiers.",
+            "type": "string",
+            "linkTo": "Library"
+        }
+    })
+    def replicate_libraries(self, request, registry, root, dataset, library=None):
```
keenangraham: Are registry and root used here?

src/encoded/types/experiment.py
```diff
@@ -543,3 +597,45 @@ def __ac_local_roles__(self):
...
+        for rep_uuid in experiment.get_rev_links('replicates'):
+            rep_props = root.get_by_uuid(rep_uuid).upgrade_properties()
+            # Only care (check and add to the list) about non-deleted technical
+            # replicates of this replicate, meaning belonging to the same
+            # biological replicate.
+            if (rep_props['biological_replicate_number'] != biological_replicate_number
+                or rep_props['status'] == 'deleted'):
+                continue
+            if rep_props['technical_replicate_number'] < technical_replicate_number:
+                # Found smaller technical replicate, libraries will be
```
keenangraham: Why?

src/encoded/types/experiment.py
```diff
@@ -382,29 +411,50 @@ def superseded_by(self, request, superseded_by):
         "type": "string"
     })
     def replication_type(self, request, replicates=None, assay_term_name=None):
+        # ENCD-4251 loop through replicates and select one replicate, which has
+        # the smallest technical_replicate_number, per biological replicate.
+        # That replicate should have a libraries property which, as calculated
+        # in replicate.libraries (ENCD-4251), should have collected all
+        # possible technique replicates belong to the biological replicate.
```
keenangraham: Technical.

src/encoded/types/experiment.py
```diff
@@ -382,29 +411,50 @@ def superseded_by(self, request, superseded_by):
...
+        # TODO: change this once we remove technical_replicate_number.
+        bio_rep_dict = {}
+        for rep in replicates:
+            replicateObject = request.embed(rep, '@@object')
```
keenangraham: Does this have to be camelCase?

src/encoded/types/experiment.py
```diff
@@ -382,29 +411,50 @@ def superseded_by(self, request, superseded_by):
...
-                    biosample_donor_list.append(biosampleObject.get('donor'))
-                    biosample_number_list.append(replicateObject.get('biological_replicate_number'))
-                    biosample_species = biosampleObject.get('organism')
-                    biosampleTypeObject = request.embed(
-                        biosampleObject['biosample_ontology'],
-                        '@@object'
-                    )
-                    biosample_type = biosampleTypeObject.get('classification')
+        for replicateObject in bio_rep_dict.values():
+            if 'libraries' in replicateObject and replicateObject['libraries']:
```
keenangraham: `if replicateObject.get('libraries')`

src/encoded/types/experiment.py
```diff
@@ -382,29 +411,50 @@ def superseded_by(self, request, superseded_by):
...
+                biosamples = request.select_distinct_values(
+                    'biosample', *replicateObject['libraries']
+                )
+                if biosamples:
+                    for b in biosamples:
+                        biosampleObject = request.embed(b, '@@object')
```
keenangraham: CamelCase.

src/encoded/types/experiment.py
```diff
@@ -382,29 +411,50 @@ def superseded_by(self, request, superseded_by):
...
+                        biosample_donor_list.append(
+                            biosampleObject.get('donor')
+                        )
+                        biosample_number_list.append(
+                            replicateObject.get('biological_replicate_number')
```
keenangraham: What happens if this doesn't exist for some reason and you append None?

src/encoded/types/experiment.py
yunhailuo: Overall, we use calculation to link all technical replicates of a biological replicate to only the first non deleted replicate. In the future, when we upgrade, all other technical replicates will be "deleted" and only this one will have the submittable property "libraries" filled.

src/encoded/types/experiment.py
yunhailuo: No. I tried to keep it consistent with other properties, like biosample_summary, here and consistent with old code. Any suggestions?

src/encoded/types/experiment.py
yunhailuo: "biological_replicate_number" is required for replicate objects here.
Biosample object not having donor should be handled properly by the old code below, right?
https://github.com/ENCODE-DCC/encoded/blob/ENCD-4251-tech-rep-yl/src/encoded/types/experiment.py#L478-L495

src/encoded/types/file.py
keenangraham: ?

src/encoded/types/experiment.py
keenangraham: Let's break this pattern if you can make your property self-consistent.

PR #2581, opened 2019-02-05
ENCD-3877-annotation-download-btn

zoldello:


src/encoded/batch_download.py
```diff
@@ -216,13 +247,123 @@ def peak_metadata(context, request):
...
+            if cart.get('elements'):
+                param_list['@id'] = cart['elements']
+    else:
+        try:
+            elements = request.json.get('elements')
+        except ValueError:
+            pass
+        else:
+            param_list['@id'] = elements
+    path = '{}?{}'.format(search_path, urlencode(param_list, True))
```
hitz: I think you need to add limit=all here if there are > 25 results.

src/encoded/batch_download.py
```diff
@@ -216,13 +247,123 @@ def peak_metadata(context, request):
...
+            file.get('file_size', ''),
+            file.get('lab', {}).get('title', ''),
+            file.get('md5sum', ''),
+            file.get('dbxrefs', ''),
+            file.get('href', ''),
+            file.get('assembly', ''),
+            file.get('controlled_by', ''),
+            file.get('status', ''),
+            file.get('restricted', ''),
+        ])
```
hitz: you don't do the audit mapping here?  
Also, I think you also need to add
`                    if prop in ['files.paired_with', 'files.derived_from']:
                        # chopping of path to just accession
                        if len(temp):
                            new_values = [t[7:-1] for t in temp]
                            temp = new_values
`
I think because we have file.derived_from in the list of fields.

src/encoded/batch_download.py
```diff
@@ -216,13 +247,123 @@ def peak_metadata(context, request):
...
+
+def _get_graph(schema_type, request, ids):
+    """
+    Get a graph based on schema type and id.
+
+        :param schema_type: schema type.
+        :param request: request object from pyramid.
+        :param ids: ids.
+    """
+    datasets = ''.join([''.join(['&dataset=', id]) for id in ids])
```
hitz: how big a search do you create before this is too slow?    Like what about this query: `https://www.encodeproject.org/search/?type=Annotation&encyclopedia_version=4`

src/encoded/batch_download.py
```diff
@@ -216,13 +247,123 @@ def peak_metadata(context, request):
...
+
 @view_config(route_name='metadata', request_method='GET')
 def metadata_tsv(context, request):
     param_list = parse_qs(request.matchdict['search_params'])
+    type_param = ('type' in param_list and param_list['type'][0] and param_list['type'][0]) or None
     if 'referrer' in param_list:
         search_path = '/{}/'.format(param_list.pop('referrer')[0])
     else:
         search_path = '/search/'
+    is_annotation = type_param and type_param.lower() == 'annotation'
```
hitz: What happens if type not in ('annotation', 'experiment')???

src/encoded/helpers/helper.py
```diff
@@ -1,5 +1,9 @@
 from encoded.vis_defines import vis_format_url
 
+BATCH_DOWNLOAD_DOC_TYPES = [
```
hitz: I strongly dislike using helper.py ... for anything.    I feel like it was created as a shim for refactoring news.py and region_search.py.     Should batch_download.py be a subclass of a searchView?

src/encoded/batch_download.py
zoldello: @hitz I assume you mean param_list['all']. If correct, this is a chicken-egg problems because param_list is needed to get results so results cannot be made dependent on param_list.

src/encoded/batch_download.py
zoldello: Do you mean the code, line 347 is slow or do you mean the query size is slow? If it is the code, I can use for-loop rather than list comprehension. I know "join" is faster than using '+'

src/encoded/batch_download.py
zoldello: Great observation. You get a HTTP 400 response.

src/encoded/batch_download.py
hitz: Hmm... I *suspect* it might be slow, was it tested?   How long does it take to create a metadata.tsv for an annotation search with a large number of datasets?   Test it for 10/100/1000/all Annotations.   I think the total number is currently >6,000 and it will probably grow.   Demo is off, but you should check it (it is a good habit).

src/encoded/batch_download.py
```diff
@@ -216,13 +247,123 @@ def peak_metadata(context, request):
...
+    Get annotation metadata.
+
+        :param request: Pyramid's request.
+        :param search_path: Search url.
+    """
+    param_list = parse_qs(request.matchdict['search_params'])
+    param_list['limit'] = ['all']
+    # Handle metadata.tsv lines from cart-generated files.txt.
+    # TODO: When rewriting this file, refactor this duplicated cart code.
+    cart_uuids = param_list.get('cart', [])
```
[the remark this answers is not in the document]
hitz: the carts part not line 258!  See below!

src/encoded/batch_download.py
hitz: oh I see it was added here: C/encoded/blob/1ae7afda60c56b732a3512df1a4e0d677d771916/src/encoded/batch_download.py#L258, I missed it when you moved it to the function.

src/encoded/helpers/helper.py
zoldello: We can have a discussion on the merits of helper/util functions at some point if so desired. But for now, refactoring the code to eliminate the need for the helper seems tangential to solving this ticket. Still, is it very important to you to address it now?

src/encoded/batch_download.py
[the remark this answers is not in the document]
zoldello: Acknowledged.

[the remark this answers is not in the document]
zoldello: addressed.

src/encoded/batch_download.py
zoldello: Opps. Added audit and other columns.

src/encoded/batch_download.py
zoldello: I assume this does not need any further action.

src/encoded/helpers/helper.py
zoldello: I assume this does not need further action at this time; but is technical debt in our virtual technical debt bucket, on the back of the teal-colored huge elephant in the office :)

hitz: I defer to @caseylitton on that.

src/encoded/batch_download.py
zoldello: I am still investigating this. There are problems with even 3000 I do not see in experiment.

src/encoded/batch_download.py
```diff
@@ -216,13 +250,112 @@ def peak_metadata(context, request):
...
+        datasets,
+    )
+    results = request.embed(path, as_user=True)
+    graph = results['@graph']
+    return graph
+
 @view_config(route_name='metadata', request_method='GET')
 def metadata_tsv(context, request):
     param_list = parse_qs(request.matchdict['search_params'])
+    type_param = ('type' in param_list and param_list['type'][0] and param_list['type'][0]) or None
```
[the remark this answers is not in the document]
zoldello: Good catch. I refactored it to make it cleaner.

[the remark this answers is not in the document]
zoldello: This I think is a dup of a previous comment.

src/encoded/helpers/helper.py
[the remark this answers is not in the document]
zoldello: Done

src/encoded/batch_download.py
```diff
@@ -216,13 +250,112 @@ def peak_metadata(context, request):
...
+    annotation_graph = results['@graph']
+    ids = [g['@id'] for g in annotation_graph]
+    file_graph = _get_graph('file', request, ids)
+    software_graph = _get_graph('software', request, ids)
+    header = [header for header in _tsv_mapping_annotation]
+    header.extend([prop for prop in _audit_mapping])
+    fout = io.StringIO()
+    writer = csv.writer(fout, delimiter='\t')
+    writer.writerow(header)
+    for file in file_graph:
```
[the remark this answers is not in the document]
zoldello: Done

caseylitton:
Closed because jira status is stalled

PR #2584, opened 2019-02-05
ENCD-4227-add-PLAC-seq-to-assays

khzlin:


src/encoded/commands/generate_ontology.py
```diff
@@ -175,6 +175,7 @@
         'OBI:0001859': ['DNA accessibility'],  # 'OBI:0000870', / OBI:0000424  / FAIRE-seq
         'OBI:0002042': ['3D chromatin structure'],  # 'OBI:0000870' (Hi-C)
         'OBI:0001848': ['3D chromatin structure'],  # ChIA-PET / OBI:000870
+        'NTR:0000513': ['3D chromatin structure'], # 'NTR:0000513', / PLAC-seq
```
jahilton: You don't need to add NTRs to the slim_shims list here. Adding the NTR object in the ntr_assays lists accomplishes the desired shimming.

I think you should also ditch all the changes to this file. I will include it in my ontology update branch (ENCD-4298/ENCD-4439). Removing them won't impact your demo at all since you didn't run the generate-ontology script

PR #2589, opened 2019-02-07
ENCD-4419 use biosample ontology for audit

p-sud:


src/encoded/audit/experiment.py
```diff
@@ -2758,32 +2758,16 @@ def audit_experiment_biosample_term(value, system, excluded_types):
...
-            if bs_type != term_type:
+            bs_type = biosample.get('biosample_ontology', {}).get('@id')
+            bs_name = biosample.get('biosample_ontology', {}).get('name')
+            experiment_bs_type = value.get('biosample_ontology', {}).get('@id')
+            experiment_bs_name = value.get('biosample_ontology', {}).get('name')
+            if bs_type != experiment_bs_type:
                 detail = 'Experiment {} '.format(value['@id']) + \
                          'contains a library {} '.format(lib['@id']) + \
-                         'prepared from biosample type \"{}\", '.format(bs_type) + \
+                         'linked to biosample type \"{}\", '.format(bs_name) + \
```
caseylitton: Since we're changing this line let's clean up the string formating.  No slashes and one format.  Use double quotes for the outer most and single quotes for the internal ones.

p-sud: Sounds good, I'll edit and update the PR

PR #2592, opened 2019-02-12
ENCD-4449: use top level class files property for Annotation

hitz:
I think I still know how to do this?

### DIFF
--- a/src/encoded/types/dataset.py
+++ b/src/encoded/types/dataset.py
@@ -330,6 +330,27 @@ class Annotation(FileSet, CalculatedVisualize):
     }
 
     @calculated_property(schema={
+        "title": "Files",
+        "type": "array",
+        "items": {
+            "type": "string",
+            "linkTo": "File",
+        },
+    })
+    def files(self, request, original_files, status):
+        if status in ('released', 'archived'):
+            return paths_filtered_by_status(
+                request, original_files,
+                include=('released', 'archived'),
+            )
+        else:
+            return paths_filtered_by_status(
+                request, original_files,
+                exclude=('revoked', 'deleted', 'replaced'),
+            )
+
+
+    @calculated_property(schema={
         "title": "Superseded by",
         "type": "array",
         "items": {


src/encoded/types/dataset.py
```diff
@@ -328,6 +328,27 @@ class Annotation(FileSet, CalculatedVisualize):
...
 
+    @calculated_property(schema={
+        "title": "Files",
+        "type": "array",
+        "items": {
+            "type": "string",
+            "linkTo": "File",
+        },
+    })
+    def files(self, request, original_files, status):
```
caseylitton: Can we not call this files.  Generic names should be avoided when possible.

src/encoded/types/dataset.py
```diff
@@ -328,6 +328,27 @@ class Annotation(FileSet, CalculatedVisualize):
...
+        if status in ('released', 'archived'):
```
caseylitton: might be better to put these sets in vars to avoid hardcoding in two locations.
```
good_status = ('released', 'archived')
bad_status = ('revoked', 'deleted', 'replaced')
if status in good_status:
    etc...
else:
    etc...

src/encoded/types/dataset.py
hitz: has to be files for API backwards compatibility.

src/encoded/types/dataset.py
hitz: I could refactor this, it would probably have to be refactored in types/experiment.py and even elsewhere.

hitz:
Yeah the problem starts with the fact that Experiments probably should be FileSets but they pre-date them by at least 1 epoch.    Series are also very ad-hoc and don't provide the value they should (as structured groupings of Experiments or Annotations).

src/encoded/types/dataset.py
zoldello: @caseylitton I assume Ben's answers satisfies your code change request, so I will resolve this.

src/encoded/types/dataset.py
zoldello: @caseylitton This was update as per your review, and just this method.

PR #2593, opened 2019-02-12
ENCD-4408 mobile menu readability

emro:
Should fix two style bugs on mobile:
(1) the sub-categories under the main links are very difficult to read, the contrast is not high enough
(2) when you click on a menu link to see the sub-categories, it is not obvious that the menu is scrollable (I made the menu the full height to avoid this problem)

### DIFF
--- a/src/encoded/static/scss/bootstrap/_navbar.scss
+++ b/src/encoded/static/scss/bootstrap/_navbar.scss
@@ -48,6 +48,7 @@
 // content for the user's viewport.
 
 .navbar-collapse {
+  max-height: $navbar-collapse-max-height;
   overflow-x: visible;
   padding-right: $navbar-padding-horizontal;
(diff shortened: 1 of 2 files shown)


src/encoded/static/scss/bootstrap/_navbar.scss
```diff
@@ -48,7 +48,6 @@
 // content for the user's viewport.
 
 .navbar-collapse {
-  max-height: $navbar-collapse-max-height;
```
forresttanaka: Could this instead be overridden in, maybe, src/encoded/scss/encoded/modules/_navbar.scss? I’ve been considering the files in the bootstrap directory as non-modifiable since they come from the bootstrap team.

emro: Yes, see update. Thank you!

PR #2596, opened 2019-02-13
SNO-80 typeahead for facets

emro:
- New facet typeahead component allows users to search for a facet term from a long list of terms by typing it into a searchbar
- Component is implemented for 5 facets: "Biosample name", Assay", "Target of assay", "Organ" and "Cell"
- Both non-typeahead and typeahead facets now display selected filters
- Both non-typeahead and typeahead facets now have a maximum height and allow scrolling, with shading at the top and bottom of the term lists to indicate overflow
- Horizontal facets in matrix and summary views are now vertical facets
- Ordering of facets is changed slightly, to be: Assay category, Assay, Experiment status, Project, RFA, Genome assembly, Target category, Target of assay, Organism, Biosample type, Biosample name, Organ, Cell, Life stage, Biosample treatment, etc.


emro:
A couple of questions for @forresttanaka:
- I made this a separate component because I thought it would be relatively different from the Facet component but I am now thinking it could just be one component with different options. In particular, the selected filters and scroll components are duplicated in both components.
- Are there ways I could make this more accessible? Right now I have aria-label and role attribute but maybe there are other good strategies.
- Do the tests I added seem good or is there something else I should add?

emro:
One other potential issue: Long facets are reaching a limit at 200. I am not sure if that should be part of this PR. @caseylitton might know?

src/encoded/static/components/search.js
```diff
@@ -723,11 +723,31 @@ class Facet extends React.Component {
...
         // Bind `this` to non-React methods.
         this.handleClick = this.handleClick.bind(this);
+        this.scrollEvent = this.scrollEvent.bind(this);
     }
 
     handleClick() {
         this.setState(prevState => ({ facetOpen: !prevState.facetOpen }));
     }
+    
+    scrollEvent(e) {
```
forresttanaka: I’m getting a lot of ESLint/AirBnB errors in VSCode starting here and going on for many lines. They’ll all have to be cleaned up. In this particular case, `scrollEvent` has to be moved out to its own top-level function because it doesn’t refer to `this`, so it can’t exist in a class unless you make it a `static` method. But if you move it to its own function, then both Facet and TypeaheadFacet can share it.

forresttanaka:
> but I am now thinking it could just be one component with different options

If you can combine them practically, I think that would be a lot better. The amount of duplication between the two components makes maintenance look a little scary.

> Are there ways I could make this more accessible?

I think at this stage, this is fine. We could write a ticket to audit the accessibility of these facets and update them then.

> Do the tests I added seem good or is there something else I should add?

They look good to me. @hitz might be a better person to ask.

src/encoded/static/components/matrix.js
```diff
@@ -195,32 +193,28 @@ class Matrix extends React.Component {
                 <div>
                     <div className="panel data-display main-panel">
                         <div className="row matrix__facet--horizontal">
-                            <div className="col-sm-5 col-md-4 col-lg-3 sm-no-padding" style={{ paddingRight: 0 }}>
+                            <div className="matrix-header" style={{ paddingRight: 0 }}>
```
forresttanaka: Out of curiosity, could the inline style be merged into the `matrix-header` class, or is this the easier way to go?

src/encoded/static/components/search.js
```diff
@@ -711,30 +711,48 @@ TypeTerm.propTypes = {
...
         // Bind `this` to non-React methods.
-        this.handleClick = this.handleClick.bind(this);
+        this.handleSearch = this.handleSearch.bind(this);
     }
 
-    handleClick() {
-        this.setState(prevState => ({ facetOpen: !prevState.facetOpen }));
+    handleSearch(event) {
+        // search term entered by the user
+        const filterVal = String(event.target.value.toLowerCase().replace(/ /g, '').replace(/[^\w\s]/gi, ''));
```
forresttanaka: Not a regex expert, but I feel like the \s in the second replace would also take care of deleting spaces done by the first replace. And technically, I doubt you need the `i` in `gi` since you’ve already lowercased the string, though I don't know if that has any real performance difference.

src/encoded/static/components/search.js
```diff
@@ -711,30 +711,48 @@ TypeTerm.propTypes = {
...
+
+        // which facet terms match the search term entered by the user
+        let terms = this.props.facet.terms.filter((term) => {
+            if (term.doc_count > 0) {
+                const termKey = term.key.toLowerCase().replace(/ /g, '').replace(/[^\w\s]/gi, '');
```
forresttanaka: Since this is the same transform as you do above, maybe make a one-line stand-alone arrow function private to search.js that does this transformation? Just to avoid having to make the same change in two different places in the future.

src/encoded/static/components/search.js
```diff
@@ -783,34 +804,72 @@ class Facet extends React.Component {
...
+                                        {/* Only show bottom shading when list of results overflows */}
+                                        <div className={`shading ${(this.state.filteredTerms.length < displayedTermsCount) ? 'hide-shading' : ''}`} />
+                                    </div>
+                                }
                             </div>
-                        : null}
+                        :
+                            <div>
+                                {/* If the user has not searched, we will display the full set of facet terms */}
+                                {((terms.length && terms.some(term => term.doc_count)) || (field.charAt(field.length - 1) === '!')) ?
```
forresttanaka: Our AirBnB config requires `terms.length` to be compared against an integer instead of truthyness. https://github.com/airbnb/javascript#comparison--shortcuts

src/encoded/static/components/search.js
```diff
@@ -966,6 +1020,9 @@ export class FacetList extends React.Component {
         // non-negated versions inhabit.
         const negationFilters = filters.filter(filter => filter.field.charAt(filter.field.length - 1) === '!');
 
+        // Certain facets are typeahead for easy searching and they are listed here
+        const typeaheadList = ['biosample_ontology.term_name', 'assay_title', 'target.label', 'biosample_ontology.organ_slims', 'biosample_ontology.cell_slims'];
```
forresttanaka: ESLint is showing `typeaheadList` is unused.

src/encoded/static/scss/encoded/modules/_facet.scss
```diff
@@ -247,6 +252,84 @@
...
+}
+
+.filter-container {
+    display: block;
+    margin: 10px auto;
+    .filter-hed {
+        display: inline-block;
+        padding: 0 5px;
+    }
+    .negationFilter {
```
forresttanaka: Camel case in a CSS class seems kind of unusual.

src/encoded/static/components/search.js
emro: I was thinking about this a little more. My original thought had been that if you type "%%%", the list of terms shouldn't be filtered at all so that was why I was ignoring non-alphabetical characters. However, then you don't get an error message if you type "seq%%%" and I feel like you probably should. What do you think makes the most sense here? I am thinking that maybe we should match all non-white space characters including symbols.

caseylitton:
closing...code is on v80 and not sure I believe this is going back to in progress and sno-78 being split out.

PR #2598, opened 2019-02-13
ENCD-4425 Update audits exp and genetic mod targets

yunhailuo:


src/encoded/audit/experiment.py
```diff
@@ -2475,6 +2475,22 @@ def audit_experiment_target(value, system, excluded_types):
...
+                    'Trimethylation',
+                    'Acetylation',
+                    'Ubiquitination',
+                    'Phosphorylation']
+    if any(mod['modification'] not in non_tag_mods
+           for mod in target.get('modifications', [])
+           if 'modification' in mod):
+        detail = 'Experiment {} has a tagged target {}.'.format(value['@id'],
+                                                                target['@id'])
+        yield AuditFailure('tagged target', detail,
```
caseylitton: this can be one line.

src/encoded/audit/experiment.py
```diff
@@ -2483,8 +2499,24 @@ def audit_experiment_target(value, system, excluded_types):
                                     'CRISPR genome editing followed by RNA-seq']:
         return
 
-    # Check that target of experiment matches target of antibody
     for rep in value['replicates']:
+        # Check target of experiment matches target of genetic modifications
+        if 'library' in rep and 'biosample' in rep['library']:
```
caseylitton: This can be a bit simpler
```
        biosample = rep.get('library', {}).get('biosample', {})
        if biosample:
            etc...
```

src/encoded/audit/experiment.py
```diff
@@ -2483,8 +2499,24 @@ def audit_experiment_target(value, system, excluded_types):
...
+            biosample = rep['library']['biosample']
+            for modification in biosample.get('applied_modifications', []):
```
caseylitton: modification is a dict so we shouldn't default to list

src/encoded/audit/genetic_modification.py
```diff
@@ -3,6 +3,21 @@
...
 
+
+@audit_checker('GeneticModification', frame=['modified_site_by_target_id'])
+def audit_genetic_modification_target(value, system):
+    '''
+    Genetic modifications should targeting (modified_site_by_target_id) a wild
+    type gene, i.e. a target without modifications, as defined in the schema.
+    '''
+    if value.get('modified_site_by_target_id', {}).get('modifications'):
+        detail = ('Genetic modification {} is targeting {} which has already '
```
caseylitton: I'm just using this as an example but there are mixed code formatting styles all over the place.  
```
        detail = (
            'Genetic modification {} is targeting {} which has already '
            'been modified.'.format(
                value['@id'], 
                value['modified_site_by_target_id']['@id'],
            )
        )
```
this style if more widely applicable for various use cases.  
-End parens/brackets dropped to the next line
-if you drop one arg/item to the next line the put them all in a line with trailing commas
-pulling stuff to the left makes more readable code and allows for longer lines

We don't adhere to this, so just FYI.  There are many styles, but we should try not to mixed them.

src/encoded/tests/test_audit_experiment.py
```diff
@@ -3298,3 +3298,39 @@ def test_audit_experiment_histone_characterized_no_primary(testapp,
     res = testapp.get(base_experiment['@id'] + '@@index-data')
     assert any(error['category'] == 'antibody not characterized to standard'
                for error in collect_audit_errors(res))
+
+
+def test_audit_experiment_tag_target(testapp, experiment, ctcf):
+    tag_target = testapp.post_json(
+        '/target',
+        {'genes': [ctcf['uuid']],
```
caseylitton: Another example, the outer parens are dropped to the next line so we should follow this pattern in the code block...
```
    tag_target = testapp.post_json(
        '/target',
        {
            'genes': [ctcf['uuid']],
            'modifications': [{'modification': 'eGFP'}],
            'label': 'eGFP-CTCF',
            'investigated_as': ['recombinant protein'],
        }
    ).json['@graph'][0]
```

src/encoded/audit/experiment.py
```diff
@@ -2475,6 +2475,22 @@ def audit_experiment_target(value, system, excluded_types):
...
 
+    # Experiment target should be untagged after ENCD-4425
+    non_tag_mods = ['Methylation',
+                    'Monomethylation',
+                    'Dimethylation',
+                    'Trimethylation',
+                    'Acetylation',
+                    'Ubiquitination',
+                    'Phosphorylation']
+    if any(mod['modification'] not in non_tag_mods
```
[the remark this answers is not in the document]
yunhailuo: Wait. Target has a property "modifications" which is a list of dict. Each dict can have a key 'modification'. Is this making sense?

src/encoded/audit/experiment.py
yunhailuo: Again, "applied_modifications" is a list of dict.

yunhailuo:
Fix as looking for only one modification target match for each modified biosample (biosample without applied_modifications will be skipped) @gabdank.

src/encoded/audit/experiment.py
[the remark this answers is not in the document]
caseylitton: my fault, yea it's list of dicts makes sense.

src/encoded/audit/experiment.py
caseylitton: yes, reading that wrong

src/encoded/audit/experiment.py
```diff
@@ -2475,6 +2475,20 @@ def audit_experiment_target(value, system, excluded_types):
...
 
+    # Experiment target should be untagged
+    non_tag_mods = ['Methylation',
+                    'Monomethylation',
+                    'Dimethylation',
+                    'Trimethylation',
+                    'Acetylation',
+                    'Ubiquitination',
+                    'Phosphorylation']
+    if any(mod['modification'] not in non_tag_mods
```
gabdank: for "audit experiment target" the category should be also something like "inconsistent experiment target" instead of "tagged target"

PR #2606, opened 2019-02-19
ENCD-4462 fixing the AB characterization view bug

gabdank:


keenangraham:
Codacy trailing whitespace.

src/encoded/types/characterization.py
```diff
@@ -84,7 +80,11 @@ class DonorCharacterization(Characterization):
 class BiosampleCharacterization(Characterization):
     item_type = 'biosample_characterization'
     schema = load_schema('encoded:schemas/biosample_characterization.json')
-
+    STATUS_ACL = {
```
keenangraham: Looks okay but only issue is the code duplication. Instead of defining STATUS_ACL in these subobjects can we just add the reviewer permissions `Allow, 'role.lab_reviewer', 'edit'`, `Allow, 'role.lab_reviewer', 'review'`to the top-level `ALLOW_LAB_SUBMITTER_EDIT` ACL here https://github.com/ENCODE-DCC/encoded/blob/6435e72155cee1d5e48cea9761ccb3368117a659/src/encoded/types/base.py#L55? The in progress characterizations will then have the correct permissions if review_lab is defined, and won't change if it isn't. Also the AB characterization status ALCs (https://github.com/ENCODE-DCC/encoded/blob/6435e72155cee1d5e48cea9761ccb3368117a659/src/encoded/types/base.py#L153) won't get overwritten by defining a new dictionary.

gabdank:
@keenangraham I have refactored according to your suggestions

PR #2607, opened 2019-02-19
ENCD-3877-4449-annotation download btn

zoldello:


src/encoded/helpers/helper.py
```diff
@@ -79,19 +85,19 @@ def search_result_actions(request, doc_types, es_results, position=None):
...
+            doc_types in BATCH_DOWNLOAD_DOC_TYPES and
+            any(
                 bucket['doc_count'] > 0
-                for bucket in aggregations['files-file_type']['files-file_type']['buckets']):
-            actions['batch_download'] = request.route_url(
-                'batch_download',
-                search_params=request.query_string
-            )
-
+                for bucket in aggregations['files-file_type']['files-file_type']['buckets'])):
```
caseylitton: This sytanx is really confusing, drop the condition end bracket to the next line

src/encoded/batch_download.py
```diff
@@ -496,4 +599,4 @@ def generate_rows():
         downloadtime.minute
     )
     request.response.app_iter = generate_rows()
-    return request.response
+    return request.response
```
caseylitton: Remove this change.

src/encoded/batch_download.py
```diff
@@ -174,6 +204,76 @@ def make_audit_cell(header_column, experiment_json, file_json):
     return ', '.join(list(set(data)))
 
 
+def _get_annotation_metadata(request, search_path, param_list):
```
caseylitton: Where did this function come from?

zoldello: I made it. I could not cleanly augment the experiment metadata functionality with annotation. So, I made a function just for annotation, which is what you see.

src/encoded/batch_download.py
zoldello: Done.

src/encoded/batch_download.py
```diff
@@ -232,6 +232,10 @@ def _get_annotation_metadata(request, search_path, param_list):
         software = [s for s in result_graph.get('software_used', {})]
         software_set = ', '.join([s['software']['title'] for s in software])
         for result_file in result_files:
+            is_file_restricted = restricted_files_present(result_file)
+            is_file_type_excluded = ('files.file_type' in param_list) and (result_file['file_type'] not in param_list['files.file_type'])
+            if is_file_restricted or is_file_type_excluded:
```
caseylitton: Since these conditions use OR let separate them.
```
if restricted_files_present(result_file):
    continue
if ('files.file_type' in param_list) and (result_file['file_type'] not in param_list['files.file_type']):
    continue
```

src/encoded/batch_download.py
```diff
@@ -232,6 +232,10 @@ def _get_annotation_metadata(request, search_path, param_list):
         software = [s for s in result_graph.get('software_used', {})]
         software_set = ', '.join([s['software']['title'] for s in software])
         for result_file in result_files:
+            is_file_restricted = restricted_files_present(result_file)
+            is_file_type_excluded = ('files.file_type' in param_list) and (result_file['file_type'] not in param_list['files.file_type'])
```
caseylitton: This should be able to be written without '('files.file_type' in param_list)' by using dict.get().  Possibly with a negation in the final condition.

zoldello: Done

src/encoded/batch_download.py
zoldello: Done

PR #2612, opened 2019-02-23
ENCD-3422-datasets-table-publication

zoldello:


src/encoded/static/components/publication.js
```diff
@@ -26,7 +28,64 @@ const PublicationComponent = (props, reactContext) => {
     ];
 
     const crumbsReleased = (context.status === 'released');
-
+    const accessions = context.datasets.reduce((dataset1, dataset2) => `${dataset1}&accession=${dataset2.accession}`, '');
+    const sort = (a, b) => a - b;
+    const datasetsColumns = {
```
forresttanaka: This is a pretty large assignment that I would normally put into this module’s global space so that it only gets assigned once at page load, or sometimes as a property of the class (which you don’t have since this is a functional component) if you need the object namespaced. Not a lot can cause a re-render of a Publication page, but logging in while showing a Publication page will cause this assignment to happen several times as different context and props get updated for the page. If this object used props or state directly, you’d be forced to put it here, but `SortTable` column objects are always self-contained, and any data they use gets passed at render time.

src/encoded/static/components/publication.js
```diff
@@ -45,6 +104,21 @@ const PublicationComponent = (props, reactContext) => {
                 </div>
             : null}
 
+            {context.datasets && context.datasets.length ?
+                <SortTablePanel title="Datasets">
+                    <FetchedData ignoreErrors>
```
forresttanaka: Since this is new code, you can remove `ignoreErrors`. That’s something we used to use to get around an occasional rendering problem, but that problem went away (I don’t remember why it went away).

src/encoded/static/components/publication.js
```diff
@@ -45,6 +104,21 @@ const PublicationComponent = (props, reactContext) => {
...
+                        <Param name="datasets" url={datasetsUrl} />
+                        <DatasetsComponent columns={datasetsColumns} />
+                    </FetchedData>
+                </SortTablePanel>
+            : null }
+
+            {context.datasets && context.datasets.length ?
```
forresttanaka: What is this section for? It looks like it would do a GET request for an Experiment search, and then throw away the result. Does the result get directed somewhere else I’m not aware of?

src/encoded/static/components/publication.js
```diff
@@ -26,7 +28,64 @@ const PublicationComponent = (props, reactContext) => {
...
+            title: 'Target',
+        },
+        assay_term_name: {
+            getValue: (dataset) => {
+                // assay_term_name can be a string or an array
+                if (typeof (dataset.assay_term_name) === 'string') {
+                    return dataset.assay_term_name;
+                }
+                return (dataset.assay_term_name || []).join(', ');
+            },
```
forresttanaka: Getting a trailing space error in ESLint on this line.

src/encoded/static/components/publication.js
```diff
@@ -45,6 +104,21 @@ const PublicationComponent = (props, reactContext) => {
...
+                    <FetchedData ignoreErrors>
+                        <Param name="datasets" url={datasetsUrl} />
+                        <DatasetsComponent columns={datasetsColumns} />
+                    </FetchedData>
+                </SortTablePanel>
+            : null }
+
+            {context.datasets && context.datasets.length ?
+                <FetchedData ignoreErrors>
+                    <Param name="experiments" url={`/search/?type=Experiment&target.investigated_as!=control`} />
```
forresttanaka: You can just use double quotes instead of braces and backticks here. ESLint is giving a quotes error here.

src/encoded/static/components/publication.js
```diff
@@ -26,7 +28,64 @@ const PublicationComponent = (props, reactContext) => {
     ];
 
     const crumbsReleased = (context.status === 'released');
-
+    const accessions = context.datasets.reduce((dataset1, dataset2) => `${dataset1}&accession=${dataset2.accession}`, '');
```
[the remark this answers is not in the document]
zoldello: corrected

src/encoded/static/components/publication.js
zoldello: corrected

src/encoded/static/components/publication.js
zoldello: corrected

src/encoded/static/components/publication.js
zoldello: Based on looking at it, it may have been me experimenting with SortTable prior to implementing this feature. I removed it as it has no use now.

src/encoded/static/components/publication.js
zoldello: Removed.

src/encoded/static/components/publication.js
zoldello: Removed.

src/encoded/static/components/publication.js
```diff
@@ -7,7 +7,63 @@ import { DbxrefList } from './dbxref';
...
-
+import { SortTablePanel, SortTable } from './sorttable';
+import { FetchedData, Param } from './fetched';
+
+
+const sort = (a, b) => a - b;
+const datasetsColumns = {
+    accession: {
+        display: dataset => <a href={dataset['@id']}>{dataset.accession}</a>,
+        sort,
```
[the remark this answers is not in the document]
zoldello: Great caught. That is the sort for numbers and not strings. I removed sort and use the default.

src/encoded/static/components/publication.js
```diff
@@ -26,7 +73,11 @@ const PublicationComponent = (props, reactContext) => {
     ];
 
     const crumbsReleased = (context.status === 'released');
-
+    const accessions = !context.datasets ? '' : context.datasets.reduce((dataset1, dataset2) => `${dataset1}&accession=${dataset2.accession}`, '');
+    const fields = ['accession', 'target', '@type', 'target', 'biosample_summary', 'lab', 'assay_term_name', 'assay', 'description', 'status'].reduce((field1, field2) => `${field1}&field=${field2}`, '');
+    const datasetsUrl = context.datasets && context.datasets.length ?
+        `/search/?${accessions}${fields}&format=json` :
```
forresttanaka: You can change this if you want; `FetchedData` specifies JSON, so `format=json` in the URL here is redundant, but it also doesn’t hurt anything.

caseylitton:
closed, code is not on current rc.

PR #2614, opened 2019-02-25
ENCD-3901 Support multiple carts per user

forresttanaka:
### Front end
* Most of the new functionality for the page that lets users manage their carts is in cart/manager.js.
* The Redux store for carts used to use an array called `carts` that held the `@ids` of every experiment in the cart. I‘ve renamed that to `elements` because I often use “cart” to mean the cart object from the database, and that was getting confusing. You’ll see this change across much of the code, and even in the Jest test for carts.
* I was sometimes not distinguishing the nuances of various Redux concepts, so some of the comment changes are to fix that.
* I moved `this.cartStore` from `<App>` to the module variable `cartStore` in cart/index.js. This allows other functions access to the cart store, and I found this widely recommended to share Redux stores among code — define it as a module variable and import it to whomever needs to read it. `cartStore` gets initialized at page load, and its contents get updated through Redux actions.
* The new use of the term “cart settings” always means the cart information stored in the browser’s localstorage, which at this stage is only the `@id` of the logged-in user’s current cart. You’ll also see a `quickSelects` array but I don’t use it; I might in the future if we get demand for it. The localstorage key is always “encode-cart-” and then the `@id` of the user it belongs to. All code related to cart settings is in cart/settings.js.
* I had been using the boolean `activeCart` to distinguish between non-logged-in and logged-in carts, but found that confusing. You’ll now see `cartType` which can have the values “OBJECT” for displays using a saved cart object with `@id`, “MEMORY” for unsaved carts in memory used by logged-out users, and “ACTIVE” for the current cart object that logged-in users are manipulating and displaying. You’ll generally see an elimination to the term “active cart.”
* cart/database.js looks like a new file, but it’s just modified and renamed from cart/save.js, as it doesn’t just have to do with saving carts.
* We now use all three possible statuses for a cart object. “current” for any cart available for use, “deleted” for a deleted cart, and “disabled” for a user’s auto-save cart.
* One small change to `SortTable` to allow custom CSS for table headers, and a similar one to `ModalBody`. Another change to the modals is to more closely confirm to a11y.
### Back end
* I moved `_get_carts_by_user` from types/cart.py to cart_view.py to join its close twin `get_cart_objects_by_user` that both cart_view.py and cart.py need. I’m not sure if that’s the right direction for the import though. I could see a couple other cases of files in the types directory importing from the parent directory, and none in reverse, so that’s all I based that on.


forresttanaka:
All the Javascript codacy warnings are because their ESLint configuration conflicts with ours. I fixed one Python one, but don‘t understand the other one.

src/encoded/schemas/cart.json
```diff
@@ -23,12 +23,19 @@
             "description": "The name of the cart",
             "type": "string"
         },
+        "identifier": {
+            "title": "URI-compatible cart identifier",
+            "description": "User-specified identifer for this cart that can be used in a URI for the cart.",
+            "comment": "This can only contain non-space characters, and must be unique througout the system.",
```
emro: small typo: "throughout" not "througout"

src/encoded/cart_view.py
```diff
@@ -1,14 +1,26 @@
...
+    config.add_route('cart-manager', '/cart-manager{slash:/?}')
     config.scan(__name__)
 
 
+def get_cart_objects_by_user(request, userid):
+    request.datastore = 'database'
+    return [
+        request.embed(request.resource_path(v), '@@object')
+        for k, v in request.registry[COLLECTIONS]['cart'].items()
+        if v.properties['submitted_by'] == userid and (v.properties['status'] != 'deleted' or 'group.admin' in request.effective_principals)
```
caseylitton: This ones a little long too, can you expand with parens?

src/encoded/cart_view.py
```diff
@@ -1,14 +1,26 @@
 from pyramid.view import view_config
 from collections import OrderedDict
+from snovault import COLLECTIONS
+from pyramid.httpexceptions import HTTPBadRequest
```
caseylitton: It's better to group "like" imports...so all pyramid or all the snovault imports are in the same block.  pep8 says to put a space between groups of imports, we don't have a pattern of leaving spaces, but keeping the organized into groups makes it more readable. 

Also, I didn't find this in pep8, but I'll usually leave two spaces between imports, top level vars, and the first function declaration.

src/encoded/types/cart.py
```diff
@@ -87,12 +90,46 @@ def get_or_create_cart_by_user(context, request):
...
-    carts = _get_carts_by_user(request, userid)
-    if not carts:
-        cart = _create_cart(request, user)
+    cart_name = None
+    all_carts = get_cart_objects_by_user(request, userid)
+    countable_carts = [cart for cart in all_carts if cart['status'] != 'deleted' and cart['status'] != 'disabled']
+    cart_atids = [cart['@id'] for cart in all_carts] if all_carts else []
+    cart_status = None
+    cart_name = None
+    try:
```
caseylitton: I'm not sure we need these try/catches.  Can we try initializing the cart_status and cart_name vars directly from the request.json.get.  Default the .get('name', '') with the empty string to the strip won't fail.  If the keys don't exist both should be falsey.

src/encoded/types/cart.py
```diff
@@ -48,32 +54,29 @@ class Collection(Item.Collection):
...
-            'submitted_by': str(user.uuid),
-            'status': 'current',
-            'name': '{} cart'.format(user_props['title']),
-            'elements': []
-        }
+        initial_cart,
     )
-    return request.resource_path(cart)
+    cart_path = request.resource_path(cart)
+    cart.properties['current'] = cart_path
```
keenangraham: What is this doing?

src/encoded/types/cart.py
```diff
@@ -87,12 +90,46 @@ def get_or_create_cart_by_user(context, request):
...
+        # Error if more non-deleted/non-disabled carts than allowed per user.
+        if len(countable_carts) >= CART_USER_MAX:
+            msg = 'Users cannot have more than {} carts'.format(CART_USER_MAX)
+            raise HTTPBadRequest(explanation=msg)
+        elif next((cart for cart in all_carts if cart['status'] != 'deleted' and cart['status'] != 'disabled' and cart['name'].strip().upper() == cart_name.upper()), None):
+            msg = 'A cart with the name "{}" already exists'.format(cart_name)
+            raise HTTPBadRequest(explanation=msg)
+
+    # Create a cart if PUT request, or GET request when none exist
+    cart = None
```
keenangraham: Suspicious number of Nones here.

src/encoded/types/cart.py
```diff
@@ -48,32 +54,29 @@ class Collection(Item.Collection):
...
+        initial_cart,
     )
-    return request.resource_path(cart)
+    cart_path = request.resource_path(cart)
+    cart.properties['current'] = cart_path
+    return cart_path
 
 
-@view_config(context=Cart.Collection, request_method='GET', permission='save-carts', name='get-cart')
+@view_config(context=Cart.Collection, request_method=('GET', 'PUT'), permission='save-carts', name='get-cart')
```
keenangraham: It might be cleaner if we make a separate GET and PUT view.

src/encoded/types/cart.py
```diff
@@ -87,12 +90,46 @@ def get_or_create_cart_by_user(context, request):
     user = request.registry[COLLECTIONS]['user'].get(userid)
     if not user:
         raise HTTPBadRequest()
-    carts = _get_carts_by_user(request, userid)
-    if not carts:
-        cart = _create_cart(request, user)
+    cart_name = None
+    all_carts = get_cart_objects_by_user(request, userid)
+    countable_carts = [cart for cart in all_carts if cart['status'] != 'deleted' and cart['status'] != 'disabled']
```
keenangraham: Sometimes easier to read if you split into multiple lines:
```python
[
    cart
    for cart in all_carts
    if cart['status'] not in ['deleted',  'disabled']
]
```

src/encoded/types/cart.py
```diff
@@ -87,12 +90,46 @@ def get_or_create_cart_by_user(context, request):
...
+    cart_atids = [cart['@id'] for cart in all_carts] if all_carts else []
+    cart_status = None
+    cart_name = None
+    try:
+        cart_status = request.json.get('status')
```
keenangraham: Why are we getting JSON from request here?

src/encoded/types/cart.py
```diff
@@ -31,11 +36,12 @@
...
     item_type = 'cart'
     schema = load_schema('encoded:schemas/cart.json')
+    name_key = 'identifier'
 
     STATUS_ACL = {
         'current': [(Allow, 'role.owner', ['view', 'edit'])] + ALLOW_CURRENT,
-        'deleted': DELETED,
-        'disabled': ONLY_ADMIN_VIEW,
+        'deleted': [(Allow, 'role.owner', ['edit'])] + DELETED,
+        'disabled': [(Allow, 'role.owner', ['view', 'edit'])],
```
keenangraham: It might be a good idea to have `+ ONLY_ADMIN_VIEW` at the end of this. We should also make sure there are permission tests for this.

src/encoded/cart_view.py
```diff
@@ -1,14 +1,26 @@
...
 
 def includeme(config):
     # config.scan()
-    # config.add_route('carts', '/carts{slash:/?}')
     config.add_route('cart-view', '/cart-view{slash:/?}')
+    config.add_route('cart-manager', '/cart-manager{slash:/?}')
     config.scan(__name__)
 
 
+def get_cart_objects_by_user(request, userid):
```
keenangraham: We shouldn't override this function with status and permission filtering to encourage reuse later. (In other words since the function is called `get_cart_objects_by_user` it should just get the cart objects by user.)

src/encoded/types/cart.py
forresttanaka: The first GET request to @@get-cart passes no JSON, so right now I broke into PDB on this line and get:

```
(Pdb) request
<Request at 0x1241bfe10 GET http://localhost:6543/carts/@@get-cart>
(Pdb) request.json
*** simplejson.scanner.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
(Pdb)
```

And if I let it go, I get a 400 error on the GET request. I’ll try @keenangraham’s suggestion of breaking out the GET and PUT to work around this, so that the PUT just expects JSON and the GET never does. I assume that would answer Keenan’s question just below this one as well.

src/encoded/schemas/cart.json
forresttanaka: Corrected

src/encoded/types/cart.py
```diff
@@ -87,12 +90,46 @@ def get_or_create_cart_by_user(context, request):
...
+        pass
+
+    # User writing a new cart; check for cart overflow and naming conflicts
+    if request.method == 'PUT' and cart_status != 'disabled':
+        # PUT creates new cart even if user already has carts. 'name' required in body.
+        # Error if more non-deleted/non-disabled carts than allowed per user.
+        if len(countable_carts) >= CART_USER_MAX:
+            msg = 'Users cannot have more than {} carts'.format(CART_USER_MAX)
+            raise HTTPBadRequest(explanation=msg)
+        elif next((cart for cart in all_carts if cart['status'] != 'deleted' and cart['status'] != 'disabled' and cart['name'].strip().upper() == cart_name.upper()), None):
```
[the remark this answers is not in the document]
forresttanaka: Broke up the line, but not sure my indenting is preferred.

src/encoded/cart_view.py
forresttanaka: Changed with another change.

src/encoded/cart_view.py
forresttanaka: Changed to group them and add blank lines.

src/encoded/types/cart.py
forresttanaka: Changed.

src/encoded/cart_view.py
forresttanaka: Changed, and added a separate function for filtering.

src/encoded/types/cart.py
forresttanaka: I separated GET and PUT views which I think addresses the other things. Thanks; that really makes things look better.

src/encoded/types/cart.py
forresttanaka: I added `+ ONLY_ADMIN_VIEW` and it seems to work properly still, and pushed this change.

forresttanaka: Added a couple (enough?) permissions tests for disabled carts.

src/encoded/types/cart.py
```diff
@@ -96,3 +108,47 @@ def get_or_create_cart_by_user(context, request):
         '@type': ['result'],
         '@graph': carts or [cart]
     }
+
+
+@view_config(context=Cart.Collection, request_method='PUT', permission='save-carts', name='put-cart')
+def create_cart_by_user(context, request):
+    userid = [
```
keenangraham: The common code between get and put could be put into separate function.

src/encoded/types/cart.py
```diff
@@ -96,3 +108,47 @@ def get_or_create_cart_by_user(context, request):
...
+        if cart['status'] != 'disabled'
+    ]
+    cart_atids = [cart['@id'] for cart in visible_user_carts] if visible_user_carts else []
+    cart_status = request.json.get('status', None)
+    cart_name = request.json.get('name', '').strip()
+    # User writing a new cart; check for cart overflow and naming conflicts
+    if len(countable_carts) >= CART_USER_MAX:
+        msg = 'Users cannot have more than {} carts'.format(CART_USER_MAX)
+        raise HTTPBadRequest(explanation=msg)
+    elif next((cart for cart in visible_user_carts if (cart['status'] != 'deleted' and
```
keenangraham: I think you can use `any` here, e.g.: 
```python
elif any(
    [
        cart
        for cart in visible_user_carts
        if cart_conditions
    ]
)

File: src/encoded/types/cart.py [hunk 15]
```diff
@@ -96,3 +108,47 @@ def get_or_create_cart_by_user(context, request):
...
+        raise HTTPBadRequest()
+    cart_name = None
+    user_carts = get_cart_objects_by_user(request, userid)
+    visible_user_carts = filter_carts_to_visible(request, user_carts)
+    countable_carts = [
+        cart
+        for cart in visible_user_carts
+        if cart['status'] != 'disabled'
+    ]
+    cart_atids = [cart['@id'] for cart in visible_user_carts] if visible_user_carts else []
```
keenangraham: I don't think `if visible_user_carts else []` condition is necessary here because visible_user_carts will just be an empty cart if it doesn't exist. So cart_atids will be an empty list if you iterate over an empty list.

src/encoded/types/cart.py
forresttanaka: Is there a big risk/benefit advantage to using `any` here?

src/encoded/types/cart.py
caseylitton: thx, I'm pushing this format too.  They tend to get long and it's easier to add more conditions later on.

src/encoded/types/cart.py
```diff
@@ -85,8 +75,7 @@ def _create_cart(request, user, name=None, identifier=None, status=None):
     return cart_path
 
 
-@view_config(context=Cart.Collection, request_method='GET', permission='save-carts', name='get-cart')
-def get_or_create_cart_by_user(context, request):
+def _get_user_carts(request, blocked_statuses=[]):
```
keenangraham: The _get_user_carts versus get_cart_objects_by_user function is still confusing to me. Probably there needs to be a _get_userid function with the first part of this logic: 

```python
userid = [
        p.replace('userid.', '')
        for p in request.effective_principals
        if p.startswith('userid.')
    ]
    if not userid:
        raise HTTPBadRequest()
    else:
        userid = userid[0]
    user = request.registry[COLLECTIONS]['user'].get(userid)
    if not user:
        raise HTTPBadRequest()
```

And the get_cart_objects_by_user can be called in both files (and _get_user_carts can be deleted).

src/encoded/cart_view.py
```diff
@@ -59,19 +52,18 @@ def cart_manager(context, request):
         raise HTTPBadRequest()
     else:
         userid = userid[0]
```
keenangraham: You can also use _get_userid function here instead of duplicating code.

src/encoded/types/cart.py
```diff
@@ -99,47 +88,46 @@ def get_or_create_cart_by_user(context, request):
...
+    carts = get_cart_objects_by_user(request, userid, blocked_statuses)
+    return carts, user
+
+
+@view_config(context=Cart.Collection, request_method='GET', permission='save-carts', name='get-cart')
+def get_or_create_cart_by_user(context, request):
+    carts, user = _get_user_carts(request, ['disabled', 'deleted '])
     if not carts:
         cart = _create_cart(request, user)
+        cart_atids = None
```
keenangraham: This None seems like it could be cleaned up a bit. One suggestion:
```python
def get_or_create_cart_by_user(context, request):
    carts, user = _get_user_carts(request, ['disabled', 'deleted '])
    carts = [cart['@id'] for cart in carts]
    if not carts:
        cart = _create_cart(request, user)
    request.response.status = 200
    return {
        'status': 'success',
        '@type': ['result'],
        '@graph': carts or [cart]
    }
```

src/encoded/types/cart.py
```diff
@@ -99,47 +88,46 @@ def get_or_create_cart_by_user(context, request):
...
-    cart_name = None
-    user_carts = get_cart_objects_by_user(request, userid)
-    visible_user_carts = filter_carts_to_visible(request, user_carts)
-    countable_carts = [
-        cart
-        for cart in visible_user_carts
-        if cart['status'] != 'disabled'
-    ]
-    cart_atids = [cart['@id'] for cart in visible_user_carts] if visible_user_carts else []
+    user_carts, user = _get_user_carts(request, ['deleted'] if not 'group.admin' in request.effective_principals else [])
```
keenangraham: Again inline if/else in a function call.

keenangraham: Also they are called `carts` in GET and `user_carts` in PUT. Would be nice to make them consistent.

src/encoded/types/cart.py
```diff
@@ -99,47 +88,46 @@ def get_or_create_cart_by_user(context, request):
...
+    cart_atids = [cart['@id'] for cart in user_carts]
     cart_status = request.json.get('status', None)
     cart_name = request.json.get('name', '').strip()
     # User writing a new cart; check for cart overflow and naming conflicts
+    if cart_status != 'disabled':
+        countable_carts = [
+            cart
+            for cart in user_carts
+            if cart['status'] != 'disabled' and cart['status'] != 'deleted'
```
keenangraham: `if cart['status'] not in ['disabled', 'deleted']`

src/encoded/types/cart.py
```diff
@@ -99,47 +88,46 @@ def get_or_create_cart_by_user(context, request):
...
+        ]
+    else:
```
keenangraham: Just set countable_carts = [] at the top of the function to avoid this else statement.

src/encoded/types/cart.py
```diff
@@ -99,47 +88,46 @@ def get_or_create_cart_by_user(context, request):
...
+        # Creating an auto-save cart, so ignore existing carts
+        countable_carts = []
     if len(countable_carts) >= CART_USER_MAX:
         msg = 'Users cannot have more than {} carts'.format(CART_USER_MAX)
         raise HTTPBadRequest(explanation=msg)
-    elif next((cart for cart in visible_user_carts if (cart['status'] != 'deleted' and
+    elif next((cart for cart in user_carts if (cart['status'] != 'deleted' and
```
keenangraham: @caseylitton points out you don't have to use elif here since previous if raises error. The `any` and `all` keywords are pretty standard and imply that you are checking condition of items in list, which is what you are doing here. The `next` keyword implies that you are yielding a value in order to do something with it, which is confusing in this case because you aren't doing anything with the item. It can work well if you want to return first item from list if it exists, but aren't sure if it does.

src/encoded/types/cart.py
```diff
@@ -48,51 +54,94 @@ class Collection(Item.Collection):
...
+    if len(countable_carts) >= CART_USER_MAX:
+        msg = 'Users cannot have more than {} carts'.format(CART_USER_MAX)
+        raise HTTPBadRequest(explanation=msg)
+    elif next((cart for cart in user_carts if (cart['status'] != 'deleted' and
+            cart['status'] != 'disabled' and
+            cart['name'].strip().upper() == cart_name.upper())), None):
+        msg = 'A cart with the name "{}" already exists'.format(cart_name)
+        raise HTTPBadRequest(explanation=msg)
+    cart_identifier = request.json.get('identifier', None)
+    cart = _create_cart(request, user, cart_name, cart_identifier, cart_status)
```
keenangraham: @caseylitton also points out we shouldn't be passing keyword arguments as position arguments.

```python
cart = _create_cart(
   request,
   user,
   name=cart_name,
   identifier=cart_identifier,
   status=cart_status
)
```

src/encoded/types/cart.py
```diff
@@ -48,51 +54,94 @@ class Collection(Item.Collection):
...
         raise HTTPBadRequest()
-    carts = _get_carts_by_user(request, userid)
+    return user
+
+
+@view_config(context=Cart.Collection, request_method='GET', permission='save-carts', name='get-cart')
+def get_or_create_cart_by_user(context, request):
+    userid = _get_userid(request)
+    user = _get_user(request, userid)
+    carts = get_cart_objects_by_user(request, userid, ['disabled', 'deleted '])
```
keenangraham: Similarly should pass as keyword argument:

```python
carts = get_cart_objects_by_user(
    request,
    userid,
    blocked_statuses=['disabled', 'deleted ']
)
```

src/encoded/types/cart.py
forresttanaka: What's the problem with else?

keenangraham: It's just not really doing much here. The if statement catches everything but disabled. In my mind it makes sense to start with an empty cart and add to it based on the condition, rather than having forking paths.

src/encoded/types/cart.py
forresttanaka: I think you’re looking at old code. There is no _get_user_carts, and there is a _get_userid

src/encoded/types/cart.py
```diff
@@ -48,51 +54,94 @@ class Collection(Item.Collection):
...
+    }
+
+
+@view_config(context=Cart.Collection, request_method='PUT', permission='save-carts', name='put-cart')
+def create_cart_by_user(context, request):
+    userid = _get_userid(request)
+    user = _get_user(request, userid)
+    user_carts = get_cart_objects_by_user(request, userid, ['deleted'] if not 'group.admin' in request.effective_principals else [])
+    cart_atids = [cart['@id'] for cart in user_carts]
+    cart_status = request.json.get('status', None)
```
keenangraham: Get returns None by default so no need to specify for most of these.

src/encoded/types/cart.py
keenangraham: Well that helps with the confusion. Sorry about that.

src/encoded/types/cart.py
forresttanaka: I made this change, but for the record, I don’t like that I’m assigning to countable_carts potentially twice when before I was always assigning it once.

src/encoded/cart_view.py
```diff
@@ -16,22 +16,15 @@ def includeme(config):
...
 
 
-def get_cart_objects_by_user(request, userid):
+def get_cart_objects_by_user(request, userid, blocked_statuses=[]):
     request.datastore = 'database'
     return [
         request.embed(request.resource_path(v), '@@object')
         for k, v in request.registry[COLLECTIONS]['cart'].items()
-        if v.properties['submitted_by'] == userid
+        if v.properties['submitted_by'] == userid and not v.properties['status'] in blocked_statuses
```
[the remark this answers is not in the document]
forresttanaka: Changed

src/encoded/cart_view.py
```diff
@@ -59,19 +52,18 @@ def cart_manager(context, request):
         raise HTTPBadRequest()
     else:
         userid = userid[0]
-    user_carts = get_cart_objects_by_user(request, userid)
-    visible_user_carts = filter_carts_to_visible(request, user_carts)
+    user_carts = get_cart_objects_by_user(request, userid, ['deleted'] if not 'group.admin' in request.effective_principals else [])
```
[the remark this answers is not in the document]
forresttanaka: Changed.

src/encoded/types/cart.py
forresttanaka: Changed, though I didn’t like going through a loop when I didn’t have to, that case would loop zero times so it’s probably not a penalty. Not my preferred choice though.

src/encoded/types/cart.py
forresttanaka: Yeah, changed.

src/encoded/types/cart.py
forresttanaka: Changed.

src/encoded/types/cart.py
forresttanaka: Changed and tested, and it seems to work the same as before. I agree that makes more sense.

src/encoded/types/cart.py
forresttanaka: I couldn’t find confirmation this is true, but oh well.

src/encoded/types/cart.py
forresttanaka: Changed

src/encoded/types/cart.py
forresttanaka: Changed

src/encoded/types/cart.py
```diff
@@ -48,46 +55,44 @@ class Collection(Item.Collection):
...
+
+
+@view_config(context=Cart.Collection, request_method='GET', permission='save-carts', name='get-cart')
+def get_or_create_cart_by_user(context, request):
+    userid = get_userid(request)
+    user = _get_user(request, userid)
+    carts = get_cart_objects_by_user(
+        request,
+        userid,
+        blocked_statuses=['disabled', 'deleted ']
```
keenangraham: Extra space.

src/encoded/types/cart.py
```diff
@@ -96,3 +101,51 @@ def get_or_create_cart_by_user(context, request):
...
         '@graph': carts or [cart]
     }
+
+
+@view_config(context=Cart.Collection, request_method='PUT', permission='save-carts', name='put-cart')
+def create_cart_by_user(context, request):
+    userid = get_userid(request)
+    user = _get_user(request, userid)
+    blocked_statuses = ['deleted'] if 'group.admin' not in request.effective_principals else []
+    carts = get_cart_objects_by_user(request, userid, blocked_statuses)
```
keenangraham: Blocked_statuses also needs to be passed as keyword argument as above.

src/encoded/types/cart.py
```diff
@@ -96,3 +101,51 @@ def get_or_create_cart_by_user(context, request):
...
+            if cart['status'] not in ['disabled', 'deleted']
+        ]
+    if len(countable_carts) >= CART_USER_MAX:
+        msg = 'Users cannot have more than {} carts'.format(CART_USER_MAX)
+        raise HTTPBadRequest(explanation=msg)
+    name_conflict = any(
+        [
+            cart
+            for cart in carts
+            if (cart['status'] != 'deleted' and
```
keenangraham: `cart['status'] not in ['deleted', 'disabled']`

src/encoded/types/cart.py
keenangraham: Pyramid uses request object from Webob that sets `.json` and `.json_body` to the same thing: https://github.com/Pylons/webob/blob/b5297529ef5a8dd3dd84c6cfa41afd101acfa8a2/src/webob/request.py#L739

This just uses the Python JSON library that handles input as follows:
```
JSON	Python
object	dict
array	list
string	unicode
number (int)	int, long
number (real)	float
true	True
false	False
null	None
```

So you should be getting a dict back based on what you have PUT to the endpoint. Python dicts have the get method which you are using, defined as follows:

> get(key[, default])
> Return the value for key if key is in the dictionary, else default. If default is not given, it defaults to None, so that this method never raises a KeyError.

https://docs.python.org/3/library/stdtypes.html#dict.get

forresttanaka: Ah thanks, that helps. I thought that .get was something specific to request.json, not a general dict thing.

src/encoded/types/cart.py
forresttanaka: Damn.

src/encoded/types/cart.py
forresttanaka: OK

src/encoded/types/cart.py
forresttanaka: Changed.

src/encoded/types/cart.py
```diff
@@ -96,3 +101,50 @@ def get_or_create_cart_by_user(context, request):
...
+    if cart_status != 'disabled':
+        countable_carts = [
+            cart
+            for cart in carts
+            if cart['status'] not in ['disabled', 'deleted']
+        ]
+    if len(countable_carts) >= CART_USER_MAX:
+        msg = 'Users cannot have more than {} carts'.format(CART_USER_MAX)
+        raise HTTPBadRequest(explanation=msg)
+    name_conflict = any(
```
keenangraham: Sorry one more change with this.

I realized that we are just filtering a list here and if anything is remaining in the list when you check `if name_conflict` it will evaluate to True.

So either:
1. We remove `any` and just evaluate the list comprehension.
2. We remove the list comprehension and pass `any` a generator comprehension.

Potentially the benefit of #2 is that the expression will stop evaluating as soon as it reaches a true value, instead of iterating through the whole list.

I don't really have a strong preference either way but we shouldn't leave it how it is now.

src/encoded/types/cart.py
```diff
@@ -96,3 +101,48 @@ def get_or_create_cart_by_user(context, request):
...
+        user,
+        name=cart_name,
+        identifier=cart_identifier,
+        status=cart_status
+    )
+    request.response.status = 200
+    return {
+        'status': 'success',
+        '@type': ['result'],
+        '@graph': [cart] if cart else cart_atids,
```
keenangraham: When does cart not exist?

src/encoded/types/cart.py
```diff
@@ -48,46 +55,44 @@ class Collection(Item.Collection):
...
+
+
+@view_config(context=Cart.Collection, request_method='GET', permission='save-carts', name='get-cart')
+def get_or_create_cart_by_user(context, request):
+    userid = get_userid(request)
+    user = _get_user(request, userid)
+    carts = get_cart_objects_by_user(
+        request,
+        userid,
+        blocked_statuses=['disabled', 'deleted']
```
keenangraham: Just to confirm these should be filtered for all users including admin?

forresttanaka: This only gets called at page reload, and only to get the current usable carts or to make one, so it doesn’t matter if you’re admin or submitter — you don’t want unusable carts. The cart manager display uses the cart manager view which is different.

src/encoded/types/cart.py
keenangraham: ?

forresttanaka:
Changes this morning based on QA findings. Most of the big-looking changes are just indenting changes because I had to convert one functional component to a React class.

src/encoded/static/components/cart/manager.js
```diff
@@ -60,13 +61,20 @@ CurrentCartButtonComponent.propTypes = {
     cart: PropTypes.object.isRequired,
     /** @id of the current cart */
     current: PropTypes.string.isRequired,
+    /** True if cart operation in progress */
+    operationInProgress: PropTypes.bool,
     /** Current user object from session_properties */
     user: PropTypes.object,
     /** Dispatch function for setting the current cart in the cart Redux store */
     onCurrentCartClick: PropTypes.func.isRequired,
+    /** FUnction to call when switching operation starts */
```
emro: minor typo "FUnction" should be "Function"

src/encoded/static/components/cart/actions.js
```diff
@@ -244,7 +244,7 @@ export const setCartStatus = status => (
  * @return {Promise} Resolves to the updated cart object
  */
 export const setCartNameIdentifierAndSave = ({ name, identifier }, cart, user, fetch) => (
-    (dispatch, getState) => (
+    dispatch => (
```
emro: Why did this change?

src/encoded/cart_view.py
```diff
@@ -30,7 +30,7 @@ def get_userid(request):
 def get_cart_objects_by_user(request, userid, blocked_statuses=[]):
     request.datastore = 'database'
     return [
-        request.embed(request.resource_path(v), '@@object')
+        dict(v.properties, **{'@id': request.resource_path(v)})
```
emro: I'm not sure what this line does, maybe @caseylitton can review?

forresttanaka: `request.embed` takes a very long time to execute, and all we need is the `@id`. So this line gets the `@id` of the cart object and puts it into the `@id` property of the cart item. This is to get around that the cart item doesn’t have an @id nor uuid, so `request.embed` adds it but is extremely slow. This way we get the `@id` quickly

src/encoded/static/components/cart/actions.js
forresttanaka: getState isn’t used in this function, so it generated an ESlint error I hadn’t noticed before.

src/encoded/cart_view.py
forresttanaka: With four carts with hundreds of items each, this was taking 5.6 seconds with request.embed. With this change, it took 283ms

emro: Oh wow, 5.6 seconds each! 283 ms is much more reasonable.

PR #2617, opened 2019-02-26
ENCD-4474 model restriction map

p-sud:


p-sud:
"In dependencies you should put minItems 1 for restriction enzymes array"

Wouldn't it be better to put this under the schema "properties"? Otherwise you'd need it in two places.

PR #2621, opened 2019-02-27
ENCD-4217 Rendering error series objects

emro:
Series objects with life_stage value not present caused a rendering error.
Now tables will insert a null value for those cases.
This ticket also fixes another related error uncovered during debugging: headers were displayed for tables that did not exist (due to related datasets not being viewable).


caseylitton:
Closing pr and moving back to in progress. Pr is not rebased on origin/dev at the time of making the PR. and the demo code is not the same and PR.

PR #2623, opened 2019-02-28
ENCD-4473-add-MPRA-assay

khzlin:


jahilton:
Please also update the STARR-seq experiment in the data inserts - should also be functional characterization assay

src/encoded/tests/test_batch_download.py
```diff
@@ -73,7 +73,7 @@ def test_batch_download_report_download(testapp, workbook):
     #     b'', b'', b'', b'', b'', b'', b'', b'', b'',
     #     b'', b'', b'', b'', b'', b'', b''
     # ]
-    assert len(lines) == 48
+    assert len(lines) == 49
```
caseylitton: This is being updated in 
https://github.com/ENCODE-DCC/encoded/pull/2627/files

One will need to rebase on the other depending on which is first.

PR #2625, opened 2019-03-01
ENCD-4484 Allow deployment set ssh id file

caseylitton:


src/encoded/commands/deploy.py
```diff
@@ -194,15 +194,13 @@ def tag_ec2_instance(instance, tag_data, elasticsearch, cluster_name):
     return instance
 
 
-def read_ssh_key():
-    home = expanduser("~")
-    ssh_key_path = home + '/' + '.ssh/id_rsa.pub'
-    ssh_keygen_args = ['ssh-keygen', '-l', '-f', ssh_key_path]
+def read_ssh_key(identity_file):
+    ssh_keygen_args = ['ssh-keygen', '-l', '-f', identity_file]
```
zoldello: Where do you do some kind of validation on identify_file? You can only trust data an object owns and does not share.

src/encoded/commands/deploy.py
```diff
@@ -232,7 +230,7 @@ def _get_bdm(main_args):
 def get_user_data(commit, config_file, data_insert, main_args):
     cmd_list = ['git', 'show', commit + config_file]
     config_template = subprocess.check_output(cmd_list).decode('utf-8')
-    ssh_pub_key = read_ssh_key()
+    ssh_pub_key = read_ssh_key(main_args.identity_file)
```
zoldello: I have not ran this myself so this may be an odd question. The variable- main_args is set in main(). However, it seems internal to that method. It may be wise to push it up to the class-level or __init__, so reads know it is class-wide.

caseylitton: It has to be instantiated in main to get the command line args.  Ideally this script would subclass most of what is in main(), then we instantiate some "TBD" DeployClass where the main_args would be the initialization arguments.

yea, handing main_args though all the functions is a little weird.

src/encoded/commands/deploy.py
zoldello: I got the answer. Validation is done implicitly in subprocess.check_output

src/encoded/commands/deploy.py
zoldello: I got the answer. main() passes main-args around. It is safe to assume main() is called prior to any other object getting called.

PR #2626, opened 2019-03-01
ENCD-4483 Cleanup deploy output

caseylitton:


src/encoded/commands/deploy.py
```diff
@@ -373,26 +395,58 @@ def _get_run_args(main_args, instances_tag_data):
...
+    if main_args.elasticsearch == 'yes' and run_args['count'] > 1:
+        if cluster_master and run_args['master_user_data']:
+            is_cluster_master = True
+            output_list.append('Creating Elasticsearch Master Node for cluster')
         else:
-            instances_tag_data['name'] = tmp_name
+            is_cluster = True
+            output_list.append('Creating Elasticsearch cluster')
+    created_cluster_master = False
+    for i, instance in enumerate(instances):
```
zoldello: Rather than "enumerate(instances)", can you use instances.keys()? I do not see where you use "instance" from enumerate(instances)

src/encoded/commands/deploy.py
```diff
@@ -373,26 +395,58 @@ def _get_run_args(main_args, instances_tag_data):
...
+            elif is_cluster:
+                output_list.append('  # %s' % instance.id)
+            elif not is_cluster:
+                output_list.extend(_get_instance_output(instances_tag_data))
             instance.wait_until_exists()
             tag_ec2_instance(instance, instances_tag_data, main_args.elasticsearch, main_args.cluster_name)
-            print('ssh %s.%s.encodedcc.org' % (instances_tag_data['name'], domain))
-            print('https://%s.demo.encodedcc.org' % instances_tag_data['name'])
+    for output in output_list:
+        print(output)
```
zoldello: Not Significant: Do you need additional information along with the print-out..? More information in these of a small explanation.

src/encoded/commands/deploy.py
```diff
@@ -373,26 +395,58 @@ def _get_run_args(main_args, instances_tag_data):
...
-                instances_tag_data['name'] = "{}{}".format(tmp_name[0:-1], 'master')
-            else:
-                print('Creating Elasticsearch cluster')
-                instances_tag_data['name'] = "{}{}".format(tmp_name, i)
+    instances_tag_data['domain'] = 'production' if main_args.profile_name == 'production' else 'instance'
+    output_list = ['']
+    is_cluster_master = False
+    is_cluster = False
+    if main_args.elasticsearch == 'yes' and run_args['count'] > 1:
+        if cluster_master and run_args['master_user_data']:
```
zoldello: Not significant: Are you guaranteed to have "master_user_data'" in run_args? Otherwise, this will blow up.

src/encoded/commands/deploy.py
```diff
@@ -275,9 +296,10 @@ def _get_instances_tag_data(main_args):
         sys.exit(1)
     instances_tag_data['username'] = getpass.getuser()
     if instances_tag_data['name'] is None:
+        instances_tag_data['short_name'] = _short_name(instances_tag_data['branch'])
         instances_tag_data['name'] = nameify(
             '%s-%s-%s' % (
```
zoldello: Not significant: It would be nice to have Python 3.6 or higher, where we can use literal string interpolation (f-string.)

src/encoded/commands/deploy.py
```diff
@@ -275,9 +296,10 @@ def _get_instances_tag_data(main_args):
         sys.exit(1)
     instances_tag_data['username'] = getpass.getuser()
     if instances_tag_data['name'] is None:
+        instances_tag_data['short_name'] = _short_name(instances_tag_data['branch'])
```
zoldello: Not significant: It would be nice to have Python 3.6 or higher, where we can use literal string interpolation (f-string.)

src/encoded/commands/deploy.py
```diff
@@ -275,9 +296,10 @@ def _get_instances_tag_data(main_args):
...
         instances_tag_data['name'] = nameify(
             '%s-%s-%s' % (
-                instances_tag_data['branch'],
+                instances_tag_data['short_name'],
```
zoldello: Not significant: It would be nice to have Python 3.6 or higher, where we can use literal string interpolation (f-string.)

src/encoded/commands/deploy.py
caseylitton: yea, that was preexisting.  I didn't want to change too much.

src/encoded/commands/deploy.py
caseylitton: no, all info should be in there already.  Is just collects the prints and outputs once.

src/encoded/commands/deploy.py
caseylitton: this also was preexisting.  It just moved lines.

src/encoded/commands/deploy.py
caseylitton: once day.

PR #2627, opened 2019-03-02
ENCD-4217 Rendering error series objects

emro:
Series objects with life_stage value not present caused a rendering error.
Now tables will insert a null value for those cases.
This ticket also fixes another related error uncovered during debugging: headers were displayed for tables that did not exist (due to related datasets not being viewable).

### DIFF
--- a/src/encoded/tests/test_batch_download.py
+++ b/src/encoded/tests/test_batch_download.py
@@ -73,7 +73,7 @@ def test_batch_download_report_download(testapp, workbook):
     #     b'', b'', b'', b'', b'', b'', b'', b'', b'',
     #     b'', b'', b'', b'', b'', b'', b''
     # ]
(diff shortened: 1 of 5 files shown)


src/encoded/tests/test_batch_download.py
```diff
@@ -73,7 +73,7 @@ def test_batch_download_report_download(testapp, workbook):
     #     b'', b'', b'', b'', b'', b'', b'', b'', b'',
     #     b'', b'', b'', b'', b'', b'', b''
     # ]
-    assert len(lines) == 48
+    assert len(lines) == 49
```
caseylitton: This is being updated in 
https://github.com/ENCODE-DCC/encoded/pull/2623/files

One will need to rebase on the other depending on which is first.

PR #2628, opened 2019-03-02
ENCD-4467 Numerical facets sorted by value not frequency

emro:
Numerical facets are sorted in ascending order.
Date facets are sorted with most recent at the top.
Ranges of numbers are not sorted.

### DIFF
--- a/src/encoded/static/components/search.js
+++ b/src/encoded/static/components/search.js
@@ -754,6 +754,22 @@ class Facet extends React.Component {
             // The term exists, but without a key, so don't allow its display.'
             return false;
         });
+
+        // Sort numerical terms by value not by frequency
+        // This should ultimately be accomplished in the back end, but the front end fix is much simpler so we are starting with that
+        if (terms.length > 0) {
+            // For straightforward numerical facets, just sort by value
+            // We have to check the full list for now (until schema change) because some lists contain both numerical and string terms ('Encyclopedia version' under Annotations) and we do not want to sort those by value
+            const numericalTest = (a, b) => !isNaN(a) && !isNaN(b);
+            const isNumerical = terms.map(term => term.key).reduce(numericalTest);
+            if (isNumerical === true) {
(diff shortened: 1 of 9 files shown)


src/encoded/static/components/search.js
```diff
@@ -754,6 +754,22 @@ class Facet extends React.Component {
             // The term exists, but without a key, so don't allow its display.'
             return false;
         });
+
+        // Sort numerical terms by value not by frequency
+        // This should ultimately be accomplished in the back end, but the front end fix is much simpler so we are starting with that
+        if (terms.length > 0) {
+            // For straightforward numerical facets, just sort by value
+            // We have to check the full list for now (until schema change) because some lists contain both numerical and string terms ('Encyclopedia version' under Annotations) and we do not want to sort those by value
+            const numericalTest = (a, b) => !isNaN(a) && !isNaN(b);
```
forresttanaka: I think you can replace these two lines with `terms.every` unless this is doing something else I don’t see. And in the `every` callback just return whether a single `term.key` value is not NaN. As much as I love `reduce`, it was kind of hard to get my mind around this map/reduce, where `every` would seem more obvious what it’s checking.

src/encoded/static/components/search.js
```diff
@@ -754,6 +754,22 @@ class Facet extends React.Component {
...
+            const isNumerical = terms.map(term => term.key).reduce(numericalTest);
+            if (isNumerical === true) {
```
forresttanaka: Don’t really need to check against `true`. The ESLint rule AirBnB uses is: http://airbnb.io/javascript/#comparison--shortcuts.

emro: I originally had the test written without '=== true' but I was finding that isNumerical was returning the term name for facets with only one term name. For instance, "Library depleted in" was returning "rRNA" for isNumerical instead of true or false. Your suggestion of using "every" fixes this problem in addition to being a clearer solution.

PR #2630, opened 2019-03-04
ENCD-4407 Update column facets BK to BR - jen

jlzamanian:
Column and facet titles for 6 json objects were updated to be consistent with other objects, the path, and to be more descriptive.


caseylitton:
closed as it was merged with 4407 master

PR #2632, opened 2019-03-05
SNO-80 Typeahead for facets

emro:
- New facet typeahead component allows users to search for a facet term from a long list of terms by typing it into a searchbar
- Typeahead component is implemented for 5 facets: "Biosample name", Assay", "Target of assay", "Organ" and "Cell"
- Typeahead facets have "long" type option to capture full list of possible terms for typeahead facets with >200 terms. This is implemented for 2 typeahead facets: "Target of assay" and "Biosample name"
- Both non-typeahead and typeahead facets now display selected filters
- Both non-typeahead and typeahead facets now have a maximum height and allow scrolling, with shading at the top and bottom of the term lists to indicate overflow
- Horizontal facets in matrix and summary views are now vertical facets
- Ordering of facets is changed slightly, to be: Assay category, Assay, Experiment status, Project, RFA, Genome assembly, Target category, Target of assay, Organism, Biosample type, Biosample name, Organ, Cell, Life stage, Biosample treatment, etc.


src/encoded/static/components/search.js
```diff
@@ -711,30 +711,56 @@ TypeTerm.propTypes = {
...
+        let terms = this.props.facet.terms.filter((term) => {
+            if (term.doc_count > 0) {
+                const termKey = sanitizedString(term.key);
+                if (termKey.match(filterVal)) {
+                    return term;
+                }
+                return false;
+            }
+            return false;
+        });
```
forresttanaka: Since this works (and since I saw it before and didn’t say anything) I would consider this feedback optional. `handleSearch` is called with every keystroke and what it essentially does is see if event.target.value is in the terms. `this.props.facet.terms` doesn’t change unless the page re-renders, so I’m thinking maybe the filtered, sanitized terms should be cached in the constructor, and then handleSearch just needs to see if the sanitized `event.target.value` is in this cache or not.

forresttanaka:
Discussed my suggested change with @emro, and I think my suggested change carries a lot of risk, and it’s up to her if she wants to pursue that or do that later. I’ll approve to clear this out for now.

emro:
I had a double declaration of a style for horizontal facets, so I deleted one.

PR #2633, opened 2019-03-05
ENCD-4305 Turn Superseded by into links on exp file pages

forresttanaka:
This ticket reformats the “Superseded by” line to show more obviously, and to link to their corresponding experiment pages. This affects experiment and file pages.

While testing this, I found some file test data I had made in the past used a non-existent property of the file object (`superseded_by`). I also removed those to prevent confusion, as their values were empty arrays anyway.


src/encoded/static/components/typeutils.js
```diff
@@ -296,23 +296,32 @@ export function fileStatusList(session, sessionProperties) {
...
-
     // Make array of supersedes accessions
     let supersedes = [];
     if (context.supersedes && context.supersedes.length) {
         supersedes = context.supersedes.map(supersede => globals.atIdToAccession(supersede));
     }
 
-    if (supersededBys.length > 0 || supersedes.length > 0) {
+    let supersededByOutput = null;
+    if (context.superseded_by && context.superseded_by.length > 0) {
```
zoldello: This is minor and can be ignored if desired. Lines 306 and 320 repeat the statement "context.superseded_by && context.superseded_by.length > 0". That can be refactored into a variable.

src/encoded/static/components/typeutils.js
```diff
@@ -296,23 +296,32 @@ export function fileStatusList(session, sessionProperties) {
...
+                    <span key={supersededBy}>
+                        {index > 0 ? <span>, </span> : null}
+                        <a href={supersededBy}>{globals.atIdToAccession(supersededBy)}</a>
+                    </span>
+                ))}
+            </h4>
+        );
+    }
+
+    if ((context.superseded_by && context.superseded_by.length > 0) || supersedes.length > 0) {
```
zoldello: This is minor and can be ignored if desired. If you like terse code, you can place this line under the previous if-statement with the if condition of- (supersedes.length > 0)

src/encoded/static/components/typeutils.js
```diff
@@ -296,23 +296,32 @@ export function fileStatusList(session, sessionProperties) {
...
-    }
-
     // Make array of supersedes accessions
     let supersedes = [];
     if (context.supersedes && context.supersedes.length) {
         supersedes = context.supersedes.map(supersede => globals.atIdToAccession(supersede));
     }
 
-    if (supersededBys.length > 0 || supersedes.length > 0) {
+    let supersededByOutput = null;
```
zoldello: This is fine and me just thinking and typing up philosophy that might be wrong. I believe "undefined" means the variable has not been initialized and has never been used and null means it is initialized to nothing by some operation. The variable seems to fall under the former rather than the later. However, I have always seen people set the initial value to null

forresttanaka: I think it’s closer to the former because `supersededByOutput` gets assigned a React function if needed, and our standard has been to output null when there’s nothing to display. The React team does consider both null and undefined for outputting nothing https://reactjs.org/docs/jsx-in-depth.html#booleans-null-and-undefined-are-ignored

src/encoded/static/components/typeutils.js
forresttanaka: I agree with this, and have updated the code.

PR #2642, opened 2019-03-12
ENCD-4518 Fix javascript error for sorting dates

emro:
Fix JavaScript error on Safari and Firefox
Evidently using "new Date()" causes non-deterministic behavior and using moment.js prevents this behavior

### DIFF
--- a/src/encoded/static/components/search.js
+++ b/src/encoded/static/components/search.js
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import queryString from 'query-string';
 import _ from 'underscore';
+import moment from 'moment';
 import url from 'url';
 import { svgIcon } from '../libs/svg-icons';
 import { Modal, ModalHeader, ModalBody, ModalFooter } from '../libs/bootstrap/modal';
@@ -765,7 +766,7 @@ class Facet extends React.Component {
         // Make a list of terms for this facet that should appear, by filtering out terms that
         // shouldn't. Any terms with a zero doc_count get filtered out, unless the term appears in
         // the search result filter list.
-        const terms = facet.terms.filter((term) => {
+        const unsortedTerms = facet.terms.filter((term) => {
             if (term.key) {
                 // See if the facet term also exists in the search result filters (i.e. the term
                 // exists in the URL query string).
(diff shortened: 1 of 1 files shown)


src/encoded/static/components/search.js
```diff
@@ -786,11 +787,18 @@ class Facet extends React.Component {
...
+            terms = _.sortBy(unsortedTerms, obj => moment(obj.key, 'YYYY-MM-DD').utc());
+        } else if (field.match('month')) {
+            terms = _.sortBy(unsortedTerms, obj => moment(obj.key, 'MMMM, YYYY').utc());
+        } else if (field.match('year')) {
+            terms = _.sortBy(unsortedTerms, obj => moment(obj.key, 'YYYY').utc());
         // For straightforward numerical facets, just sort by value
-        } else if (terms.every(numericalTest)) {
-            terms.sort((a, b) => (a.key - b.key));
+        } else if (unsortedTerms.every(numericalTest)) {
+            terms = _.sortBy(unsortedTerms, obj => +obj.key);
```
forresttanaka: What’s the “+” in the iteratee for?

src/encoded/static/components/search.js
```diff
@@ -786,11 +787,18 @@ class Facet extends React.Component {
         // We have to check the full list for now (until schema change) because some lists contain both numerical and string terms ('Encyclopedia version' under Annotations) and we do not want to sort those by value
         const numericalTest = a => !isNaN(a.key);
         // For date facets, sort by date
-        if (field.match('date') || field.match('month') || field.match('year')) {
-            terms.sort((a, b) => (new Date(b.key) - new Date(a.key)));
+        let terms = [];
+        if (field.match('date')) {
+            terms = _.sortBy(unsortedTerms, obj => moment(obj.key, 'YYYY-MM-DD').utc());
+        } else if (field.match('month')) {
+            terms = _.sortBy(unsortedTerms, obj => moment(obj.key, 'MMMM, YYYY').utc());
```
forresttanaka: Looks like this format of `moment` returns an object instead of a single value. Maybe add `.toISOString()` to the end to return an ISO-formatted string which is sortable. I _think_ that would work for `sortBy`.

forresttanaka: This might be clearer than the momentjs docs. https://stackoverflow.com/questions/44813743/returning-a-value-using-moment-js#answer-44813773

src/encoded/static/components/search.js
emro: That was just to confirm that obj.key was a number not a string.

emro: (It is a number so it's not necessary, it's just a habit because sometimes JS interprets numbers as strings and then there can be very unexpected behavior.)

forresttanaka: That sounds vaguely familiar now, and it’s actually ECMA sanctioned: http://www.ecma-international.org/ecma-262/5.1/#sec-11.4.7

src/encoded/static/components/search.js
emro: Somehow, it is sorting the objects, but sorting single values is much clearer.

src/encoded/static/components/search.js
emro: Oh whoops, I didn't see this till now. Should I add it back?

forresttanaka: Oh, it’s up to you. I just didn’t know about it.

PR #2651, opened 2019-03-16
Isolate session / session_properties dependencies to App

lrowe:
(Draft because I just noticed a bug clicking 'Share Cart' with an in memory cart, I need to investigate some more to see where I introduced the bug.)

This PR moves all of the dependencies on session or session_properties into explicit context values:

* loggedIn
* adminUser
* userDisplayName
* userURI

I'm in the process of standing up a local encoded instance with a different authentication configuration. This change helps isolate my changes to the App component.


forresttanaka:
Couple general questions to start with: What prompted this change? When do you plan on putting this in? I have a couple branches set for PR soon that would be affected by this change.

lrowe:
@forresttanaka I don't plan to merge until it is convenient for you. My prompt was making things work with a different authentication plugin (my pyramid_iap package which is essentially JWT auth for Google Cloud's Identity Aware Proxy) which means the auth.userid value is not present in the session cookie so various parts of the UI were disabled. Almost all usage of session is checking loggedIn status.

Switching the code to use an explicit `loggedIn` context value seemed to simplify things so thought I'd do the same for the session_properties usage. This would potentially make it possible to render logged in pages server-side to avoid the extra requests to session_properties / session and potentially simplify UI code that currently has to be robust to not yet known logged in values.

If you think this seems like a good idea, it may be worth merging in stages, starting with everything up to 73b2af1 (Add loggedIn, adminUser, userDisplayName, userURI context values), whatever parts of da4cd54 (Use context loggedIn instead of session) don't conflict with your ongoing work and 
103fa72 (Avoid session usage in FetchedData) and then merging the rest after in flight changes are in. In the meantime new work could start depending on the new explicit context values.

forresttanaka:
I see. I like not having to use context variables so much for this, so that’s really cool.

lrowe:
The new values are still context variables, but they're explicit so the business logic about whether a user is logged in or an admin is isolated to a single place instead of spread throughout the codebase. I do think it's worth giving the current context variables another look to decide if they really need to be context variables (portal could probably be passed to Navigation as props for instance.)

lrowe:
Fixed the bug clicking share cart, it now correctly creates a cart on page load / login.

hitz:
@lrowe @cricketsloan @forresttanaka what is the status of this?  Should I make a ticket?

lrowe:
I think this is still useful and is the basis for server rendering for logged in pages which I now have working. I'll work on rebasing.

PR #2654, opened 2019-03-18
ENCD-4489 add imputation annotation_type

zmyers23:
Added annotation_type enum 'imputation' and made annotation_type required.

### DIFF
--- a/src/encoded/tests/datafixtures.py
+++ b/src/encoded/tests/datafixtures.py
@@ -549,7 +549,8 @@ def publication_data(testapp, lab, award):
 def annotation_dataset(testapp, lab, award):
     item = {
         'award': award['@id'],
-        'lab': lab['@id']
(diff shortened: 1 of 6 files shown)


src/encoded/tests/datafixtures.py
```diff
@@ -549,7 +549,8 @@ def publication_data(testapp, lab, award):
 def annotation_dataset(testapp, lab, award):
     item = {
         'award': award['@id'],
-        'lab': lab['@id']
+        'lab': lab['@id'],
+        'annotation_type': 'imputation'
```
caseylitton: Don't change this unless you have to fix the branch in some other way,

It's more standard to add a trailing comma to lists like this.
```
   item = {
        'award': award['@id'],
        'lab': lab['@id'],
        'annotation_type': 'imputation',
}
```

PR #2660, opened 2019-03-18
ENCD-3422-datasets-table-publication

zoldello:


src/encoded/static/components/publication.js
```diff
@@ -7,7 +7,54 @@ import { DbxrefList } from './dbxref';
...
+    lab: {
+        getValue: dataset => (dataset.lab || {}).title,
+        title: 'Lab',
+    },
+    biosample_summary: {
+        getValue: (dataset) => {
+            if (typeof (dataset.biosample_summary) === 'string') {
+                return dataset.biosample_summary;
+            }
+            return (dataset.biosample_summary || []).map(b => b.term_name).join(', ');
```
forresttanaka: Just wanted to confirm `experiment.biosample_summary` can be an array of objects? I skimmed over experiment.py and don’t see where that happens, and all other uses of `experiment.biosample_summary` in front-end code assumes it’s a string. If it can be an array, it seems like those have to be changed too.

src/encoded/static/components/publication.js
```diff
@@ -26,7 +73,11 @@ const PublicationComponent = (props, reactContext) => {
     ];
 
     const crumbsReleased = (context.status === 'released');
-
+    const accessions = !context.datasets ? '' : context.datasets.reduce((dataset1, dataset2) => `${dataset1}&accession=${dataset2.accession}`, '');
```
forresttanaka: Maybe just change to `context.datasets` without the `!` so you can reverse the conditions? I had to read over this twice to realize it wasn’t a mistake.

src/encoded/static/components/publication.js
```diff
@@ -26,7 +73,11 @@ const PublicationComponent = (props, reactContext) => {
...
+    const fields = ['accession', 'target', '@type', 'target', 'biosample_summary', 'lab', 'assay_term_name', 'assay', 'description', 'status'].reduce((field1, field2) => `${field1}&field=${field2}`, '');
```
forresttanaka: Could this be done in the module name space? The result wouldn’t ever change until the code is changed, so putting it in the module name space would let this happen only at page reload instead of every time a publication page is rendered.

src/encoded/static/components/publication.js
```diff
@@ -7,7 +7,54 @@ import { DbxrefList } from './dbxref';
...
+import { FetchedData, Param } from './fetched';
+
+
+const datasetsColumns = {
+    accession: {
+        display: dataset => <a href={dataset['@id']}>{dataset.accession}</a>,
+        title: 'Accession',
+    },
+    type: {
+        getValue: datasets => (datasets['@type'] || [{}])[0], // only the first matters
```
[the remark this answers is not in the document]
zoldello: @forresttanaka I forgot the sample case, but I did get a case where it was without @type. I used [{}][0] to try to be consistent with the dataset['@type'][0], which gives an object.

src/encoded/static/components/publication.js
```diff
@@ -7,7 +7,54 @@ import { DbxrefList } from './dbxref';
...
+        title: 'Type',
+    },
+    target: {
+        getValue: dataset => (dataset.target || {}).label,
```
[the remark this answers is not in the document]
zoldello: @forresttanaka The way I did it is easier for me to understand. However, I can use the way you did if your way is easier for everyone else.

src/encoded/static/components/publication.js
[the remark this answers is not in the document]
forresttanaka: That gives an object? I see as an example:

```
    "@type": [
        "ReferenceEpigenome",
        "Series",
        "Dataset",
        "Item"
    ],
```

Seems like in this case, dataset['@type'][0] would give the string "ReferenceEpigenome" wouldn’t it?

src/encoded/static/components/publication.js
zoldello: @forresttanaka I will reverse it.

src/encoded/static/components/publication.js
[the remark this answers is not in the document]
forresttanaka: If you find your way easier to understand, I would leave it.

src/encoded/static/components/publication.js
zoldello: Done

src/encoded/static/components/publication.js
```diff
@@ -38,16 +38,11 @@ const datasetsColumns = {
...
         title: 'Lab',
     },
     biosample_summary: {
-        getValue: (dataset) => {
-            if (typeof (dataset.biosample_summary) === 'string') {
-                return dataset.biosample_summary;
-            }
-            return (dataset.biosample_summary || []).map(b => b.term_name).join(', ');
-        },
+        getValue: dataset => dataset.biosample_summary,
```
forresttanaka: Since you’re getting the property with the same name as they key (`biosample_summary`) you don’t need `getValue`. Just `title` is enough because SortTable will figure it out.

src/encoded/static/components/publication.js
```diff
@@ -90,7 +86,7 @@ const PublicationComponent = (props, reactContext) => {
                 <Citation {...props} />
             </div>
 
-            {context.abstract || context.data_used || (context.datasets && context.datasets.length) || (context.identifiers && context.identifiers.length) ?
+            {context.abstract || context.data_used || (context.identifiers && context.identifiers.length) ?
```
forresttanaka: Oh, could you change that to `context.identifiers.length > 0`? A lot of times ESLint plugin misses that, but our ESLint config likes integer comparisons to be against 0 instead of truthy.

src/encoded/static/components/publication.js
```diff
@@ -7,7 +7,51 @@ import { DbxrefList } from './dbxref';
...
+    accession: {
+        display: dataset => <a href={dataset['@id']}>{dataset.accession}</a>,
+        title: 'Accession',
+    },
+    type: {
+        getValue: datasets => (datasets['@type'] || [''])[0], // only the first matters
+        title: 'Type',
+    },
+    target: {
+        getValue: dataset => (dataset.target ? dataset.target.label : ''),
```
forresttanaka: You didn’t have to change this if you preferred your way, but OK. My personal preferences aren’t gospel, just to be clear. I just found it unusual.

src/encoded/static/components/publication.js
zoldello: Done

src/encoded/static/components/publication.js
zoldello: Done

src/encoded/static/components/publication.js
```diff
@@ -26,7 +69,10 @@ const PublicationComponent = (props, reactContext) => {
     ];
 
     const crumbsReleased = (context.status === 'released');
-
+    const accessions = context.datasets ? context.datasets.reduce((dataset1, dataset2) => `${dataset1}&accession=${dataset2.accession}`, '') : '';
+    const datasetsUrl = context.datasets && context.datasets.length ?
```
forresttanaka: One more for ESLint’s sake: compare .length against 0.

PR #2661, opened 2019-03-18
ENCD-4507 Handle long cart operations

forresttanaka:
* Changed the name of the `cartPatch` function to `cartUpdate` to more accurately reflect its function in case we add a real PATCH function later. Was hoping to do that in this branch but tests didn’t work. cartPatch’s functionality hasn’t changed — just its name.
* I had used a local in-progress state for certain cart-manager operations because I was afraid of having multiple simultaneous use of the inProgress flag belonging to the cart Redux store. But I’m not worried about that anymore since I had time to look over it and make sure I avoid it. All inProgess use now uses that flag in the cart Redux store, so it gets shared across the whole site.
* Many of the changes simply add an inProgress check to controls/buttons that need to be disabled while an operation is in progress.
* The change to browserfeat lets us detect that we’re running on a browser that doesn’t support the SVG animation that spins the spinner. That explains why svg-icons.js has two spinners defined: `spinnerStatic` for MS Edge and IE11, and `spinner` for browsers with good behavior.
* Shared cart pages (cartType === OBJECT) and the cart manager show static objects, but these need to update if a cart operation completes while the user has them up in their browser. I now check for cart operations to finish, then load the updated context for the page through a GET request. My earlier solution was to navigate to the page you’re viewing, but it sometimes caused a GET request and other times caused a reload which caused other cart operations to happen which caused another navigation to the same page causing another cart operation and so on.
* Detecting cart operations completing removed the need for the `cartRetrieve` function to use `datastore=database`, because we now know to retrieve the cart object again after we know it has updated.


hitz:
@emro I don't think the tooltip message is particularly critical; to be revised if we get questions about "why can't I share this cart"

PR #2662, opened 2019-03-18
ENCD-4524 update audit matrix

emro:
We removed horizontal facets from the matrix page as part of the typeahead facet changes but forgot to also remove them from the audit matrix.
One other small change: the matrix pages had an empty "search-header-control" space which is now not created when not needed.

### DIFF
--- a/src/encoded/static/components/search.js
+++ b/src/encoded/static/components/search.js
@@ -1063,7 +1063,7 @@ export class FacetList extends React.Component {
         return (
             <div className={`box facets${addClasses ? ` ${addClasses}` : ''}`}>
                 <div className={`orientation${this.props.orientation === 'horizontal' ? ' horizontal' : ''}`}>
-                    {(context && clearButton) ?
+                    {(context || clearButton) ?
                         <div className="search-header-control">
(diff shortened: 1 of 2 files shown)


src/encoded/static/components/search.js
```diff
@@ -1063,14 +1063,16 @@ export class FacetList extends React.Component {
...
                 <div className={`orientation${this.props.orientation === 'horizontal' ? ' horizontal' : ''}`}>
-                    <div className="search-header-control">
-                        {context ? <DocTypeTitle searchResults={context} wrapper={children => <h1>{children} {docTypeTitleSuffix}</h1>} /> : null}
-                        {clearButton ?
-                            <div className="clear-filters-control">
-                                <a href={context.clear_filters}>Clear Filters <i className="icon icon-times-circle" /></a>
-                            </div>
-                        : null}
-                    </div>
+                    {(context && clearButton) ?
```
forresttanaka: Should this be `&&` or `||`? I’m not sure what the logic is supposed to be, but if it’s supposed to be `&&` then you can remove both conditionals inside the "search-header-control" div since you’ll know they’re both true.

emro: Oh yes, good catch. Should be ||. Thank you!

PR #2664, opened 2019-03-19
ENCD-4345-matrix-refactor

zoldello:


src/encoded/tests/test_search.py
```diff
@@ -238,8 +238,8 @@ def test_set_filters():
             }
         }
     }
-    result = {'filters': []}
-    used_filters = set_filters(request, query, result)
+
```
hitz: Do we actually need to repeat all these snovault/search tests in encoded?  It seems like we should just append the ones that are used in encoded specific search sub modules.

caseylitton:
Closing because the plan has changed.  Per last encd dev meeting.

PR #2665, opened 2019-03-20
ENCD-4414 Allow users to add exp to carts from series

forresttanaka:
Most changes are in cart/add_multiple.js. The existing `CartAddAll` component is for adding all search results to the cart, and got renamed to `CartAddAllSearch` to distinguish it from the new `CartAddAllElements` component which has the much simpler job of adding an array of dataset objects to the cart. It gets called from the “Add all items to cart” button that appears in any type of Series page.

### DIFF
--- a/src/encoded/static/components/cart/add_multiple.js
+++ b/src/encoded/static/components/cart/add_multiple.js
@@ -146,10 +146,10 @@ CartAddAllSearchComponent.mapDispatchToProps = (dispatch, ownProps) => ({
     setInProgress: enable => dispatch(cartOperationInProgress(enable)),
 });
 
-const CartAddAllInternal = connect(CartAddAllSearchComponent.mapStateToProps, CartAddAllSearchComponent.mapDispatchToProps)(CartAddAllSearchComponent);
+const CartAddAllSearchInternal = connect(CartAddAllSearchComponent.mapStateToProps, CartAddAllSearchComponent.mapDispatchToProps)(CartAddAllSearchComponent);
 
 export const CartAddAllSearch = (props, reactContext) => (
-    <CartAddAllInternal searchResults={props.searchResults} session={reactContext.session} fetch={reactContext.fetch} />
+    <CartAddAllSearchInternal searchResults={props.searchResults} session={reactContext.session} fetch={reactContext.fetch} />
 );
 
 CartAddAllSearch.propTypes = {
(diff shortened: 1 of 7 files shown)


src/encoded/static/components/cart/add_multiple.js
```diff
@@ -128,39 +128,126 @@ CartAddAllComponent.propTypes = {
...
     session: ownProps.session,
 });
-const mapDispatchToProps = (dispatch, ownProps) => ({
+CartAddAllSearchComponent.mapDispatchToProps = (dispatch, ownProps) => ({
     addAllResults: elementsForCart => dispatch(addMultipleToCartAndSave(elementsForCart, !!(ownProps.session && ownProps.session['auth.userid']), ownProps.fetch)),
     setInProgress: enable => dispatch(cartOperationInProgress(enable)),
 });
 
-const CartAddAllInternal = connect(mapStateToProps, mapDispatchToProps)(CartAddAllComponent);
+const CartAddAllInternal = connect(CartAddAllSearchComponent.mapStateToProps, CartAddAllSearchComponent.mapDispatchToProps)(CartAddAllSearchComponent);
```
emro: Can we change "CartAddAllInternal" to be "CartAddAllSearchInternal"? Just for consistency.
We currently have:
- CartAddAllSearchComponent
- CartAddAllInternal
- CartAddAllSearch
for adding all elements, and then:
- CartAddAllElementsComponent
- CartAddAllElementsInternal
- CartAddAllElements
for adding elements from an array of datasets.

src/encoded/static/components/dataset.js
```diff
@@ -1279,6 +1303,18 @@ export class SeriesComponent extends React.Component {
...
         const experimentList = context.related_datasets.filter(dataset => dataset.status !== 'revoked' && dataset.status !== 'replaced' && dataset.status !== 'deleted');
 
+        // If we display a table of related experiments, have to render the control to add all of
+        // them to the current cart.
+        let addAllToCartControl;
+        if (experimentList.length > 0) {
+            addAllToCartControl = (
+                <div className="experiment-table__header">
+                    <h4 className="experiment-table__title">{`Experiments in ${seriesTitle} ${context.accession}`}</h4>
+                    <CartAddAllElements elements={experimentList} css="experiment-table__control" />
```
emro: I wonder if you could add styles to the <CartAddAllElements> component without passing class names to the component. Maybe you could do something like .experiment-table__header { button { flex: 1 1 auto } }. I'm not sure if this will work for every case, or even if it's simpler, this is just a suggestion/comment.

PR #2669, opened 2019-03-20
ENCD-4530 Fix duplicate alternate accessions

forresttanaka:
* The biggest change is in the `Supersede` component of typeutils.js. I cleaned up the code to all be more consistent and easy to maintain. I also renamed `Supersede` to `ReplacementAccessions` to more accurately reflect its current role.
* I also modified a restricted-file display in the header of file pages so that it stops using CSS styles I removed in this ticket’s predecessor.

### DIFF
--- a/src/encoded/static/components/typeutils.js
+++ b/src/encoded/static/components/typeutils.js
@@ -295,44 +295,39 @@ export function fileStatusList(session, sessionProperties) {
 }
 
 
-// Display supersedes/superseded_by lists.
-export const Supersede = ({ context }) => {
+/**
+ *  Display supersedes/superseded_by/alternate_accessions lists.
+ */
+export const ReplacementAccessions = ({ context }) => {
     const alternateAccessions = context.alternate_accessions || [];
+    const supersededByAtIds = context.superseded_by || [];
+    const supersedes = (context.supersedes && context.supersedes.map(supersedesAtId => globals.atIdToAccession(supersedesAtId))) || [];
 
-    // Make array of supersedes accessions
-    let supersedes = [];
-    if (context.supersedes && context.supersedes.length) {
-        supersedes = context.supersedes.map(supersede => globals.atIdToAccession(supersede));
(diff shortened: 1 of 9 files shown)


src/encoded/static/components/typeutils.js
```diff
@@ -295,44 +295,39 @@ export function fileStatusList(session, sessionProperties) {
 }
 
 
-// Display supersedes/superseded_by lists.
-export const Supersede = ({ context }) => {
+/**
+ *  Display supersedes/superseded_by/alternate_accessions lists.
+ */
+export const ReplacementAccessions = ({ context }) => {
```
zoldello: Not a big deal
- Need JSDoc?

forresttanaka: I haven’t been putting that on React component comments because we already document props in propTypes. Not sure if JSDoc has been updated for React components.

zoldello: Not a big deal
Maybe it is just me, but I see nothing really being replaced.

forresttanaka: Accessions are being replaced by other accessions — alternates, supersedes.

src/encoded/static/components/typeutils.js
```diff
@@ -295,44 +295,39 @@ export function fileStatusList(session, sessionProperties) {
...
+                        ))}
+                    </h4>
+                : null}
+                {supersedes.length > 0 ?
+                    <h4 className="replacement-accessions__supersedes">Supersedes {supersedes.join(', ')}</h4>
+                : null}
             </div>
         );
     }
     return null;
```
zoldello: Not a big deal

This is  just my personal taste. It makes sense to me as empty components needs to be set to null. But not everyone will get at first why you need to return null rather than just do nothing (which will return undefined.) Maybe check for opposite of if-statement and return null if that is true. Otherwise, proceed as normal..?

PR #2670, opened 2019-03-20
ENCD-4170-exclude-restricted-files

zoldello:


src/encoded/batch_download.py
```diff
@@ -84,6 +84,7 @@ def includeme(config):
     ('Platform', ['files.platform.title']),
```
zoldello: @caseylitton  What code change is still required?

caseylitton: I suggest slacking people when you need to remind them about code review

src/encoded/batch_download.py
```diff
@@ -523,6 +531,8 @@ def restricted_files_present(exp_file):
         return True
     return False
 
+def is_no_file_available(exp_file):
```
caseylitton: Two spaces before and after top level functions.

src/encoded/batch_download.py
```diff
@@ -130,9 +131,12 @@ def includeme(config):
     ('Derived from', ['files.derived_from']),
     ('S3 URL', ['files.cloud_metadata']),
     ('Size', ['files.file_size']),
+    ('No File Available', ['file.no_file_available']),
     ('Restricted', ['files.restricted'])
 ])
 
+_excluded_columns = ('Restricted', 'No File Available')
+
```
caseylitton: top space between top level vars and functions.

src/encoded/batch_download.py
zoldello: @caseylitton Acknowledged

PR #2671, opened 2019-03-21
ENCD-4517 Add UI for GM introduced elements

forresttanaka:
This branch branches off ENCD-4464-add-GM-category-enums which is QA Completed as I type this. Might need to clear up conflicts after that gets merged, but I’m not sure.

With this change, I now check for GM modification site property names in multiple places, so I moved existing code to a function and module-level variable so they can be shared.

### DIFF
--- a/src/encoded/static/components/genetic_modification.js
+++ b/src/encoded/static/components/genetic_modification.js
@@ -53,6 +53,22 @@ IntroducedTags.propTypes = {
 };
 
 
+// All possible genetic modification object modification site property names. This might need to
+// change if the genetic_modification.json schema changes.
+const modificationSitePropNames = ['modified_site_by_target_id', 'modified_site_by_coordinates', 'modified_site_by_sequence', 'modified_site_nonspecific'];
+
+
+/**
+ * Determine if a genetic modification object includes any modification site properties.
+ * @param {object} geneticModification GM object
+ * @return {boolean} True if `geneticModification` includes any modification site properties.
+ */
+const hasModificationSiteProps = (geneticModification) => {
+    const gmKeys = Object.keys(geneticModification);
+    return modificationSitePropNames.some(siteType => gmKeys.indexOf(siteType) !== -1);
+};
+
+
 // Render the modification site items into a definition list.
 const ModificationSiteItems = (props) => {
     const { geneticModification, itemClass } = props;
@@ -101,7 +117,7 @@ const ModificationSiteItems = (props) => {
         ),
     };
 
-    const elements = ['modified_site_by_target_id', 'modified_site_by_coordinates', 'modified_site_by_sequence', 'modified_site_nonspecific'].map((siteType) => {
+    const elements = modificationSitePropNames.map((siteType) => {
         if (geneticModification[siteType]) {
             return <div key={siteType}>{renderers[siteType](geneticModification)}</div>;
         }
@@ -129,15 +145,18 @@ ModificationSiteItems.defaultProps = {
 // render into the GM summary panel as its own section;
 const ModificationSite = (props) => {
     const { geneticModification } = props;
-    const itemClass = globals.itemClass(geneticModification, 'view-detail key-value');
 
-    return (
-        <div>
-            <hr />
-            <h4>Modification site</h4>
-            <ModificationSiteItems geneticModification={geneticModification} itemClass={itemClass} />
-        </div>
-    );
+    if (hasModificationSiteProps(geneticModification)) {
+        const itemClass = globals.itemClass(geneticModification, 'view-detail key-value');
+        return (
+            <div>
+                <hr />
+                <h4>Modification site</h4>
+                <ModificationSiteItems geneticModification={geneticModification} itemClass={itemClass} />
+            </div>
+        );
+    }
+    return null;
 };
 
 ModificationSite.propTypes = {
@@ -561,6 +580,13 @@ class GeneticModificationComponent extends React.Component {
                                         </div>
                                     : null}
 
+                                    {context.introduced_elements ?
+                                        <div data-test="introduced-elements">
+                                            <dt>Introduced elements</dt>
+                                            <dd>{context.introduced_elements}</dd>
+                                        </div>
+                                    : null}
+
                                     {context.zygosity ?
                                         <div data-test="zygosity">
                                             <dt>Zygosity</dt>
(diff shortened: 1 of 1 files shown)


src/encoded/static/components/genetic_modification.js
```diff
@@ -129,15 +145,18 @@ ModificationSiteItems.defaultProps = {
...
+        const itemClass = globals.itemClass(geneticModification, 'view-detail key-value');
+        return (
+            <div>
+                <hr />
+                <h4>Modification site</h4>
+                <ModificationSiteItems geneticModification={geneticModification} itemClass={itemClass} />
+            </div>
+        );
+    }
+    return null;
```
zoldello: Not big deals
1. You may need to comment on why you are returning null. Not everyone may realize you need null with components instead of undefined (which functions/methods return by default).

2. My understanding is that code should check if it can do what it needs to do early. If it detects it cannot, then it does not try to proceed. An alternate approach would be to check for (!hasModificationSiteProps(geneticModification)). If this true, return null. If not true, the if statement will be skipped and the code will proceed as normal.

src/encoded/static/components/genetic_modification.js
```diff
@@ -693,7 +719,7 @@ GeneticModificationSummary.columns = {
     method: { title: 'Method' },
     site: {
         title: 'Site',
-        display: item => <ModificationSiteItems geneticModification={item} itemClass={'gm-table-modification-site'} />,
+        display: item => (hasModificationSiteProps(item) ? <ModificationSiteItems geneticModification={item} itemClass={'gm-table-modification-site'} /> : null),
```
zoldello: Not big deal
Am I wrong, or does line have code you repeated in line 149 hasModificationSiteProps(...)? It may be clearer to put hasModificationSiteProps() inside ModificationSiteItems. The component ModificationSiteItems is called and the caller let ModificationSiteItems worry about whether or not it can display

src/encoded/static/components/genetic_modification.js
forresttanaka: For #2, what would the difference be? In your version, has hasModificationSiteProps would be called and if it returns false, the rest of the code is skipped and null is returned. In my version, isn’t the same thing true? None of the code would be executed in my version either.

src/encoded/tests/test_schema_genetic_modification.py
```diff
@@ -328,3 +348,24 @@ def test_tale_replacement_properties(testapp, tale_replacement, source):
     tale_replacement.update({'introduced_sequence': 'TTATCGATCGATTTGAGCATAGAAATGGCCGATTTATATGCCCGA'})
     res = testapp.post_json('/genetic_modification', tale_replacement, expect_errors=True)
     assert res.status_code == 201
+
+
+def test_mpra_properties(testapp, mpra):
```
zoldello: Not a big deal

- I normally see asserts at the end rather at the middle.
- If you need multiple asserts, that may be a sign you need multiple unit test. A unit test supposed to test a single functionality (unit) and not multiple actions.

forresttanaka: This branch branches off ENCD-4464-add-GM-category-enums, so that’s not my code. I only modified genetic_modification.js

src/encoded/schemas/genetic_modification.json
```diff
@@ -92,31 +99,42 @@
...
                         {
-                            "required": ["introduced_tags"],
-                            "properties": {
-                                "category": {
-                                    "enum": ["insertion"]
-                                },
-                                "purpose": {
-                                    "enum": ["tagging"]
+                            "anyOf": [
+                            {
```
zoldello: Need fixing
This is not properly formatted.

forresttanaka: This branch branches off ENCD-4464-add-GM-category-enums, so that’s not my code. I’ll pull in any changes from that branch if that happens.

src/encoded/static/components/genetic_modification.js
forresttanaka: For #1, I think people understand that, and we do that _everywhere_. React also says to do this, so people can just refer to the [React documentation](https://reactjs.org/docs/conditional-rendering.html#preventing-component-from-rendering). We would have to add this same comment to about 90 places in our code.

src/encoded/schemas/genetic_modification.json
zoldello: Got it. I misread the coloring.

forresttanaka:
I’m not sure how to update the PR title. This is for the only ticket that addresses the UI impact of the new GM property.

src/encoded/schemas/genetic_modification.json
jlzamanian: Could you please explain this formatting issue? And the others alluded to? I'm the one who made the changes.

forresttanaka:
Do you mean because this is a branch off ENCD-4464-add-GM-category-enums instead of a branch off dev?

src/encoded/schemas/genetic_modification.json
jlzamanian: OK. I see this formatting issue now. You mean the { in line 108 and the option after it should have an additional level of indentation.

caseylitton:
@forresttanaka  Right, this ticket is composed of two JIRA tickets, so both ticket numbers should be in the title.  Or the feature we intend on dropping should be squashed into one commit with a revert tag.  I suggest doing the title thing in either case because it's more likely the reviewers read the PR title even if they don't check the commit history.  

There are two situations.  

1. Feature-A needs Feature-B but Feature-B is being merged via its own PR.  Then Feature-A pr should have one commit like, '[Revert] ENCD-#### feature-b-msg'.  Then when the code reviewer looks at the commit history they know to only review feature-b changes.  And to make comments on feature-a in the feature-a pr.

The code reviewer,  @zoldello, should actually block the PR by requesting a change for this specifically.  And not approve the PR until the code is corrected after QA.  We can still move throubh QA while not approving the PR in github.  It actually blocks me from accidentally merging until the commits are removed.  Also, JIRA tickets shouldn't be moved to 'merge ready' until all comments have been cleared on github so there is some redundancy.

2. OR Feature-A needs Feature-B and we decided to include both features in one PR.  Then PR-A-B will have two commits, one for each feature.

I don't expect everyone to have known this or how to handle it.  We don't do it very often so the procedure is not documented.

forresttanaka:
I couldn't really follow any of this, so I suggest closing this PR. Once ENCD-4464-add-GM-category-enums gets merged into dev, then I’ll start this over.

caseylitton:
@forresttanaka 

Always: If there are other JIRA tickets in your PR you need to add that ticket number to the PR title.  ENCD-###1-###2

Then depending on how we want to handle the extra code in the PR,

Option 1: If you're making a PR with more than one ticket each feature should be squashed into one commit.  Where each commit has a ticket number.

Option 2: If you intend on dropping the other ticket from your PR before merging then flag that commit message with a revert tag.  [REVERT] ENCD-####D This other feature I need.

PR #2672, opened 2019-03-21
ENCD-4349 restrict gm reagents

p-sud:


src/encoded/upgrade/genetic_modification.py
```diff
@@ -185,4 +185,78 @@ def genetic_modification_5_6(value, system):
...
+                for source, regex in sources.items():
+                    if re.match(regex, identifier) is not None:
+                        matching.append((source_name, source, identifier))
+            if not matching:
+                unmatched_reagents += '{} '.format(json.dumps(reagent))
+            else:
+                if len(matching) >= 2:
+                    # Bring the identifier with matching source to the reagent to the top
+                    matching.sort(key=lambda match: int(match[1] in match[0]), reverse=True)
+                    # matching = [i for i in matching if i[1] in i[0]]
```
caseylitton: extra comment?

src/encoded/upgrade/genetic_modification.py
```diff
@@ -185,4 +185,78 @@ def genetic_modification_5_6(value, system):
...
+        if unmatched_reagents:
+            date = time.strftime('%m/%d/%Y')
+            message = 'On {}, the following reagents were removed due to invalid identifer: '.format(date)
+            if 'notes' in value:
+                new_notes = value['notes'] + ' ' + message + unmatched_reagents
+            else:
+                new_notes = message + unmatched_reagents
+            new_notes = new_notes.strip()
+            value['notes'] = new_notes
+        del value['reagents']
```
caseylitton: This is kinda weird hanging out here all alone.  Do we delete the reagents if there are not any new_reagents?  Maybe we should delete them at the top, I think there is a way to pop from a dict instead of 'reagents = value.get('reagents')'  and then delete.

src/encoded/upgrade/genetic_modification.py
```diff
@@ -185,4 +185,78 @@ def genetic_modification_5_6(value, system):
...
+        if new_reagents:
+            value['reagents'] = new_reagents
```
caseylitton: I wonder if this can be created in the for loop, we're removing reagent from value, so we could just value[reagents].append(new_reagent).  The var 'new_reagents' isn't really used.

src/encoded/upgrade/genetic_modification.py
```diff
@@ -185,4 +185,78 @@ def genetic_modification_5_6(value, system):
...
+                if len(matching) >= 2:
+                    # Bring the identifier with matching source to the reagent to the top
+                    matching.sort(key=lambda match: int(match[1] in match[0]), reverse=True)
+                    # matching = [i for i in matching if i[1] in i[0]]
+                _, source, identifier = matching[0]
+                new_identifier = ('{}:{}'.format(source, identifier))
+                new_reagent['identifier'] = new_identifier
+                new_reagents.append(new_reagent)
+        # Dump the unmatched identifier reagents to the notes
+        if unmatched_reagents:
```
caseylitton: This block of code is different, it creates value['notes'] based on unmatched reagents.  But it sitting in between code that acts on value['reagents'].  But they are not dependent on each other.  I might finish with the value reagents and then create the value notes at the end.

src/encoded/upgrade/genetic_modification.py
```diff
@@ -185,4 +185,78 @@ def genetic_modification_5_6(value, system):
...
+    # https://encodedcc.atlassian.net/browse/ENCD-4349
+    import re
+    import copy
+    import time
+    import json
+
+    conn = system['registry'][CONNECTION]
+
+    sources = {
+        'addgene': r'^\d{5,6}$',
```
caseylitton: Did you look at re.pattern?  I think you can create all these regex experssion in a pattern the then use them directly....
```
if sources[source_from_reagent].match(regex, identifier) is not None:
   etc...
```

src/encoded/upgrade/genetic_modification.py
```diff
@@ -185,4 +185,78 @@ def genetic_modification_5_6(value, system):
...
+            new_reagent = copy.deepcopy(reagent)
+            matching = []
+            source_from_reagent = None
+            for source in sources:
+                if source in source_name:
+                    source_from_reagent = source_name
+                    break
+            # Preferentially add prefix to reagent that matches the reagent source
+            if source_from_reagent is not None:
+                regex = sources[source_from_reagent]
```
caseylitton: I think this block could be added to the above for directly.  Then if the for loop exits without setting source_from_reagent we can 
```
if not source_from_reagent:
    for source, regex in sources.items():
        ....2nd for loop check
```
Also, we don't need to break in the second for loop like we do in the first?  The second could add multiple values.

And do we need to specifically check is not None in all these cases?

'regex' is not a good variable name, pretty generic.  'src_regex' would be better.

src/encoded/upgrade/genetic_modification.py
```diff
@@ -185,4 +185,78 @@ def genetic_modification_5_6(value, system):
...
+                if re.match(regex, identifier) is not None:
+                    matching.append((source_name, source_from_reagent, identifier))
+            else:
+                for source, regex in sources.items():
```
caseylitton: I wonder if we need to for loop for the same dict.  Maybe we do it in one, where we add 'matching.append((source_name, source, identifier))' but if we find 'if source in source_name:' then override matching with 'matching.append((source_name, source_from_reagent, identifier))' and then break.

I don't know, seems like it might have the same outcome.

src/encoded/upgrade/genetic_modification.py
p-sud: In this case, I want to preserve the mapping between source name and the appropriate regular expression. It could be done with pattern, but then I would have to name the groups in the regex to preserve the information (?P<id>), then extract the group names indirectly via Pattern.groupindex. This seems like a more straightforward way to do keep the source information

src/encoded/upgrade/genetic_modification.py
p-sud: 1. & 2. ) The loops definitely could be consolidated. The second loop shouldn't break because it is possible for the identifier to match multiple and we want to catch all of them.
3. ) re.match returns None if there is no match, so checking 'if re.match(...)' should suffice. I'll change this
4. ) I'll change regex to src_regex.

src/encoded/upgrade/genetic_modification.py
p-sud: It's probably a good idea to consolidate it into one for loop over the dict, I'll change it.

src/encoded/upgrade/genetic_modification.py
p-sud: Yeah it's obsolete code from when I was trying to get the upgrade to work. I'll delete it.

src/encoded/upgrade/genetic_modification.py
p-sud: Rearranged

src/encoded/upgrade/genetic_modification.py
p-sud: Yeah we still want to delete them if there aren''t any new ones. .pop is a good idea, I'll add that in.

src/encoded/upgrade/genetic_modification.py
p-sud: In some cases we don't want value to have a "reagents" key at all. To be able to append you'd at least need value[reagents] to be an empty list. I think this way is cleaner than adding it in the for loop would be.

src/encoded/audit/genetic_modification.py
```diff
@@ -73,3 +73,28 @@ def audit_genetic_modification_reagents(value, system):
...
             yield AuditFailure('missing genetic modification reagents', detail, level='WARNING')
+
+
+@audit_checker('GeneticModification', frame='object')
+def audit_genetic_modification_reagent_source(value, system):
+    reagents = value.get('reagents', [])
+    for reagent in reagents:
+        source = reagent['source']
+        identifier = reagent['identifier']
+        source_name = source.split('/')[2]
```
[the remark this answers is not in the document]
gabdank: @caseylitton source.split('/')[2] relies on the identifier structure of source objects, and the embedding of the source object in genetic_modification. I assume @p-sud can add a check that will prevent the exception in case of identifier structure change or embedding change in the future.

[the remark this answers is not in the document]
p-sud: Yeah I'll change this to get the source object from the using its @id in the reagent, and then use the source's 'name' instead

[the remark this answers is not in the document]
p-sud: I'm no longer using source_name = source.split('/')[2], I've changed the audit to request.embed the source object then access source.name. source_name_from_identifier is fine as is because it is guaranteed by the schema to include a colon in the middle, and disallow other colons.

PR #2673, opened 2019-03-21
ENCD-4535 long rna qc

p-sud:


jahilton:
A couple files don't end with empty lines

jahilton:
As an FYI, ISO-seq is being changed to 'long read RNA-seq' during this same release. Should probably go ahead and make those changes on this branch.

p-sud:
@jahilton the assay name and newline issues should be fixed

src/encoded/schemas/long_read_rna_mapping_quality_metric.json
```diff
@@ -0,0 +1,48 @@
...
+        },
+        "full_length_non_chimeric_read_count": {
+            "title": "Full length non-chimeric read count",
+            "description": "Quantity of full-length reads not mapping chimerically of the genome",
+            "type": "integer",
+            "minimum": 0
+        },
+        "mapping_rate": {
+            "title": "Mapping rate",
+            "description": "Percentage of reads mapping to the genome",
```
ottojolanki: This description is inaccurate and leads one to think that this would be a number between 0 and 100.

src/encoded/schemas/long_read_rna_mapping_quality_metric.json
```diff
@@ -0,0 +1,48 @@
...
+        { "$ref": "mixins.json#/standard_status"},
+        { "$ref": "mixins.json#/assay" }
+   ],
+    "properties": {
+        "schema_version": {
+            "default": "1"
+        },
+        "full_length_non_chimeric_read_count": {
+            "title": "Full length non-chimeric read count",
+            "description": "Quantity of full-length reads not mapping chimerically of the genome",
```
ottojolanki: This description would be better stated as: "Quantity of reads that are full-length and do not contain a chimeric adaptor arrangement."

src/encoded/schemas/long_read_rna_mapping_quality_metric.json
p-sud: Changed to "Proportion of reads..."

src/encoded/schemas/long_read_rna_mapping_quality_metric.json
p-sud: Updated description, thanks

PR #2675, opened 2019-03-22
ENCD-4158 Let apache cfg set on deploy

caseylitton:
This allows apache config file to be set on deployment depending on the role or the newly added --apache-cfg arg.  This fix the region indexer config issue too.   

This is helpful for the remote index worker as those deployments have special configs too.

TODO; See if we can inherit the conf that is similar between the files.

### DIFF
--- a/etc/no-region-apache.conf
+++ b/etc/no-region-apache.conf
@@ -1,0 +1,208 @@
+KeepAliveTimeout 75
+
+
+# The socket directory must be readable by the daemon process user
+WSGISocketPrefix /var/run/wsgi
+WSGIDaemonProcess encoded user=encoded group=encoded processes=6 threads=1 display-name=encoded-app
+# No need for embedded interpreters
+WSGIRestrictEmbedded On
+# Pass the authorization header so basic auth works
+WSGIPassAuthorization On
+
+# Indexer. Configure first to avoid catchall '/'
+WSGIDaemonProcess encoded-indexer user=encoded group=encoded processes=1 threads=1 display-name=encoded-indexer
+WSGIScriptAlias /_indexer /srv/encoded/parts/production-indexer/wsgi process-group=encoded-indexer application-group=%{GLOBAL}
+
+# visindexer. Configure first to avoid catchall '/'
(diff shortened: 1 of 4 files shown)


etc/no-region-apache.conf
```diff
@@ -0,0 +1,208 @@
+KeepAliveTimeout 75
```
hitz: I admit the name "no-region" threw me for a second.

etc/no-region-apache.conf
```diff
@@ -0,0 +1,208 @@
...
+
+<Directory /srv/encoded/parts/production-visindexer>
+    Order deny,allow
+    Allow from all
+    <IfModule access_compat_module>
+        Require all granted
+    </IfModule>
+</Directory>
+
+<Directory /srv/encoded/parts/production-regionindexer>
```
hitz: Should this be in the no region?  it will just be ignored I suppose since it's not configured above.

src/encoded/commands/deploy.py
```diff
@@ -541,23 +540,6 @@ def main():
 
 def parse_args():
 
-    def check_region_index(value):
```
hitz: glad this particular bit of absurdity is gone.

src/encoded/commands/deploy.py
```diff
@@ -640,6 +621,15 @@ def hostname(value):
             args.candidate = False
         elif args.candidate:
             args.role = 'candidate'
+    # Set apache conf based on role and/or argument
+    available_cfgs = ['no-region', 'encoded']
```
hitz: I find this a little confusing.   Should it maybe be encoded-no-region?

caseylitton: yea, then the default could be an empty string.  encoded-apache.cfg, encoded-noregion-apache.cfg, encoded-worker-apache.cfg.

etc/no-region-apache.conf
caseylitton: yea, we don't need that.  there could be more too.

caseylitton:
This was most likely updated in updates to deploy.  Closing.

PR #2676, opened 2019-03-23
SNO-85-facet-reappearing

zoldello:


buildout.cfg
```diff
@@ -61,7 +61,7 @@ webtest = git https://github.com/Pylons/webtest.git
 WSGIProxy2 = git https://github.com/lrowe/WSGIProxy2.git
 zope.sqlalchemy = git https://github.com/zopefoundation/zope.sqlalchemy.git
 pytest-bdd = git https://github.com/lrowe/pytest-bdd.git branch=allow-any-step-order
-snovault = git https://github.com/ENCODE-DCC/snovault.git rev=1.0.26
+snovault = git https://github.com/ENCODE-DCC/snovault.git rev= SNO-85-facet-reappearing
```
forresttanaka: I think this should be “branch=SNO-85-facet-reappearing.” It didn’t build locally until I changed that, though I guess long term it doesn’t matter.

forresttanaka:
If it’s OK, I’d like to postpone code reviewing the JS changes until after the snovault part of this is code reviewed, in case that feedback results in JSON changes that need JS changes. The JS changes aren’t huge.

hitz:
@forresttanaka SNO-85/backend in snovault reviewed.

forresttanaka:
@hitz Got it — thanks

src/encoded/static/components/search.js
```diff
@@ -656,6 +656,22 @@ const Term = (props) => {
...
+                {statusFacet ? <Status item={term} badgeSize="small" css="facet-term__status" noLabel /> : null}
+                <a className="facet-term__item" href={vhref} onClick={href ? onFilter : null}>
+                    <div className="facet-term__text">
+                        {em ? <em>{title}</em> : <span>{title}</span>}
+                    </div>
+                </a>
+            </li>
+        );
+    }
+
```
forresttanaka: I think this is good, but this return value is a near duplicate of the first several lines of the following return value. Do you think there’s a way to reduce this duplication, or would that make things _more_ complicated and difficult to maintain? I’ll go with your judgement here. This duplication looks like a maintenance hassle, but that’s OK if the alternative is worse.

buildout.cfg
caseylitton: yea, this is not the correct format.  @zoldello Please look at 

https://github.com/ENCODE-DCC/encoded/pull/2667/files

for correct format.

buildout.cfg
```diff
@@ -61,7 +61,8 @@ webtest = git https://github.com/Pylons/webtest.git
 WSGIProxy2 = git https://github.com/lrowe/WSGIProxy2.git
 zope.sqlalchemy = git https://github.com/zopefoundation/zope.sqlalchemy.git
 pytest-bdd = git https://github.com/lrowe/pytest-bdd.git branch=allow-any-step-order
-snovault = git https://github.com/ENCODE-DCC/snovault.git rev=1.0.27
+snovault = git https://github.com/ENCODE-DCC/snovault.git branch=SNO-85-facet-reappearing
+
```
caseylitton: remove extra line

src/encoded/static/components/search.js
zoldello: I went with the option of having two separate code; that can enable them to change independently.

buildout.cfg
zoldello: Fixed.

buildout.cfg
zoldello: removed

PR #2678, opened 2019-03-26
Kcl/encd 4517 add introduced elements ui

caseylitton:


caseylitton:
testing only for screen shots

PR #2679, opened 2019-03-26
ENCD-4503 micro RNA qc

gabdank:


jahilton:
Many changed files don't end with an empty line

jahilton:
Is there a specific reason why assay_term_name wasn't added as a facet to schemas/micro_rna_quality_metric.json?

src/encoded/schemas/micro_rna_quantification_quality_metric.json
```diff
@@ -0,0 +1,41 @@
...
+        { "$ref": "mixins.json#/assay" }
+   ],
+    "properties": {
+        "schema_version": {
+            "default": "1"
+        },
+        "expressed_mirnas": {
+            "title": "Expressed miRNAs",
+            "description": "Number of miRNAs expressed",
+            "comment": "miRNAs expressed < 2 cpms are filtered",
```
ottojolanki: If this comment is needed, I think come clarification may be good. "miRNAs expressed < 2 cpms are not counted as expressed" or something like that.

forresttanaka:
The e828ec88-8f74-437f-8bf7-8535ee34819e experiment in src/encoded/tests/data/inserts/experiment.json has an extra comma at the end of the last property that causes an error during local instance indexing (and I expect Travis CI too). The extra comma prevents experiments from indexing, and causes validation errors in objects referring to them.

PR #2680, opened 2019-03-26
ENCD-4535 long rna QC metric

gabdank:


src/encoded/schemas/long_read_rna_quality_metric.json
```diff
@@ -0,0 +1,151 @@
...
+                "properties": {
+                    "replicate": {
+                        "title": "Replicate",
+                        "description": "Isogenic replicate",
+                        "type": "string",
+                        "linkTo": "Replicate"
+                    },
+                    "mapping_rate": {
+                        "title": "Mapping rate",
+                        "type":  "number",
```
ottojolanki: I think the mapping rate should be a float between 0 and 1.0

gabdank: adressed

gabdank:
@p-sud is making his branch for this change, it will be PR'ed

PR #2681, opened 2019-03-26
ENCD-4513 add collections links to menu

emro:
'Collections' sub-drop-down updated to include 4 items:
- 'ENCORE' --> matrix/?type=Experiment&status=released&internal_tags=ENCORE
- 'ENTEx' --> matrix/?type=Experiment&status=released&internal_tags=ENTEx
- 'SE Stem Cell Consortium' --> matrix/?type=Experiment&status=released&internal_tags=SESCC
- also, move the 'Reference Epigenome' that is currently under 'Data' into the 'Collections' subset
In the very unlikely event that a user manually enters an internal tag that does not exist, a "not found" image will be displayed.


emro:
I feel like I've tried a lot of different solutions for the Chrome issue not displaying the "Not Found" image. I suggest we move forward with this ticket without that fix.

src/encoded/static/components/matrix.js
```diff
@@ -197,7 +197,10 @@ class Matrix extends React.Component {
                                 <div className="row">
                                     <div className="col-sm-11">
                                         <div>
-                                            <h1>{context.title}</h1>
+                                            <h1>
+                                                <div className="matrix-title">{context.title}</div>
```
forresttanaka: You’ll need to break this up somehow. `<h1>` elements [can only contain phrasing elements](https://w3c.github.io/html-reference/h1.html#h1) and a div [isn’t one](https://w3c.github.io/html-reference/common-models.html#common.elem.phrasing).

src/encoded/static/scss/encoded/modules/_matrix.scss
```diff
@@ -146,3 +146,30 @@ table.matrix th.group-all-groups-cell {
 .matrix-general-search {
     margin-bottom: 10px;
 }
+
+.matrix-header {
+    .col-sm-11 {
+        h1 {
+            display: flex;
+            align-items: flex-end;
```
forresttanaka: So that “Experiment matrix” will be aligned with the bottom row of internal_tags badges?

emro: Yeah, exactly.

emro:
I was looking at the demo to see the duplicate React key error in the JS console and I don't see it. But I agree that I should, so I'm confused about this. This is a case where we should see the error, right? https://encd-4513-b944a0852-emmao.demo.encodedcc.org/matrix/?type=Experiment&status=released&internal_tags=ENCORE&internal_tags=ENCORE

forresttanaka:
Yeah, the duplicate keys don’t seem to happen on the demo, so I’m not sure why I saw that. I think we can assume that’s not really a problem.

PR #2684, opened 2019-03-27
ENCD-4510 Fix human donor children calc prop

jlzamanian:
The code that calculates the children property was instead calculating the parents themselves. 

I fixed it in src/encoded/types/donor.py so that children are calculated properly and added a file (src/encoded/tests/test_types_donor.py) with a test to make sure the children are indeed calculated and placed in the children property of the specified parents.

### DIFF
--- a/src/encoded/tests/test_types_donor.py
+++ b/src/encoded/tests/test_types_donor.py
@@ -2,7 +2,7 @@ import pytest
 
 
 @pytest.fixture
-def parent(testapp, award, lab, human):
+def parent_human_donor(testapp, award, lab, human):
     item = {
         'lab': lab['@id'],
         'award': award['@id'],
@@ -12,7 +12,7 @@ def parent(testapp, award, lab, human):
 
 
 @pytest.fixture
-def child(testapp, award, lab, human):
+def child_human_donor(testapp, award, lab, human):
     item = {
         'lab': lab['@id'],
         'award': award['@id'],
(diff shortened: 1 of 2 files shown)


src/encoded/tests/test_types_donor.py
```diff
@@ -0,0 +1,29 @@
+import pytest
+
+
+@pytest.fixture
+def parent(testapp, award, lab, human):
```
keenangraham: More descriptive name. `parent_donor`?

src/encoded/tests/test_types_donor.py
```diff
@@ -0,0 +1,29 @@
...
+    item = {
+        'lab': lab['@id'],
+        'award': award['@id'],
+        'organism': human['@id']
+    }
+    return testapp.post_json('/human_donor', item).json['@graph'][0]
+
+
+@pytest.fixture
+def child(testapp, award, lab, human):
```
keenangraham: More descriptive name.

src/encoded/tests/test_types_donor.py
```diff
@@ -0,0 +1,29 @@
...
+    item = {
+        'lab': lab['@id'],
+        'award': award['@id'],
+        'organism': human['@id']
+    }
+    return testapp.post_json('/human_donor', item).json['@graph'][0]
+
+
+# patch parent to child, check that children are a calculated property for the parent
```
keenangraham: Unnecessary comment.

src/encoded/tests/test_types_donor.py
```diff
@@ -0,0 +1,29 @@
...
+def test_types_calculated_children(testapp, parent, child):
+    testapp.patch_json(child['@id'],
+                             {'parents': [parent['@id']]}).json['@graph'][0]
+    res = testapp.get(parent['@id'] + '@@index-data')
+    assert res.json['object']['children'] == [child['@id']]
```
keenangraham: Doesn't `res.json['children']` work?

src/encoded/tests/test_types_donor.py
jlzamanian: I changed the name to parent_human_donor. 

I included human because donor.py has calculated properties for donors from many organisms, not just human.

src/encoded/tests/test_types_donor.py
jlzamanian: I changed the name to child_human_donor.

src/encoded/tests/test_types_donor.py
jlzamanian: I removed the comment.

src/encoded/tests/test_types_donor.py
jlzamanian: It seems like it should, but removing ['object'] results in KeyError: 'children'.

Other tests also use the res.json['object']['property'] form.

keenangraham: Oh, yeah, I see. It is because you are getting the `'@@index-data'` endpoint.

PR #2685, opened 2019-03-27
ENCD-3919-index-vis-blobs-on-deploy

caseylitton:


src/encoded/visualization.py
```diff
@@ -1136,9 +1136,12 @@ def stringify(self):
         return self.ucsc_trackDb()
 
 
-def vis_cache_add(request, dataset):
+def vis_cache_add(request, dataset, is_vis_indexer=False):
```
hitz: is this ever called without is_vis_indexer=True...    Just not sure what case we are trying to handle here.
Principle of minimum effect?

caseylitton: yea, I didn't want to remove it this late in the sprint cycle.  Doesn't look like it's used anywhere else, but who knows...maybe regulome uses it or some third party.

PR #2695, opened 2019-04-01
ENCD-4412 Add Clear Cart item to cart menu

forresttanaka:
The most significant change is the refactorization of `CartClear`, which used to handle the trigger button as well as the modal it triggered. The modal now gets rendered by the new `CartClearModal` component so that both a now-separate button (now called `CartClearButton`) as well as the new menu item can trigger the Clear Cart warning modal.

Because “Clear Cart” would be the third item in the cart menu needing the current cart name, and because we now have a non-selectable menu item in the “Data” menu (“Collections”), now seemed the time to have a non-selectable menu item displaying the current cart name in the cart menu so I could remove them from each of “View cart,” “Share cart,” and “Clear cart.” I implemented this using a `<span>` within the menu `<li>`, so I had to add the `<span>` to the menu CSS, including an inline-block style so that this `<span>` gets the same margins as other menu items.

### DIFF
--- a/src/encoded/static/components/__tests__/cart-test.js
+++ b/src/encoded/static/components/__tests__/cart-test.js
@@ -20,6 +20,7 @@ const initialCart = {
     elements: [],
     name: 'Third cart',
     current: '/carts/9570dbda-191d-42c3-8ac9-f479ca55f6f6/',
+    inProgress: false,
 };
 const mockStore = configureStore();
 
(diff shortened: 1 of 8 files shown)


forresttanaka:
Hang on; found a mobile styling problem.

forresttanaka:
Mobile styling fixed and Travis passed.

PR #2702, opened 2019-04-05
ENCD-4560 update typeaheads after selections

emro:
Two bug fixes here:
(1) When user types in a typeahead, and then chooses another facet, the typeahead facet terms should update according to the new filter (previously they did not). This is accomplished by moving the filtering into the render() function and by only keeping the search term, not the full filtered list, in state.
(2) Previously typeahead search bars would show or not show depending on how many terms were available and this introduced bugs (confusing behavior) if you searched on them and then selected/deselected other filters, because searches could be in state but not displayed. Now we will show the search bars for all typeaheads no matter how many terms there are.

### DIFF
--- a/src/encoded/static/components/search.js
+++ b/src/encoded/static/components/search.js
@@ -721,8 +721,8 @@ class Facet extends React.Component {
 
         // Set initial React commponent state.
         this.state = {
-            filteredTerms: null,
             initialState: true,
+            searchTerm: '',
         };
 
         // Bind `this` to non-React methods.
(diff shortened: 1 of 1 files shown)


emro:
This is a pretty minor case, but when users enter a search term with * (or \, but I haven't added that yet), this causes errors so I strip them out of this.state.searchTerm. So right now if you type "****", it shows up in the search bar but none of the results are filtered because I'm ignoring *s and replacing them with empty strings. If I make `value={this.state.searchTerm}`, then every time you type a *, I replace it with empty space and you get nothing in the search bar no matter how many *s you type. Which is a little weird. No one should try to type *s anyway. Does that seem like ok behavior? Is there some better way to handle this?

PR #2703, opened 2019-04-05
ENCD-4548 fix fallback images on collection pages

emro:
Collections pages display a collections logo. In the case where someone attempts to search for a collections page that does not exist, we have a fallback image. This fallback image was not getting displayed on some browsers. The new "ImageWithFallback" component fixes this issue by ensuring that the onError function is loaded.

### DIFF
--- a/src/encoded/static/components/objectutils.js
+++ b/src/encoded/static/components/objectutils.js
@@ -683,15 +683,57 @@ InternalTags.defaultProps = {
 };
 
 /**
+ * Display image with fallback for non-existent image links
+ */
+export class ImageWithFallback extends React.Component {
+    constructor() {
+        super();
+
+        // Initialize image src and alt tags to be empty to ensure component mounts
+        this.state = {
+            imageUrl: '',
+            imageAlt: '',
+        };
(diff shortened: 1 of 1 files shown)


src/encoded/static/components/objectutils.js
```diff
@@ -682,16 +682,58 @@ InternalTags.defaultProps = {
...
+        // Initialize image src and alt tags to be empty to ensure component mounts
+        this.state = {
+            imageUrl: '',
+            imageAlt: '',
+        };
+        this.onError = this.onError.bind(this);
+    }
+
+    // Once the component has mounted, update image src and alt tag
+    componentDidMount() {
```
forresttanaka: I think you might need to comment about this needing to be done at mount so we don’t accidentally change it later, making this fix ineffective. Normally you’d set state with props in the constructor, and I worry I or someone else will see this code and think, “Oh, this can be done easier in the constructor” and change it, breaking the fix.

emro: Good point, hopefully this is clearer!

PR #2711, opened 2019-04-12
ENCD-4559-x-limit-zero-results

zoldello:


src/encoded/helpers/helper.py
```diff
@@ -103,8 +103,10 @@ def search_result_actions(request, doc_types, es_results, position=None):
 
 class View_Item:
     def __init__(self, request, search_base):
+        # strip out x.limit=<number> and y.limit=<number>; as they are not used
+        search_base_without_limit = re.sub(r'&[x|y]\.limit\=\d*', '', search_base, flags=re.IGNORECASE)
```
caseylitton: Do we always want to throw out x limit or just when it's empty?  

Also, we only want this to happen the limit comes after an ampersand?  If it's the first parameters limit will comes after a question mark.

src/encoded/helpers/helper.py
```diff
@@ -103,8 +103,10 @@ def search_result_actions(request, doc_types, es_results, position=None):
 
 class View_Item:
     def __init__(self, request, search_base):
+        # strip out x.limit=<number> and y.limit=<number>; as they are not used
```
caseylitton: this comment just says what the code does.  It's not a bad idea to explain why we're doing this, but it's weird to have a comment at the top of init.  Maybe put the sub in a function and then explain in there.  Just don't explain what the next line of code does unless it's really confusing.

zoldello: @caseylitton I took out the comment. Code supposed to explain itself or be re-written to. With that in mind, the comment was unnecessary as one could read the code.

src/encoded/helpers/helper.py
zoldello: Good questions. 

We only want to get rid of x/y limit when they are not needed. So, the variable: "search_base_without_limit" is used in some but not all cases (see lines after 107) 

Yes, x/y limit always come after an ampersand. In the narrow case where this code is used, x/y limit are never the first parameter. The first parameter is typically "type", like "type=Experiment"

caseylitton: The ticket is not clear on whether we want to always remove limit or only when it breaks the site (i.e. empty limit).  Please check with Ben and update the JIRA to be more clear.  Do we always remove it in certain circumstances or only when it's empty.  To me the problem is that 'limit=' where limit=1 doesn't.  But that is a product question.

***

I don't see a good reason to always assume limit is after &.  Seems like an unnecessary qualification.

src/encoded/helpers/helper.py
caseylitton: Okay, but again it is not a bad idea to put it in a function and explain why we're doing it.  Just don't explain what the code does.

src/encoded/helpers/helper.py
caseylitton: per convo, I was misunderstanding the feature.  But we did decided to make is ? or &

caseylitton: also seems to complicated to fix, so we're assuming x limit is not first.

PR #2714, opened 2019-04-15
ENCD-4565 remove back slashes from search terms

emro:
Back slashes entered in the typeahead search bars cause console errors. In order to prevent this, we will remove back slashes from search terms. We will also remove forward slashes for consistency handling slashes.

### DIFF
--- a/src/encoded/static/components/search.js
+++ b/src/encoded/static/components/search.js
@@ -730,7 +730,11 @@ TypeTerm.propTypes = {
 };
 
 // Sanitize user input and facet terms for comparison: convert to lowercase, remove white space and asterisks (which cause regular expression error)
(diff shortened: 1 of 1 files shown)


emro:
I like writing it in just one line, it's more concise. I wrote it slightly differently but I think it accomplishes the same replacement.

forresttanaka:
Is it better to remove anything that’s not a letter or number from the search, or is that too restrictive?

emro:
I talked to Jason this question, whether it was better to exclude special characters or only allow numbers and letters, and he thought the list of special characters was short enough that this made more sense. Also it sounds like they are thinking about renaming one of the assays 'polyA(-) RNA-seq' and I think this solution will work well for cases like that. Users can type the ()s and they'll just get ignored.

PR #2716, opened 2019-04-16
ENCD-4577 Add documents on characterization page

yunhailuo:


yunhailuo:
> I think you inadvertently removed the 'new line' at the end of the file

Thanks a lot. Corrected.

PR #2717, opened 2019-04-16
[REVERT] Manually install Java 211

caseylitton:
Temporary change that downloads java 8 211 from our aws prod s3 bucket and manually installs.  Java 8 is behind a user login and we need to update to java 11, since 9/10 are not major releases.


caseylitton:
closing for https://github.com/ENCODE-DCC/encoded/pull/2718

PR #2718, opened 2019-04-16
ENCD-4581 Temp install java 8 through aws

caseylitton:
Temporary change that downloads java 8 211 from our aws prod s3 bucket and manually installs. Java 8 is behind a user login and we need to update to java 11, since 9/10 are not major releases.


caseylitton:
Restarted non bdd tests, not sure what was going on there

PR #2720, opened 2019-04-18
SNO-89 Fix backoff error key in indexer

caseylitton:


hitz:
Indexing unit tests have serious failure.

PR #2722, opened 2019-04-19
SNO-88-fix-user-search-count

zoldello:


zoldello:
@hitz Ready for code review

PR #2730, opened 2019-04-24
ENCD-4545 Update visualization browser list

forresttanaka:
Removed the `setFilters` method because while debugging this, I found it made things _more_ confusing. This component has other one-line methods that I think _do_ help readability, but this one hurt it.

Most of this ticket is in the new method, `resetCurrentBrowser`, which changes the `currentBrowser` state when the chosen one becomes unavailable.


forresttanaka:
Now do the same check for an empty browser list as the code it replaces. Also added an optional files parameter to `resetCurrentBrowser` for the cases in which the `files` state gets set just before, and so `this.state.files` wouldn’t be up to date until the next render.

forresttanaka:
Oh, and to me `currentFilterValue` only holds `<select>` values, so I find the connection between them clearer with …`value` in the name.

PR #2734, opened 2019-04-25
Fix import_data to use changed load_all in #2351

lrowe:


lrowe:
ENCD-5098

lrowe:
I've identified the commit causing the failure but need to investigate why it should be a problem (PATCH-ing on second pass of import data feels like the 'correct' thing to do.)

PR #2736, opened 2019-04-29
ENCD-4504 miRNA audits

p-sud:


p-sud:
Updated test to include check that audits are not being triggered when qc values are above thresholds

src/encoded/loadxl.py
```diff
@@ -62,6 +62,8 @@
     'bismark_quality_metric',
     'cpg_correlation_quality_metric',
     'chipseq_filter_quality_metric',
+    'micro_rna_quantification_quality_metric',
```
caseylitton: flagging for removal of revert commit

PR #2741, opened 2019-04-30
ENCD-4593 Split ChIP-seq

yunhailuo:


src/encoded/schemas/changelogs/target.md
```diff
@@ -1,5 +1,9 @@
 ## Changelog for target.json
 
+### Minor changes since schema version 10
+
+* *investigated_as* enum changed to delete "control", "nucleotide modification", "other post-translational modification", and to add "Co-factor", "Cohesin", "DNA replication", "DNA repair", "RNA polymerase complex".
```
gabdank: I think that should be slightly reworded to be: "enum list was modified. The following terms were removed .. and the following terms were added ..."

gabdank: And "control" was not deleted from the list

yunhailuo: Fixed with the new commit.

yunhailuo:
> @yunhailuo ENCD-3108 has to be merged before this PR?

Yes, that's the plan.

yunhailuo:
> @yunhailuo can you please add a test that checks the proper choice of preffered_name in ChIP experiment?

Added. Can you check if that's good enough, @gabdank ?

yunhailuo:
This is a "Multi-Jira Ticket Pull Requests". #2742 should get in first, @caseylitton .

src/encoded/schemas/changelogs/target.md
```diff
@@ -1,5 +1,18 @@
 ## Changelog for target.json
 
+### Schema version 11
```
caseylitton: flagging for removal of revert commit

PR #2746, opened 2019-05-04
ENCD-3642-fix-file-format-download

keenangraham:
Quick fix for batch_download and metadata endpoint when plus is in file type.

### DIFF
--- a/src/encoded/batch_download.py
+++ b/src/encoded/batch_download.py
@@ -230,7 +230,7 @@ def _get_annotation_metadata(request, search_path, param_list):
         urlencode(param_list, True),
         ''.join(fields)
     )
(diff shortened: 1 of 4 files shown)


forresttanaka:
Is there a simple description of under what conditions you would have to double encodedURIComponent?

keenangraham:
Yes. It's to make encoding consistent with the batch_download link generated from search pages (e.g. _batch_download_ field here https://www.encodeproject.org/search/?type=Experiment&status=released&status=released&files.file_type=bed+bed3%2B&format=json), which happens to be double quoted (probably by accident). This is ugly solution to deal with the first line in both batch_download and metadata endpoints (`param_list = parse_qs(request.matchdict['search_params'])`) which double unquotes.

PR #2748, opened 2019-05-07
ENCD-4361-validate-schema-set-status

keenangraham:
This validates schema on status transition.

### DIFF
--- a/src/encoded/tests/test_status_trigger.py
+++ b/src/encoded/tests/test_status_trigger.py
@@ -641,3 +641,23 @@ def test_set_status_skip_acl_on_not_available_file(testapp, content, mocker, fil
     assert File.set_public_s3.call_count == 0
     res = testapp.get(file['@id'])
     assert res.json['status'] == 'released'
+
+
+def test_set_status_validation_error_on_content_error_details(testapp, file):
+    testapp.patch_json(file['@id'], {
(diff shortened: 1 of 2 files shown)


src/encoded/tests/test_status_trigger.py
```diff
@@ -641,3 +641,23 @@ def test_set_status_skip_acl_on_not_available_file(testapp, content, mocker, fil
...
+
+
+def test_set_status_no_validation_error_on_content_error_details(testapp, file):
+    testapp.patch_json(file['@id'], {
+        'status': 'content error',
+        'content_error_detail': 'Fastq file contains bad sequence'
+    })
+    r = testapp.get(file['@id'] + '@@raw')
+    assert 'content_error_detail' in r.json
+    testapp.patch_json(file['@id'] + '@@set_status?update=true&validate=false', {'status': 'uploading'}, status=200)
```
hitz: I thought you were trying to remove the content_error_detail?

keenangraham: I would actually have to clear content_error_detail and change status simultaneously (which has been requested in another ticket). This tests that the validate function is what is causing the initial patch to fail (because of the content error status/content_error_details dependency), which is bypassed when you specify `validate=false`.

PR #2749, opened 2019-05-07
ENCD-4522-fixed-issue-with-news-date

zoldello:


src/encoded/static/components/news.js
```diff
@@ -31,10 +31,12 @@ NewsPreviewItem.propTypes = {
 // Display a list of news preview items from search results.
 const NewsPreviews = (props) => {
     const { items } = props;
-    if (items && items.length) {
+    const newsItems = items && items.length ? [...items] : [];
```
forresttanaka: Only one little thing: this runs afoul of the [AirBnB explicit-comparison](https://github.com/airbnb/javascript#comparison--shortcuts) requirement.

src/encoded/static/components/news.js
```diff
@@ -31,10 +31,12 @@ NewsPreviewItem.propTypes = {
...
+    newsItems.sort((item1, item2) => new Date(item2.date_created) - new Date(item1.date_created));
+    if (newsItems.length > 0) {
```
forresttanaka: Might as well move this sort inside the following `if` so we don’t have to call sort on a potentially zero-length array.

PR #2752, opened 2019-05-09
ENCD-4619 add experiment_classification to data insert

jahilton:
From ENCD-4558, a merge conflict wasn't handled correctly. This fixes it by adding back a required property to an insert.


caseylitton:
I restarted the tests also, 830am thursday.

PR #2757, opened 2019-05-12
SNO-88-fix-user-search-count

zoldello:


caseylitton:
closing, make new branch with correct number

PR #2758, opened 2019-05-13
SNO-93-user-count-does-not-match

zoldello:


caseylitton:
@hitz They should be reviewed.  I've seen people build demos with the wrong sno branch and say they are correct.  Also we won't know until looking whether there are changes beside buildout or not. 
 And encode changes may be added after code review or qa that didn't initially exist.  We shouldn't assume or say there is no need for code review.

The reviewer should verify the branch is correct and then do nothing if sno branch is the only update.  If there are other changes or changes get added later in the process we'll use the github approval?

Then there is the additional question of what the process should be.  This won't be changed until next release though.  As of now the encoded branch will be merged.

PR #2764, opened 2019-05-15
ENCD-4609 Exclude fccs institutional certification

p-sud:
… certification audit


p-sud:
My idea was that we could potentially use "component" in the future to indicate "mapping", "data analysis", etc to indicate other types of awards we need to do stuff with. I did consider a boolean but it's not extensible.

PR #2769, opened 2019-05-16
SNO-92 Fix batch upgrade logging

caseylitton:
test_admin_user failed for bdd

same indexing issue for non bdd


caseylitton:
going in with sno 100 https://github.com/ENCODE-DCC/encoded/pull/2805/commits

PR #2771, opened 2019-05-16
SNO-97-matrix-update

zoldello:


zoldello:
Still in development

src/encoded/viewconfigs/target_matrix.py
```diff
@@ -0,0 +1,66 @@
+"""
```
caseylitton: Remove all unused imports from file.

zoldello:
Closing PR. Still in prototype mode.

PR #2773, opened 2019-05-17
ENCD-4639 Organize demo cloud yaml

caseylitton:


cloud-build/run-script.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
```
caseylitton: Testing only, not to be checked in

cloud-config.yml
```diff
@@ -96,12 +96,26 @@ bootcmd:
 - cloud-init-per once mkswap-swapfile mkswap /swapfile
```
hitz: Should probably cross check with @lrowe 's https://github.com/ENCODE-DCC/encoded/pull/2652 for python 3.7/ES6+

caseylitton: This should only change the cloud-config yaml.  lrowes didn't have any config/deploy like changes.

cloud-build/post-install-install.sh
```diff
@@ -0,0 +1,28 @@
...
+a2enmod ssl
+a2ensite encoded.conf
+a2dissite 000-default
+a2enconf logging
+a2disconf charset
+a2disconf security
+a2disconf localized-error-pages
+a2disconf other-vhosts-access-log
+a2disconf serve-cgi-bin
+if test "%(ROLE)s" = "demo"
```
hitz: I wonder, given our experience with ENCD-4613 if the batchupgrade should be broken out.  I am 99% sure the order doesn't matter, just that it's supposed to run before the machine reboots with these apache config changes.

caseylitton: Yea, that's a good idea.  If it's easy I'll through that in.

PR #2780, opened 2019-05-23
ENCD-4541 Add multiple facets to cart view pages

forresttanaka:
The most memory-impactful change involves collecting all file objects from the experiment search results for all experiments in the cart. Before this branch, I do the same search but throw out the file information. Because I handle the facet selection manually in this branch, I have to track all files so I know which to include and exclude. These aren’t complete file objects — just the five properties we facet. But in extreme cases, this can include close to 400,000 of these small file objects.

### DIFF
--- a/src/encoded/batch_download.py
+++ b/src/encoded/batch_download.py
@@ -446,7 +446,7 @@ def metadata_tsv(context, request):
 def batch_download(context, request):
     # adding extra params to get required columns
     param_list = parse_qs(request.matchdict['search_params'])
-    param_list['field'] = ['files.href', 'files.file_type', 'files.restricted']
+    param_list['field'] = ['files.href', 'files.restricted'] + [k for k, v in param_list.items() if k.startswith('files.')]
     param_list['limit'] = ['all']
 
     experiments = []
@@ -503,7 +503,7 @@ def batch_download(context, request):
 
     files = [metadata_link]
     for exp_file in exp_files:
-        if not file_type_param_list(exp_file, param_list):
+        if not files_prop_param_list(exp_file, param_list):
             continue
         elif restricted_files_present(exp_file):
             continue
@@ -521,10 +521,12 @@ def batch_download(context, request):
     )
 
 
-def file_type_param_list(exp_file, param_list):
-    if 'files.file_type' in param_list:
-        if not exp_file['file_type'] in param_list.get('files.file_type', []):
-            return False
+def files_prop_param_list(exp_file, param_list):
+    for k, v in param_list.items():
+        if k.startswith('files.'):
+            file_prop = k[len('files.'):]
+            if file_prop in exp_file and exp_file[file_prop] not in v:
+                return False
     return True
 
 
--- a/src/encoded/batch_download.py
+++ b/src/encoded/batch_download.py
@@ -397,9 +397,9 @@ def metadata_tsv(context, request):
                             'files.output_type']
 
             for f in experiment_json['files']:
-                if 'files.file_type' in param_list:
-                    if f['file_type'] not in param_list['files.file_type']:
-                        continue
+                # If we're looking for a file type but it doesn't match, ignore file
+                if not files_prop_param_list(f, param_list):
+                    continue
                 if restricted_files_present(f):
                     continue
                 if is_no_file_available(f):
--- a/src/encoded/static/components/cart/cart.js
+++ b/src/encoded/static/components/cart/cart.js
@@ -7,6 +7,7 @@ import { connect } from 'react-redux';
 import _ from 'underscore';
 import Pager from '../../libs/bootstrap/pager';
 import { Panel, PanelBody, PanelHeading } from '../../libs/bootstrap/panel';
+import { tintColor } from '../datacolors';
 import { itemClass, encodedURIComponent, parseAndLogError } from '../globals';
 import { requestObjects, DisplayAsJson } from '../objectutils';
 import { ResultTableList } from '../search';
@@ -18,12 +19,14 @@ import CartMergeShared from './merge_shared';
 
 /** Number of dataset elements to display per page */
 const PAGE_ELEMENT_COUNT = 25;
-/** File facet fields to display */
+/** File facet fields to display in order of display */
 const displayedFacetFields = [
+    'output_type',
     'file_type',
+    'assembly',
+    'lab.title',
+    'status',
 ];
-/** File facet field to use to count the total file count */
-const displayedFacetCountingField = 'file_type';
 
 
 /**
@@ -44,7 +47,8 @@ class CartSearchResults extends React.Component {
     }
 
     componentDidUpdate(prevProps) {
-        if (prevProps.currentPage !== this.props.currentPage || !_.isEqual(prevProps.elements, this.props.elements) || (prevProps.loggedIn !== this.props.loggedIn)) {
+        const { currentPage, elements, loggedIn } = this.props;
+        if (prevProps.currentPage !== currentPage || !_.isEqual(prevProps.elements, elements) || (prevProps.loggedIn !== loggedIn)) {
             this.retrievePageElements();
         }
     }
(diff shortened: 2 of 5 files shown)


src/encoded/batch_download.py
```diff
@@ -521,10 +521,12 @@ def batch_download(context, request):
...
-    if 'files.file_type' in param_list:
-        if not exp_file['file_type'] in param_list.get('files.file_type', []):
-            return False
+def files_prop_param_list(exp_file, param_list):
+    for k, v in param_list.items():
+        if k.startswith('files.'):
+            file_prop = k[len('files.'):]
+            if file_prop in exp_file and exp_file[file_prop] not in v:
+                return False
     return True
```
emro: Maybe someone else should review the Python changes?

src/encoded/static/components/cart/cart.js
```diff
@@ -229,67 +378,141 @@ const addToAccumulatingFacets = (accumulatingResults, currentResults, facetField
...
+     */
+    handleExpanderEvent(e) {
+        this.props.expanderClickHandler(this.props.field, e.altKey);
     }
 
     render() {
-        const { facet, selectedFacetTerms } = this.props;
+        const { facet, field, title, expanded, selectedFacetTerms } = this.props;
+        const maxTermCount = Math.max(...Object.keys(facet).map(term => facet[term]));
+        const sortedTerms = Object.keys(facet).sort();
```
emro: I like having the terms sorted alphabetically.

src/encoded/static/components/cart/cart.js
```diff
@@ -312,21 +543,61 @@ Facet.defaultProps = {
...
             /** Tracks facet loading progress */
-            facetLoadProgress: 0,
+            facetLoadProgress: null,
         };
+
+        // Initialize the expanded state of every facet; only the first one is expanded by default.
+        const expandedStates = {};
+        displayedFacetFields.forEach((field, index) => {
+            expandedStates[field] = index === 0;
+        });
```
emro: I don't know if this is better, but the default `expandedStates` could probably be defined next to `displayedFacetFields`. Also it could just be defined explicitly since the first one is hard-coded to be the only expanded one. If we'll want to be able to redefine which facets are expanded, there could be a variable for the open ones that could be checked against here but I'm not sure if there's a good place to define that / if we'll want to be changing that.

src/encoded/static/components/cart/cart.js
```diff
@@ -411,65 +683,164 @@ class FileFacets extends React.Component {
...
+            const nextExpandedStates = Object.assign({}, state.expanded);
+            nextExpandedStates[field] = !nextExpandedStates[field];
+            return { expanded: nextExpandedStates };
+        });
+    }
+
+    render() {
+        const { selectedTerms, termClickHandler } = this.props;
+
+        // Based on carted files and currently selected facet terms, generate a facet object for
```
emro: I think you mean `cart` not `carted`?

src/encoded/static/components/cart/cart.js
```diff
@@ -411,65 +683,164 @@ class FileFacets extends React.Component {
...
+        this.setState((state) => {
+            if (altKey) {
+                // Alt key held down, so expand or collapse *all* facets.
+                const nextExpandedState = !state.expanded[field];
+                const expandedStates = {};
+                displayedFacetFields.forEach((facetField) => {
+                    expandedStates[facetField] = nextExpandedState;
+                });
+                return { expanded: expandedStates };
+            }
```
emro: This is cool! When we build the advanced facet search we should do this too. And possibly this should also be implemented for the FAQ.

src/encoded/static/components/cart/cart.js
```diff
@@ -93,53 +97,182 @@ CartSearchResults.defaultProps = {
 
 
 /**
- * Display one term of the File Format facet.
+ * Display a count of the total number of files selected for download given the current facet term
+ * selections. Shows the facet-loading progress bar if needed.
+ */
```
emro: I like all of the visual cues on this component to let you know when everything is done loading / when the number of files is updated.

src/encoded/batch_download.py
forresttanaka: Yeah, I’ll pass it by Ben.

src/encoded/static/components/cart/cart.js
```diff
@@ -93,53 +97,182 @@ CartSearchResults.defaultProps = {
...
+            <li className="cart-facet-term">
+                <div
+                    className="cart-facet-term__item"
+                    role="button"
+                    tabIndex="0"
+                    id={`cart-facet-term-${term}`}
+                    onKeyDown={this.handleKeyDown}
+                    onClick={this.handleTermClick}
+                    aria-pressed={selected}
+                    aria-label={`${termCount} files ${term}`}
```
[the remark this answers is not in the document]
forresttanaka: I had it more detailed until a very late commit because, since aria-labels are read as speech, I thought they should be shorter so a visually impaired person doesn’t get annoyed. These are for individual facet terms that could get to be a very long list. I haven’t been able to find a general philosophy about aria-labels though, probably because it depends on the situation.

Right now with Voice Over for macOS, it says, “Twelve hundred twenty-two files alignments toggle button” or “Twelve hundred twenty-two files alignments selected toggle button” if the user had selected it. Other text-to-speech engines could say it differently.

src/encoded/static/components/cart/cart.js
forresttanaka: I’ll give that some thought and update this later. If I make `expandedStates` a module-level variable (I think that’s what you mean?) I’d have to clone it here since we have to give `this.state` its own objects each time. Here I’m generating a new object each time.

src/encoded/static/components/cart/cart.js
forresttanaka: I mean “files that are included in the cart,” so “carted files.” Is that weird?

forresttanaka:
@hitz One small Python change to check.

src/encoded/static/components/cart/cart.js
forresttanaka: I’d like to keep it this way. Right now if you change the order of `displayedFacetFields`, the first one gets opened by default with no other changes, and that already happened when Ben wanted the order changed earlier. If we decide to have specific ones opened by default, I’d prefer to make `displayedFacetFields` an array of objects that includes a boolean set to true for the opened-by-default ones, instead of a separate object that parallels `displayedFacetFields`.

src/encoded/static/components/cart/cart.js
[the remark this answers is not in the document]
emro: I can definitely see the benefit to keeping it short, I just thought '1222 files alignments' was confusing. Maybe just adding a preposition 'files of alignments' would make it a little more clear??

src/encoded/static/components/cart/cart.js
emro: OK! Sounds good.

src/encoded/static/components/cart/cart.js
emro: I think 'cart files' could also mean files in the cart, but 'carted files' works!

src/encoded/tests/test_batch_download.py
```diff
@@ -133,11 +136,13 @@ def test_batch_download_lookup_column_value(lookup_column_value_item, lookup_col
...
 
 @pytest.mark.parametrize("test_input,expected", [
-    (file_type_param_list(exp_file_1, param_list_2), True),
-    (file_type_param_list(exp_file_1, param_list_1), True),
-    (file_type_param_list(exp_file_2, param_list_1), False),
+    (files_prop_param_list(exp_file_1, param_list_2), True),
+    (files_prop_param_list(exp_file_1, param_list_1), True),
+    (files_prop_param_list(exp_file_2, param_list_1), False),
+    (files_prop_param_list(exp_file_3, param_list_3), True),
+    (files_prop_param_list(exp_file_1, param_list_3), False),
```
keenangraham: Would you mind adding a test for weird file_types like `bigBed bed3+`.

forresttanaka: I’m sure I could. Let me know what I have to do. I have a CSS change to put in as well.

keenangraham: Also I don't think the files_prop_param_list function call in the parametrize decorator is a good pattern to use. The function will be evaluated at decoration time instead of test time and it causes a bunch of duplicate files_prop_param_list strings. Instead just pass parameters and call the function in the test: 
```
@pytest.mark.parametrize("exp_file, param_list, expected", [
    (exp_file_1, param_list_2, True),
    (exp_file_1, param_list_1, True),
])
def test_files_prop_param_list(exp_file, param_list, expected):
    assert file_type_param_list(exp_file, param_list) == expected
```

keenangraham: > Let me know what I have to do.
Since the urlencoding is the main issue I think it needs more than unit test. 

You should actually make a `testapp.get` request to batch_download endpoint, using the URL that would be generated by cart, and assert that the file is listed in the returned body. See here for framework:
https://github.com/ENCODE-DCC/encoded/blob/5c2b446a05a969ac89460107810d9fa010ad41bd/src/encoded/tests/test_batch_download.py#L109

There's already a bed3+ file (ENCFF880XNW) in data inserts: https://github.com/ENCODE-DCC/encoded/blob/5c2b446a05a969ac89460107810d9fa010ad41bd/src/encoded/tests/data/inserts/file.json#L2572

So you could load workbook in your test, make get request with various file filters included, and assert that ENCFF880XNW is or is not in the body.

`testapp.get('/batch_download/type=Experiment&files.file_type=bed%2Bbed3%252B')`

I would like to see this with several different filters applied. For example `&files.accession=ENCFF880XNW` or `&files.accession!=ENCFF880XNW` (in which case the file shouldn't be in the TSV). 

By the way will negation (!=) work here!?

forresttanaka: I feel like overall batch_download fixes shouldn’t be in a ticket about cart facets. This fix is in there so that the number of files downloaded isn’t completely different from the number of files promised by the front-end facet code. Unrelated issues should go into one or more other tickets.

forresttanaka: > Also I don't think the files_prop_param_list function call in the parametrize decorator is a good pattern to use.

I think this should also be addressed in another ticket so that `test_restricted_files_present` can be fixed as well.

forresttanaka:
Newest CSS-only commit addresses the too-low checkmarks in Firefox and the wide gaps between facet terms in Edge and IE11. Both changes tested in Chrome, Firefox, Safari, Edge, and IE11.

src/encoded/static/components/cart/cart.js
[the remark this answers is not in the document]
forresttanaka: I reversed the words, so now it says things like, “Twenty-three alignments files selected toggle button” (if it’s selected).

src/encoded/batch_download.py
```diff
@@ -446,7 +446,7 @@ def metadata_tsv(context, request):
 def batch_download(context, request):
     # adding extra params to get required columns
     param_list = parse_qs(request.matchdict['search_params'])
-    param_list['field'] = ['files.href', 'files.file_type', 'files.restricted']
+    param_list['field'] = ['files.href', 'files.restricted'] + [k for k, v in param_list.items() if k.startswith('files.')]
```
hitz: Do we even need to specify .href and .restricted here?

forresttanaka: I wasn’t sure, but I was afraid they might have some purpose not obvious to me, since they didn’t _appear_ to be used before either.

src/encoded/tests/test_batch_download.py
hitz: I agree with @forresttanaka here; it was karthik's weird pattern and outside the scope of this ticket.

I will add your comments to the batch_download test ticket.

PR #2782, opened 2019-05-24
ENCD-4656 fix stupid bdd tests on travis

hitz:
pulling this to trigger travis

### DIFF
--- a/src/encoded/tests/features/forms.feature
+++ b/src/encoded/tests/features/forms.feature
@@ -7,7 +7,7 @@ Feature: Edit forms
         And I click the element with the css selector ".icon-gear"
         And I click the link to "/antibodies/ENCAB728YTO/#!edit"
         And I wait for the content to load
-        And I wait for 5 seconds
+        And I wait for 8 seconds
         Then I should see an element with the css selector "form.rf-Form"
         When I fill in "antigen_description" with "It's not a very nice antigen"
         And I press "save"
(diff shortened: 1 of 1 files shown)


forresttanaka:
There’s also this kind of line defined in browsersteps.py, though I don’t think I’ve ever used it and am not sure it helps:

```
@then(parse('I should see an element with the css selector "{css}" within {timeout:d} seconds'))
def should_see_element_with_css_within_timeout(browser, css, timeout):
    assert browser.is_element_present_by_css(css, wait_time=timeout), u'Element not found'
```

I don’t know what wait_time is by default.

hitz:
Does that test fail often?

forresttanaka:
The “I should see an element with the css selector…” is the kind of test you’re adding an extra three seconds before, so now that I see the version of the same test but with a delay, I wonder if that would work too.

forresttanaka:
I’ll give that a try in another branch. I just got that CSS timeout error in a different place in the BDD test, in fact, a place I’d never seen it happen before.

hitz:
yeah this one timed out waiting to load... maybe add a timeout here.?

forresttanaka:
My (n=1) test went well until near the end when it timed out waiting for a form to load. I’m trying another experiment with changing the default timeout time. Travis started on that a few minutes ago #shootinginthedark

PR #2795, opened 2019-06-03
ENCD-4668 Update moment parsing string

forresttanaka:
For some reason I had an incorrect moment() parsing string, and for some reason that now causes this problem, as well as a deprecation warning in the console. This one fix gets rid of the “Invalid date” facet terms as well as the console warning.

### DIFF
--- a/src/encoded/static/components/news.js
+++ b/src/encoded/static/components/news.js
@@ -242,7 +242,7 @@ const DateFacet = (props) => {
                         <li key={term.key} className="news-facet__item">
                             <a href={termHref} className={termFilter ? 'selected' : ''}>
                                 <span className="news-facet__item-title">
-                                    {moment(term.key).format('MMMM YYYY')}&nbsp;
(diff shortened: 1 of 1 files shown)


src/encoded/static/components/news.js
```diff
@@ -242,7 +242,7 @@ const DateFacet = (props) => {
                         <li key={term.key} className="news-facet__item">
                             <a href={termHref} className={termFilter ? 'selected' : ''}>
                                 <span className="news-facet__item-title">
-                                    {moment(term.key).format('MMMM YYYY')}&nbsp;
+                                    {moment(term.key, ['MMMM, YYYY', 'YYYY-MM']).format('MMMM YYYY')}&nbsp;
```
emro: I think the brackets here are unnecessary (though it's also good as is).

emro: Ignore this comment, it's good as is!

forresttanaka:
Closing because I PRed ENCD-4662 which gets rid of the /news/ page altogether.

PR #2796, opened 2019-06-04
ENCD-4676-add-nagios-role-deployment

caseylitton:


src/encoded/commands/deploy.py
```diff
@@ -208,6 +208,9 @@ def tag_ec2_instance(instance, tag_data, elasticsearch, cluster_name):
     ]
     if elasticsearch == 'yes':
         tags.append({'Key': 'elasticsearch', 'Value': elasticsearch})
+        # This if for integration with nagios server.
+        # Only used on production.
```
caseylitton: Used on test also

PR #2803, opened 2019-06-10
ENCD-4106 Remove kyoto from aliases and dbxrefs

zmyers23:
Removed the enum 'kyoto' from the FlyDonor's dbxrefs and aliases.


src/encoded/schemas/changelogs/fly_donor.md
```diff
@@ -1,5 +1,9 @@
 ## Changelog for fly_donor.json
 
+### Schema version 10
+
+* The enum "kyoto" was removed from *aliases* and *dbxrefs* namespaces
```
yunhailuo: Need a note for being replaced by DGGR?

src/encoded/tests/test_upgrade_donor.py
```diff
@@ -227,3 +236,11 @@ def test_upgrade_mouse_donor_10_11(root, upgrader, mouse_donor_10):
     value = upgrader.upgrade(
         'mouse_donor', mouse_donor_10, current_version='10', target_version='11')
     assert 'parent_strains' not in value
+
+
+def test_upgrade_fly_donor_9_10(root, upgrader, fly_donor_9):
+    value = upgrader.upgrade('fly_donor', fly_donor_9, current_version='9', target_version='10')
+    assert 'kyoto' not in value['aliases'][0]
+    assert 'Kyoto' not in value['dbxrefs'][0]
+    assert value['schema_version'] == '10'
```
yunhailuo: And need to check if there DGGR with correct "ID".

src/encoded/upgrade/donor.py
```diff
@@ -217,3 +217,26 @@ def fly_worm_donor_8_9(value, system):
...
+
+
+@upgrade_step('fly_donor', '9', '10')
+def fly_donor_9_10(value, system):
+    # https://encodedcc.atlassian.net/browse/ENCD-4106
+    if 'aliases' in value:
+        for alias in value['aliases']:
+            if alias[0:6] == 'kyoto:':
+                new_alias = alias.replace(
+                    'kyoto:', 'dggr:')
```
yunhailuo: I will give `count=1`: https://docs.python.org/3.4/library/stdtypes.html#str.replace

I don't think there will be cases where "kyoto:" is in the "ID" but just to be safe and specific.

Same below.

src/encoded/upgrade/donor.py
```diff
@@ -226,7 +226,7 @@ def fly_donor_9_10(value, system):
         for alias in value['aliases']:
             if alias[0:6] == 'kyoto:':
                 new_alias = alias.replace(
-                    'kyoto:', 'dggr:', 1)
+                    'kyoto:', 'encode:', 1)
```
yunhailuo: I'd say do `encode:kyoto_` as Idan suggested so that he can later on check the demo and see if there is really nothing like that.

PR #2807, opened 2019-06-12
Encd-4659-batch-download-unit-test

zoldello:


src/encoded/tests/test_batch_download.py
```diff
@@ -135,6 +135,39 @@ def test_batch_download_lookup_column_value(lookup_column_value_item, lookup_col
...
+    ('/batch_download/type=Experiment&files.file_type=bed bed6+', 'http://localhost/metadata/type%3DExperiment%26files.file_type%3Dbed%20bed6%2B/metadata.tsv'),
+    ('/batch_download/type=Annotation&organism.scientific_name!=Homo sapiens&organism.scientific_name=Mus musculus', 'http://localhost/metadata/type%3DAnnotation%26organism.scientific_name%21%3DHomo%20sapiens%26organism.scientific_name%3DMus%20musculus/metadata.tsv\nhttp://localhost/files/ENCFF015OKV/@@download/ENCFF015OKV.bed.gz'),
+])
+def test_batch_download_files(testapp, workbook, test_url, expected):
+    response = testapp.get(test_url)
```
hitz: Do you think the 'batch_download' part should be abstracted from the test_url?  I can't decide.

src/encoded/tests/test_batch_download.py
```diff
@@ -135,6 +135,39 @@ def test_batch_download_lookup_column_value(lookup_column_value_item, lookup_col
...
...763145717\tMichael Snyder, Stanford\tfd942e78195f509cf0454ec7621eaa7a\t\thttp://localhost/files/ENCFF000LBB/@@download/ENCFF000LBB.csqual.gz\t\t\tApplied Biosystems SOLiD System 3 Plus\t\treleased\t\t\texperiment not submitted to GEO, mismatched status\tmissing documents, unreplicated experiment\t\nENCFF000LBA\tcsfasta\treads\tENCSR751YPU\tDNase-seq\tUBERON:0001891\tmidbrain\ttissue\tHomo sapiens\t\t\t\t\t\t\t\t\t\t\tDNA\t\t\t\t\tFalse\t2015-08-31\tRoadmap\t\t\t\t1\t1\t25\t\tpaired-ended\t1\tENCFF000LAZ\t\t2307361344\tMichael Snyder, Stanford\t51046aaf12d18c9c314d524f4599c98f\t\thttp://localhost/files/ENCFF000LBA/@@download/ENCFF000LBA.csfasta.gz\t\t\tApplied Biosystems SOLiD System 3 Plus\t\treleased\t\t\texperiment not submitted to GEO, mismatched status\tmissing documents, unreplicated experiment\t\nENCFF000LAZ\tcsfasta\treads\tENCSR751YPU\tDNase-seq\tUBERON:0001891\tmidbrain\ttissue\tHomo sapiens\t\t\t\t\t\t\t\t\t\t\tDNA\t\t\t\t\tFalse\t2015-08-31\tRoadmap\t\t\t\t1\t1\t25\t\tpaired-ended\t2\tENCFF000LBA\t\t2314662239\tMichael Snyder, Stanford\t20125f316dafa4daed390afae59e7bb2\t\thttp://localhost/files/ENCFF000LAZ/@@download/ENCFF000LAZ.csfasta.gz\t\t\tApplied Biosystems SOLiD S
```
hitz: the expected value is thorough but very hard to change.    Like you could test the first line last line and number of lines or something.

In fact this will break if the number of Experiments in the test set changes.

src/encoded/tests/test_batch_download.py
```diff
@@ -146,6 +179,22 @@ def test_files_prop_param_list(test_input, expected):
     assert test_input == expected
 
 
+@pytest.mark.parametrize('test_url,expected', [
```
hitz: same, this test is very fragile in terms of the test set, which changes slowly but it does change.

src/encoded/tests/test_batch_download.py
zoldello: Thinking about it, yes. I can make a dictionary of tuples to extract away test_url and expected; and place them in a separate file. It makes the test more readable and hide the huge text.

src/encoded/tests/test_batch_download.py
zoldello: Before I do that, how does this seem to you- Unit test supposed to be brittle. If not, they may not catch subtle changes that ought to be caught. That logic is fine to me, but is it fine to the rest of the world? :)

src/encoded/tests/test_batch_download.py
zoldello: Same logic as before- Do you think it is "pythonic" to have fragile test?

src/encoded/tests/test_batch_download.py
```diff
@@ -135,6 +423,34 @@ def test_batch_download_lookup_column_value(lookup_column_value_item, lookup_col
...
+])
+def test_batch_download_meta_files(testapp, workbook, test_url, expected_row_1, expected_row_n):
+    # compare only ids, as it is the only essential field
+    response = testapp.get(test_url)
+    target = response.body.decode('utf-8').strip().split('\n')
+    assert expected_row_1[:11] == target[1][:11]
```
hitz: this should be commented.

PR #2808, opened 2019-06-13
ENCD-4471-add-prod-id-source-treatment

meenakshikagda:


src/encoded/tests/test_schema_treatment.py
```diff
@@ -0,0 +1,16 @@
+import pytest
+
+@pytest.fixture
+def test_treatment_with_bad_prod (testapp,treatment, source):
```
gabdank: @meenakshikagda I don't think you need a fixture. I would suggest in the function to use "treatment" and modify it by patching in "product_id" and "source" validating that the patching works.

gabdank (changes requested):

gabdank (approved):

gabdank (approved):

PR #2810, opened 2019-06-14
ENCD-4671 port tests circleci

paul-sud:


p-sud:
Two commits in here: one for adding config using oracle jdk, then the second that replaces oracle jdk with openjdk

.circleci/config.yml
```diff
@@ -0,0 +1,92 @@
+version: 2.1
+
+orbs:
+    browser-tools: circleci/browser-tools@0.1.0
+
+executors:
+  encoded-executor:
+    docker:
+      - image: circleci/buildpack-deps:trusty
```
hitz: just to check, I think this is 14.04 what we run in production... it's hard to say because the ami cloud-config uses is pretty old (again).  @clitton?

README.md
```diff
@@ -2,7 +2,7 @@
 -------------------------------------------------------
 
 
-![Build status](https://travis-ci.org/ENCODE-DCC/encoded.svg?branch=master)
+[![CircleCI](https://circleci.com/gh/ENCODE-DCC/encoded/tree/dev.svg?style=svg)](https://circleci.com/gh/ENCODE-DCC/encoded/tree/dev)
```
hitz: I wonder if there is a way to tie this badge to the current branch?  We don't really want to edit the README twice every time.

hitz (approved):
I don't think my comments are relevant to ENCD-4671

README.md
p-sud: You'd probably have to interpolate the branch into the readme

.circleci/config.yml
caseylitton: I think trusty is 14.04.X  And is updated the x as needed.  Now its 14.04.06.

README.md
caseylitton: we have a cloud config update coming that builds cloud configs using python string interpolation.  If we include the travis/circle config in that process then we can do stuff like this.

PR #2812, opened 2019-06-14
ENCD-4662 Experiment series

yunhailuo:


src/encoded/schemas/experiment_series.json
```diff
@@ -0,0 +1,211 @@
...
+    "mixinProperties": [
+        { "$ref": "mixins.json#/schema_version" },
+        { "$ref": "mixins.json#/uuid" },
+        { "$ref": "mixins.json#/accession" },
+        { "$ref": "mixins.json#/aliases" },
+        { "$ref": "mixins.json#/attribution" },
+        { "$ref": "mixins.json#/submitted" },
+        { "$ref": "mixins.json#/notes" },
+        { "$ref": "mixins.json#/submitter_comment" },
+        { "$ref": "series.json#/properties" }
```
hitz: missing dataset.json/#properties?

src/encoded/schemas/experiment_series.json
```diff
@@ -0,0 +1,211 @@
...
+                    }
+                }
+            ]
+        }
+    },
+    "properties": {
+        "schema_version": {
+            "default": "1"
+        },
+        "accession": {
```
hitz: this is in dataset.json

src/encoded/schemas/experiment_series.json
```diff
@@ -0,0 +1,211 @@
...
+            "accessionType": "SR"
+        },
+        "award": {
+            "default": "5854a1dc-4c50-44b3-ba58-0a6f4f399297",
```
hitz: is this normally how we do this?

src/encoded/schemas/experiment_series.json
```diff
@@ -0,0 +1,211 @@
...
+            "default": "cb0ef1f6-3bd3-4000-8636-1c5b9f7000dc",
+            "linkEnum": [
+                "cb0ef1f6-3bd3-4000-8636-1c5b9f7000dc"
+            ],
+            "permission": "import_items"
+        },
+        "related_datasets": {
+            "minItems": 1
+        },
+        "status": {
```
hitz: also in dataset

src/encoded/schemas/experiment_series.json
```diff
@@ -0,0 +1,211 @@
...
+            "title": "Biosample classification"
+        },
+        "biosample_ontology.organ_slims": {
+            "title": "Organ"
+        },
+        "biosample_ontology.term_name": {
+            "title": "Biosample term name"
+        },
+        "contributing_awards.project": {
+            "title": "Project"
```
hitz: Remove from facets

src/encoded/schemas/experiment_series.json
```diff
@@ -0,0 +1,211 @@
...
+        },
+        "contributing_awards.rfa": {
```
hitz: REmove from facets

src/encoded/schemas/experiment_series.json
```diff
@@ -0,0 +1,211 @@
...
+            "title": "RFA"
+        },
+        "contributors.title": {
+            "title": "Lab"
```
hitz: I think we might want to remove this one as well, but if we don't the title should be "Contributors"

src/encoded/schemas/experiment_series.json
```diff
@@ -0,0 +1,211 @@
...
+            "title": "Cell"
+        },
+        "biosample_ontology.classification": {
+            "title": "Biosample classification"
+        },
+        "biosample_ontology.organ_slims": {
+            "title": "Organ"
+        },
+        "biosample_ontology.term_name": {
+            "title": "Biosample term name"
```
hitz: Should be
`        "biosample_ontology.term_name" : {
            "title": "Biosample term name",
            "type": "typeahead",
            "length": "long"
        },
`

src/encoded/schemas/experiment_series.json
```diff
@@ -0,0 +1,211 @@
...
+            "title": "Project"
+        },
+        "contributing_awards.rfa": {
+            "title": "RFA"
+        },
+        "contributors.title": {
+            "title": "Lab"
+        },
+        "date_released": {
+            "title": "Month released"
```
hitz: This should be "Date Released"

src/encoded/schemas/experiment_series.json
```diff
@@ -0,0 +1,211 @@
...
+        "organism.scientific_name": {
+            "title": "Organism"
+        },
+        "status": {
+            "title": "Status"
+        },
+        "target.investigated_as": {
+            "title": "Target category"
+        },
+        "target.label": {
```
hitz: Should be:
        "target.label": {
            "title": "Target of assay",
            "type": "typeahead",
            "length": "long"
        },

src/encoded/schemas/experiment_series.json
```diff
@@ -0,0 +1,211 @@
...
+            "title": "Assay name"
+        },
+        "assay_title": {
+            "title": "Assay title"
+        },
+        "biosample_ontology.term_name": {
+            "title": "Biosample term name"
+        },
+        "contributors.title": {
+            "title": "Lab"
```
hitz: Should be "Contributing Labs"

src/encoded/types/dataset.py
```diff
@@ -666,3 +666,69 @@ class ReferenceEpigenome(Series):
...
+    item_type = 'experiment_series'
+    schema = load_schema('encoded:schemas/experiment_series.json')
+    name_key = 'accession'
+    embedded = [
+        'biosample_ontology',
+        'contributing_awards',
+        'contributors',
+        'organism',
+        'related_datasets.lab',
+        'related_datasets.replicates.library.biosample',
```
hitz: Do we need this and biosample_ontology?  Probably but I am not sure

src/encoded/schemas/experiment_series.json
yunhailuo: That's on purpose to not have other properties. Should I have them, like `alternate_accessions`,  `dbxrefs`, `internal_tags`, and have `status` defined that way?

src/encoded/schemas/experiment_series.json
yunhailuo: No. But we do it in dependency: https://github.com/ENCODE-DCC/encoded/blob/dev/src/encoded/schemas/biosample.json#L50-L65
The reason is I think we can/should restrict it to just one value. But I could be wrong about it.

src/encoded/schemas/experiment_series.json
yunhailuo: Will users want to filter this series based on who's in this series?

src/encoded/types/dataset.py
yunhailuo: This is just for the following two in columns. I don't see they are very important.
```
        "related_datasets.replicates.library.biosample.age_display": {
            "title": "Biosample age",
            "type": "string"
        },
        "related_datasets.replicates.library.biosample.life_stage": {
            "title": "Life stage",
            "type": "array"
        },
```

src/encoded/schemas/experiment_series.json
```diff
@@ -0,0 +1,211 @@
...
+        "contributors.institute_name": 1.0,
+        "contributors.institute_label": 1.0,
+        "contributors.title": 1.0,
+        "notes": 1.0,
+        "organism.name": 1.0,
+        "organism.scientific_name": 1.0,
+        "organism.taxon_id": 1.0,
+        "contributors.institute_name": 1.0,
+        "contributors.institute_label": 1.0,
+        "contributors.title": 1.0,
```
forresttanaka: These three "contributors" lines repeat a few lines above

src/encoded/schemas/experiment_series.json
```diff
@@ -0,0 +1,211 @@
...
+            "enum" : [
+                "in progress",
+                "submitted",
+                "released",
+                "archived",
+                "deleted",
+                "replaced",
+                "revoked"
+            ],
+            "permission": "import_items"
```
forresttanaka: This repeats the "permission" line a few lines above.

src/encoded/schemas/experiment_series.json
yunhailuo: Repeat?

src/encoded/schemas/experiment_series.json
yunhailuo: Repeat? This one is for status. Others are for other properties.

forresttanaka: This is a repeat of "permission": "import_items" line a few lines above. VSCode's JSON linter marks it.

src/encoded/schemas/experiment_series.json
forresttanaka: "contributors.institute_name": 1.0, and the next two lines are already there a few lines above. VSCode marks these.

src/encoded/types/dataset.py
hitz: no we need those for batch_download, I forgot.

src/encoded/schemas/experiment_series.json
hitz: yes they are in there twice.

src/encoded/schemas/experiment_series.json
yunhailuo: Sorry. Saw that. Thanks.

src/encoded/schemas/experiment_series.json
yunhailuo: Sorry. Saw that. Thanks.

src/encoded/schemas/experiment_series.json
hitz: @gabdank do we need to actually set the default award in the schema?...   I don't think we do.  It's not a strict requirement that we own them all, just that we will likely do so.

src/encoded/schemas/experiment_series.json
forresttanaka: I agree with @hitz. Since ExperimentSeries is a kind of dataset, I was already surprised to find it didn’t inherit from the dataset schema, even if some properties aren’t used.

src/encoded/schemas/experiment_series.json
gabdank: I don't think we should set up a default award in the schema. The reason it is used in biosample - was to restrict model organism properties to non-human and not mouse organisms.

src/encoded/static/components/experiment_series.js
```diff
@@ -0,0 +1,356 @@
...
+                                        <dd>{context.assay_term_name.join(', ')}</dd>
+                                    </div>
+                                : null}
+
+                                {terms.length > 0 || speciesRender ?
+                                    <div data-test="biosamplesummary">
+                                        <dt>Biosample summary</dt>
+                                        <dd>
+                                            {terms.length > 0 ? <span>{terms.join(' and ')} </span> : null}
+                                            {speciesRender ? <span>({speciesRender})</span> : null}
```
emro: I'm not sure that this wrapper `span` is necessary, since the component itself is wrapped in a `span` but also it is fine and then it matches the syntax above. Just a thought because you end up with a lot of nested `span`s.

forresttanaka:
Yeah, I’m not sure Wranglers would have an issue with "Experiments in experiment series” since it represents the type of objects within the new "experiment series" object.

src/encoded/static/components/experiment_series.js
forresttanaka: Not sure how to avoid the wrapper `<span>`. I changed it to:

```
{speciesRender ? { speciesRender } : null}
```

This was the only syntax I could make that would avoid an outright ESLint error, but then I crashed the server with a run-time error.

src/encoded/loadxl.py
```diff
@@ -50,6 +50,7 @@
     'organism_development_series',
     'replication_timing_series',
     'aggregate_series',
+    'experiment_series',
```
caseylitton: Flag to squash commits into the two ticket number

PR #2816, opened 2019-06-17
ENCD-4415 Add GM perturbation property

jenjou:


jenjou:
Renamed property and added new test_types_genetic_modification.py file.

PR #2820, opened 2019-06-19
ENCD-4511-make-pool-table

zoldello:


zoldello:
@forresttanaka Great catches. Fixed.

PR #2824, opened 2019-06-21
ENCD-4703-test-submitter-user-circleci-fix

p-sud:


forresttanaka:
In case it “counts” and you need to save time, I had two successful runs in a row with this fix in my React upgrade branch (and I’ll back out that change; I just wanted to test it in a branch that I know had consistent failures). But still, the more successful runs you have time for, the better.

p-sud:
Fixed the PR name, going to rerun tests. Thanks for the extra datapoints @forresttanaka

p-sud:
Bdd tests passed three consecutive runs

PR #2825, opened 2019-06-21
ENCD-4704 Fix frozen column for Safari

forresttanaka:
Safari requires a browser prefix for position:sticky. Also a couple other small adjustments to fix small Safari-specific rendering issues. Those fixes do not affect other browsers.


forresttanaka:
For v87rc2

PR #2829, opened 2019-06-24
ENCD-4699 Fix Javascript crash display series file tables

forresttanaka:
FileTable can be called from FileGalleryRenderer for experiments and annotations, but it can also be called from DatasetFiles where visualizations aren’t relevant to objects using DatasetFiles so it passes nothing for the required browserOptions.browserFileSelectHandler, and it fails prop validation. This change removes this property as a _requirement_, and checks whether it exists, and hides the Visualize column from the table if it doesn’t.

### DIFF
--- a/src/encoded/static/components/filegallery.js
+++ b/src/encoded/static/components/filegallery.js
@@ -179,6 +179,7 @@ export class FileTable extends React.Component {
             // TODO: (1)Move .hide to FileTable.procTableColumns declaration
             // (2) move showReplicateNumber to meta
             FileTable.procTableColumns.biological_replicates.hide = () => !showReplicateNumber;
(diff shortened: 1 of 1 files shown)


zoldello:
Note: The bdd failure is a known issue. It will be ignored during code review.

zoldello:
@forresttanaka- This is minor and can be ignored if need be. You solved the problem- Make something not required, not required.  But in addition, you added an unrelated solution to a different problem in line 182. Will that impact another ticket?

forresttanaka:
We now know with some datasets (like Series) we don’t have a visualization property, so we shouldn‘t display a property in the table that only relates to visualizations when we display objects of that type.

PR #2830, opened 2019-06-24
[EPIC] ENCD-4716 Update cloud init deployment

caseylitton:
pr for the cloud config update.  

Probably easier to explain in person, but if you do look, start with the 
 _test_cloud_configs function in the deploy script.   

There are a two kinds of deployments.  demo and frontend.  The frontend requires an es cluster to be built.  There are two kinds of es clusters supported.  One has an elected master/head node and the other requires a separate head node to be deployed along with the 5 data clusters.

This test compares the prebuilt configs( in encoded/cloud-config/prebuilt-config-yamls) with the ones that get built on the fly(in encoded/cloud-config/config-build-files).  The diff is clean at the moment so the build process works.    HOWEVER I need to test the different builds and double check all the lines from the current cloud configs were copied over correctly.

The difference between the prebuilt configs and the current v86 cloud configs is that the runcmd has been organized into scripts.  Both the prebuilt and on the fly build configs use the same deploy  run scripts, encoded/cloud-config/deploy-run-scripts.  These run commands could be used to install from a prebuilt ami although that is not part of this feature.


hitz:
Oracle JDK download failed.   Maybe we should switch to OpenJDK?

src/encoded/commands/deploy.py
```diff
@@ -625,10 +604,7 @@ def hostname(value):
     parser.add_argument('--redis-port', default=6379, help="Redis Port.")
     parser.add_argument('--set-region-index-to', type=check_region_index,
                         help="Override region index in yaml to 'True' or 'False'")
-    parser.add_argument('--spot-instance', action='store_true', help="Launch as spot instance")
-    parser.add_argument('--spot-price', default='0.70', help="Set price or keep default price of 0.70")
-    parser.add_argument('--teardown-cluster', default=None,
-                        help="Takes down all the cluster launched from the branch")
+    parser.add_argument('--use-local-config', action='store_true', help="Use local config file in current directory")
```
hitz: Is the normal use_case to use the local ones or to build new ones?

src/encoded/commands/deploy.py
```diff
@@ -441,119 +294,243 @@ def _wait_and_tag_instances(main_args, run_args, instances_tag_data, instances,
...
+
+
 def main():
     # Gather Info
     main_args = parse_args()
+    if main_args.dry_run_cc:
+        # Testing Only
+        if _test_cloud_configs(main_args.branch):
+            sys.exit(0)
+        sys.exit(5)
```
hitz: oooh exit code 5... fancy.  Not sure it has meaning to anyone though.

src/encoded/commands/deploy.py
```diff
@@ -602,13 +579,15 @@ def hostname(value):
...
     parser.add_argument('-n', '--name', default=None, type=hostname, help="Instance name")
     parser.add_argument('--dry-run-aws', action='store_true', help="Abort before ec2 requests.")
+    parser.add_argument('--dry-run-cc', action='store_true', help="Abort after building cloud config.")
     parser.add_argument('--single-data-master', action='store_true',
             help="Create a single data master node.")
-    parser.add_argument('--check-price', action='store_true', help="Check price on spot instances")
     parser.add_argument('--cluster-name', default=None, help="Name of the cluster")
     parser.add_argument('--cluster-size', default=2, help="Elasticsearch cluster size")
+    parser.add_argument('--conf-dir', default='./cloud-config', help="Location of cloud build config")
     parser.add_argument('--elasticsearch', default=None, help="Launch an Elasticsearch instance")
```
hitz: Do we want to change this to a boolean now? instead of a text string == "yes"?

src/encoded/commands/deploy.py
```diff
@@ -441,119 +294,243 @@ def _wait_and_tag_instances(main_args, run_args, instances_tag_data, instances,
...
+            "Name": run_args['iam_role'],
+        },
+        Placement={
+            'AvailabilityZone': main_args.availability_zone,
+        },
+    )
+    _wait_and_tag_instances(main_args, run_args, instances_tag_data, instances)
+    if 'master_user_data' in run_args and main_args.single_data_master:
+        # ES MASTER instance when deploying elasticsearch data clusters
+        if run_args['master_user_data'] and run_args['count'] > 1 and main_args.elasticsearch == 'yes':
```
hitz: this is the independent master/

cloud-config/deploy-run-scripts/es-wait-install.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
caseylitton: there are two kinds of cluster deployments.  Both have a frontend and either
 
1. wait es cluster, that waits for a head es node.
2. elect es cluster, that elects a head es node.

Naming could be improved but would need be holistic.

cloud-config/deploy-run-scripts/es-head-install.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+# Setup elasticsearch data node config
+# root user
+# apt deps:
+#   java
+#   elasticsearch with apt_source and key
```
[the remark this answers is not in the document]
caseylitton: You're right, The difference between the files is only a variable, but there is not a clear location to put the variable.  There are a few layers.  We can optimize this once we're more clear on the structure.

src/encoded/commands/deploy.py
```diff
@@ -271,10 +122,16 @@ def get_user_data(commit, config_file, data_insert, main_args):
     )
     data_insert['S3_AUTH_KEYS'] = auth_keys_dir
     data_insert['REDIS_PORT'] = main_args.redis_port
-    user_data = config_template % data_insert
+    user_data = config_yaml % data_insert
```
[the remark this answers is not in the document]
caseylitton: yes, we've always done this in the deploy script.   Python string replace.

src/encoded/commands/deploy.py
caseylitton: build new ones on the fly is the default.  We could go either way at this point, but I have some rough tests to diff the build and local.  They come back clean.

src/encoded/commands/deploy.py
caseylitton: nope, I just picked different numbers for different failures.

src/encoded/commands/deploy.py
caseylitton: yes, but we haven't used or tested it in a while.  so I'll see if I can build one.

cloud-config/deploy-run-scripts/es-elect-install.sh
```diff
@@ -0,0 +1,62 @@
...
+# apt deps:
+#   java
+#   elasticsearch with apt_source and key
+
+CLUSTER_NAME=$1
+ES_MASTER=$2
+ES_DATA=$3
+MIN_MASTER_NODES=$4
+
+# Add options elasticsearch.yml
```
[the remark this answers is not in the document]
caseylitton: the comment hash?  In vim its easy.

src/encoded/commands/deploy.py
caseylitton: oh the elasticsearch line you mean, yea we can do that finally.

cloud-config/deploy-run-scripts/es-elect-install.sh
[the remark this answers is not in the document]
hitz: The "echo a line of code" >> file 
for every line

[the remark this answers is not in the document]
caseylitton: yes, it would be better to make the files directly.

https://encodedcc.atlassian.net/browse/ENCD-4714

cloud-config/deploy-run-scripts/es-head-install.sh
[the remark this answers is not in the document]
caseylitton: Went over the structure with Keenan and it seemed to make sense.  I'll merge these files into one with a new argument.

PR #2836, opened 2019-06-26
ENCD-4685-matrix-vertical-scroll

zoldello:


zoldello:
Want to see how UI looks on a demo before requesting code review.

zoldello:
@emro I had the same in the back of my mind. Since you mentioned it, it is important. I will experiment and get back to you.

src/encoded/static/components/matrix.js
```diff
@@ -449,7 +449,16 @@ class MatrixPresentation extends React.Component {
...
             }
 
-            // Category does exist in array, so remove it.
+            // Category does exist in array
+            // Move close to header
+            const header = document.querySelector(`#${category.replace(/\s/g, '_')}`);
+            header.scrollIntoView({
+                behavior: 'smooth',
+                block: 'end',
+                inline: 'nearest',
```
emro: Is `inline: 'nearest'` important? It doesn't seem like there is any horizontal scrolling here.

zoldello: No, also, "inline" is set to "nearest" by default. I removed it.

src/encoded/static/components/matrix.js
```diff
@@ -170,6 +170,28 @@ const analyzeSubCategoryData = (subCategoryData, columnCategoryType) => {
...
+ *
+ * @returns height of navbar on first call and restores same value until page is garbage collected
+ */
+const getNavbarHeight = () => {
+    if (_navbarHeight === null) {
+        const navbar = document.querySelector('#navbar');
+        _navbarHeight = navbar ? navbar.getBoundingClientRect().height : 0;
+    }
+    return _navbarHeight;
+};
```
forresttanaka: I’m _personally_ not a fan of using IDs to get the height of React-rendered elements, but I won’t gate this review on that because this is way simpler than the way I would have done it, so I think this method has a valid argument in its favor. Since we have the `<Navigation>` component which can find out the height of the menu using React refs, and can feed that back to `<App>` which can then pass this information in React context (maybe along with other info we don’t yet use, like footer height, navigation color, cart menu visible, etc.). Maybe that’s a mechanism best done in another ticket.

PR #2837, opened 2019-06-26
ENCD-4705-test-submitter-user-bdd-test-fix

zoldello:


zoldello:
No longer needed; issue fixed by Paul S in a prior release.

PR #2841, opened 2019-07-02
ENCD-3567 Upgrade React JS 15.6.0

forresttanaka:
No changes from the first PR — just rebase to v87.

### DIFF
--- a/src/encoded/static/components/genetic_modification.js
+++ b/src/encoded/static/components/genetic_modification.js
@@ -22,7 +22,7 @@ import { BiosampleTable, DonorTable } from './typeutils';
 const IntroducedTags = (props) => {
     const { geneticModification } = props;
 
-    if (geneticModification.introduced_tags && geneticModification.introduced_tags.length) {
+    if (geneticModification.introduced_tags && geneticModification.introduced_tags.length > 0) {
         // Generate an array of React components, each containing one epitope_tags display from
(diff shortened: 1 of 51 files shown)


src/encoded/static/components/genetic_modification.js
```diff
@@ -642,13 +642,28 @@ GeneticModificationComponent.propTypes = {
...
     auditDetail: PropTypes.func.isRequired, // Audit HOC function to display audit details
+    session: PropTypes.object, // Login information from <App>
+};
+
+GeneticModificationComponent.defaultProps = {
+    session: null,
+};
+
+const GeneticModificationInternal = (props, reactContext) => (
+    <GeneticModificationComponent {...props} session={reactContext.session} />
```
emro: This looks good, but I was just curious - is there a reason you didn't add `reactContext.session` to `GeneticModificationComponent` and instead made a wrapper?

forresttanaka: It’s because of a change to React 16 in which componentDidUpdate doesn’t receive information about changing React context anymore. The old componentDidUpdate received a third parameter for the previous React context which I used to see if login information changed. Without that parameter, I had to convert context.session from a React context property to a regular property so componentDidMount could compare that instead, so that new wrapper is there to do that conversion.

emro: That makes sense! Thank you!

PR #2842, opened 2019-07-02
ENCD-4697 add biosample summary to experiment series

yunhailuo:


src/encoded/static/components/experiment_series.js
```diff
@@ -171,12 +169,13 @@ const ExperimentSeriesComponent = (props, reactContext) => {
...
 
-                                {terms.length > 0 || speciesRender ?
+                                {(context.biosample_summary && context.biosample_summary.length > 0) || speciesRender ?
                                     <div data-test="biosamplesummary">
                                         <dt>Biosample summary</dt>
                                         <dd>
-                                            {terms.length > 0 ? <span>{terms.join(' and ')} </span> : null}
-                                            {speciesRender ? <span>({speciesRender})</span> : null}
+                                            {speciesRender ? <span>{speciesRender}</span> : null}
+                                            &nbsp;
```
forresttanaka: This inserts a non-breaking space even if `speciesRender` is null/undefined. That would be visible as an odd space before the `biosample_summary` in that case. Like before, you can just put a space before the previous `</span>` and it should work in all cases.

biosample summary

yunhailuo: Thanks

yunhailuo:
One test added.

src/encoded/tests/test_types_experiment_series.py
```diff
@@ -0,0 +1,110 @@
...
+    res = testapp.get(base_experiment_series['@id']+'@@index-data')
+    assert res.json['object']['biosample_summary'] == [
+        'S2R+ nuclear fraction treated with ethanol'
+    ]
+    testapp.patch_json(
+        base_experiment_series['@id'],
+        {'related_datasets': [experiment_1['@id'], experiment_2['@id']]}
+    )
+    res = testapp.get(base_experiment_series['@id']+'@@index-data')
+    print(res.json['object']['biosample_summary'])
```
gabdank: I think @yunhailuo you need to remove the print statement

yunhailuo: Sorry, that's really bad. Fixed.

PR #2843, opened 2019-07-02
ENCD-4701 Update STAR quality metric for chimeric reads

yunhailuo:


yunhailuo:
> @yunhailuo please add minimum zero value to "number of chimeric reads"
> do you think there is a way to check percentage in terms of formatting? right now I can put any string there...which is not the best setup.

Patched for both.

PR #2849, opened 2019-07-09
ENCD-4608 ENCODE4 antibody review

yunhailuo:


src/encoded/schemas/biosample_characterization.json
```diff
@@ -26,6 +26,15 @@
         "schema_version": {
             "default": "11"
         },
+        "antibodies": {
```
gabdank: Why does it have to be an array? why one biosample characterization will have multiple antibodies?

yunhailuo: In principle, a blot can have multiple antibodies marking different proteins...

src/encoded/static/components/typeutils.js
```diff
@@ -460,3 +461,84 @@ ControllingExperiments.defaultProps = {
...
+
+/**
+ * Display a count of biosample characterizations in the footer, with a link to the corresponding search if needed.
+ */
+const BiosampleCharacterizationTableFooter = ({ items, total, url }) => (
+    <div>
+        <span>Displaying {items.length} of {total} </span>
+        {items.length < total ? <a className="btn btn-info btn-xs pull-right" href={url}>View all</a> : null}
+    </div>
+);
```
forresttanaka: React 16 has a new feature to replaced these extra `<div>`s — only needed for JSX parsing — with `<React.Fragment>`...`</React.Fragment>`. These serve the same purpose as the `<div>`s but don’t generate any actual HTML elements.

src/encoded/static/components/typeutils.js
```diff
@@ -460,3 +461,84 @@ ControllingExperiments.defaultProps = {
...
+    // clone the array with just that specified number of elements.
+    const biosampleCharacterizations = limit > 0 && limit < items.length ? items.slice(0, limit) : items;
+
+    return (
+        <div>
+            <SortTablePanel title={title}>
+                <SortTable list={biosampleCharacterizations} columns={biosampleCharacterizationTableColumns} footer={<BiosampleCharacterizationTableFooter items={biosampleCharacterizations} total={total} url={url} />} />
+            </SortTablePanel>
+        </div>
+    );
```
forresttanaka: We don’t need these extra `<div>`s either, but you can just delete them. The `<SortTablePanel>` component does the job just fine.

yunhailuo:
> All the Javascript looks good! I just have a couple small changes as we try to reduce the number of now-unneeded extra `<div>`s. I’m _assuming_ Idan reviewed the Python, but let me know if I should.

Thanks. Both suggestions on `<div>` tags are great. Just JS is fine.

PR #2850, opened 2019-07-15
ENCD-3904-remove-quotation-marks

zoldello:


buildout.cfg
```diff
@@ -215,4 +215,4 @@ recipe = collective.recipe.cmd
 on_install = true
 on_update = true
 cmds =
-    curl -o annotations.json https://s3-us-west-1.amazonaws.com/encoded-build/annotations/annotations_2017_10_10.json
+    curl -o annotations.json https://s3-us-west-1.amazonaws.com/encoded-build/annotations/annotations_2019_07_11.json
```
caseylitton: This doesn't look related to the ticket.  Why is an annotations update in this pr?

zoldello: @caseylitton I updated the name to signify the file is updated (despite time stamp); to follow the apparent convention. I can keep the current name if you like.

hitz: @caseylitton the bug that appears in ENCD-3904 is caused by extra " in the file that is created by the script.  So this PR does both; fixes the script (writes a new file) and corrects the download of the fixed file.

caseylitton: Are you sure?  The two annotation files are the exact same

annotations_2017_10_10.json
annotations_2019_07_11.json

hitz: @caseylitton it turns out the old file was "illegally" updated recently... we will restore it.

src/encoded/commands/generate_annotations.py
```diff
@@ -85,7 +85,7 @@ def human_single_annotation(r):
 
         # Adding gene synonyms to autocomplete
         if r['Synonyms'] is not None and r['Synonyms'] != '':
-            synonyms = [x.strip(' ') + species for x in r['Synonyms'].split(',')]
+            synonyms = [x.strip(' ').strip('\"') + species for x in r['Synonyms'].split(',')]
```
caseylitton: I thought a made a comment on here earlier.  

This is not really a bug fix.  The slash-quote issue is not in the original tsv so we're adding it in the generate annotations process.  This only undoes the original issue, which is probably the quoted lists in the original tsv being decoded here.

https://github.com/ENCODE-DCC/encoded/blob/dev/src/encoded/commands/generate_annotations.py#L190

Did you look into fixing this issue at the source, get_rows_from_file or human_annotations?

buildout.cfg
zoldello: I updated "annotations_2017_10_10.json" to be the same content as prior. I could not revert as that was no enabled before.

zoldello:
@caseylitton Are the current changes acceptable?

caseylitton:
@zoldello It doesn't look like both of my comments have been answered/considered.

src/encoded/commands/generate_annotations.py
zoldello: @caseylitton I missed your comment about "address the root cause." Excellent point. I am looking into now.

zoldello: In this file (generate_annotations.py), line 191 and 192, a double quote-string is at the beginning and end of the Synonym (like- '"apple"'). So, what I did was remove them and added a comment to try to make that clear.

hitz: actually I think this fix at this line IS correct because he has to parse strings that look like "A, B, C" into [A,B,C].  So he needs to strip the \"

However field 3, "Approved Name" ALSO has (sometimes extraneous \" in it.  So either every field has to be stripped or Name has to be stripped as well.

zoldello: Addressed. Stripped double quotes from all fields.

src/encoded/commands/generate_annotations.py
```diff
@@ -189,7 +189,9 @@ def get_rows_from_file(file_name, row_delimiter):
     response = requests.get(file_name)
     rows = response.content.decode('utf-8').split(row_delimiter)
     header = rows[0].split('\t')
-    zipped_rows = [dict(zip(header, row.split('\t'))) for row in rows[1:]]
+    # remove the leading and ending double-quote string
```
caseylitton: Sorry I didn't explain throughly when we talked.  I don't think this is what we're doing here. 

The original tsv file
```
https://www.encodeproject.org/files/ENCFF277WZC/@@download/ENCFF277WZC.tsv
```
 has quoted lists like
```
"BNaC2, hBNaC2"
```
The \\" gets added when we create the annotations json.  The r.strip('\\"') removes a literal \\".  Not the quotes around the list.  

This is what I wanted you to check.  Where does the \\" get added as WE create the annotations json.  Trace the code from main(), but I expect the issue is in 'get_rows_from_file' function.  It could happen in the get request on line 189 or in the decode on line 190.   Please find out where the slash gets added.  Once we determine that we can decide where to fix it, or to add a comment explaining the issue.

caseylitton: I'm looking into this a little more, but we shouldn't change the code without uploading the file again.  In any case please return the feature fix the original location in synonyms.  I just wanted a more clear comment OR a fix in a better location.  Not both.

caseylitton: @zoldello Per convo, I thought an actual slash quote was inserted into the json file during creation.  I'll see if I can find out why I assumed that.

hitz: no they are in the source file, I checked with @zoldello: (snippet)
`HGNC:26859	RBMXL3	"RNA binding motif protein, X-linked like 3"	Approved	CXorf55	FLJ40249	Xq23	AK097568	139804	ENSG00000175718	NM_001145346
HGNC:2686	DAZLA~withdrawn	"symbol withdrawn, see DAZL"	Symbol Withdrawn							
HGNC:26861	C2orf73	chromosome 2 open reading frame 73	Approved		FLJ40298	2p16.2	"BC031669, AK097617"	129852	ENSG00000177994	NM_001100396
HGNC:26862	HSF5	heat shock transcription factor 5	Approved		FLJ40311	17q23.2	BC033020	124535		XM_064190
`

hitz: sometimes column 3 has quotes (this was missed in first version).  Column 6 often has "" around a list of items that is split.  Both cases the " should be stripped.

PR #2851, opened 2019-07-15
SNO-106-fix-spec-char-500-error

zoldello:


zoldello:
@keenangraham Do you have any thing you need me to update with this PR?

PR #2855, opened 2019-07-18
ENCD-4724 md5 uniqueness adjustment

gabdank:


src/encoded/types/file.py
```diff
@@ -139,7 +139,9 @@ def __name__(self):
     def unique_keys(self, properties):
         keys = super(File, self).unique_keys(properties)
         if properties.get('status') != 'replaced':
-            if 'md5sum' in properties:
+            file_output_category = self.schema['output_type_output_category'].get(properties.get('output_type'))
```
keenangraham: Are you sure you can't properties.get('output_category')?

src/encoded/types/file.py
```diff
@@ -139,7 +139,9 @@ def __name__(self):
...
+            if ((properties.get('lab') != 'a558111b-4c50-4b2e-9de8-73fd8fd3a67d' or
```
keenangraham: It would be nice if this were wrapped in function with an informative named that returned True/False. Something like `if is_md5sum_contrained_file():` You would probably have to pass in properties and output_category etc. but would allow for future extension. It would also be nice if you defined constant at the top `CHERRY_LAB_UUID = 'a558111b-4c50-4b2e-9de8-73fd8fd3a67d'` (or whatever it is) and use the variable name instead of the uuid so it's clearer what it is.

src/encoded/types/file.py
gabdank: replaced by the calculated 'output_category'

src/encoded/types/file.py
gabdank: created external function following your suggestion

gabdank:
@keenangraham I have addressed your PR review comments

src/encoded/types/file.py
```diff
@@ -66,6 +66,14 @@ def property_closure(request, propname, root_uuid):
     return seen
 
 
+def is_md5sum_constrained_file(file_properties, unique_md5_output_categories, uniform_processing_pipeline_uuid):
+    if ((file_properties.get('lab') != uniform_processing_pipeline_uuid or
```
keenangraham: This could still be cleaned up a bit:

```
def is_md5sum_constrained():
    conditions = [
        properties.get('lab') != ENCODE_PROCESSING_PIPELINE_UUID,
        properties.get('output_category') == RAW_DATA
    ]
    if any(conditions):
        return True
    return False
```

src/encoded/types/file.py
```diff
@@ -66,6 +66,14 @@ def property_closure(request, propname, root_uuid):
     return seen
 
 
+def is_md5sum_constrained_file(file_properties, unique_md5_output_categories, uniform_processing_pipeline_uuid):
```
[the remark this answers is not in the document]
gabdank: done

src/encoded/types/file.py
```diff
@@ -135,11 +143,15 @@ def __name__(self):
         if properties.get('status') == 'replaced':
             return self.uuid
         return properties.get(self.name_key, None) or self.uuid
-
+    
     def unique_keys(self, properties):
         keys = super(File, self).unique_keys(properties)
+        ENCODE_PROCESSING_PIPELINE_UUID = "a558111b-4c50-4b2e-9de8-73fd8fd3a67d"
```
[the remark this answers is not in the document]
gabdank: done

src/encoded/types/file.py
```diff
@@ -135,11 +143,15 @@ def __name__(self):
...
         if properties.get('status') != 'replaced':
-            if 'md5sum' in properties:
+            if is_md5sum_constrained_file(
+                properties,
+                ['raw data'],
```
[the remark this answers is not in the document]
gabdank: done

src/encoded/types/file.py
```diff
@@ -66,6 +66,14 @@ def property_closure(request, propname, root_uuid):
     return seen
 
 
+def is_md5sum_constrained_file(file_properties, unique_md5_output_categories, uniform_processing_pipeline_uuid):
+    if ((file_properties.get('lab') != uniform_processing_pipeline_uuid or 
+        file_properties.get('output_category') in unique_md5_output_categories) and 
+        'md5sum' in file_properties):
+        return True
```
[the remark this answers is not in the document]
gabdank: done

src/encoded/types/file.py
gabdank: thanks, done

src/encoded/types/file.py
```diff
@@ -136,12 +150,19 @@ def __name__(self):
...
+
     def unique_keys(self, properties):
         keys = super(File, self).unique_keys(properties)
         if properties.get('status') != 'replaced':
             if 'md5sum' in properties:
                 value = 'md5:{md5sum}'.format(**properties)
-                keys.setdefault('alias', []).append(value)
+                if self.registry[CONNECTION].get_by_unique_key('alias', value):
+                    if file_is_md5sum_constrained(properties):
+                        value = 'md5:{md5sum}'.format(**properties)
```
keenangraham: Why are you setting this three times?

src/encoded/types/file.py
```diff
@@ -136,12 +150,19 @@ def __name__(self):
...
         return properties.get(self.name_key, None) or self.uuid
 
+
     def unique_keys(self, properties):
         keys = super(File, self).unique_keys(properties)
         if properties.get('status') != 'replaced':
             if 'md5sum' in properties:
                 value = 'md5:{md5sum}'.format(**properties)
-                keys.setdefault('alias', []).append(value)
+                if self.registry[CONNECTION].get_by_unique_key('alias', value):
```
keenangraham: Why don't you try to set the value and then catch the exception if it already exists. Then determine whether to raise the exception or continue based on if the file is md5sum constrained.

src/encoded/types/file.py
gabdank: removed unnecessary value='md5'... lines

src/encoded/types/file.py
gabdank: I don't think the exception can be caught from here, as the keys go into postgres later

PR #2862, opened 2019-07-25
ENCD-4740-fix-broken-report-link

zoldello:


src/encoded/viewconfigs/views.py
```diff
@@ -65,15 +65,23 @@ def _get_doc_types(context, request):
 
 def _get_search_views(view_instance, context, request):
     doc_types = _get_doc_types(context, request)
+    types = request.registry[TYPES]
     views = []
     # TODO: Fix using protected members
     # pylint: disable=protected-access
     view_item = View_Item(view_instance._request, view_instance._search_base)
     # TODO: Move into SearchView after doc_types check
+    subtypes = []
```
keenangraham: Instead of computing subtypes in separate try/except block (and pulling in types again), why can't you just do something like: 
```
if len(doc_types) == 1:
    if doc_types[0] in view_instance._types:
        type_info = view_instance._types[doc_types[0]]
        if len(type_info.subtypes) <= 1:
            views.append(view_item.tabular_report)
```

zoldello: @keenangraham The reason why I did not do this is because the code blows up if doc_type[0] is invalid (e.g http://localhost:8000/search/?type=DatasetABC). But I will look at this again as pulling in types again does not seem right

zoldello: @keenangraham  Never mind. Although this changes when the code blows up, it is the same exception.

PR #2868, opened 2019-07-30
ENCD-4752 read_structure for fastq files

yunhailuo:


src/encoded/schemas/file.json
```diff
@@ -1386,6 +1394,39 @@
...
+            "type": "array",
+            "uniqueItems": true,
+            "items": {
+                "title": "Sequence element",
+                "description": "One element in the read with start and end coordinates.",
+                "type": "object",
+                "additionalProperties": false,
+                "properties": {
+                    "sequence_element": {
+                        "title": "Element Name",
```
gabdank: "Element Name" probably should be "Element name"

PR #2869, opened 2019-07-31
ENCD-4398-new-login-type

zoldello:


src/encoded/auth0.py
```diff
@@ -96,6 +93,40 @@ def forget(self, request):
...
 
 
+def signup(request):
+    """
+    Create new user.
+
+    :param request: Pyramid request object
+    """
+    domain = 'encode.auth0.com'
+    url = "https://{domain}/userinfo?access_token={access_token}".format(domain=domain, access_token=request.json['accessToken'])
```
keenangraham: What if `accessToken` isn't in JSON?

src/encoded/auth0.py
```diff
@@ -96,6 +93,40 @@ def forget(self, request):
...
+
+    users = request.registry[COLLECTIONS]['user']
+    user_data = user_data_request.json()
+
+    if not user_data['email_verified']:
+        raise HTTPBadRequest(explanation='Unverified e-mail')
+
+    user_info = {
+        'email': user_data['email'],
+        'first_name': user_data.get('given_name', '') or user_data.get('first_name', '') or user_data.get('name', ''),
```
keenangraham: So someone could make an account without a name?

src/encoded/auth0.py
```diff
@@ -96,6 +93,40 @@ def forget(self, request):
...
+        'last_name': user_data.get('family_name', '') or user_data.get('last_name', ''),
+        'groups': [],
+        'submits_for': [],
+        'viewing_groups': [],
+        'status': 'current',
+    }
+    user = create_item(users.type_info, request, user_info)
```
keenangraham: What if there's already an account with the same email address?

src/encoded/auth0.py
```diff
@@ -109,10 +140,9 @@ def login(request):
     else:
         namespace, userid = login.split('.', 1)
 
+    # create new user account if one does not exist
     if namespace != 'auth0':
-        request.session.invalidate()
-        request.response.headerlist.extend(forget(request))
-        raise LoginDenied()
+        userid = signup(request)
```
keenangraham: Does this automatically create a new account for you?

src/encoded/auth0.py
```diff
@@ -96,6 +93,40 @@ def forget(self, request):
...
+    :param request: Pyramid request object
+    """
+    domain = 'encode.auth0.com'
+    url = "https://{domain}/userinfo?access_token={access_token}".format(domain=domain, access_token=request.json['accessToken'])
+    user_data_request = requests.get(url)
+
+    if user_data_request.status_code != 200:
+        raise HTTPBadRequest(explanation='Could not get user data')
+
+    users = request.registry[COLLECTIONS]['user']
```
keenangraham: Maybe place this closer to create_item if it's only used for `users.type_info`?

src/encoded/auth0.py
```diff
@@ -96,6 +93,40 @@ def forget(self, request):
...
+        'first_name': user_data.get('given_name', '') or user_data.get('first_name', '') or user_data.get('name', ''),
+        'last_name': user_data.get('family_name', '') or user_data.get('last_name', ''),
+        'groups': [],
+        'submits_for': [],
+        'viewing_groups': [],
+        'status': 'current',
+    }
+    user = create_item(users.type_info, request, user_info)
+    if not user:
+        raise HTTPServerError(explanation='user not created')
```
keenangraham: Is there a test for this?

src/encoded/static/components/app.js
```diff
@@ -496,7 +499,14 @@ class App extends React.Component {
             this.setState({ profilesTitles });
         });
     }
-
+    /**
+    * Login with exisitin user or create new user and then login
```
keenangraham: Typo.

src/encoded/auth0.py
zoldello: Let me think about that a little. My idea is to let the code blow up as the user has an invalid request object. This is the way it currently works.

src/encoded/auth0.py
zoldello: Great point. Let me think about this too. Github just provides a name object; no first or last name. The ENCODE db does not blow up if there is no name. So, I tried to fine a good solution given this. Since you are raising a flag, my solution is not clear and may need to be revised.

src/encoded/auth0.py
zoldello: The user is passed through. I spoke with Idan about this and he said the behavior is expected. So, to be sure we are on the same page, If I am registered with "me@me.com" on linkedin, I can enter use "me@me.com" on google to enter the system if I have a google account. I seems strange to me too.

src/encoded/auth0.py
zoldello: Yes. If you are validated by an authentication system like google or linkedin but not registered in encoded, an account is created for you.

src/encoded/static/components/app.js
zoldello: Typo? Where?

src/encoded/auth0.py
zoldello: I will look into it.

src/encoded/auth0.py
zoldello: Technically yes; but I may need a more explicit one since you need to ask.

src/encoded/static/components/app.js
keenangraham: existin -> existing.

src/encoded/auth0.py
zoldello: Done.

src/encoded/auth0.py
zoldello: Good pointed. I added unit test.

src/encoded/static/components/app.js
zoldello: Corrected.

src/encoded/auth0.py
```diff
@@ -96,6 +92,47 @@ def forget(self, request):
...
+
+    if user_data_request.status_code != 200:
+        raise HTTPBadRequest(explanation='Could not get user data')
+
+    user_data = user_data_request.json()
+
+    if user_data['email_verified'] is not True:
+        raise HTTPBadRequest(explanation='Unverified e-mail')
+
+    if user_data.get('name', ''):
```
keenangraham: Seems like this code might break easily? What if someone has two last names? Also are we really sure we want to add 'N/A' to the database?

src/encoded/tests/test_auth0.py
```diff
@@ -44,10 +90,93 @@ def test_login_unknown_user(anontestapp, auth0_encode_user_token):
...
+@mock.patch('requests.get', return_value=mock_requests_get(
+    url='',
+    json_data={
+        'email_verified': True,
+        'email': 'fakeemail@email.com',
+        'first_name': 'fakefirstname',
+        'last_name': 'fakelastname',
+    },
+    status_code=200,))
+def test_login_verify_signup_is_called_when_creating_account(mock_get, signup, remember):
```
keenangraham: Nice tests.

src/encoded/auth0.py
```diff
@@ -96,6 +92,47 @@ def forget(self, request):
...
 
 
+def _signup(request):
+    """
+    Create new user.
+
+    :param request: Pyramid request object
+    """
+    domain = 'encode.auth0.com'
+    access_token = request.json['accessToken']
```
keenangraham: This still seems problematic if the request doesn't have accessToken.

src/encoded/auth0.py
keenangraham: My main concern with this code is that it could accidentally create a bunch of spam accounts. Should definitely try to break it in QA.

src/encoded/auth0.py
zoldello: Ok, I will do a check and blow up if accessToken is not found. But it also my be a good to leave it because if accessToken is not available, the user has an invalid request or is trying to hack the system; so the code should blow up.

src/encoded/auth0.py
zoldello: This is tricky.

src/encoded/auth0.py
zoldello: I now check for accessToken and raise an exception if not available

src/encoded/auth0.py
keenangraham: In any case this logic should be split into its own function and unit tested.

src/encoded/types/user.py
```diff
@@ -80,7 +80,11 @@ def __ac_local_roles__(self):
         },
     }, category='page')
     def access_keys(self, request):
-        if not request.has_permission('view_details'):
+        properties = self.properties
+        is_admin = 'groups' in properties and 'admin' in properties['groups']
```
zoldello: Lines 84 - 87 do not seem right. But I could not think of a better way.

src/encoded/types/access_key.py
```diff
@@ -79,6 +82,14 @@ class Collection(Item.Collection):
 @view_config(context=AccessKey.Collection, permission='add', request_method='POST',
              validators=[validate_item_content_post])
 def access_key_add(context, request):
+    principals = effective_principals(request)
+    has_permission = any([
+        principal
+        for principal in principals
+        if principal in ('group.admin') or principal.split('.')[0] in ('submits_for', 'viewing_group',)
```
zoldello: This does not seem right. But I could not think of a better way.

src/encoded/static/components/app.js
```diff
@@ -188,6 +190,30 @@ class Timeout {
...
+            <div>
+                <p>
+                    You are about to create an account in ENCODED. Please, have a look at out <a href="https://www.stanford.edu/site/terms/">terms of service</a> and our <a href="https://www.stanford.edu/site/privacy/">privacy policy</a>.
+                </p>
+            </div>
+        </ModalBody>
+        <ModalFooter
+            closeModal={closeModal}
+            submitBtn={signup}
+            submitTitle="Proceed"
```
zoldello: "Proceed" or "Continue"?

src/encoded/tests/test_permissions.py
```diff
@@ -228,21 +228,21 @@ def test_users_view_basic_authenticated(submitter, authenticated_testapp):
     res = authenticated_testapp.get(submitter['@id'])
     assert 'title' in res.json
     assert 'email' not in res.json
-    assert 'access_keys' not in res.json
+    assert not res.json.get('access_keys', '')
```
keenangraham: This seems like a suboptimal change. `assert 'access_keys' not in res.json` is more idiomatic Python for checking existence of a key in a dict. Also `assert not res.json.get('access_keys', '')` returns `None` by default so doesn't need the default empty string.

src/encoded/upgrade/user.py
```diff
@@ -11,10 +11,11 @@ def user_0_3(value, system):
 @upgrade_step('user', '3', '4')
 def user_3_4(value, system):
     # http://redmine.encodedcc.org/issues/4743
-    value['viewing_groups'] = [
-        'ENCODE3' if v == 'ENCODE' else v
-        for v in value['viewing_groups']
-    ]
+    if 'viewing_groups' in value:
```
keenangraham: You shouldn't change anything in this file. The upgrades have already been applied to specific schema versions (current user version is 8) and stored in the database. Changing things here is tampering with the historical record of migrations and won't have any effect on current schema version.

src/encoded/auth0.py
```diff
@@ -96,6 +97,54 @@ def forget(self, request):
...
+    if user_data['email_verified'] is not True:
+        raise HTTPBadRequest(explanation='Unverified e-mail')
+
+    if user_data.get('name', ''):
+        name_split = user_data.get('name', '').split(' ')
+        name_length = len(name_split)
+        name1 = name_split[0] if name_length >= 1 else user_data['email']
+        name2 = name_split[-1] if name_length >= 2 else user_data['email']
+
+    user_info = {
```
keenangraham: You should try splitting your code out into functions that do exactly one thing and can be unit tested. Same goes for the name parsing logic above.

src/encoded/auth0.py
```diff
@@ -96,6 +97,54 @@ def forget(self, request):
...
+        'email': user_data['email'],
+        'first_name': user_data.get('given_name', '') or user_data.get('first_name', '') or name1,
+        'last_name': user_data.get('family_name', '') or user_data.get('last_name', '') or name2,
+        'groups': [],
+        'submits_for': [],
+        'viewing_groups': [],
+        'status': 'current',
+    }
+    request.validated = user_info
```
keenangraham: Why?

keenangraham: You are just manually inserting this value without actually validating it.

src/encoded/auth0.py
```diff
@@ -96,6 +97,54 @@ def forget(self, request):
...
+    users = request.registry[COLLECTIONS]['user']
+    user = collection_add(users, request, None)
```
keenangraham: The default argument is already None so this is redundant: `def collection_add(context, request, render=None):` Also you should use keyword arguments to specify keyword arguments. Also the user object you are adding still isn't getting validated.

src/encoded/types/access_key.py
keenangraham: Yeah this isn't how ACLs work. I think the correct approach is changing who is assigned add permission here: https://github.com/ENCODE-DCC/encoded/blob/fa8ff2d166b89bbce4933005adca2bbf5f552702/src/encoded/types/access_key.py#L41 Currently it is `Authenticated` which is a Pyramid catchall for anyone who is logged in.

src/encoded/types/user.py
keenangraham: This again should be assigned in ACLs. And in any case this is already `view_details` is already restricted to admin group so this is unnecessary.

hitz: Should maybe this be checked in access_key.AccessKey.access_key_add()

src/encoded/auth0.py
```diff
@@ -96,6 +101,56 @@ def forget(self, request):
         return []
 
 
+@view_config(context=User.Collection, request_method='POST', permission=NO_PERMISSION_REQUIRED, name='sign-up')
```
keenangraham: You should make a new permission (maybe called `signup`) and add it to this list of permissions that are assigned to everyone: https://github.com/ENCODE-DCC/encoded/blob/eea4bf93e65f8368a54b16acb098e6daea1cfa2a/src/encoded/root.py#L56

Then make this view use the `signup` permission instead of NO_PERMISSION_REQUIRED. This is similar to how the search/report/matrix views have a `search` permission.

src/encoded/auth0.py
```diff
@@ -96,6 +101,56 @@ def forget(self, request):
...
+        user_data['email']
+    )
+    user_info = {
+        'email': user_data['email'],
+        'first_name': user_data.get('given_name', '') or user_data.get('first_name', '') or first_name,
+        'last_name': user_data.get('family_name', '') or user_data.get('last_name', '') or last_name,
+    }
+    validate_request(context.type_info.schema, request, user_info)
+    if request.errors:
+        raise HTTPUnprocessableEntity(json={'errors':request.errors}, content_type='application/json')
```
keenangraham: This should raise a validation error.

src/encoded/auth0.py
```diff
@@ -96,6 +101,56 @@ def forget(self, request):
...
+    user = collection_add(context, request, user_info)
+    if not user or user['status'] != 'success':
+        raise HTTPInternalServerError(explanation='user not created')
+    return HTTPCreated(explanation='User created')
+
+
+def _get_first_and_last_names_from_name(name, email=''):
+    if not name or not name.strip():
+        return '', ''
```
keenangraham: There's something weird about returning empty strings.

src/encoded/static/components/app.js
```diff
@@ -188,6 +190,30 @@ class Timeout {
 }
 
 
+const EulaModal = ({ closeModal, signup }) => { return (
+    <Modal>
+        <ModalHeader title="Creating a new account" closeModal={closeModal} />
+        <ModalBody>
+            <div>
+                <p>
+                    You are about to create an account in ENCODED. Please, have a look at out <a href="https://www.stanford.edu/site/terms/">terms of service</a> and our <a href="https://www.stanford.edu/site/privacy/">privacy policy</a>.
```
keenangraham: `Please,` doesn't need the comma.

src/encoded/auth0.py
```diff
@@ -96,6 +101,56 @@ def forget(self, request):
...
 
+@view_config(context=User.Collection, request_method='POST', permission=NO_PERMISSION_REQUIRED, name='sign-up')
+def signup(context, request):
+    """
+    Create new user.
+
+    :param request: Pyramid request object
+    """
+    domain = 'encode.auth0.com'
+    access_token = request.json.get('accessToken', '')
```
keenangraham: You shouldn't fill in a default empty string here.

src/encoded/tests/test_auth0.py
```diff
@@ -34,7 +48,61 @@ def auth0_encode_user_profile(auth0_access_token):
...
+    return result
 
+
+@pytest.fixture(scope='session')
+def mock_validate_request(schema, request, user_inf):
+    return ''
+
+
+@pytest.fixture(scope='session')
+def mock_collection_add_return_None(users, request, render):
```
keenangraham: Function name shouldn't have capital None.

src/encoded/auth0.py
```diff
@@ -3,11 +3,16 @@
 from snovault.validation import ValidationFailure
 from snovault.validators import no_validate_item_content_post
 from operator import itemgetter
+from snovault.crud_views import collection_add
+from snovault.schema_utils import validate_request
 from pyramid.authentication import CallbackAuthenticationPolicy
-import requests
+from .types.user import User
```
keenangraham: Use full path here.

src/encoded/auth0.py
```diff
@@ -96,6 +101,56 @@ def forget(self, request):
...
+    user_data = user_data_request.json()
+    if user_data['email_verified'] is not True:
+        raise HTTPBadRequest(explanation='Unverified e-mail')
+    first_name, last_name = _get_first_and_last_names_from_name(
+        user_data.get('name', ''),
+        user_data['email']
+    )
+    user_info = {
+        'email': user_data['email'],
+        'first_name': user_data.get('given_name', '') or user_data.get('first_name', '') or first_name,
```
keenangraham: This still needs testing.

zoldello:
@forresttanaka The Javascript part is ready for review.

src/encoded/upgrade/user.py
keenangraham: Still not sure why there are changes here? You shouldn't touch these.

src/encoded/auth0.py
keenangraham: Still needs testing.

src/encoded/static/components/app.js
keenangraham: `account in ENCODED` Maybe `You are about to create an ENCODE account.` ENCODED isn't a thing.

keenangraham: Typo: `Please have a look at out`

Maybe `Please have a look at the terms of service and privacy policy.`

src/encoded/tests/test_auth0.py
```diff
@@ -1,5 +1,20 @@
 import pytest
 import requests
+from unittest import mock
+from snovault.crud_views import create_item
```
keenangraham: You're not actually using this anywhere?

src/encoded/tests/test_access_key.py
```diff
@@ -72,8 +72,8 @@ def test_access_key_principals(anontestapp, execute_counter, access_key, submitt
     ]
 
 
-def test_access_key_self_create(anontestapp, access_key, submitter):
-    extra_environ = {'REMOTE_USER': str(submitter['email'])}
+def test_access_key_self_create(anontestapp, admin):
```
keenangraham: Shouldn't a submitter still have permission to create an access_key? Why did you change it to admin?

src/encoded/tests/test_auth0.py
zoldello: Hold over from old code. Removed.

src/encoded/static/components/app.js
zoldello: I meant "our" instead of "out". I changed the text to reflect the idea you have.

src/encoded/upgrade/user.py
zoldello: I missed that. My revert went wrong and I misread the diff changes prior to completion. It is fixed

src/encoded/static/components/app.js
```diff
@@ -497,12 +530,48 @@ class App extends React.Component {
         });
     }
 
-    handleAuth0Login(authResult, retrying) {
+    closeSignupModal() {
+        this.setState({ eulaModalVisibility: false });
+    }
+
+    signup(authResult) {
+        const accessToken = authResult.accessToken;
```
forresttanaka: Are you guaranteed to be able to get accessToken here? authResult is initialized to an empty string.

src/encoded/static/components/app.js
```diff
@@ -1123,6 +1196,12 @@ class App extends React.Component {
                                     </div>
                                 </Provider>
                             </div>
+                            {this.state.eulaModalVisibility ?
+                                <EulaModal
+                                    closeModal={this.closeSignupModal}
+                                    signup={() => this.signup(this.state.authResult)}
```
forresttanaka: Is there a benefit to using an inline function to call an internal method, instead of just calling the internal method directly?

src/encoded/static/components/schema.js
```diff
@@ -629,7 +629,8 @@ globals.contentViews.register(SchemaPage, 'JSONSchema');
 // add a new object of that type if you're logged in.
 const AllSchemasPage = (props, reactContext) => {
     const { context } = props;
-    const loggedIn = !!(reactContext.session && reactContext.session['auth.userid']);
+    const reactContextUser = reactContext.session_properties ? reactContext.session_properties.user : {};
+    const hasLab = !!(reactContextUser && reactContextUser.lab && reactContextUser.lab.status === 'current');
```
forresttanaka: I think technically `reactContextUser` will always be truthy, because if session_properties is nothing, then you set reactContextUser to an empty object, which is truthy.

src/encoded/static/components/user.js
```diff
@@ -206,6 +206,11 @@ const User = (props) => {
         { id: 'Users' },
     ];
     const crumbsReleased = (context.status === 'released');
+    const hasSubmitterForRights = context.submits_for && context.submits_for.length > 0;
+    const isAdmin = context.groups && context.groups.includes('admin');
+    const hasViewGroupRights = context.viewing_groups && context.viewing_groups.length > 0;
+    const hasAccessKeyRights = (hasSubmitterForRights || isAdmin || hasViewGroupRights) && context.access_keys;
```
forresttanaka: Does `context.access_keys` have a possibility of not existing? I realize the existing code does this same test, but I’m wondering if it’s ever not truthy?

zoldello: It should always exist and the code should blow up if it does not. But never say never and to make things clearer, so I will check for existence.

forresttanaka: You’re already checking if it exists. I was wondering if you need to.

src/encoded/static/components/schema.js
zoldello: Switching between Python and Javascript... Python evaluate to the key-value nature of object-{} has the opposite effect. I will fix this.

zoldello:
@hitz @forresttanaka  Are there any changes you require to this ticket? The status still say you have requested changes but I do not see any.

forresttanaka:
@zoldello I did request changes, but it looks like you addressed them all. I need to finish a convoluted rebase, and then I’ll look at this again in a bit.

src/encoded/static/components/schema.js
```diff
@@ -627,7 +627,12 @@ globals.contentViews.register(SchemaPage, 'JSONSchema');
 // add a new object of that type if you're logged in.
 const AllSchemasPage = (props, reactContext) => {
     const { context } = props;
```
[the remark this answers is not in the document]
zoldello: I saw this late. Since this is optional and I already have a rebased code-based ready for QA, I will address this is there are bugs on this ticket found or otherwise during a future enhancement.

zoldello:
@forresttanaka @hitz There was new Javascript changes. For this ticket @jenjou Brought up a good point. We need to hide the add-button on all schema items "Materials & Methods" -> "Schemas", tag,

src/encoded/static/components/audit.js
```diff
@@ -335,13 +335,13 @@ AuditCounts.propTypes = {
     /** True to wrap counts in a div with appropriate CSS class; false if taken care of outside */
     useWrapper: PropTypes.bool,
     /** True if user is logged in */
```
forresttanaka: Could you update the comment?

src/encoded/static/components/audit.js
```diff
@@ -363,8 +363,11 @@ export const auditDecor = AuditComponent => class extends React.Component {
     }
 
     auditIndicators(audits, id, options) {
-        const { session, search } = options || {};
-        const loggedIn = !!(session && session['auth.userid']);
+        const { session, search, sessionProperties } = options || {};
+        const reactContextUser = sessionProperties ? sessionProperties.user : null;
+        const isAdmin = reactContextUser && reactContextUser.admin;
+        const isSubmitter = reactContextUser && reactContextUser.lab && reactContextUser.lab.status === 'current' && reactContextUser.submits_for && reactContextUser.submits_for.length > 0;
+        const isAuthorized = isAdmin || isSubmitter;
```
forresttanaka: I feel like this isAdmin/isSubmitter/isAuthorized pattern will be needed in many places, and the tests (particularly isSubmitter) aren’t obvious. I’m thinking they could be three functions in globals.js, or one function in globals.js returning three booleans as an object, or maybe done once in app.js and supplied in context so people can just use things like `reactContext.isSubmitter` or something like that.

src/encoded/static/components/biosample.js
```diff
@@ -71,7 +71,7 @@ class BiosampleComponent extends React.Component {
                     </div>
                     <ItemAccessories item={context} audit={{ auditIndicators: this.props.auditIndicators, auditId: 'biosample-audit' }} />
                 </header>
-                {this.props.auditDetail(context.audit, 'biosample-audit', { session: this.context.session, except: context['@id'] })}
+                {this.props.auditDetail(context.audit, 'biosample-audit', { session: this.context.session, sessionProperties: this.context.session_properties, except: context['@id'] })}
```
forresttanaka: If you like, you can delete the `except` property whereever it gets passed to `auditDetail` because `auditDetail` doesn’t use it anymore. It’s harmless to pass in; just unneeded, so it’s up to you.

src/encoded/static/components/biosample_type.js
```diff
@@ -119,11 +120,11 @@ const ListingComponent = ({ context: result, auditIndicators, auditDetail }, rea
             </div>
             <div className="result-item__meta">
                 <div className="type meta-title">Biosample Type</div>
-                {auditIndicators(result.audit, result['@id'], { session: reactContext.session, search: true })}
+                {auditIndicators(result.audit, result['@id'], { session: reactContext.session, sessionProperties: reactContext.session_properties, search: true })}
             </div>
             <PickerActions context={result} />
         </div>
-        {auditDetail(result.audit, result['@id'], { session: reactContext.session, except: result['@id'], forcedEditLink: true })}
+        {auditDetail(result.audit, result['@id'], { session: reactContext.session, sessionProperties: reactContext.session_properties, except: result['@id'], forcedEditLink: true })}
```
forresttanaka: Like the `except` property, `forcedEditLink` isn’t used anymore, so you can delete this property wherever it occurs if you like.

src/encoded/static/components/biosample.js
zoldello: Would you have an issue if I keep it? There are 30 files modified and a lot of changes... Removing it when it is actually needed may cause a hard-to-detect problem?

forresttanaka: No issues. It’s harmless. It could be in a later audit ticket.

src/encoded/static/components/audit.js
forresttanaka: Actually, is this isAdmin correct? the `admin` property is a property of sessionProperty, not user.

zoldello: I addressed this in the code-refactor

src/encoded/static/components/audit.js
zoldello: Done.

zoldello:
@forresttanaka I fixed the suggested change except the one I said I would not. I also adjusted the status bar permissio.

src/encoded/static/components/schema.js
```diff
@@ -627,7 +629,12 @@ globals.contentViews.register(SchemaPage, 'JSONSchema');
...
-    const loggedIn = !!(reactContext.session && reactContext.session['auth.userid']);
+    const reactContextUser = reactContext.session_properties ? reactContext.session_properties.user : null;
+    const canAddSchema = !!(
+        reactContextUser && reactContextUser.lab && reactContextUser.lab.status === 'current' &&
+        reactContextUser.submits_for && reactContextUser.submits_for.length > 0 &&
+        reactContextUser.submits_for.includes(reactContextUser.lab['@id'])
+    );
```
forresttanaka: Looks like you can use globals.getRoles() which does most of the same tests and check for submitter and admin here, but I won’t gate the code review on it since this does the job, but it duplicates some code.

src/encoded/static/components/status.js
```diff
@@ -110,6 +110,8 @@ const defaultObjectStatuses = {
             'archived',
             'revoked',
         ],
+        unprivileged: [
+        ],
```
forresttanaka: Is this still needed? Sort of odd this is the only use of `unprivileged` in this file, but it’s empty.

forresttanaka:
@hitz is listed as a reviewer, so I’ll keep the ticket in Code Review Ready.

src/encoded/static/components/audit.js
```diff
@@ -363,17 +364,24 @@ export const auditDecor = AuditComponent => class extends React.Component {
...
 
         if (auditsDisplayed(audits, session)) {
             // Calculate the class of the indicator button based on whether the audit detail panel
             // is open or not.
             const indicatorClass = `audit-indicators btn btn-sm${this.state.auditDetailOpen ? ' active' : ''}${search ? ' audit-search' : ''}`;
+            const auditItems = Object.keys(audits);
+
+            // special case for unauthorized users
+            if (!isAuthorized && auditItems.length === 1 && auditItems[0] === 'INTERNAL_ACTION') {
+                return;
```
forresttanaka: ESLint error. Turns out, if you return something in one case, you need to return something in _all_ cases from that function or method. So you need to return null here.

src/encoded/static/components/filegallery.js
```diff
@@ -275,6 +282,8 @@ FileTable.propTypes = {
     setInfoNodeVisible: PropTypes.func,
     /** User session */
     session: PropTypes.object,
+    /** User session properties */
+    session_properties: PropTypes.object,
```
forresttanaka: ESLint error. You’re getting session_properties from contextTypes, so you don’t use this one.

src/encoded/static/components/form.js
```diff
@@ -11,6 +11,7 @@ import { FileInput, ItemPreview, ObjectPicker } from './inputs';
 import Layout from './layout';
 import DropdownButton from '../libs/ui/button';
 import { DropdownMenu } from '../libs/ui/dropdown-menu';
+import * as globals from './globals';
```
forresttanaka: ESLint ereror: You’re already importing from globals.js above, so you’ll need to resolve this.

src/encoded/static/components/audit.js
zoldello: Fixed. Weird. Returning nothing returns undefined. Plus, I cannot get ESlint to warn me about this.

src/encoded/static/components/filegallery.js
zoldello: Fixed. I did not get the exception initially.

src/encoded/static/components/form.js
zoldello: Fixed. Weird. Did not see this initially.

PR #2870, opened 2019-07-31
ENCD-4768 Add matching md5sum property

jenjou:


src/encoded/audit/file.py
```diff
@@ -373,6 +373,29 @@ def audit_file_in_correct_bucket(value, system):
...
+
+    for matching_file in value.get('matching_md5sum'):
+        if matching_file.get('md5sum') != value.get('md5sum'):
+            detail = 'File {} is listed as having a matching md5 sum '.format(matching_file['@id']) + \
+                     'as file {}, but the files have different md5 sums.'.format(value['@id'])
+            yield AuditFailure('Incorrect matching_md5sum', detail, level='ERROR')
+        else:
+            detail = 'The md5 sum of file {} '.format(value['@id']) + \
+                     'matches that of file {}.'.format(matching_file['@id'])
+            yield AuditFailure('Matching md5 sums', detail, level='WARNING')
```
yunhailuo: I'm wondering what will happen for multiple matching files. I understand we need a warning to flag this file having matching files. But if we do this in the loop and there are 10 matching files, are we going to get 10 warnings or 1. And do we want to get 10 or 1 by design? @jenjou and @gabdank

src/encoded/audit/file.py
```diff
@@ -373,6 +373,29 @@ def audit_file_in_correct_bucket(value, system):
...
+    Files with md5 sums matching other files should be marked with a WARNING audit.
+    If the other files are listed as matching but in fact have different md5 sums,
+    the file should be flagged with an ERROR for incorrect metadata.
+    '''
+
+    checked_statuses = ['released', 'revoked', 'archived', 'in progress']
+    if 'matching_md5sum' not in value or value.get('status') not in checked_statuses:
+        return
+
+    for matching_file in value.get('matching_md5sum'):
```
yunhailuo: This is just a comment and probably a long nonsense.

It's interesting case here. So 1) this `if` above makes sure you have `matching_md5sum` and 2) the schema below makes sure you would get a list, at least an empty one, anyway. Those two then guarantee you won't get `None` from `value.get('matching_md5sum')` which will trigger error for the `for` loop. Personally, I don't want to have a code depends on code at other places or far away. Like if schema changed or something, there is no guarantee you won't get `None` or trigger error. So I like the `get` here but you could also have the default being list. On the contrary, though, an error here could be good to tell whoever changes codes in the future to check back here. So I don't really know and I'm not really recommending further changes here.

src/encoded/audit/file.py
jenjou: My understanding was we wanted a separate warning for each file, so it's written to raise 10 warnings instead of collecting them into 1. But I'd like to double check with @gabdank this is what we want.

src/encoded/audit/file.py
p-sud: The audit test would fail if the schema got updated, so I think this is OK. The developer would know that they need to update the audit.

src/encoded/audit/file.py
jenjou: Changed this section, please view the updated file.

src/encoded/audit/file.py
```diff
@@ -373,6 +373,41 @@ def audit_file_in_correct_bucket(value, system):
...
+            detail = 'File {} is listed as having a matching md5 sum '.format(file['@id']) + \
+                     'as file {}, but the files have different md5 sums.'.format(value['@id'])
+            yield AuditFailure('Incorrect matching_md5sum', detail, level='ERROR')
+        else:
+            matching_files.append(file['@id'])
+
+    if len(matching_files) > 2:
+        matching_files_joined = '{}, and {}'.format(
+            ', '.join(matching_files[:-1]),
+            matching_files[-1:][0]
```
yunhailuo: Why not just `matching_files[-1]`?

src/encoded/audit/file.py
```diff
@@ -373,6 +373,41 @@ def audit_file_in_correct_bucket(value, system):
...
+        )
+    else:
+        matching_files_joined = ' and '.join(matching_files)
+
+    detail = 'The md5 sum of file {} '.format(value['@id']) + \
+             'matches that of file(s) {}.'.format(matching_files_joined)
+    yield AuditFailure('Matching md5 sums', detail, level='WARNING')
```
yunhailuo: Should this be optional if `matching_files` is empty?

src/encoded/audit/file.py
jenjou: You are right, I'll update this

src/encoded/audit/file.py
jenjou: I'll add a case for len(matching_files)= 0.

PR #2871, opened 2019-08-01
ENCD-4646-add-search-btn-top-nav

zoldello:


src/encoded/static/components/navigation.js
```diff
@@ -211,7 +211,7 @@ const Search = (props, context) => {
     const id = url.parse(context.location_href, true);
     const searchTerm = id.query.searchTerm || '';
     return (
-        <form className="navbar-form navbar-right" action="/search/">
+        <form className="navbar-form navbar-right navbar-search-form" action="/search/">
```
emro: I think instead of adding a class here, you could also have applied the flexbox styles to `navbar-form` because it is the only element on Encode with that class name.

src/encoded/static/scss/encoded/modules/_forms.scss
```diff
@@ -90,10 +96,12 @@ input[type="text"], input[type="password"], input[type="date"], input[type="date
 
 .search-wrapper {
 	position: 0;
+	background-color: white;
+	border-radius: 20px;
+	padding: 2px !important;
```
emro: I'm not sure that you need the `!important` here.

src/encoded/static/scss/encoded/modules/_forms.scss
```diff
@@ -105,6 +113,18 @@ input[type="text"], input[type="password"], input[type="date"], input[type="date
...
 
+.navbar-search-form {
+    display: -webkit-box;
+    display: -ms-flexbox;
+    display: flex;
+}
+
+.search-button {
+	background-color: white;
+    border: 0 none white;
```
emro: You could just use `border: none` here. (Or `border: 0`)

src/encoded/static/scss/encoded/modules/_forms.scss
```diff
@@ -90,10 +96,12 @@ input[type="text"], input[type="password"], input[type="date"], input[type="date
 
 .search-wrapper {
 	position: 0;
```
emro: This line could be deleted as part of these changes. It's giving an "invalid property" error.

src/encoded/static/scss/encoded/modules/_forms.scss
```diff
@@ -105,6 +113,18 @@ input[type="text"], input[type="password"], input[type="date"], input[type="date
 	}
 }
 
+.navbar-search-form {
+    display: -webkit-box;
+    display: -ms-flexbox;
+    display: flex;
```
emro: These styles could potentially be added to `navbar-form`.

src/encoded/static/scss/encoded/modules/_forms.scss
```diff
@@ -63,21 +63,27 @@ input[type="text"], input[type="password"], input[type="date"], input[type="date
...
+	height: 25px;
 	padding: 3px 26px 4px 14px;
 	padding-top: 3px;
 	padding-bottom: 4px;
 	margin-top: 3px;
+	border: none;
+	box-shadow: none !important;
+	font-family: inherit;
+	font-size: 1rem;
+	font-weight: normal;
```
emro: It's ok to leave these, but I think lines 76-78 aren't necessary. `#navbar-search` will have `font-size: 1rem` and `font-weight: normal` already, from parent elements and `<input>` elements are already set to have `font-family: inherit`. If it seems clearer to have these explicitly set, though, that's also good.

src/encoded/static/components/navigation.js
zoldello: I can if you like. I did not originally because navbar-right is part of twitter bootstrap. I am normally hesitant to attach new style to 3rd party tools as it can lead to issues during version updates or in removal.

src/encoded/static/scss/encoded/modules/_forms.scss
zoldello: Good point. Corrected. I refactored the code from somewhere else and failed to realize these were not needed.

src/encoded/static/scss/encoded/modules/_forms.scss
zoldello: Good point. Corrected. I think I needed it at one point but that is no longer the case.

src/encoded/static/scss/encoded/modules/_forms.scss
zoldello: I did not do it originally because navbar-form is part of twitter bootstrap and modifying its styles make updates even more complication. But I can if you like.

src/encoded/static/scss/encoded/modules/_forms.scss
zoldello: Good point. Corrected.

src/encoded/static/components/navigation.js
emro: Ah, yes, that is a good point! I agree that it is better with a new class name.

src/encoded/static/scss/encoded/modules/_forms.scss
emro: It's good the way you wrote it!

src/encoded/static/scss/encoded/modules/_forms.scss
emro: @zoldello, @forresttanaka, and I discussed this, and the branch to get rid of bootstrap gets rid of this class entirely so this will be resolved when that branch is merged.

PR #2872, opened 2019-08-01
ENCD-4773 fix browser styles on IE 11

emro:
This branch should fix 3 issues on IE 11:
(1) There should be no light gray background on the facet buttons (aside from the bars that show how many results there are).
(2) The tabs and their facets should not wrap except at narrow screen sizes.
(3) A message should be displayed on the genome browser tab on IE 11 suggesting that users upgrade their browsers because Valis does not support IE 11.

Coding note: I thought about putting the browser detection code in filegallery.js but I think it makes more sense in genome_browser.js because that is where the other error message is ("There are no visualizable results.") and because we still want to display the browser tab, we just don't want to display the browser in the tab.

### DIFF
--- a/src/encoded/static/components/genome_browser.js
+++ b/src/encoded/static/components/genome_browser.js
@@ -130,53 +130,55 @@ class GenomeBrowser extends React.Component {
     }
 
     componentDidUpdate(prevProps, prevState) {
-        if (this.state.contig !== prevState.contig && !(this.state.disableBrowserForIE)) {
-            if (this.state.visualizer) {
-                this.state.visualizer.setLocation({ contig: this.state.contig, x0: this.state.x0, x1: this.state.x1 });
-            }
-        }
-
-        if (this.props.assembly !== prevProps.assembly && !(this.state.disableBrowserForIE)) {
-            // Determine pinned files based on genome, filter and sort files, compute and draw tracks
-            this.setGenomeAndTracks();
-            // Clear the gene search
-            this.setState({ searchTerm: '' });
-        }
-
-        // If the parent container changed size, we need to update the browser width
-        if (this.props.expanded !== prevProps.expanded && !(this.state.disableBrowserForIE)) {
-            setTimeout(this.drawTracksResized, 1000);
-        }
-
-        if (!(_.isEqual(this.props.files, prevProps.files)) && !(this.state.disableBrowserForIE)) {
-            let newFiles = [];
-            let files = [];
-            let domain = `${window.location.protocol}//${window.location.hostname}`;
-            if (domain.includes('localhost')) {
-                domain = domainName;
-                newFiles = [...this.state.pinnedFiles, ...dummyFiles];
-            } else {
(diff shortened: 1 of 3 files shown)


src/encoded/static/components/genome_browser.js
```diff
@@ -118,30 +120,35 @@ class GenomeBrowser extends React.Component {
...
+            this.setState({ disableBrowserForIE: true });
+        } else {
+            // Determine pinned files based on genome, filter and sort files, compute and draw tracks
+            this.setGenomeAndTracks();
+        }
     }
 
     componentDidUpdate(prevProps, prevState) {
-        if (this.state.contig !== prevState.contig) {
+        if (this.state.contig !== prevState.contig && !(this.state.disableBrowserForIE)) {
```
forresttanaka: Would it be better to have just one check for !(this.state.disableBrowserForIE) in an `if` that wrapped around all this, or the opposite check that returns immediately? It seems like you skip everything in this method if this.state.disableBrowserForIE is true anyway. Also, I do wonder if componentDidUpdate would _ever_ get called if componentDidMount doesn’t call setGenomeAndTracks(), and so you wouldn’t need this check at all? I get it if you’d like to check anyway, for safety.

src/encoded/static/components/genome_browser.js
```diff
@@ -475,7 +482,12 @@ class GenomeBrowser extends React.Component {
                         <div ref={(div) => { this.chartdisplay = div; }} className="valis-browser" />
                     </div>
                 :
-                    <div className="browser-error valis-browser">There are no visualizable results.
+                    <div>
```
forresttanaka: With React 16, you can replace some of these `<div>`s with `<React.Fragment>` to reduce the number of levels of DOM nodes — if CSS styles within don’t rely on these `<div>`s.

src/encoded/static/components/genome_browser.js
```diff
@@ -475,7 +482,12 @@ class GenomeBrowser extends React.Component {
...
+                        {(this.state.disableBrowserForIE) ?
+                            <div className="browser-error valis-browser">The genome browser does not support Internet Explorer. Please upgrade your browser to Edge in order to visualize files on ENCODE.</div>
```
forresttanaka: This is a pure taste thing, but I usually find I can remove “in order” and keep the same meanning and clarity in two fewer words.

src/encoded/static/components/genome_browser.js
emro: Awesome!!

src/encoded/static/components/genome_browser.js
emro: Oh yes, that seems easier. I remember thinking about that and then adding it to every conditional instead, but I think this way there will only be one check if the browser is IE and that's better. Also, this check is definitely necessary. The component updates (1) when you change the screen size, say by clicking on the arrow that expands or collapses the facets, (2) every time you select a facet or assembly, and (3) when you click from tab to tab. So everything will break if we don't keep this!

PR #2877, opened 2019-08-01
ENCD-4730-change-menu-item-names-for-matrices

p-sud:


p-sud:
Good catch. We're not consistent with that, but I think it would be good to make "Matrix" lowercase.

PR #2884, opened 2019-08-05
ENCD-4779 fix default tracks

emro:
This branch fixes several issues with the genome browser:
- The browser does not display for cases where the file list is filtered down to one result.
- We were not displaying all available DNA and gene tracks.
- We renamed the genome and gene tracks:
     --  genome track should be "GRCh38" or "mm10" etc
     -- genes should be "GENCODE V29" for GRCh38 and hg19 and should be "GENCODE M21" on mm10


hitz:
Please resolve conflicts and resubmit PR.

PR #2885, opened 2019-08-05
ENCD-4729-fix-add-minus-btn-on-graph

zoldello:


src/encoded/static/components/graph.js
```diff
@@ -644,9 +666,9 @@ export class Graph extends React.Component {
...
-                                    <td className="zoom-indicator"><i className="icon icon-plus" /></td>
+                                    <td className="zoom-indicator"><button onClick={() => this.changeZoom(-1)} touchstart={() => this.changeZoom(-1)}><i className="icon icon-minus" /></button></td>
+                                    <td className="zomm-controller"><input type="range" className="zoom-slider" ref={this.slider} min={minZoom} max={maxZoom} value={this.state.zoomLevel === null ? 0 : this.state.zoomLevel} onChange={this.rangeChange} onInput={this.rangeChange} onDoubleClick={this.rangeDoubleClick} onMouseUp={this.rangeMouseUp} onMouseDown={this.rangeMouseDown} /></td>
+                                    <td className="zoom-indicator"><button onClick={() => this.changeZoom(1)} touchstart={() => this.changeZoom(1)}><i className="icon icon-plus" /></button></td>
```
forresttanaka: These should be `onTouchStart` instead of `touchstart`, and in fact on dev builds you get an error in the Javascript console:

Screen Shot 2019-08-05 at 15 09 01

But I don’t think you need to react to touch events at all. Touch devices should give an onClick event when you tap these buttons. Touch-event handlers are very rarely needed.

zoldello: Good catch on the touchstarts. Touchstarts and click are slightly different but not enough to justitfy both. I removed touchstart. I increased the change rate as Jennifer J also preferred larger increments. The lack of auto-repeat zooming seems fine now after talking with Jennifer J. I can change it if user feedback suggest it is needed.

PR #2886, opened 2019-08-05
ENCD-4791 remove Quick View

emro:
Now that we have an embedded quick view browser (Valis), there is no need for a link to a browser in a new page.


emro:
I thought that I would need to make changes in snovault but I couldn't find code in snovault that seemed relevant and this does remove the Quick View links. If there is code I should also remove from snovault I can do that as well.

PR #2891, opened 2019-08-13
ENCD-4802 update ChIP-seq QMs

yunhailuo:


hitz:
seems innocuous enough to skip a demo

PR #2893, opened 2019-08-14
ENCD-4811 Change breadcrumb pluralization

forresttanaka:
For v89rc2.

### DIFF
--- a/src/encoded/static/components/experiment.js
+++ b/src/encoded/static/components/experiment.js
@@ -284,7 +284,15 @@ const ExperimentComponent = ({ context, auditIndicators, auditDetail }, reactCon
     // Determine whether object is Experiment or FunctionalCharacterizationExperiment.
     const experimentType = context['@type'][0];
     const isFunctionalExperiment = experimentType === 'FunctionalCharacterizationExperiment';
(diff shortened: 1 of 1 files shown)


src/encoded/static/components/experiment.js
```diff
@@ -284,7 +284,15 @@ const ExperimentComponent = ({ context, auditIndicators, auditDetail }, reactCon
...
-    const displayType = isFunctionalExperiment ? 'Functional Characterization Experiment' : 'Experiment';
+    let displayType;
+    let displayTypeBreadcrumbs;
+    if (isFunctionalExperiment) {
+        displayTypeBreadcrumbs = 'Functional Characterization Experiments';
+        displayType = 'Functional Characterization Experiment';
+    } else {
+        displayTypeBreadcrumbs = 'Experiments';
+        displayType = 'Experiment';
+    }
```
emro: Just a thought: you could have displayTypeBreadCrumbs = displayType + 's' or something or potentially append the "s" in line so that you'd just have the one variable, but this is definitely the most readable solution, I think it's the best.

forresttanaka: That was the first thought to cross my mind, but since it doesn’t work in every case (Antibody - Antibodies if it ever came to that) I felt better doing the method that would work every time.

PR #2909, opened 2019-08-29
ENCD-4819 Use not-allowed cursor on disabled inputs and buttons

forresttanaka:


src/encoded/static/scss/encoded/_base.scss
```diff
@@ -257,6 +257,11 @@ $disabled-color-factor: 20%;
 	}
 }
 
+input[disabled],
+button[disabled] {
+    cursor: not-allowed;
+}
```
emro: Looks good! Just a question: it could also work to write `input: disabled, button: disabled` -- is it generally the Encode style to use the brackets instead?

forresttanaka: I don’t think we have a set style for this.

PR #2913, opened 2019-09-02
ENCD-4079-fix-npm-audit-errors

zoldello:


package.json
```diff
@@ -22,71 +22,75 @@
...
     "redux-mock-store": "^1.5.1",
-    "sass-loader": "^4.1.1",
-    "string-replace-loader": "^1.0.5",
-    "style-loader": "^0.13.1",
-    "url-loader": "^0.5.7",
-    "webpack": "^1.14.0"
+    "sass-loader": "^8.0.0",
+    "string-replace-loader": "^2.2.0",
+    "style-loader": "^1.0.0",
+    "uglifyjs-webpack-plugin": "^2.2.0",
```
hitz: Do we still need uglifyjs?

zoldello: Done.

forresttanaka:
Just to confirm, do any global CLI tools (things installed with `npm install -g …`) need to be installed or updated, or are all the needed changes in the branch itself?

forresttanaka:
File association graphs don’t render in this branch. The errors stem from dagre-d3. Not sure if upgrading dagre-d3 will fix this, but my guess is it won’t. Need to investigate what the incompatibility is.

Screen Shot 2019-09-23 at 11 01 24

zoldello:
Demo is building, so far it looks good.

webpack.config.js
```diff
@@ -15,18 +16,22 @@ const PATHS = {
 
 const plugins = [];
 // don't include momentjs locales (large)
-plugins.push(new webpack.IgnorePlugin(/^\.\/locale$/, [/moment$/]));
+plugins.push(new webpack.IgnorePlugin(/^\.\/locale$/, /moment$/));
```
emro: This may have a conflict with the moment js update.

zoldello: Great point. But it is a river I will cross when I get to that river (aka when fixing an upcoming merge-conflict battle with dev). Are you okay with this approach?

emro:
It looks like the issue @forresttanaka identified has been resolved, the association graphs look good now as far as I can tell. While I was looking at the graphs on the demo, I noticed a 404 error for the spinner.gif. This might be due to the demo not being done indexing?

hitz:
>  While I was looking at the graphs on the demo, I noticed a 404 error for the spinner.gif. This might be due to the demo not being done indexing?

Probably not... I have seen this before, I think we might have an old reference to an old spinner that isn't used.

emro:
It looks like ENCD-4050 is merged so this needs a rebase.

zoldello:
@emro I built the demo around 11:00 AM, 10/03/2019.  I fixed the issue where the spinner was not appearing. It is also added the spinner image inline as opposed to being fetch as needed. I rebase on v91 RC2. The previous version had issues with Javascript and css bundling and minification. I fixed these.

package.json
```diff
@@ -22,72 +22,79 @@
...
-    "chart.js": "^2.5.0",
+    "auth0-lock": "^11.17.2",
+    "brace": "^0.11.1",
+    "chart.js": "^2.8.0",
     "ckeditor": "file:node_shims/ckeditor",
     "create-react-class": "^15.5.2",
     "d3": "^3.4.6",
-    "dagre-d3": "git+https://github.com/ENCODE-DCC/dagre-d3.git#v1.2.0",
-    "dalliance": "git+https://github.com/ENCODE-DCC/dalliance.git#v1.0",
+    "dagre-d3": "git+https://github.com/ENCODE-DCC/dagre-d3.git",
```
emro: I'm not sure if this would be better or if it matters very much, but you could make a release for your changes to `dagre-d3` and include the release number here.

zoldello: I can, but the style seems to have changed to move the changes to master, to make a release number be redundant. Should I still add it?

forresttanaka: Yeah, I agree with @emro on this; I know I mentioned it was probably OK to use master, but I’m not feeling that comfortable with that. I think tagging dagre-d3 with a release would be better in case we want to update master, but not cause new builds to change at that moment.

zoldello: Done. I created WRAN-1363 and WRAN-1364 with a comment in it signifying they are ENCODE-specific.

emro:
@forresttanaka and @hitz do either of you want changes made or should I move the ticket to QA?

forresttanaka:
@zoldello The dagre dependency is still loading graphlib/graphlib 2.1.7 from the graphlib org, along with your ENCODE-DCC/graphlib 1.2.3. Do you see that as an issue?

zoldello:
@forresttanaka Yes, that is fine. The purpose of the ticket is to make the npm-audit error-demons go away; and not so much to upgrade packages because we ought to (even though we ought to in the future). So, there will be cases where things like dagre 0.8.4 has a dependency on graph 2.17 while a different package has a dependency on our fork of graphlib. It is fine as long as no code is broken. In a different ticket, maybe during the d3-upgrade, we can address upgrading all SVG/canvas related packages. Is this an acceptable answer?

PR #2914, opened 2019-09-03
ENCD-4050-replace-momentjs

zoldello:


package-lock.json
```diff
@@ -5,22 +5,22 @@
   "requires": true,
   "dependencies": {
     "@babel/runtime": {
-      "version": "7.2.0",
-      "resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.2.0.tgz",
-      "integrity": "sha512-oouEibCbHMVdZSDlJBO6bZmID/zA/G/Qx3H1d3rSNPTD+L8UNKvCat7aKWSJ74zYbm5zWGh0GQN0hKj8zYFTCg==",
+      "version": "7.5.5",
```
hitz: @zoldello After talking to @forresttanaka and @emro these package upgrades should be deferred and done as a separate ticket; please take them out if they are not dependencies of day.js

zoldello: @hitz Are you referring to packages in package-lock.json? If yes, it is updated by npm and it pull in a new package if a minor or patch update requires one. I recommend considering it a black box. Now, for package.json, I only added the needed package.

webpack.config.js
```diff
@@ -25,6 +23,9 @@ let chunkFilename = '[name].js';
 let styleFilename = './css/[name].css';
 
 if (env === 'production') {
+    // don't include momentjs locales (large)
+    plugins.push(new webpack.IgnorePlugin(/^\.\/locale$/, [/moment$/]));
```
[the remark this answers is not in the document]
zoldello: Great point. I need to think about this. I left this in for chart.js. I need to test if this is significant. Also, I may be wise to move it out of env === 'production'.

src/encoded/static/components/award.js
```diff
@@ -1687,8 +1687,8 @@ MilestonesTable.propTypes = {
  */
 const fillDates = (sortedDateTerms, startDate) => {
     // The new array limits go between `startDate` and the current date.
-    const startDateMoment = moment(startDate, 'YYYY-MM');
-    const endDateMoment = moment();
+    const startDateMoment = dayjs(startDate, 'YYYY-MM');
+    const endDateMoment = dayjs();
```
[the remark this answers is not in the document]
zoldello: I did not think so because "Moment" is still a word for time. But, since you see a problems, I can see how others may see a problem too. I will update the variable names.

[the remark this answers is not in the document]
zoldello: Done.

package-lock.json
zoldello: I see optional:true was added. I believe it is harmless (https://stackoverflow.com/a/55373854/178550)

package-lock.json
```diff
@@ -4587,12 +4609,14 @@
...
-          "dev": true
+          "dev": true,
+          "optional": true
         },
         "yallist": {
           "version": "3.0.2",
           "bundled": true,
-          "dev": true
+          "dev": true,
+          "optional": true
```
emro: @forresttanaka and I have been noticing that `"optional": true` is added sometimes when new packages are installed and it looks like it is a known bug: https://github.com/npm/npm/issues/20926#issuecomment-397372131. We've been removing them because they don't seem important and we're trying to reduce changes in package-lock. I think maybe if you run `npm install` again they might be removed.

zoldello: @emro I threw the kitchen-sink at it and optional:true always stayed. I think it is due to an update in npm (https://stackoverflow.com/a/55373854/178550.) I can manually remove it if desired.

emro: OK, it looks like this is necessary, we should keep it.

zoldello:
@hitz Do you have changes you need made to this ticket?

zoldello:
@hitz @emro I had to re-review a review because github did not recognize your approval if you gave it.

zoldello:
@hitz Do you have any changes you want made?

PR #2915, opened 2019-09-04
ENCD-4824 Make Safari search box styling the same as others

forresttanaka:


src/encoded/static/scss/encoded/modules/_facet.scss
```diff
@@ -334,11 +334,10 @@
     margin: 5px 0;
     .icon-search {
         margin-right: 5px;
-        float: left;
-        flex-shrink: 0;
+        flex: 0 0 auto;
     }
     .searchform {
-        flex-grow: 1;
+        flex: 1 0 auto;
```
emro: I'm going to make a note on the ticket with a screenshot - I think this line causes a problem at certain widths.

forresttanaka: Changed to allow growing _and_ shrinking. I tested in Safari, FF, Chrome, IE11, and Edge.

PR #2928, opened 2019-09-19
ENCD-4803-new-search-views

keenangraham:
Several changes from ENCD-4803 epic. Includes switching old search views to v1, implementing v2 views in their place, and creating new matrix, audit, and button frontend code.

Points to SNO-113-make-audit-view.


forresttanaka:
* The `filterTerm` property to `<ViewControls>` exists for the (currently) sole situation in the old /audit/ matrix page where ?type=Experiment produces nothing in filters, and we have to assume it exists. This might be removed in the future when this situation no longer exists.
* Modal rendering update
    * With React 16, the existing mechanism to focus on modal elements when a modal appears no longer worked, so there was no way to tab into modals when they appeared, making them inaccessible, including the browser selector modal involved with this branch. The external API to use modals has not changed.
    * I inserted a new empty “modal-root” div in app.js where modal DOM elements get inserted.
    * The <Modal> component has split off its rendering and tab-key-handling code into a new <ModalElement> component that uses the new-in-React-16 ReactDOM.createPortal function that’s designed for modals.

src/encoded/search_views.py
```diff
@@ -0,0 +1,252 @@
+from pyramid.view import view_config
+
+from snovault.elasticsearch.searches.interfaces import AUDIT_TITLE
```
zoldello: This is not significant. Why are these (from snovault.elasticsearch.searches.interfaces) not group together?

src/encoded/search_views.py
```diff
@@ -0,0 +1,252 @@
...
+from snovault.elasticsearch.searches.interfaces import MATRIX_TITLE
+from snovault.elasticsearch.searches.interfaces import REPORT_TITLE
+from snovault.elasticsearch.searches.interfaces import SEARCH_TITLE
+from snovault.elasticsearch.searches.interfaces import SUMMARY
+from snovault.elasticsearch.searches.interfaces import SUMMARY_TITLE
+from snovault.elasticsearch.searches.fields import AuditMatrixWithFacetsResponseField
```
zoldello: This is not significant. Why are these (from snovault.elasticsearch.searches.fields ) not group together?

src/encoded/search_views.py
```diff
@@ -0,0 +1,252 @@
...
+    'File',
+    'Lab'
+]
+
+
+@view_config(route_name='search', request_method='GET', permission='search')
+def search(context, request):
+    # Note the order of rendering matters for some fields, e.g. AllResponseField and
+    # NotificationResponseField depend on results from BasicSearchWithFacetsResponseField.
+    fr = FieldedResponse(
```
zoldello: (Minor)The variable name does not confirm to the snake case naming style and is not meaningful

src/encoded/search_views.py
```diff
@@ -0,0 +1,252 @@
...
+        _meta={
```
zoldello: (Minor) Why does this have underscore?

src/encoded/search_views.py
```diff
@@ -0,0 +1,252 @@
...
+            SortResponseField(),
+            DebugQueryResponseField()
+        ]
+    )
+    return fr.render()
+
+
+@view_config(route_name='searchv2_raw', request_method='GET', permission='search')
+def searchv2_raw(context, request):
+    fr = FieldedResponse(
```
zoldello: (Minor) Would it be a good idea to place the declaration of things like _meta outside of the signature of FieldedResponse?

src/encoded/search_views.py
```diff
@@ -0,0 +1,252 @@
...
+            RawSearchWithAggsResponseField(
+                default_item_types=DEFAULT_ITEM_TYPES
+            )
+        ]
+    )
+    return fr.render()
+
+
+@view_config(route_name='searchv2_quick', request_method='GET', permission='search')
+def searchv2_quick(context, request):
```
zoldello: I understand the v2 part. But why is there quick and basic?

forresttanaka: I believe quick and raw are options if needed/desired (I don’t remember what they do). The UI only uses basic.

forresttanaka:
Since Keenan won’t be back until October 7, I think we should treat all back-end feedback as non-blocking on the code review, unless you find something that really should block it, and then we’ll move to the next sprint unless it’s a trivial change.

I’ll address any front-end feedback of course. I have added three changes from QA feedback after Keenan put this up for code review.

hitz:
Yeah, I'll review any back end suggestions myself.

src/encoded/static/components/auditmatrixv1.js
```diff
@@ -0,0 +1,740 @@
+import React from 'react';
```
hitz: This the one we are going to delete (next sprint) right?

src/encoded/viewconfigs/matrix.py
```diff
@@ -187,7 +187,7 @@ def preprocess_view(self):
         Main function to construct query and build view results json
         * Only publicly accessible function
         '''
-        matrix_route = self._request.route_path('matrix', slash='/')
+        matrix_route = self._request.route_path('matrixv1', slash='/')
```
hitz: Oh this is the old matrix.py

src/encoded/static/components/auditmatrixv1.js
forresttanaka: Deleted next sprint, yes.

src/encoded/search_views.py
zoldello: I mean the variable name "fr". It would be hard to know what it means later on in the code. It is not a standard acronym.

src/encoded/search_views.py
zoldello: Kinda, for a private variable. But "_meta" is not a private variable.

src/encoded/static/components/auditmatrixv1.js
```diff
@@ -0,0 +1,740 @@
...
+    handleOnScroll(e) {
+        this.handleScrollIndicator(e.target);
+    }
+
+    /**
+     * Show a scroll indicator depending on current scrolled position.
+     * @param {object} element DOM element to apply shading to
+     */
+    handleScrollIndicator(element) {
+        // Have to use a "roughly equal to" test because of an MS Edge bug mentioned here:
```
zoldello: Is this still important in that Microsoft is dumping trident-based MS Edge with a chromium-based Edge?

src/encoded/static/components/auditmatrixv1.js
```diff
@@ -0,0 +1,740 @@
...
+    let query;
+    if (rowCategoryBucket.key === 'no_audits') {
+        switch (subCategoryQuery) {
+        case 'no audits':
+            query = 'audit.WARNING.category%21=%2A&audit.NOT_COMPLIANT.category%21=%2A&audit.ERROR.category%21=%2A&audit.INTERNAL_ACTION.category%21=%2A';
+            break;
+        case 'no errors':
+            query = 'audit.ERROR.category%21=%2A';
+            break;
+        case 'no errors, compliant, and no warnings':
```
zoldello: Do you need to consider the case like: 'no errors, no warnings and compliant '? Like, will ordering always be like this?

src/encoded/static/components/datatable.js
```diff
@@ -139,7 +139,7 @@ const DataTable = ({ tableData }) => {
 
                                 // Cell's colSpan can be a specific number, 0 (full width), or
                                 // undefined (default of 1).
-                                let cellColSpan = (typeof cell === 'object') && cell.colSpan;
+                                let cellColSpan = (typeof cell === 'object' && cell !== null) && cell.colSpan;
```
zoldello: (Minor) Out of curiosity, why are you comparing cell to object; why not something like Node or Element?

src/encoded/static/components/view_controls.js
```diff
@@ -0,0 +1,325 @@
...
+};
+
+
+/**
+ * Displays view control buttons appropriate for the given search results.
+ */
+export const ViewControls = ({ results, filterTerm }) => {
+    // Add the hard-coded type filter if given.
+    let filters = [];
+    if (filterTerm) {
```
zoldello: (Minor) Maybe use ternary operator?

src/encoded/static/components/view_controls.js
```diff
@@ -0,0 +1,325 @@
...
+        }
+
+        // No download button if "type=" for an allowed type doesn't exist in query string.
+        const hasDownloadDocType = results.filters.some(filter => filter.field === 'type' && BATCH_DOWNLOAD_DOC_TYPES.includes(filter.term));
+        if (!hasDownloadDocType) {
+            return null;
+        }
+
+        // No download button if no files.
+        const hasFiles = results.facets.some(facet => facet.field === 'files.file_type' && facet.total > 0);
```
zoldello: (Minor). Maybe facet.total should be first?

src/encoded/static/components/vis_defines.js
```diff
@@ -191,3 +202,260 @@ const browserOrder = [
...
+ */
+const ucscUrlGenerator = (assembly, batchHubUrl) => {
+    const details = ucscAssemblyDetails[assembly];
+    if (details) {
+        return {
+            visualizer: 'UCSC',
+            url: `http://genome.ucsc.edu/cgi-bin/hgTracks?hubClear=${batchHubUrl}&db=${details.mappedAssembly}`,
+        };
+    }
+    return null;
```
zoldello: (Minor) Is this needed in that functions return undefined by default.

src/encoded/static/components/__tests__/search-controls.js
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
+import { mount } from 'enzyme';
+import _ from 'underscore';
```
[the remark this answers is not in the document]
forresttanaka: I made ENCD-4864 to address this. I have no issues converting to lodash, though I don’t know what’s involved. My understanding right now is this wouldn’t cause many difficulties.

src/encoded/static/components/auditmatrix.js
```diff
@@ -0,0 +1,617 @@
...
+    'audit.NOT_COMPLIANT.category',
+    'audit.ERROR.category',
+    'audit.INTERNAL_ACTION.category',
+];
+
+/** Maximum number of selected items that can be visualized. */
+const VISUALIZE_LIMIT = 500;
+
+
+/**
```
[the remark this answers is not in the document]
forresttanaka: Yeah, I’m debating with myself on that. I find the JSDoc syntax for simple constants a little ponderous, but maybe I should just suck it up. We don’t currently have any requirements for JSDoc, though anyone who wants to change the front-end coding requirements can propose anything for discussion and adoption.

src/encoded/static/components/auditmatrix.js
```diff
@@ -0,0 +1,617 @@
...
+ *
+ * @return {object} Summary information about given matrix data:
+ *     {
+ *         {number} maxSubCategoryValue: Maximum value in all cells in all subcategory rows
+ *         {number} minSubCategoryValue: Minimum value in all cells in all subcategory rows
+ *     }
+ */
+const analyzeSubCategoryData = (subCategoryData, columnCategoryType) => {
+    let maxSubCategoryValue = 0;
+    let minSubCategoryValue = Number.MAX_VALUE;
```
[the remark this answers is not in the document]
forresttanaka: Both `maxSubCategoryValue` and `minSubCategoryValue` have “impossible” values initially just as initial conditions, so we know they’ve been changed in the loop below.

src/encoded/static/components/auditmatrix.js
```diff
@@ -0,0 +1,617 @@
...
+    subCategoryData.forEach((rowData) => {
+        const rowDataValues = rowData[columnCategoryType].buckets.map(rowItem => rowItem.doc_count);
+
+        // Update min and max values found within all subcategories of the given category.
+        const prospectiveMax = Math.max(...rowDataValues);
+        if (maxSubCategoryValue < prospectiveMax) {
+            maxSubCategoryValue = prospectiveMax;
+        }
+        const prospectiveMin = Math.min(...rowDataValues.filter(value => value));
+        if (minSubCategoryValue > prospectiveMin) {
```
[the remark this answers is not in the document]
forresttanaka: Initially, `minSubCategoryValue` is guaranteed to have a value greater than (or in an extremely unlikely case, equal to) `prospectiveMin`, and this successful initial test changes `minSubCategoryValue` to one of the matrix values.

src/encoded/static/components/auditmatrix.js
```diff
@@ -0,0 +1,617 @@
...
+let _navbarHeight = null;
+
+/**
+ * Get height of the navbar.
+ *
+ * @return {number} Height of navbar on first call and restores same value until page is garbage
+ *                  collected
+ */
+const getNavbarHeight = () => {
+    if (_navbarHeight === null) {
```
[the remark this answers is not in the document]
forresttanaka: It doesn’t look like _navbarHeight can be undefined. I think this code comes from the auto-scroll you worked on? Am I wrong about that?

src/encoded/static/components/auditmatrixv1.js
forresttanaka: That would be up to wranglers. They came up with these categories.

src/encoded/static/components/datatable.js
forresttanaka: The fix is that in Javascript, `typeof null` returns “object.” I added the `cell !== null` to take care of that case.

src/encoded/static/components/view_controls.js
forresttanaka: Yeah, it could.

src/encoded/static/components/view_controls.js
forresttanaka: Would that be better? The hope being the first test fails more than the second so that the test would short circuit, and I think `facet.field === 'files.file_type'` would fail more than `facet.total > 0`.

src/encoded/static/components/vis_defines.js
forresttanaka: I’m not a big fan of returning nothing from a function that’s supposed to return something, but I have no issues with those who do that.

buildout.cfg
```diff
@@ -61,7 +61,7 @@ webtest = git https://github.com/Pylons/webtest.git
 WSGIProxy2 = git https://github.com/lrowe/WSGIProxy2.git
 zope.sqlalchemy = git https://github.com/zopefoundation/zope.sqlalchemy.git
 pytest-bdd = git https://github.com/lrowe/pytest-bdd.git branch=allow-any-step-order
-snovault = git https://github.com/ENCODE-DCC/snovault.git rev=1.0.34
+snovault = git https://github.com/ENCODE-DCC/snovault.git branch=SNO-113-make-audit-view
```
hitz: this should have been a distinct commit.  Since I am about to do the RC I can patch over it.

PR #2934, opened 2019-09-26
ENCD-4822-update-hgnc-file

bonitalam:


src/encoded/commands/generate_annotations.py
```diff
@@ -43,6 +43,8 @@ def assembly_mapper(location, species, input_assembly, output_assembly):
     except:
         return('', '', '')
     else:
+        if 'mappings' not in new_response:
+            print(new_response)
```
gabdank: @bonitalam can you please take out this condition and print statement, it was ok for troubleshooting, but should not remain in the code

PR #2935, opened 2019-09-30
ENCD-4871 Fix summary page experiment graph

forresttanaka:
The issue is that a function that mutated its argument in moment.js does not in dayjs.


hitz:
I have to pull this in after ENCD-4803? @forresttanaka

forresttanaka:
The order between those branches doesn’t matter — they’re completely independent.

PR #2937, opened 2019-10-01
ENCD-4722 install Valis as component

emro:
We want to load the Valis genome visualizer as a component.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,18 +14,18 @@
       }
     },
     "@babel/core": {
-      "version": "7.5.5",
-      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.5.5.tgz",
-      "integrity": "sha512-i4qoSr2KTtce0DmkuuQBV4AuQgGPUcPXMr9L5MyYAtk06z068lQ10a4O009fe5OB/DfNV+h+qqT7ddNV8UnRjg==",
+      "version": "7.6.4",
+      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.6.4.tgz",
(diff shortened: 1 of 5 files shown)


package-lock.json
```diff
@@ -4136,20 +4603,17 @@
...
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "console-control-strings": {
           "version": "1.1.0",
           "bundled": true,
-          "dev": true,
-          "optional": true
```
emro: I can go through and undo these changes if they're not desired.

webpack.config.js
```diff
@@ -52,6 +52,7 @@ const loaders = [
             path.resolve(__dirname, 'node_modules/superagent'),
         ],
         loader: 'babel',
+        query: { compact: false },
```
emro: [Here](https://stackoverflow.com/questions/29576341/what-does-the-code-generator-has-deoptimised-the-styling-of-some-file-as-it-e) is why I added this.

package-lock.json
forresttanaka: My impression right now is that this is just the new way of things in npm world.

forresttanaka: 👍

webpack.config.js
forresttanaka: 👍

emro:
@forresttanaka I think you approved this because it was in QA Ready but somehow Github didn't have it labeled as approved.

forresttanaka:
I remember reviewing this. Maybe I forgot to approve it at that time? I’m not sure.

PR #2938, opened 2019-10-01
ENCD-4874-change-apache-conf

hitz:
This is to be hotfixed on 10/1/2019


hitz:
Not sure if we should close this or not as it's superseded by other work.

caseylitton:
Closing there is an arg in deploy.py that allows you to set the apache workers.

PR #2941, opened 2019-10-02
ENCD-4880 Restore disabled button styling on certain buttons

forresttanaka:
Use background-color regardless of what button/link is being disabled, instead of opacity.


src/encoded/static/scss/encoded/_base.scss
```diff
@@ -258,9 +259,7 @@ $disabled-color-factor: 20%;
 	}
 }
 
-input[disabled],
-button[disabled],
-button[aria-disabled=true] {
+input[disabled] {
```
emro: We don't want the 🚫showing up on hover for disabled buttons?

PR #2949, opened 2019-10-04
ENCD-4858 Remove bolded line below certain raw data tables

forresttanaka:
I redid Raw Data table border drawing so that it doesn’t work like Raw Sequencing Data table border drawing, because I had wrongly assumed before that they’re basically the same.

While Raw _Sequencing_ Data tables can have multiple pairs that share a replicate/library combination, and therefore should have bold lines between the pairs, Raw Data tables can’t have multiple groups that share a replicate/library combination — the groups are _defined_ by a shared replicate/library. Therefore, you can’t have a bold line that only underlines a group of files; the bold line has to separate groups and should therefore span the entire row.

So you’ll now see the bold line between groups spanning the entire row, and never see a bold line between files, as you can see in Raw Sequencing Data tables.

You’ll also never see a bold line at the bottom of groups, whether ungrouped files appear below them or not.

### DIFF
--- a/src/encoded/static/components/filegallery.js
+++ b/src/encoded/static/components/filegallery.js
@@ -306,7 +306,6 @@ FileTable.defaultProps = {
     setInfoNodeId: null,
     setInfoNodeVisible: null,
     session: null,
-    session_properties: null,
     adminUser: false,
     schemas: null,
     noDefaultClasses: false,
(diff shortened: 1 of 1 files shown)


src/encoded/static/components/filegallery.js
```diff
@@ -779,15 +778,19 @@ class RawFileTable extends React.Component {
...
+                                const groupBottom = j < groupKeys.length - 1 ? 'group-bottom' : '';
 
                                 // Render each file's row, with the biological replicate and library
                                 // cells only on the first row.
                                 return groupFiles.sort((a, b) => (a.title < b.title ? -1 : 1)).map((file, i) => {
-                                    const groupBottom = i === (groupFilesLength - 1) ? 'group-bottom' : '';
+                                    // Determine whether to draw a bottom border based on whether
+                                    // this is the last file in a group (do draw) or not (don't
+                                    // draw) and it's not within the last group (don't draw).
+                                    const fileBottom = (i === groupFiles.length - 1 && groupBottom) ? 'group-bottom' : '';
```
emro: I found the naming a little confusing here. `groupBottom` indicates if the bottom of the group should have a bolded line (and is not within the bottom / last group) whereas I would expect it to mean that it's the bottom group (and therefore should not have a bolded line). I think it's fine to leave it as is, though.

PR #2951, opened 2019-10-08
ENCD-4891 Fix region search Javascript crash

forresttanaka:
Not yet sure if this will force a v9rc4 or not. I think it can go into a hotfix for rc3.

The root cause is I forgot about region search, so region_search.js used the older, now-non-existent batch download code. This updates it to use the newer code. I also had to modify search.js because `clear_filters` doesn't exist in region search results.

### DIFF
--- a/src/encoded/static/components/region_search.js
+++ b/src/encoded/static/components/region_search.js
@@ -1,11 +1,11 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import url from 'url';
-import { BrowserSelector } from './objectutils';
 import { Panel, PanelBody } from '../libs/ui/panel';
 import { FacetList, Listing } from './search';
 import { FetchedData, Param } from './fetched';
 import * as globals from './globals';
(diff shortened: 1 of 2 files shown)


src/encoded/static/components/region_search.js
```diff
@@ -1,11 +1,11 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import url from 'url';
-import { BrowserSelector } from './objectutils';
```
emro: I think that `BrowserSelector` from `objectutils.js` isn't used anywhere else, so the component could probably be deleted.

forresttanaka: I already deleted `BrowserSelector` from objectutils.js last release I think. It only exists in vis_defines.js now.

emro: Got it, sounds good

hitz:
superseded by #2955

PR #2955, opened 2019-10-08
ENCD-4891 Region search crash fix plus position

forresttanaka:
Same as ENCD-4891-region-crash branch but with a `results.coordinates` processor to generate a “chr1:012345678-012345678” string to add to region search visualization URLs for UCSC. Adds that processing to `BrowserSelector` and changes `ucscUrlGenerator` to use that string.

### DIFF
--- a/src/encoded/static/components/vis_defines.js
+++ b/src/encoded/static/components/vis_defines.js
@@ -269,17 +269,15 @@ const ucscAssemblyDetails = {
 /**
  * Generate a batch hub URL for UCSC based on the given assembly.
  * @param {string} assembly Assembly to use for the generated URL
- * @param {object} {
- *      visualizer {string} "UCSC" visualizer name to use in UI
- *      url {string} URL to pass to UCSC to visualize the given assembly
- * }
+ * @param {string} batchHubUrl Batch hub URL
+ * @param {string} position Optional region search position string
  */
-const ucscUrlGenerator = (assembly, batchHubUrl) => {
+const ucscUrlGenerator = (assembly, batchHubUrl, position) => {
     const details = ucscAssemblyDetails[assembly];
     if (details) {
         return {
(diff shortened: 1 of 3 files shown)


src/encoded/static/components/vis_defines.js
```diff
@@ -374,12 +369,24 @@ export class BrowserSelector extends React.Component {
...
 
+        // If coordinates are given, construct a "position" query string parameter expanded by 200
+        // bp in either direction.
+        let position = '';
+        if (results.coordinates) {
+            const matches = results.coordinates.match(/(.+):(\d+)-(\d+)/);
+            if (matches) {
+                const lowCoordinate = Number(matches[2]);
+                const highCoordinate = Number(matches[3]);
+                position = `${matches[1]}:${lowCoordinate - 200}-${highCoordinate + 200}`;
```
emro: Optional suggestions: for the genome browser, I don't have a range hard-coded, I take the length of the coordinate range and divide it by 2 and add and subtract it from the high and low coordinates to get the display range. So that's an option. The other thing you could consider is making `200` a parameter and listing it at the top of the component.

forresttanaka: Yeah, making 200 a constant crossed my mind but…I didn’t feel like it. I followed the back end adding and subtracting 200, but if people want to change that, it won’t be difficult of course. Given Ben’s comment on Slack and the complete lack of automated testing (Keenan opened a ticket to add region search QANCODE tests yesterday) it seems like region search gets little 💙.

hitz: yes a "scalable" range is better than a fixed 400 would be better, but as I said on slack, this code never really made it out of beta.   It just doesn't really do what it's supposed to do and the ES database should be calved off a la regulomeDB.

PR #2958, opened 2019-10-09
ENCD-4849-fix-description-misalignment-in-docs

zoldello:


zoldello:
@emro Added height to prevent description box from changing size. I tried "rem" instead of "px", but the icon to the far right got misaligned on certain screen sizes.

emro:
@zoldello I can't remember what was in the previous commit that I reviewed and I'm a little confused by the comments in the ticket. What does this commit fix that the other one didn't?

PR #2960, opened 2019-10-10
ENCD-4731 Add audits for experiment series

jenjou:


src/encoded/audit/experiment_series.py
```diff
@@ -0,0 +1,167 @@
...
+			if dataset['target'] in targets:
+				targets[dataset['target']].append(dataset['@id'])
+			else:
+				targets[dataset['target']] = [dataset['@id']]	
+
+		if 'replicates' in dataset:
+			for replicate in dataset['replicates']:
+				if replicate['status'] not in ['deleted'] and \
+				'library' in replicate and 'biosample' in replicate['library']:
+					biosample = replicate['library']['biosample']
```
yunhailuo: I'm not sure but
1. Should the status check against `excluded_statuses`? I mean replicate can be "revoked" (cannot be "replaced" though)?
2. Should we check status on library and biosample as well?

src/encoded/audit/experiment_series.py
```diff
@@ -0,0 +1,167 @@
...
+	treatments = {'no treatment': []}
+	genetic_modifications = {'no genetic modification': []}
+
+	filtered_datasets = [dataset for dataset in value['related_datasets'] if dataset['status'] not in excluded_statuses]
+
+	for dataset in filtered_datasets:
+		if dataset['assay_title'] in assays:
+			assays[dataset['assay_title']].append(dataset['@id'])
+		else:
+			assays[dataset['assay_title']] = [dataset['@id']]
```
yunhailuo: If you use list here and below, you might end up with non-unique list which will give you messages like
> Biosamples of Experiments ENCSR960UYN and ENCSR960UYN are from donor ENCDO271OUW.

src/encoded/audit/experiment_series.py
jenjou: Spoke with Idan and he agrees status should be checked for replicate/library/biosample. I will update this.

src/encoded/audit/experiment_series.py
jenjou: Ok. I can modify it to use a set() instead.

PR #2961, opened 2019-10-10
ENCD-4816 Summary release date source change

forresttanaka:
* The main part of this involves using `date_released` as a data source instead of the now-defunct `month_released` so that we don’t always display no data for experiment releases.
* I also removed the second parameter from dayjs() because that’s not supported unless you use the CustomParseFormat plugin, which we don’t include right now. dayjs() seems fine interpreting our dates without that.
* I removed the adjustment to date-submitted data because that often drives those numbers negative — we often release experiments submitted before the start date of a summary search, causing this to look like we’ve released more experiments than have been submitted. Now both released and submitted experiments accumulate over time with no adjustments.

### DIFF
--- a/src/encoded/static/components/award.js
+++ b/src/encoded/static/components/award.js
@@ -1746,16 +1746,12 @@ const consolidateSortedDates = (dateTerms) => {
 export const ExperimentDate = (props) => {
     const { experiments, award, panelCss, panelHeadingCss } = props;
     let releasedDates = [];
-    let submittedDates = [];
     let deduplicatedreleased = {};
-    let deduplicatedsubmitted = {};
 
     // Search experiments for date_released and date_submitted in facets
     if (experiments.facets && experiments.facets.length > 0) {
         const dateReleasedFacet = experiments.facets.find(facet => facet.field === 'date_released');
-        const dateSubmittedFacet = experiments.facets.find(facet => facet.field === 'date_submitted');
         releasedDates = (dateReleasedFacet && dateReleasedFacet.terms && dateReleasedFacet.terms.length > 0) ? dateReleasedFacet.terms : [];
-        submittedDates = (dateSubmittedFacet && dateSubmittedFacet.terms && dateSubmittedFacet.terms.length > 0) ? dateSubmittedFacet.terms : [];
     }
 
     // Take an array of date facet terms and return an array of terms sorted by date.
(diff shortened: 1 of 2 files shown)


forresttanaka:
The change from last time is that we no longer use `date_submitted` in the summary chart anymore — we only show `date_released` in the chart.

src/encoded/static/components/award.js
```diff
@@ -1791,50 +1787,35 @@ export const ExperimentDate = (props) => {
...
         let awardStartDate;
         if (award && award.start_date) {
             awardStartDate = dayjs(award.start_date).format('YYYY-MM');
         } else {
             const releasedIndex = sortedreleasedTerms.findIndex(item => item.doc_count);
-            const submittedIndex = sortedsubmittedTerms.findIndex(item => item.doc_count);
             const earliestReleased = releasedIndex > -1 ? sortedreleasedTerms[releasedIndex].key : sortedreleasedTerms[sortedreleasedTerms.length - 1];
-            const earliestSubmitted = submittedIndex > -1 ? sortedsubmittedTerms[submittedIndex].key : sortedsubmittedTerms[sortedsubmittedTerms.length - 1];
-            awardStartDate = earliestReleased < earliestSubmitted ? earliestReleased : earliestSubmitted;
+            awardStartDate = earliestReleased;
```
emro: You could just have: `const awardStartDate = releasedIndex...`

src/encoded/static/components/award.js
```diff
@@ -2045,11 +2025,6 @@ class CumulativeGraph extends React.Component {
                 data: {
                     labels: monthReleased,
                     datasets: [{
```
emro: I'm not very familiar with chartjs, but is there a simpler way to write this if there's only one dataset? Or is it better to keep it this way since it's possible that we'll want to add back submitted data at some point?

src/encoded/static/components/award.js
forresttanaka: 👍

src/encoded/static/components/award.js
forresttanaka: I’m not sure, but I’m not seeing anything significantly simpler in their samples, e.g. https://github.com/chartjs/Chart.js/blob/master/samples/charts/line/point-styles.html#L36

hitz:
Ah this is the one I accidentally did not squash merge!

PR #2962, opened 2019-10-10
ENCD-4862-fix-audit-script

bonitalam:


src/encoded/audit/experiment.py
```diff
@@ -412,7 +412,7 @@ def audit_experiment_mixed_libraries(value, system, excluded_types):
 
 def audit_experiment_pipeline_assay_details(value, system, files_structure):
     for pipeline in get_pipeline_objects(files_structure.get('original_files').values()):
-        if value.get('assay_term_name') not in pipeline['assay_term_names']:
+        if 'assay_term_names' not in pipeline or value.get('assay_term_name') not in pipeline['assay_term_names']:
```
gabdank: I am not sure how that will behave if "assay_term_names": [].

 I think it should be: 
```
pipeline_assays = pipeline.get('assay_term_names")
if not pipeline_assays or value.get('assay...') not in pipeline_assays: ...
```

src/encoded/tests/datafixtures.py
```diff
@@ -918,6 +918,17 @@ def pipeline_bam(testapp, lab, award, analysis_step_bam):
     return testapp.post_json('/pipeline', item).json['@graph'][0]
 
 
+@pytest.fixture
+def pipeline_noassay_bam(testapp, lab, award, analysis_step_bam):
```
gabdank: please rename the fixture to be something more clear, like "pipeline_without_assay_term_names"

bonitalam: Ok, will implement a clearer name.

PR #2966, opened 2019-10-11
SNO-115-delete-old-search-code

keenangraham:
Remove old search code.


src/encoded/cart_view.py
```diff
@@ -74,3 +79,13 @@ def cart_manager(context, request):
         'cart_user_max': CART_USER_MAX
     }
     return result
+
+
+@view_config(route_name='search_elements', request_method='POST')
```
hitz: Is this of more general utility?  Should it be moved out of cart_view.py?

PR #2969, opened 2019-10-14
ENCD-4894 Exclude long read RNA-seq from some audits

jenjou:


src/encoded/audit/experiment.py
```diff
@@ -3705,7 +3705,9 @@ def audit_library_RNA_size_range(value, system, excluded_types):
     if value['status'] in ['deleted', 'replaced']:
         return
 
-    if value.get('assay_term_name') == 'transcription profiling by array assay':
+    if value.get('assay_term_name') in ['transcription profiling by array assay',
+                                        'long read RNA-seq'
```
gabdank: you need trailing comma

src/encoded/tests/test_audit_experiment.py
```diff
@@ -1272,6 +1290,52 @@ def test_audit_experiment_with_RNA_library_array_size_range(testapp, base_experi
                for error in collect_audit_errors(res))
 
 
+def test_audit_experiment_with_long_read_RNA_no_size_range(testapp, base_experiment,
+                                                            base_replicate,
+                                                            base_library):
+    testapp.patch_json(base_library['@id'], {'nucleic_acid_term_name': 'RNA'})
+    testapp.patch_json(base_replicate['@id'], {'library': base_library['@id']})
```
gabdank: I think the setup of this test is not imitating real experiment, hence the audit for long read RNA is not triggered, I think if you were to substitute the long read RNA by just "RNA-seq" it would still pass. To make it really work you should do something like you did in the "test_audit_experiment_with_RNA_library_missing_read_length" test. That brings me to the following point: you will probably need to create a fixture for the experiment you want to test, and in the test body use this fixture, potentially only switching the pipeline title.

src/encoded/tests/test_audit_experiment.py
```diff
@@ -1272,6 +1290,52 @@ def test_audit_experiment_with_RNA_library_array_size_range(testapp, base_experi
                for error in collect_audit_errors(res))
 
 
+def test_audit_experiment_with_long_read_RNA_no_size_range(testapp, base_experiment,
+                                                            base_replicate,
+                                                            base_library):
```
gabdank: I think indentation here is missing one space.

PR #2970, opened 2019-10-14
ENCD-4890-library-target-size-range

bonitalam:


src/encoded/schemas/changelogs/library.md
```diff
@@ -1,5 +1,9 @@
 ## Changelog for library.json
 
+### Minor changes since schema version 9
+
+* *queried_RNP_size_range* property is added, the property captures ribonucleoprotein size range for eCLIP controls.
+
```
gabdank: it is indeed ribonucleoprotein size range, but not limited to eCLIP controls only - it could be specified in any eCLIP (and potentially other assays). Please omit the specification of eCLIP control. It can be something along the lines: "the property captures ribonucleoprotein size range targeted   in the assay, the library was constructed for."

src/encoded/schemas/library.json
```diff
@@ -282,6 +282,12 @@
             "type": "string",
             "pattern": "(^[0-9]+-[0-9]+$|^[<>][0-9]+$)"
         },
+        "queried_RNP_size_range": {
+            "title": "Queried RNP size range",
+            "description": "The measured size range of queried ribonucleoprotein (RNP) eCLIP control, in kD.",
+            "type": "string",
+            "pattern": "(^[0-9]+-[0-9]+$|^[<>][0-9]+$)"
```
gabdank: I think we should go with a restrictive regular expression and adjust it only if there will be need to do that. Please leave only ^[0-9]+-[0-9]+$ and not ^[<>][0-9]+$

src/encoded/schemas/library.json
```diff
@@ -282,6 +282,12 @@
             "type": "string",
             "pattern": "(^[0-9]+-[0-9]+$|^[<>][0-9]+$)"
         },
+        "queried_RNP_size_range": {
+            "title": "Queried RNP size range",
+            "description": "The measured size range of queried ribonucleoprotein (RNP) eCLIP control, in kD.",
```
gabdank: please shorten the description to be something like "The measured size range of queried ribonucleoprotein (RNP) in kD." and add "comment" where you can explain that for example this range could be specified in eCLIP experiments.

PR #2972, opened 2019-10-15
ENCD-4900 Add status check for matching md5 sum audit

jenjou:


src/encoded/audit/file.py
```diff
@@ -467,7 +467,7 @@ def audit_file_matching_md5sum(value, system):
                 )
             )
             yield AuditFailure('Incorrect matching_md5sum', detail, level='ERROR')
-        else:
+        elif file.get('md5sum') == value.get('md5sum') and file.get('status') in checked_statuses:
```
yunhailuo: You are having `elif` here. Is `file.get('md5sum') == value.get('md5sum')` needed here?

jenjou: Removed

PR #2975, opened 2019-10-17
ENCD-4840-Add-Donor-accession-boost-value

jnau:
Added boost values to donor subclasses - human, mouse, worm, and fly. started with value of 1.0 for accession and alternate_accessions. Added Donor to Default Item Types in search_views.py. donor.json was edited, but the change was trivial.


src/encoded/schemas/donor.json
```diff
@@ -64,9 +64,9 @@
         }
     },
     "boost_values": {
+        "accession": 1.0,
         "external_ids": 1.0,
         "aliases": 1.0,
-        "accession": 1.0,
         "alternate_accessions": 1.0
     }
```
gabdank: if there was no change I am not sure why the 'accession' moved form line 69 to 67?

jnau: Fixed this. "accession" was originally line 69, but I moved it to line 67 because I thought maybe order had something to do with priority. Obviously doesn't, and now it is moved back to line 69.

PR #2977, opened 2019-10-17
ENCD-3086-add-dataset-to-file

keenangraham:
Add some dataset properties to file via calculated_properties. Will have to see how long this takes to index.


src/encoded/types/file.py
```diff
@@ -476,6 +479,33 @@ def s3_uri(self):
...
+        return dataset.get('biosample_ontology')
+
+    @calculated_property(schema={
+        "title": "Target",
+        "type": "string",
+        "notSubmittable": True,
+    })
+    def target(self, request, dataset):
+        dataset = request.embed(dataset, '@@object')
+        return dataset.get('target')
```
hitz: I think this should just be target.label?

keenangraham: K, thought maybe there was some other interesting stuff in target but should probably try to keep it as slim as possible.

src/encoded/types/utils.py
```diff
@@ -0,0 +1,26 @@
+
```
hitz: are there other functions in types/* that could be moved here?

src/encoded/tests/test_types_utils.py
```diff
@@ -0,0 +1,130 @@
+import pytest
+
+
+@pytest.fixture
+def fastq_no_replicate(award, experiment, lab, platform1):
```
hitz: this is duplicated from test_file.py right?   Probably their should be a single fastq fixture in a shared place;  datafixtures.py?  Or if it's distinct, it should be a different name.

keenangraham: I pulled duplicates into datafixtures. Works fine.

src/encoded/types/utils.py
keenangraham: Hmm. I think I might just put these in https://github.com/ENCODE-DCC/snovault/blob/dev/src/snovault/util.py

keenangraham:
Moved utils to snovault branch.

PR #2984, opened 2019-10-22
SNO-97-new-target-matrix

zoldello:


src/encoded/search_views.py
```diff
@@ -195,6 +198,44 @@ def matrix(context, request):
     return fr.render()
 
 
+@view_config(route_name='target_matrix', request_method='GET', permission='search')
+def target_matrix(context, request):
+    assay_title_route = urllib.parse.quote(request.matchdict['assay_title_route'])
+    request.query_string += ('&assay_title=' + assay_title_route)
```
keenangraham: I know this is something @hitz recommended but I really don't like for several reasons: 

1. This is breaking encapsulation and will make maintaining and extending code on both sides (search view and hardcoded route) more difficult in the future.
2. This is also making the function do more than one thing which will make unit testing a headache (and hacking the request like this needs unit tests).
3. This is very similar to parsing the path that gets passed into batch_download which keeps breaking (we are trying to move to a query string solution).
4. The FieldedResponse matrix view is generic enough to be reused by several views (@forresttanaka is probably duplicating a lot of this already), but if we do it this way we'll just have to keep copying code over and over again.

A better design would be to create another view and manipulate the request all you want _before_ passing it into the matrix view, and make the matrix view generic enough to use whatever matrix_definition_name and facets etc. kwargs you pass in. In this case it probably makes sense to create a hardcoded _/target_matrix/tf_ and _/target_matrix/histone_ route with hardcoded query strings appended to the end (trying to parse the assay_title from the route isn't going to make a human-readable URL which I think is the whole point of doing it this way in the first place). 

So in general it should look something like this:

```
def generic_matrix(context, request, matrix_definition_name, at_type, title, facets=None):
    fr = FieldedResponse(
        BasicMatrixWithFacetsResponseField(
                default_item_types=DEFAULT_ITEM_TYPES,
                matrix_definition_name=matrix_definition_name
                facets=facets
        ),
        TitleResponseField(
                title=title
        ),
        TypeResponseField(
                at_type=at_type
        )
        ...
    )

def add_param_to_query_string(request, param):
    # Add param to request query string
    return request

@view_config(route_name='target_matrix/tf')
def target_matrix_tf(request, context):
       request = add_param_to_request('assay_title=TF Chip-seq`)
       return generic_matrix(context, request, matrix_definition_name, at_type, title, facets=None)

@view_config(route_name='target_matrix/histone')
def target_matrix_tf(request, context):
       request = add_param_to_request('assay_title=Histone`)
       return generic_matrix(context, request, matrix_definition_name, at_type, title, facets=None)
```

Then you can write good unit tests for add_param_to_query_string, as well as integration tests for the views.

forresttanaka: @keenangraham Have you documented these views somewhere? It more or less makes sense when I see an existing one, but I‘m certain I couldn’t for the life of me come up with a new one without documentation.

keenangraham: No, I've tried to make a variety of views that demonstrate different functionality, but reading snovault fields.py is probably still the best place to get an understanding of the internals. (To be a field you just have to implement a render method and return a dictionary. A lot of the fields are pretty simple.) On the plus side every function does have a unit test, and every view does have an integration test, which in a sense documents how the code should be used.

zoldello:
@forresttanaka @keenangraham  Thanks. I will look at the comments. This was no code review ready though as I expected more functionality changes (it is still a draft PR.) I will take out the draft status when it is ready for code review.

src/encoded/search_views.py
hitz: Then if we want to extend BasicMatrix ResponseField with different nestings (i.,e 2 X 1 Y or 2 X 2Y or 3 Y 2 X or whatever... is that a different Response field?

keenangraham: Groupings are defined on the type so you would just pass in a new `matrix_definition_name`.

zoldello: This sounds like a good idea. However, it does not seems flexible as every new assay title needs a newly added function.

keenangraham: You are right but in this case I think it makes sense for the following reasons:

1. Trying to avoid the pain of parsing params like this: `urllib.parse.quote(request.matchdict['assay_title_route'])` Like you said this already has different behavior if you pass in `TF%20ChIP-seq` versus `TF+ChIP-seq`. Eventually you run into a case that really doesn't work (e.g. `bed bed3+`) as we've seen in batch_download params.

2. `/targetmatrix/TF%20ChIP-seq/?type=Experiment` or `/targetmatrix/TF+ChIP-seq/?type=Experiment` is arguably worse than just being consistent and passing a query_string to the matrix endpoint: `/matrix/?type=Experiment&assay_title=TF+ChIP-seq`

3. There's only two options (tf and histone) and these aren't likely to increase given that it's a target matrix. So if we really want something that humans can type in we should make it as simple as possible. Hence:`/targetmatrix/tf` and `/targetmatric/histone`. If you want to continue to use match_dict and map tf -> TF+ChIP-seq and histone -> Histone+ChIP-seq inside the function I don't have a problem with that.

zoldello:
The jira ticket number  changed, so I had to create a new branch. The new branch is https://github.com/ENCODE-DCC/encoded/pull/2987

PR #2996, opened 2019-10-24
ENCD-4904-add-ingrid-as-user

iyoungworth:
I copy-pasted Bonita's user schema, substituted my own details, and removed the uuid line in user.json

### DIFF
--- a/src/encoded/tests/data/inserts/user.json
+++ b/src/encoded/tests/data/inserts/user.json
@@ -1155,5 +1155,30 @@
             "ENCORE"
         ],
         "uuid": "64974e05-12df-4733-ae90-b402298b142e"
+    },
+    {
+        "email": "iay@stanford.edu",
+        "first_name": "Ingrid",
+        "groups": [
+            "admin",
+            "verified"
+        ],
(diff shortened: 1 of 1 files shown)


src/encoded/tests/data/inserts/user.json
```diff
@@ -1155,5 +1155,29 @@
...
+            "/labs/j-michael-cherry/"
+        ],
+        "viewing_groups": [
+            "community",
+            "ENCODE3",
+            "ENCODE4",
+            "GGR",
+            "REMC",
+            "ENCORE"
+        ],
```
caseylitton: This trailing comma will break.  Json files are quite strict on  syntax.  There should be a way to added json syntax to your text editor.  Mine shows errors like this in red when I make them.  Maybe one of the wrangler has this setup in sublime already?    

FYI:  We recommend trailing commas in python files however.

iyoungworth: Thanks! I will remove it. I haven't finished setting up the linters in Sublime that are recommended, there might be something for JSON in those?

PR #2998, opened 2019-10-26
SNO-125-switch-batch-download-to-query-string

keenangraham:


src/encoded/region_search.py
```diff
@@ -816,6 +816,6 @@ def search_result_actions(request, doc_types, es_results, position=None):
             bucket_has_doc):
         actions['batch_download'] = request.route_url(
             'batch_download',
-            search_params=request.query_string
+            search_params='?' + request.query_string
```
hitz: I don't think region-search batchdownload works; I think it's disabled in the UI.  I can't remember if it's because it's buggy (throws 500s) or is too slow (timeouts)

src/encoded/tests/test_batch_download.py
```diff
@@ -423,34 +386,42 @@ def test_batch_download_lookup_column_value(lookup_column_value_item, lookup_col
...
...nhttp://localhost/files/ENCFF003COS/@@download/ENCFF003COS.bigBed\nhttp://localhost/files/ENCFF006EPI/@@download/ENCFF006EPI.fastq.gz\nhttp://localhost/files/ENCFF164JWT/@@download/ENCFF164JWT.fastq.gz\nhttp://localhost/files/ENCFF032OJW/@@download/ENCFF032OJW.bam\nhttp://localhost/files/ENCFF182CHA/@@download/ENCFF182CHA.bed.gz\nhttp://localhost/files/ENCFF007EPI/@@download/ENCFF007EPI.fastq.gz\nhttp://localhost/files/ENCFF001CON/@@download/ENCFF001CON.bam\nhttp://localhost/files/ENCFF003CON/@@download/ENCFF003CON.bam\nhttp://localhost/files/ENCFF001REP/@@download/ENCFF001REP.bam\nhttp://localhost/files/ENCFF003REP/@@download/ENCFF003REP.bam\nhttp://localhost/files/ENCFF008EPI/@@download/ENCFF008EPI.bam\nhttp://localhost/files/ENCFF959VGP/@@download/ENCFF959VGP.fastq.gz\nhttp://localhost/files/ENCFF477MLC/@@download/ENCFF477MLC.fastq.gz\nhttp://localhost/files/ENCFF000LBC/@@download/ENCFF000LBC.csqual.gz\nhttp://localhost/files/ENCFF000LBB/@@download/ENCFF000LBB.csqual.gz\nhttp://localhost/files/ENCFF000LBA/@@download/ENCFF000LBA.csfasta.gz\nhttp://localhost/files/ENCFF000LAZ/@@download/ENCFF000LAZ.csfasta.gz\nhttp://localhost/files/ENCFF010LAO/@@download/ENCFF010LAO.csfasta.gz'),
```
hitz: I guess you think these tests are useless?

src/encoded/tests/test_batch_download.py
```diff
@@ -423,34 +386,42 @@ def test_batch_download_lookup_column_value(lookup_column_value_item, lookup_col
...
-    ('/metadata/type=Annotation&organism.scientific_name%21=Homo+sapiens&organism.scientific_name=Mus+musculus&format=json/metadata.tsv','ENCFF015OKV\tbed enhancer predictions\tpredicted heart enhancers\tENCSR356VQT\tlong-range chromatin interactions\tSamtools, Picard\t2\tUBERON:0000948\theart\ttissue\tembryonic\t11.5\tday\tMus musculus\t\t2015-02-27\tENCODE\tJ. Michael Cherry, Stanford\ta5f849c78025c80dca58771492c97318\t\thttp://localhost/files/ENCFF015OKV/@@download/ENCFF015OKV.bed.gz\tmm10\t\treleased\t\t\t1544154147\tmissing analysis_step_run\tmismatched status, missing derived_from','File accession\tFile format\tOutput type\tDataset accession\tAnnotation type\tSoftware used\tEncyclopedia Version\tBiosample term id\tBiosample term name\tBiosample type\tLife stage\tAge\tAge units\tOrganism\tTargets\tDataset date released\tProject\tLab\tmd5sum\tdbxrefs\tFile download URL\tAssembly\tControlled by\tFile Status\tDerived from\tS3 URL\tSize\tAudit WARNING\tAudit INTERNAL_ACTION\tAudit NOT_COMPLIANT\tAudit ERROR\r',),
```
hitz: and these?

src/encoded/tests/test_batch_download.py
```diff
@@ -478,9 +449,9 @@ def test_metadata_tsv_fields(testapp, workbook):
         _tsv_mapping,
         _excluded_columns,
     )
-    r = testapp.get('/metadata/type%3DExperiment/metadata.tsv')
-    metadata_file = r.body.decode('UTF-8').split('\n')
-    actual_headers = metadata_file[0].split('\t')
-    assert len(actual_headers) == len(set(actual_headers))
+    r = testapp.get('/metadata/?type=Experiment')
```
hitz: are you changing the URL here?   There's no longer a metadata.tsv at the end?   We will have to forward them somehow.  People can have old files.txt

src/encoded/region_search.py
keenangraham: That's good. Just fixing any mention for consistency.

src/encoded/tests/test_batch_download.py
keenangraham: They are sort of useful but they aren't quoted correctly (should have probably been caught in review). They are also a mess to deal with. Changing one letter invalidated the whole string.

src/encoded/tests/test_batch_download.py
keenangraham: This also aren't quoted correctly: `/metadata/type=Experiment&biosample_ontology.term_name!=basal cell of epithelium of terminal bronchiole&format=json/metadata.tsv` All the links pass in URL encoded URLs. These have blank spaces in them.

src/encoded/tests/test_batch_download.py
keenangraham: Yes! The URL is changing. Moving from params as part of the path to query strings is by definition changing both URLs. If we keep a redirect that accepts search params we will run into the same double unquote URL file type bug that we are trying to avoid by moving to query strings. I removed the metdata.tsv ending because it's not consistent with batch_download (which also returns a file but doesn't have a batchdownload.txt/tsv ending).

hitz: So we will have to rewrite the URL at the apache level for backwards compatibility.

keenangraham: Added.

PR #3000, opened 2019-10-29
reduce overhead of teardown

Bek:


src/encoded/tests/test_indexing.py
```diff
@@ -101,13 +104,6 @@ def test_indexing_simple(testapp, indexer_testapp):
     assert res.json['txn_count'] == 1
     assert res.json['updated'] == [uuid]
     res = testapp.get('/search/?type=TestingPostPutPatch')
-    uuids = [indv_res['uuid'] for indv_res in res.json['@graph'] if 'uuid' in indv_res]
```
hitz: Why did we take this part out?

PR #3013, opened 2019-11-04
SNO-118-simple-query-string

keenangraham:


keenangraham:
> No idea what happens if you have an advancedQuery with a date AND something else, but it's not supported anyhoo.

Should work.

PR #3015, opened 2019-11-04
ENCD-2529-fix-default-required-properties

bonitalam:


src/encoded/schemas/experiment_series.json
```diff
@@ -47,8 +46,17 @@
             "default": "1"
         },
         "related_datasets": {
```
gabdank: can't you still use the seiries properties with the $ref, and only specify the "minItem" in the overwritten related_datasets? I am not sure it will work though.

src/encoded/tests/test_audit_dataset.py
```diff
@@ -1,6 +1,8 @@
 import pytest
 
 
+# References are required and set to have at least one item, no empty lists are allowed
```
gabdank: if this test is no longer going to be needed because a dependency was introduced you should delete the test, and add another test in the schema testing file to test your dependency and schema validation

src/encoded/schemas/experiment_series.json
bonitalam: I think other series use this related_datasets property such as MatchedSet and OrganismDevelopmentSeries, which have empty lists. I don't think I can remove the "default: []" from series properties? Correct me if I'm wrong

bonitalam: I tested out using it as a $ref with "minItem: 1" but the default [] still remained, so I will have to go back to my original approach.

PR #3017, opened 2019-11-06
ENCD-4932 Fix links and formatting in Experiment Series audit

jenjou:


src/encoded/audit/experiment_series.py
```diff
@@ -150,10 +173,17 @@ def audit_mismatched_properties(value, system):
 			audit_link(path_to_text(value['@id']),value['@id']))
 		for treatment_key in treatments:
 			expt_list = generate_formatted_list_of_experiments(treatments[treatment_key])
+			if treatment_key is 'no treatment':
+				link = treatment_key
+			else:
+				formatted_links = []
+				for label_id in treatment_key:
+					formatted_links.append(audit_link(label_id[0], label_id[1]))
+				link = ', '.join(formatted_links)
```
yunhailuo: I'm not sure if this can be changed but I don't feel good about this because it depends on `label_id` being a tuple with specific 0 and 1 item which is defined ~100 lines above. I worried this is not good for code maintenance. The previous version has just strings which may end up bad in text but won't break codes if the string goes wrong. I have an intention to move all `audit_link` thing up closer to `label_id` so that it gets a simple string down here. But the logic may end up even more twisted...

jenjou: Initially I was converting the treatment info into lists of strings and then joining the lists to get one long string as a dictionary key, which I then split() back down in this section. I ended up switching to tuples because I thought it would be less risky, in case someone enters a value for treatment_term_name containing the same delimiters I used to join() leading to incorrect split(). Not really sure what's the best method here. But, if we eliminate the link on the treatments, I think then I can avoid using tuples in this section.

yunhailuo: The code is not bad. It's just I felt there might be a better way. I couldn't think of a better way by far. It's OK to go this way.

PR #3021, opened 2019-11-06
ENCD-4898-add-pipeline-assay-audit

bonitalam:


src/encoded/tests/data/inserts/pipeline.json
```diff
@@ -178,5 +178,16 @@
...
+    {
+        "accession": "ENCPL001NAT",
+        "analysis_steps": [
+            "align-star-step-v-1"
+        ],
+        "status": "released",
+        "title": "RBNS pipeline",
+        "lab": "brenton-graveley",
+        "award": "U54HG007005",
+        "uuid": "d71cee3c-1d49-46f1-b4d6-45e0848a0f33"
```
gabdank: have you created this insert to show that the audit will appear on the local demo? If that is the reason I don't think you should do it. You should definitely write a test for the audit triggering in the correct circumstances, but you should not create "wrong" inserts for the sake of audit triggering.

src/encoded/audit/pipeline.py
```diff
@@ -23,5 +23,14 @@ def audit_analysis_steps_closure(value, system):
...
 
 
+@audit_checker('Pipeline', frame='object')
+def audit_pipeline_assay_term_names(value, system):
+    ''' All pipelines should have the assay_term_names property specified.
+    '''
+    if 'assay_term_names' not in value:
+        detail = ('Pipeline {} has no assay_term_names specified.'.format(audit_link(path_to_text(value['@id']), value['@id'])))
+        yield AuditFailure('missing assay_term_names', detail, level='ERROR')
+
```
gabdank: I think the audit should trigger even if you have "assay_term_names" but it is an empty list.

src/encoded/tests/data/inserts/pipeline.json
bonitalam: Will remove the insert

src/encoded/tests/data/inserts/pipeline.json
```diff
@@ -166,17 +166,5 @@
         "award": "U54HG007005",
         "title": "Test STARR-seq pipeline",
         "uuid": "5fe3fdf6-5b1a-4aca-8116-ded7acf13037"
-    },
```
gabdank: I am not sure why this pipeline was removed form the inserts?

bonitalam: I must've accidentally deleted it, will add it back

PR #3025, opened 2019-11-06
ENCD-4958 Group PR of 5 backend changes

caseylitton:
A bunch of isolated backend changes.  Some are not really testable(4908/4939) so as long as the machine doesn't fall over on deployment or usage.  We'll want to load test this and check logs(4938) . QA machine should be a cluster so we can add a new node(4937).  QA Cluster should be deployed from a fake tag(4947).  4911 can be tested directly by bonita/idan as with v91x0.

4947: Deployment bug fix that only impact bin/deploy where -b arg is a tag.
4911:  Makes upload limit 5gb.  This was already running in v91x0-test and v91x0 prod.
4937: Rare occasion we'll need to add one node to an already existing cluster.  This was used to add a new node to v91x0.
4938: There should be a new log /var/log/apache/forensic_log.log.  And a key in the current access.log to reference the incoming request in forensic_log
4939: Not really testable, process that hit memlimit will die more gracefully.

***
MOVED to https://github.com/ENCODE-DCC/encoded/pull/3037
4908: Not really testable, but was turned on in v91x0 as part of troubleshooting issues during that release period.

### DIFF
--- a/cloud-config/config-build-files/cc-parts/CC_RUNCMD_DEMO.yml
+++ b/cloud-config/config-build-files/cc-parts/CC_RUNCMD_DEMO.yml
@@ -1,13 +1,13 @@
 runcmd:
 - sudo -u ubuntu git clone %(GIT_REPO)s /home/ubuntu/encoded
-- sudo -u ubuntu git -C /home/ubuntu/encoded checkout -b %(GIT_BRANCH)s origin/%(GIT_BRANCH)s
+- sudo -u ubuntu git -C /home/ubuntu/encoded checkout -b %(GIT_BRANCH)s %(GIT_REMOTE)s/%(GIT_BRANCH)s
 - sudo -u ubuntu %(CC_DIR)s/post-boot-install.sh %(S3_AUTH_KEYS)s
 - sudo -u ubuntu %(CC_DIR)s/java11-oracle-install.sh
 - %(CC_DIR)s/redis-install.sh %(REDIS_PORT)s
 - %(CC_DIR)s/es-install.sh NONE %(JVM_GIGS)s %(ES_OPT_FILENAME)s
 - %(CC_DIR)s/cloudwatchmon-install.sh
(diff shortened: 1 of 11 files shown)


cloud-config/config-build-files/cc-parts/CC_BOOTCMD.yml
```diff
@@ -3,3 +3,4 @@ bootcmd:
 - cloud-init-per once fallocate-swapfile fallocate -l 4G /swapfile
 - cloud-init-per once chmod-swapfile chmod 600 /swapfile
 - cloud-init-per once mkswap-swapfile mkswap /swapfile
+- cloud-init-per once swapon-swapfile swapon /swapfile
```
hitz: This is the only change I am concerned with.   Possibly if we run some load testing on a cluster while the front end is using a lot of effort to index.   For sure we should ~never put real memory pressure on the front end machines.
I guess we could try simulating running the machine out of RAM?

caseylitton: We did do this on v91x0 live or test and did loading testing...It really hard to get it to use swap.  Do you want me to separate out this into its own pr?  Might be better since there is concern about it.

hitz: I think I would prefer it separate because who knows if it will ever matter but if it does we may need to revert it.

caseylitton: I'm just going to switch it out.  'll try to get a load test going tomorrow.  We can always make this change live on v93 or test.

caseylitton: The commit would have always been separate since the pr is organized, so revert would have been fine.  I separated it for testing...in case we decided not to merge then we don't have to hold off on the other 5 tickets.

PR #3026, opened 2019-11-07
ENCD-4934-add-perturbed-property-to-biosample

jenjou:


src/encoded/types/biosample.py
```diff
@@ -513,6 +513,30 @@ def summary(self, request,
...
+    })
+    def perturbed(
+        self,
+        request,
+        treatments=None,
+        genetic_modifications=None,
+        model_organism_donor_modifications=None
+    ):
+        applied_modifications = get_applied_modifications(
+            genetic_modifications, model_organism_donor_modifications)
```
yunhailuo: Is it worth having this property condition on applied_modifications so that `get_applied_modifications` doesn't have to be called twice, @keenangraham ?

src/encoded/tests/test_types_biosample.py
```diff
@@ -161,3 +161,15 @@ def test_biosample_summary_construct(testapp,
...
+def test_perturbed(
+    testapp,
+    biosample_1,
+    interference_genetic_modification,
+):
+    testapp.patch_json(biosample_1['@id'], {
+        'genetic_modifications': [interference_genetic_modification['@id']],
+        })
+    res = testapp.get(biosample_1['@id']+'@@index-data')
+    assert res.json['object']['perturbed'] is True
```
yunhailuo: Maybe add a test for False before patching, @jenjou

src/encoded/types/biosample.py
keenangraham: It looks like the calculated applied_modifications property already does this exact calculation, so you should be able to add `define=True` to applied_modifications https://github.com/ENCODE-DCC/encoded/blob/3b97e8bf8e15511899737f2c70471523ca7e4fa7/src/encoded/types/biosample.py#L270 and use it in your function signature: 
```
def perturbed(
        self,
        request,
        applied_modifications,
        treatments=None,
    ):
```

src/encoded/types/biosample.py
```diff
@@ -513,6 +513,30 @@ def summary(self, request,
...
+        is_perturbation = []
+        for modification in applied_modifications:
+            gm = request.embed(modification, '@@object')
+            is_perturbation.append(gm.get('perturbation'))
+        if treatments or True in is_perturbation:
+            return True
+        else:
+            return False
+
```
keenangraham: `return any(treatments) or any(is_perturbation)`

keenangraham: You might want to return True right away if there are treatments. Faster than always going through applied_modifications.

I think you can do this in one line but might be a little too condensed for clarity:

```
return any(treatments) or any(
    (
        request.embed(m, '@@object').get('perturbation', False)
        for m in applied_modifications
    )
)
```

src/encoded/tests/test_types_biosample.py
keenangraham: Yes, the more tests the better.

src/encoded/types/biosample.py
jenjou: Added define=True and removed the get_applied_modifications section

src/encoded/tests/test_types_biosample.py
jenjou: Added a second test for not perturbed

src/encoded/tests/test_types_biosample.py
```diff
@@ -161,3 +161,23 @@ def test_biosample_summary_construct(testapp,
...
+
+def test_is_perturbed(
+    testapp,
+    biosample_1,
+    interference_genetic_modification,
+):
+    testapp.patch_json(biosample_1['@id'], {
+        'genetic_modifications': [interference_genetic_modification['@id']],
+        })
+    res = testapp.get(biosample_1['@id']+'@@index-data')
```
keenangraham: Need a space around the `+`.

src/encoded/tests/test_types_biosample.py
```diff
@@ -161,3 +161,23 @@ def test_biosample_summary_construct(testapp,
...
         'Drosophila melanogaster '
         'female (10 days) liver tissue stably expressing C-terminal eGFP-tagged ATF4 under daf-2 promoter')
+
+
+def test_is_perturbed(
+    testapp,
+    biosample_1,
+    interference_genetic_modification,
+):
+    testapp.patch_json(biosample_1['@id'], {
```
keenangraham: Formatting nit: 
```
testapp.patch_json(
    biosample_1['@id'],
    {
        'genetic_modifications': [interference_genetic_modification['@id']]
    }
)
```

src/encoded/tests/test_types_biosample.py
```diff
@@ -161,3 +161,23 @@ def test_biosample_summary_construct(testapp,
...
+        })
+    res = testapp.get(biosample_1['@id']+'@@index-data')
+    assert res.json['object']['perturbed'] is True
+
+
+def test_is_not_perturbed(
+    testapp,
+    biosample_1,
+):
+    res = testapp.get(biosample_1['@id']+'@@index-data')
```
keenangraham: Space around `+`.

src/encoded/tests/test_types_biosample.py
```diff
@@ -161,3 +161,23 @@ def test_biosample_summary_construct(testapp,
...
+    interference_genetic_modification,
+):
+    testapp.patch_json(biosample_1['@id'], {
+        'genetic_modifications': [interference_genetic_modification['@id']],
+        })
+    res = testapp.get(biosample_1['@id']+'@@index-data')
+    assert res.json['object']['perturbed'] is True
+
+
+def test_is_not_perturbed(
```
keenangraham: What about a test for treatments and treatments/perturbed modifications together?

jenjou: Added 2 more tests

src/encoded/types/biosample.py
```diff
@@ -513,6 +513,24 @@ def summary(self, request,
...
+    })
+    def perturbed(
+        self,
+        request,
+        applied_modifications,
+        treatments=None,
+    ):
+        return bool(treatments) or any(
+            (
+                request.embed(m, '@@object').get('perturbation', False) for m in applied_modifications
```
keenangraham: Minor formatting nit. Usually worth splitting up comprehensions into multiple lines, where thing getting returned is first line, iterator is second line, and (when applicable) if statement is third line:

```
[
    d.get('x')
    for d in data
    if some_condition
]
```
Probably not worth changing here though.

jenjou: I can fix it, it should be a quick change.

PR #3031, opened 2019-11-08
ENCD-4842-new-gm-property-introduced-elements

iyoungworth:


iyoungworth:
Discussion with the wranglers ended with the decision that the word "regions" should be substituted for the word "fragments" in this case, the newest commit makes this change for changes in these four files.

PR #3034, opened 2019-11-08
ENCD-4949-ie11-polyfil-promise

zoldello:


zoldello:
The solution is a hack will be replaced when we move off @babel/polyfill (deprecated) to core.js. Along story short, @babel/polyfill has issues with Promise and I doubt it they will address it.

PR #3036, opened 2019-11-11
ENCD-4930-flag-read1-without-read2

khzlin:


src/encoded/tests/test_audit_file.py
```diff
@@ -247,6 +247,19 @@ def test_audit_file_mismatched_paired_with(testapp, file1, file4):
...
+    testapp.patch_json(file2['@id'], {
+                        'run_type': 'paired-ended', 'paired_end': '1'})
+    res = testapp.get(file2['@id'] + '@@index-data')
+    errors = res.json['audit']
+    errors_list = []
+    for error_type in errors:
+        errors_list.extend(errors[error_type])
+        print(errors_list)
+    assert any(error['category'] ==
+                'missing paired_end 2 file' for error in errors_list)
```
yunhailuo: Can you add one more section here making sure this audit will be gone once files are patched?

khzlin: done.

PR #3038, opened 2019-11-13
ENCD-4848 Display Juicebox for hic visualizations

forresttanaka:
I added a mapping of browser name from the back end to a display name. This mostly maps from something to the exact same thing, but allows us to add new mappings in the future.

Indexed demo: ~https://encd-4848-juicebox-name-57715a626-fytanaka.demo.encodedcc.org/~

### DIFF
--- a/src/encoded/static/components/vis_defines.js
+++ b/src/encoded/static/components/vis_defines.js
@@ -205,6 +205,23 @@ export const visSortBrowsers = browsers => (
 
 
 /**
+ * Map of browser to display name.
+ */
+const browserNameMap = {
+    UCSC: 'UCSC',
+    hic: 'Juicebox',
+    Ensembl: 'Ensembl',
+};
+
+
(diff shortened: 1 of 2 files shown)


src/encoded/static/components/vis_defines.js
```diff
@@ -204,6 +204,23 @@ export const visSortBrowsers = browsers => (
...
 
 
+/**
+ * Map of browser to display name.
+ */
+const browserNameMap = {
+    UCSC: 'UCSC',
+    hic: 'Juicebox',
+    Ensembl: 'Ensembl',
+};
```
emro: It would be very convenient if JavaScript guaranteed the order of `Object.keys` because then you could use the object for `browserOrder` and not have an array and an object with repeated information - just so that adding a browser wouldn't require remembering to add it to both the array and the object. Oh well!

PR #3039, opened 2019-11-14
ENCD-4949-Fix-IE11

zoldello:


package.json
```diff
@@ -86,7 +87,7 @@
     "marked": "^0.7.0",
     "pluralize": "^8.0.0",
     "prop-types": "^15.7.2",
-    "query-string": "^6.8.3",
+    "query-string": "^4.1.0",
```
zoldello: I did not experiment with any query-string 5.X.X because at this point, I am extremely risk-avert. I may do so in a future npm update.

src/encoded/static/inline.js
```diff
@@ -1,5 +1,9 @@
 'use strict';
 
+if (window && !window.Promise) {
```
zoldello: It is better to do something like
window.Promise = window.Promise || require('bluebird');
However, I am in risk-avert mode and do not want to risk changing anything unless neccesary.

PR #3041, opened 2019-11-15
ENCD-4970 Make reference-epignome cells render on IE11

forresttanaka:
This involved adding a non-breaking space to the relevant cells. Because the colors and striped SVG backgrounds are absolutely positioned, this has no visual effect otherwise. Without the non-breaking space, the height of the colored cells is zero on IE11.

### DIFF
--- a/src/encoded/static/components/matrix_reference_epigenome.js
+++ b/src/encoded/static/components/matrix_reference_epigenome.js
@@ -132,18 +132,6 @@ const generateColMap = (context) => {
 
 
 /**
- * Display a disabled cell in the matrix. Used to reduce a bit of code per cell when matrices can
- * be very large.
- */
-const DisabledCell = () => (
-    <React.Fragment>
-        <div className="matrix__disabled-cell" />
-        &nbsp;
-    </React.Fragment>
-);
-
-
-/**
  * Takes matrix data from JSON and generates an object that <DataTable> can use to generate the JSX
  * for the matrix. This is a shim between the incoming matrix data and the object <DataTable>
  * needs.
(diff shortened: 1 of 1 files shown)


forresttanaka:
Khine found that with rows containing multiple lines of text, IE11 only colors the top part of the taller-than-normal cell. So I changed it so that instead of having the link have a background color, the table cells themselves have the background color. This solves the IE11 problem without causing problems with other browsers. The problem is that on IE11, only the top-most portion of taller-than-normal cells are clickable to the corresponding search page. I haven't figured out a way to solve that, and for now I don't think it's worth worrying about.

https://encd-4970-ref-ep-ie11-pr2-fytanaka.demo.encodedcc.org/

src/encoded/static/components/matrix_reference_epigenome.js
```diff
@@ -246,10 +239,14 @@ const convertReferenceEpigenomeToDataTable = (context, expandedRowCategories, ex
                             const colIndex = colMap[colMapKey].col;
                             cells[colIndex] = {
                                 content: (
-                                    <a href={`${context.search_base}&${rowCategoryQuery}&${subCategoryQuery}&${colMap[colMapKey].query}`} style={{ backgroundColor: rowSubcategoryColor }}>
-                                        <span className="sr-only">Search {rowCategoryBucket.key}, {rowSubcategoryBucket.key} for {rowSubcategoryColCategoryBucket.key}, {cellData.key}</span>
-                                    </a>
+                                    <React.Fragment>
```
zoldello: Not important- A newer alternative to React.Fragment is to use "<>". See https://reactjs.org/docs/fragments.html#short-syntax

forresttanaka: I don’t like how `<>` looks so I’ve decided not to use it myself. Nothing wrong with others using it.

PR #3042, opened 2019-11-16
ENCD-4973 Fix raw schema tab crashing

forresttanaka:


forresttanaka:
Demo for PR: https://encd-4973-raw-schema-pr1-fytanaka.demo.encodedcc.org/

PR #3043, opened 2019-11-20
ENCD-4967-IE11-potpourri-fixes

zoldello:


src/encoded/static/components/blocks/richtext.js
```diff
@@ -25,7 +25,7 @@ export default class RichTextBlockView extends React.Component {
         // Polyfill for 'closest' for IE
         const elementPrototype = window.Element.prototype;
         if (!elementPrototype.closest) {
-            elementPrototype.closest = (s) => {
+            elementPrototype.closest = function closest(s) {
```
zoldello: the this keyword (function context) needs to come from closest-function and not the parent function. With arrow function, this-keywords comes from the parent function which is undesired. For some reason, only IE recognizes this behaviors correctly.

src/encoded/static/components/cart/batch_download.js
```diff
@@ -72,16 +72,22 @@ class CartBatchDownloadComponent extends React.Component {
...
-            // including https://blog.jayway.com/2017/07/13/open-pdf-downloaded-api-javascript/
-            const link = document.createElement('a');
-            link.href = window.URL.createObjectURL(blob);
-            link.download = filename;
-            document.body.appendChild(link);
-            link.click();
-            document.body.removeChild(link);
+            const nav = window.navigator;
+
+            // IE11 workaround (also activates on Edge-Trident but not Edge-Chromium)
```
zoldello: Edge-trident is on the way out, so I did not bother to have it skip nav.msSaveOpenBlob.

src/encoded/static/components/objectutils.js
```diff
@@ -618,18 +618,9 @@ export class ImageWithFallback extends React.Component {
         });
     }
 
-    // Display default "not found" image for non-existent image src
-    onError() {
```
zoldello: We may have to discuss this in person. IE11 throws a permission error always. I skipped this check and let the error fall through.

src/encoded/static/components/objectutils.js
```diff
@@ -618,11 +618,24 @@ export class ImageWithFallback extends React.Component {
         });
     }
 
-    // Display default "not found" image for non-existent image src
     onError() {
+        let imageAlt = null;
+        let imageUrl = null;
+
+        // IE11 has an issue where it frequently throws a "Permission denied" exception, when the image
```
zoldello: I could not figure out exactly why IE11 has the issue this comment mentions (always throws a permission exception.) So, I did the behavior in the code. IE11 has a low usage so I think it is okay. Let me know if you have an issue with it and I will think of an alternative solution

src/encoded/static/components/graph.js
```diff
@@ -650,8 +650,16 @@ export class Graph extends React.Component {
...
+
+        if (typeof (Event) === 'function') {
+            const event = new Event('input', { bubbles: true });
+            this.slider.current.dispatchEvent(event);
+        } else {
+            // Needed for IE11
+            const event = document.createEvent('Event', { bubbles: true });
+            event.initEvent('input', true, false, window, 0);
+            this.slider.current.dispatchEvent(event);
+        }
```
emro: Just curious, could we just use the code that works for IE11 for all the browsers? Or is it worth keeping both methods because eventually we'll want to drop support for IE11 and the other method is better?

src/encoded/static/components/objectutils.js
```diff
@@ -618,11 +618,24 @@ export class ImageWithFallback extends React.Component {
...
+        } else {
+            imageUrl = '/static/img/brokenImage.png';
+            imageAlt = 'Not found';
+        }
+
         this.setState({
-            imageUrl: '/static/img/brokenImage.png',
-            imageAlt: 'Not found',
+            imageUrl,
+            imageAlt,
```
emro: This is a pretty weird bug.

Just wondering if we need the setState outside the if/else. Maybe it could just be in the `else` because for the `if` case, imageAlt and imageUrl are already saved in state.

src/encoded/static/components/graph.js
zoldello: If you insist I can, as they both would work in all browsers. The thinking of the approach I used is that Event-approach is the newer way while document.createEvent is an older way. See https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Creating_and_triggering_events. So, some would say it is okay to have both so as not to force new browsers to use the old way. But again, if you insist, I can use the old way for all browsers.

src/encoded/static/components/objectutils.js
zoldello: @emro Code updated.

src/encoded/static/components/graph.js
emro: I think it's better the way you have it! Hopefully we can eventually drop the older method.

PR #3057, opened 2019-11-26
ENCD-4984-add-ENCODE3-to-scRNA-without-controls

iyoungworth:


src/encoded/audit/experiment.py
```diff
@@ -3227,8 +3227,11 @@ def audit_experiment_control(value, system, excluded_types):
         return
 
     # single cell RNA-seq in E4 do not require controls (ticket WOLD-6)
+    # single cell RNA-seq in E3 also do not require controls (ENCD-4984, WOLD-52)
     if value.get('assay_term_name') == 'single cell isolation followed by RNA-seq' and \
-            check_award_condition(value, ["ENCODE4"]):
+            check_award_condition(value, [
+                "ENCODE4",
+                "ENCODE3"]):
```
gabdank: I think you need to add trailing comma after ENCODE3 and probably "drop" parenthesis on the line below that.

iyoungworth: Would it be preferable to keep it all in one line, since it's a list of only two things? I was trying to match the style of the next check_award_condition 'if' statement a few lines later, which doesn't have the trailing comma or dropped parenthesis

iyoungworth: I've made the syntax changes to match your description, @gabdank.

PR #3070, opened 2019-12-04
ENCD-4655 remove recombinant protein category of target

yunhailuo:


src/encoded/audit/antibody_characterization.py
```diff
@@ -54,7 +54,27 @@ def audit_antibody_characterization_target(value, system):
     '''
     antibody = value['characterizes']
     target = value['target']
-    if 'recombinant protein' in target['investigated_as']:
+    # The following list should in sync with modification enum defined in the
+    # target schema.
+    tags = [
```
gabdank: this list is not in sync, there is a new tag introduced tagRFP I think

gabdank: see ENCD-4971

PR #3071, opened 2019-12-05
ENCD-4917-new-target-matrix

zoldello:


zoldello:
This is not ready for review.

forresttanaka:
Wondering if your `<TargetDataTable>` does something that the existing `<DataTable>` that every other matrix uses can't? Does `<DataTable>` need a new features? Seems like duplicate work to have `<TargetDataTable>` when `<DataTable>` could be fixed.

zoldello:
@forresttanaka I missed your comment. I created TargetDataTable because I needed to do a lot of things DataTable did not and I needed fine-grain control because I was experimenting with the different table-structures. I thought retro-fitting DataTable to meet needs was taking too many surgerical moves, so I made TargetDataTable. Do you insist I use DataTable and update it while not breaking pages?

forresttanaka:
I’m not clear what you’re doing in TargetDataTable that DataTable can't do. DataTable is very generic and has been able to render everything so far, and the target matrix seems comparatively simple, though larger, and DataTable has no size limits. I just don't like to duplicate effort if not needed, and I'm not clear why we need to duplicate effort here.

zoldello:
@forresttanaka - using DataTable.

src/encoded/static/components/matrix_target.js
```diff
@@ -0,0 +1,677 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import pluralize from 'pluralize';
+import url from 'url';
+import PubSub from 'pubsub-js';
```
zoldello: This is the pubsub thing I spoke about

src/encoded/static/components/matrix_target.js
```diff
@@ -0,0 +1,677 @@
...
+    </div>
+);
+
+TargetMatrixContent.propTypes = {
+    /** Matrix search result object */
+    context: PropTypes.object.isRequired,
+};
+
+
+class TargetTabPanel extends TabPanel {
```
zoldello: I needed fine-grain control to experiment, so I extend rather than use TabPanel. Like DataTable, I am open to trying to use TabPanel with modifications if you feel that is needed.

src/encoded/static/components/matrix_target.js
```diff
@@ -0,0 +1,677 @@
...
+                    {svgIcon('largeArrow')}
+                </div>
+                <div className="matrix__presentation-content">
+                    <div className="matrix__label matrix__label--vert"><div>{svgIcon('largeArrow')}{context.matrix.y.label}</div></div>
+                    {showOrganismRequest ? <GetTargets /> : null }
+                    <TargetTabPanel tabs={headers} selectedTab={this.state.selectedTab} tabPanelCss="matrix__data-wrapper">
+                        <TargetTabPanel tabs={this.subTabsHeaders} selectedTab={selectedSubTab} tabPanelCss="matrix__data-wrapper" handleTabClick={this.subTabClicked} fontColors={fontColors}>
+                            {targetData ?
+                                  <div className="matrix__data" onScroll={this.handleOnScroll} ref={(element) => { this.scrollElement = element; }}>
+                                      <DataTable tableData={convertTargetDataToDataTable(targetData, selectedSubTab)}> test </DataTable>
```
zoldello: Removed TargetDataTable and using DataTable as requested. I am using targetData (extracted from context) a lot because I need data structure I can easily extract information from like for search, filtering, result count and a mixture of these. That is not so easy with context.

src/encoded/static/components/datacolors.js
```diff
@@ -221,7 +221,14 @@ class DataColors {
     //         integer -- do not pass a string with a "%" sign at the end.
     colorList(keys, options) {
         let colors = [];
-        if (keys && keys.length > 0) {
+        if (options && keys && options.merryGoRoundColors && keys.length > 0) {
```
zoldello: merryground gives leeway in that the colors wrap back to the first one instead of showing default, after the color-limit is reached. I can see if you say it is YAGNI-violation.

src/encoded/search_views.py
```diff
@@ -171,29 +173,22 @@ def matrixv2_raw(context, request):
     return fr.render()
 
 
-@view_config(route_name='matrix', request_method='GET', permission='search')
-def matrix(context, request):
+def generic_matrix(request, title_response_field_params=None, type_response_field_params=None, response_field_params=None):
```
keenangraham: Let's actually not do the generic_matrix thing here. It will be cleaner just to make your own target_matrix view.

src/encoded/search_views.py
```diff
@@ -64,6 +65,7 @@ def includeme(config):
     'Lab'
 ]
 
+TARGET_MATRIX_TITLE = 'TargetMatrix'
```
keenangraham: Don't want to start putting constants here. Probably just type TargetMatrix into your view.

src/encoded/search_views.py
```diff
@@ -227,6 +222,34 @@ def reference_epigenome_matrix(context, request):
     )
     return fr.render()
 
+@view_config(route_name='matrix', request_method='GET', permission='search')
+def matrix(context, request):
```
keenangraham: Revert this to original.

src/encoded/search_views.py
```diff
@@ -259,31 +282,15 @@ def entex_matrix(context, request):
 
 @view_config(route_name='summary', request_method='GET', permission='search')
 def summary(context, request):
-    fr = FieldedResponse(
-        _meta={
-            'params_parser': ParamsParser(request)
+    return generic_matrix(
```
keenangraham: Revert this to original.

src/encoded/search_views.py
```diff
@@ -227,6 +222,34 @@ def reference_epigenome_matrix(context, request):
...
+        request,
+        title_response_field_params={'title': MATRIX_TITLE},
+        type_response_field_params={'at_type': [MATRIX_TITLE]},
+        response_field_params={'default_item_type': DEFAULT_ITEM_TYPES},
+    )
+
+
+@view_config(route_name='target_matrix', request_method='GET', permission='search')
+def target_matrix(context, request):
+    return generic_matrix(
```
keenangraham: Create your own target_matrix view here.

src/encoded/static/components/matrix_target.js
```diff
@@ -0,0 +1,677 @@
...
+                a.push(m);
+                return a;
+            }, []);
+
+        const yMatrixT = {};
+
+        yData.forEach((y) => {
+            const yKey = Object.keys(y)[0];
+            // IE11 does not support .fill. So .map is used.
+            yMatrixT[yKey] = yMatrixT[yKey] || [...Array(xMatrixLength + 1)].map(x => 0);
```
forresttanaka: You can use Array.prototype.fill on IE11 because babel polyfills it. In fact I'd recommend it here and the other place in here because they're causing ESLint errors.

forresttanaka:
With zero documentation through any of matrix_target.js this code review is going to be very difficult and time consuming. We saw this with facet.appended, where I couldn't figure out what it was for, and you didn't remember what it was for.

src/encoded/static/components/datacolors.js
zoldello: Good point. Done.

src/encoded/search_views.py
zoldello: Done.

src/encoded/static/components/matrix_target.js
zoldello: Done. I used ES5 (ecmascript 5) syntax because I think inline.js does not use babel.js to for transpiling.

src/encoded/search_views.py
zoldello: Done.

src/encoded/search_views.py
zoldello: Done.

src/encoded/tests/test_search.py
```diff
@@ -0,0 +1,86 @@
...
+    res = testapp.get('/target-matrix/?type=Experiment').json
+    assert res['@type'] == ['TargetMatrix']
+    assert res['@id'] == '/target-matrix/?type=Experiment'
+    assert res['@context'] == '/terms/'
+    assert res['notification'] == 'Success'
+    assert res['title'] == 'Target Matrix'
+    assert res['total'] > 0
+    assert 'filters' in res
+    assert 'matrix' in res
+    assert res['matrix']['x']['group_by'] == ['biosample_ontology.classification', 'biosample_ontology.term_name']
```
keenangraham: Formatting nit:

Put items in list on separate line to avoid have to scroll horizontally:

```
assert res['matrix']['x']['group_by'] == [
    'biosample_ontology.classification',
    'biosample_ontology.term_name'
]
```

src/encoded/tests/test_search.py
```diff
@@ -0,0 +1,86 @@
...
+    assert res['matrix']['x']['label'] == 'Term Name'
+    assert res['matrix']['y']['group_by'] == ['replicates.library.biosample.donor.organism.scientific_name', 'target.label']
+    assert res['matrix']['y']['label'] == 'Target'
+    assert len(res['matrix']['y']['replicates.library.biosample.donor.organism.scientific_name']['buckets']) > 0
+    assert len(res['matrix']['y']['replicates.library.biosample.donor.organism.scientific_name']['buckets'][0]['target.label']['buckets']) > 0
```
keenangraham:
```
assert len(
    res['matrix']['y']['replicates.library.biosample.donor.organism.scientific_name']['buckets'][0]['target.label']['buckets']
) > 0
```

zoldello: Done.

src/encoded/tests/test_search.py
zoldello: Done.

src/encoded/tests/test_search.py
```diff
@@ -0,0 +1,86 @@
+# Use workbook fixture from BDD tests (including elasticsearch)
```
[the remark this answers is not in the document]
zoldello: Good catch. I moved the code to the other test file and deleted test_search.py

forresttanaka:
Many, many thanks for adding comments. This should go much more smoothly.

src/encoded/static/components/matrix_target.js
```diff
@@ -0,0 +1,776 @@
...
+                    ref={this.searchBox}
+                />
+                <select name="searchOption" onChange={this.handleChange}>
+                    <option value="biosample">Biosample</option>
+                    <option value="target">Target</option>
+                </select>
+            </div>
+        );
+    }
+}
```
forresttanaka: The React team discourages the use of component inheritance, and I think that must be especially true with the emphasis on functional components with hooks. https://reactjs.org/docs/composition-vs-inheritance.html#so-what-about-inheritance

src/encoded/static/components/matrix_target.js
```diff
@@ -0,0 +1,776 @@
...
+                assayTitle: 'Histone ChIP-seq',
+                title: 'Mus musculus | Histone ChIP-seq',
+                url: 'replicates.library.biosample.donor.organism.scientific_name=Mus musculus&assay_title=Histone ChIP-seq',
+            }, {
+                organismName: 'Mus musculus',
+                assayTitle: 'TF ChIP-seq',
+                title: 'Mus musculus | TF ChIP-seq',
+                url: 'replicates.library.biosample.donor.organism.scientific_name=Mus musculus&assay_title=TF ChIP-seq',
+            },
+        ];
```
forresttanaka: You could put `header` in the global space — it’s static anyway, and in the global space this would get assigned only on page load, not on every render.

src/encoded/static/components/matrix_target.js
```diff
@@ -0,0 +1,776 @@
...
+                              :
+                                  <div className="matrix__warning">
+                                      <br />
+                                      { targetData && Object.keys(targetData).length === 0 ? 'Select an organism to view data.' : 'No data to display.' }
+                                      <br /><br /><br />
+                                  </div>
+                            }
+                            <br />
+                        </TargetTabPanel>
+                        <br />
```
forresttanaka: What are the two `<br />`s for? I don’t even see them in the generated HTML.

src/encoded/static/components/matrix_target.js
forresttanaka: I think you’ll have to use TabPanel — again the warning against React component inheritance. I think we’ll have to go this direction, especially as we phase out component classes.

src/encoded/search_views.py
zoldello: Done.

src/encoded/search_views.py
```diff
@@ -331,3 +367,4 @@ def audit(context, request):
         ]
     )
     return fr.render()
+
```
keenangraham: Can remove this extra space.

src/encoded/types/experiment.py
```diff
@@ -188,6 +188,22 @@ def superseded_by(self, request, superseded_by):
...
+                'replicates.library.biosample.donor.organism.scientific_name',
+                'target.label',
+            ],
+            'label': 'Target',
+        },
+        'x': {
+            'group_by': ['biosample_ontology.classification', 'biosample_ontology.term_name'],
+            'label': 'Term Name',
+        },
+        'viewableTabs': ['Homo sapiens', 'Mus musculus'],
```
keenangraham: 'viewableTabs' -> 'viewable_tabs'

src/encoded/tests/test_search_views.py
```diff
@@ -436,6 +436,34 @@ def test_search_views_matrix_response_no_results(workbook, testapp):
...
+    assert res['matrix']['y']['group_by'] == [
+        'replicates.library.biosample.donor.organism.scientific_name',
+        'target.label'
+    ]
+    assert res['matrix']['y']['label'] == 'Target'
+    assert len(res['matrix']['y']['replicates.library.biosample.donor.organism.scientific_name']['buckets']) > 0
+    assert len(
+        res['matrix']['y']['replicates.library.biosample.donor.organism.scientific_name']['buckets'][0]['target.label']['buckets']
+)       > 0
+    assert res['matrix']['viewableTabs'] == ['Homo sapiens', 'Mus musculus']
```
keenangraham: Shouldn't introduce camelCase into the JSON.

src/encoded/static/components/matrix_target.js
zoldello: I can show you. Go to [https://encd-4917-21525e593-phil.demo.encodedcc.org/target-matrix/?type=Experimen](https://encd-4917-21525e593-phil.demo.encodedcc.org/target-matrix/?type=Experimen)t. If you look "Select an Organism to view data", you will see the break points.

forresttanaka: OK, I see them now. I deleted both `<br>` from the DOM and didn’t see any difference though. I’m not sure what they do here.

src/encoded/static/components/matrix_target.js
forresttanaka: @zoldello explained to me the issue and I now agree that the best thing for now is to keep this as is. I wrote ENCD-5067 to address this in the future.

src/encoded/tests/test_search_views.py
zoldello: Remove the variable as it was used in old code.

src/encoded/types/experiment.py
zoldello: Remove the variable as it was used in old code.

src/encoded/search_views.py
zoldello: Remove the space

src/encoded/static/components/matrix_target.js
zoldello: Move spacing to SASS.

src/encoded/static/components/matrix_target.js
zoldello: Move to outside of a class.

src/encoded/static/components/matrix_target.js
zoldello: Moved the inheritance. TabPanel may be revisited and used in the future instead of TargetTabPanel

src/encoded/static/components/matrix_target.js
```diff
@@ -0,0 +1,776 @@
...
+    fontColors: null,
+};
+
+
+/**
+ * Used for creating modal pop up that gathers information on what organism a user wants to view.
+ *
+ */
+const GetTargets = () => (
+    <>
```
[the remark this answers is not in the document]
zoldello: removed.

src/encoded/static/components/matrix_target.js
```diff
@@ -0,0 +1,776 @@
...
+
+        return { rowContent, css };
+    });
+
+    dataTable.push(...rowData);
+
+    const rowKeys = [];
+    for (let i = 0; i < dataTable.length; i += 1) {
+        rowKeys.push(`key${i}`);
+    }
```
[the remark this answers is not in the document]
zoldello: Removed.

src/encoded/static/components/matrix_target.js
```diff
@@ -0,0 +1,776 @@
...
+
+
+/**
+ * Get search result-count
+ *
+ * @param {targetData} targetData
+ * @returns Number 0 or higher that corresponds to search count
+ */
+const getSearchResultCount = (targetData) => {
+    const count = [...targetData.dataRow]
```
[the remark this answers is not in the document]
zoldello: You are right and cloning is unnecessary.

src/encoded/static/components/matrix_target.js
```diff
@@ -0,0 +1,776 @@
...
+            .reduce((a, b) => {
+                const m = {};
+                m[b.key] = b[xGroupBy1].buckets
+                    .filter(f => f.key === subTab)
+                    .reduce((x, y) => {
+                        x.push(y[xGroupBy2].buckets
+                            .reduce((i, j) => i.concat(j), []));
+                        return x;
+                    }, []);
+                a.push(m);
```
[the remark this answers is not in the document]
zoldello: Updated.

src/encoded/static/components/datacolors.js
```diff
@@ -221,7 +221,14 @@ class DataColors {
     //         integer -- do not pass a string with a "%" sign at the end.
```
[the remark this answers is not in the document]
zoldello: Done

src/encoded/static/components/matrix_target.js
```diff
@@ -105,7 +105,7 @@ const getTargetData = (context, assayTitle, organismName) => {
         targetData[subTab] = { headerRow, dataRow, assayTitle, organismName };
     });
 
-    const subTabs = subTabList;
+    const subTabs = subTabList.sort();
```
forresttanaka: Is it OK that `subTabs` and `subTabsHeaders` will have a different order? I can’t quite tell.

src/encoded/search_views.py
```diff
@@ -42,6 +42,7 @@ def includeme(config):
     config.add_route('matrix', '/matrix{slash:/?}')
     config.add_route('reference-epigenome-matrix', '/reference-epigenome-matrix{slash:/?}')
     config.add_route('entex-matrix', '/entex-matrix{slash:/?}')
+    config.add_route('target_matrix', '/target-matrix{slash:/?}')
```
keenangraham: Might want to change this to `target-matrix` to match `entex-matrix` etc. Not sure it really matters as long as we aren't using underscores in the URL.

src/encoded/static/components/matrix_target.js
zoldello: subTabsHeaders is redundant and was removed. It is created only where it is needed and map is used. Javascript map gurantees order ([https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#Objects_and_maps_compared](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#Objects_and_maps_compared))

zoldello:
@forresttanaka There may be two or three more changes coming. There are (1) Changing word "Target" to "ChIP" (2) moving search dropdown (3)Your comment about tab on mobile. Final decisions have not been made.

src/encoded/search_views.py
```diff
@@ -218,6 +219,41 @@ def matrix(context, request):
...
+    fr = FieldedResponse(
+        _meta={
+            'params_parser': ParamsParser(request)
+        },
+        response_fields=[
+            TitleResponseField(
+                title='ChIP Seq Matrix'
+            ),
+            TypeResponseField(
+                at_type=['ChIPSeqMatrix']
```
keenangraham: Probably `ChipSeqMatrix` to match up with `HistoneChipSeqQualityMetric` @type format of this: https://www.encodeproject.org/histone-chipseq-quality-metrics/ecadc412-f9b5-42f2-8b5e-3a171f019698/?format=json

src/encoded/search_views.py
```diff
@@ -218,6 +219,41 @@ def matrix(context, request):
...
 
+@view_config(route_name='chip-seq-matrix', request_method='GET', permission='search')
+def chip_seq_matrix(context, request):
+    fr = FieldedResponse(
+        _meta={
+            'params_parser': ParamsParser(request)
+        },
+        response_fields=[
+            TitleResponseField(
+                title='ChIP Seq Matrix'
```
keenangraham: `ChIP-seq`

src/encoded/tests/test_search_views.py
```diff
@@ -436,6 +436,32 @@ def test_search_views_matrix_response_no_results(workbook, testapp):
     assert r.json['notification'] == 'No results found'
 
 
+def test_chip_seq_matrix_view(workbook, testapp):
+    res = testapp.get('/chip-seq-matrix/?type=Experiment').json
+    assert res['@type'] == ['ChIPSeqMatrix']
```
keenangraham: ChipSeqMatrix

src/encoded/tests/test_search_views.py
```diff
@@ -436,6 +436,32 @@ def test_search_views_matrix_response_no_results(workbook, testapp):
...
+    assert res['@id'] == '/chip-seq-matrix/?type=Experiment'
+    assert res['@context'] == '/terms/'
+    assert res['notification'] == 'Success'
+    assert res['title'] == 'ChIP Seq Matrix'
```
keenangraham: ChIP-seq matrix

src/encoded/search_views.py
zoldello: Done

src/encoded/search_views.py
zoldello: Done

src/encoded/tests/test_search_views.py
zoldello: Done.

src/encoded/tests/test_search_views.py
zoldello: Called this "ChIP-seq" to match the title in the backend. Let me know if you think the backend should be re-named this.

zoldello:
@forresttanaka I dismissed your review.

src/encoded/static/components/matrix_chip_seq.js
```diff
@@ -0,0 +1,886 @@
...
+                const m = {};
+                m[b.key] = b[xGroupBy1].buckets
+                    .filter(f => f.key === subTab)
+                    .reduce((x, y) => {
+                        x.push(y[xGroupBy2].buckets
+                            .reduce((i, j) => i.concat(j), []));
+                        return x;
+                    }, []);
+                return a.concat(m);
+            }, []);
```
hitz: This looks like a lot of work; I wonder if it would make sense to push this to the backend?

src/encoded/static/components/matrix_chip_seq.js
```diff
@@ -0,0 +1,886 @@
...
+        // sub chIP Seq data to display
+        const chIPSeqData = this.ChIPSeqMatrixData ? this.ChIPSeqMatrixData.chIPSeqData[selectedSubTab] : {};
+        const isMatrixLarge = chIPSeqData && chIPSeqData.dataRow && chIPSeqData.dataRow.length > MIN_SIZE_OF_LARGE_MATRIX;
+        const matrixUpdate = {
+            chIPSeqData,
+            selectedSubTab,
+            showOrganismRequest, // determines if organism modal shows
+        };
+
+        if (isMatrixLarge) {
```
hitz: Is this actually necessary to test if it's large?  If it's small it will just flash the spinner and move on, right?  Does this look stupid?

src/encoded/static/robots.txt
```diff
@@ -10,7 +10,7 @@ Disallow: /batch_hub/
 Disallow: /batch_download/
 Disallow: /metadata/
 Disallow: /entex-matrix/
-Disallow: /target-matrix/
+Disallow: /chip_seq_matrix/
```
hitz: This is an error it's chip-seq-matrix

src/encoded/static/components/matrix_chip_seq.js
zoldello: I can think about changing the format of data from elastic search. For things like subtabs-creation , searching and filtering out empty rows, to do these quickly and frequently on tabs, I need the data in a format that backend did not provide. That is why I had to dig into it like this.

src/encoded/static/components/matrix_chip_seq.js
zoldello: It's silly and hacky but I had reasons. It is a rabbit hole. There is a spinner from fetched.js, which comes for free on every backend call. For small matrices, it suffices to use the spinner from fetched.js as it does not get clobbered by the matrix-painting. That is why I do not call showSpinner here on small size-matrices

On large matrices, I have to call _.defer() which pushes execution until call stack is empty. That is computationally expensive and should only be done when necessary. fetched.js usually kills the spinner in the time it takes to paint large matrices, so the matrices in this case has to be re-started by the painting-code and stopped afterwards. That means I have to check if the matrix is large and do things different compared to when small.

src/encoded/static/components/matrix_chip_seq.js
keenangraham: From an architectural perspective we want to keep things that change a lot like frontend views isolated from things that change less like core search logic. Since this data transformation is very tied to a frontend view, and specifically an encoded frontend view, it would make sense to keep it isolated in a data transformation layer between the search API and React rendering.

The question is then should this layer be on the Python or the Javascript side. I doubt there would be any performance benefits to keeping it on the Python side and the amount of data transferred would be more or less the same as now. Also since this code is doing transformations to render tabs and rows it seems more appropriate to keep it on the JS side rather than piping it through a JSON API in Python.

So why not just make a data transformation layer in JS with reusable library code?

forresttanaka:
Pausing code review for now. @zoldello is trying something that, if it works out, I _think_ might make things much simpler.

src/encoded/static/components/matrix_chip_seq.js
hitz: well this page specifically has performance issues with large matrices (no idea if this loop has any bearing on that).

src/encoded/static/components/matrix_chip_seq.js
hitz: good enough!

src/encoded/static/robots.txt
zoldello: updated

src/encoded/static/components/matrix_chip_seq.js
zoldello: @hitz reading and processing of the data from the backend is pretty quick. It is the painting/DOM update that is the slow part.

src/encoded/static/components/matrix_chip_seq.js
```diff
@@ -750,12 +772,29 @@ class ChIPSeqMatrixPresentation extends React.Component {
...
+                _.defer(() => {
+                    this.setState({ chIPSeqData: null }, () => {
+                        this.setState({ chIPSeqData }, () => {
+                            _.defer(() => {
+                                this.setState({ spinnerActive: false });
+                            });
+                        });
+                    });
+                });
+            }
```
forresttanaka: Calling setState in the second parameter to setState worries me a little, but I haven’t been able to find anything that says you _can’t_ do this. I’ll assume it’s fine, but good to note unless you found something that says this is fine.

src/encoded/static/components/matrix_chip_seq.js
```diff
@@ -0,0 +1,886 @@
...
+
+    return matrixConfig;
+};
+
+
+/**
+ * Show or hide the spinner
+ *
+ * @param {*} active
+ * Note - Changing DOM in react is a code smell. Nevertheless, this was done
```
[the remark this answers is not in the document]
zoldello: this is old

src/encoded/static/components/matrix_chip_seq.js
zoldello: That and calling nested-defer's. I have not seen anyone else do these. However, I never read anyone that said don't and the code seems to have no trouble with it. It is not normally good to be that adventurous... But I had to get creative.

zoldello:
forced to rebase to fix merge conflict

src/encoded/search_views.py
```diff
@@ -218,6 +219,41 @@ def matrix(context, request):
...
+            TypeResponseField(
+                at_type=['ChipSeqMatrix']
+            ),
+            IDResponseField(),
+            SearchBaseResponseField(),
+            ContextResponseField(),
+            BasicMatrixWithFacetsResponseField(
+                default_item_types=DEFAULT_ITEM_TYPES,
+                matrix_definition_name='chip_seq_matrix',
+                facets=[
```
keenangraham: I don't think these facets are actually used?

zoldello: they are used.

keenangraham: Where?

zoldello: I missed read what you wrote. Yeah, it was facets are not needed as they were used in an early iteration that has since changed. BasicMatrixWithFacetsResponseField is needed though.

src/encoded/static/inline.js
```diff
@@ -16,6 +16,52 @@ if ('NodeList' in window && !NodeList.prototype.forEach) {
     };
 }
 
+// IE11 polyfill for [].fill, based on- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/fill#Polyfill
+if (!Array.prototype.fill) {
```
forresttanaka: Babel polyfills this already.

src/encoded/static/components/matrix_chip_seq.js
```diff
@@ -0,0 +1,886 @@
...
+    subTabs.forEach((subTab) => {
+        const xGroupBy1 = context.matrix.x.group_by[0];
+        const xGroupBy2 = context.matrix.x.group_by[1];
+        const headerRow = context.matrix.x[xGroupBy1].buckets.find(f => f.key === subTab)[xGroupBy2]
+            .buckets
+            .reduce((a, b) => {
+                const m = a.concat(b);
+                return m;
+            }, [])
+            .map(x => x.key);
```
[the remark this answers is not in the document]
zoldello: Done. I made it verbose because it is easier to put break points and change values to test. It is okay if it is tense.

src/encoded/static/components/matrix_chip_seq.js
```diff
@@ -0,0 +1,886 @@
...
+        yData.forEach((y) => {
+            const yKey = Object.keys(y)[0];
+            dataRowT[yKey] = dataRowT[yKey] || [...Array(headerRowLength + 1)].fill(0);
+            dataRowT[yKey][0] = yKey;
+
+            const keyDocCountPair = y[yKey].reduce((a, b) => {
+                const m = a.concat(b);
+                return m;
+            }, []);
+
```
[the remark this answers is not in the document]
zoldello: Done. I did this for the same reasons easier, it was easier to have breakpoints and debug. I made it more tense.

src/encoded/static/inline.js
```diff
@@ -16,6 +16,52 @@ if ('NodeList' in window && !NodeList.prototype.forEach) {
...
+          O[k] = value;
+          k++;
+        }
+
+        // Step 13.
+        return O;
+      }
+    });
+}
+
```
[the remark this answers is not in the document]
zoldello: Good catch.

src/encoded/static/components/matrix_chip_seq.js
```diff
@@ -0,0 +1,906 @@
...
+                        return x;
+                    }, []);
+                return a.concat(m);
+            }, []);
+
+        const dataRowT = {};
+
+        yData.forEach((y) => {
+            const yKey = Object.keys(y)[0];
+            dataRowT[yKey] = dataRowT[yKey] || [...Array(headerRowLength + 1)].fill(0);
```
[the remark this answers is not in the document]
zoldello: Left as-is. The [...] is needed for the array to be of size headerRowLength + 1, otherwise, it is a single value. Let me know if you still have concerns and we can go from there.

src/encoded/static/components/matrix_chip_seq.js
```diff
@@ -0,0 +1,906 @@
...
+        const yGroupBy2 = context.matrix.y.group_by[1];
+
+        const yData = context.matrix.y[yGroupBy1].buckets
+            .find(rBucket => rBucket.key === organismName)[yGroupBy2].buckets
+            .reduce((a, b) => {
+                const m = {};
+                m[b.key] = b[xGroupBy1].buckets
+                    .filter(f => f.key === subTab)
+                    .reduce((x, y) => {
+                        x.push(y[xGroupBy2].buckets
```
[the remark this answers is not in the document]
zoldello: Good catch. I used [...] to make a new object. Does that suffice?

src/encoded/static/components/matrix_chip_seq.js
```diff
@@ -0,0 +1,906 @@
...
+            dataRow.push(dataRowT[key]);
+        });
+
+        // Remove all rows with no data (all entries/count are zero)
+        // Remove this line if you want to show all data and not just ones with content
+        dataRow = dataRow.filter((d) => {
+            const data = [...d]; // !Imporant. Clone to avoid modifying dataRow itself
+            data.shift(); // remove first entry as it is biosample ontology classification and not a count
+            return !data.every(i => i === 0); // remove all rows with all zeros
+        });
```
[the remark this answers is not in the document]
zoldello: Done, kind of. I made this more tense. I made it verbose initially for others to understand each step. I changed this to be more tense. I used some instead of every as it fast.

src/encoded/static/inline.js
zoldello: Good catch. In my mind, babel is a transpiler. I got to remember it also polyfills

src/encoded/static/components/matrix_chip_seq.js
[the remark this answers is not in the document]
forresttanaka: Not sure what you mean. I could look at it after you push.

src/encoded/static/components/matrix_chip_seq.js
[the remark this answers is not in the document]
forresttanaka: I think it’s the same without the brackets.
image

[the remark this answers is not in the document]
zoldello: Done. I was doing [Array(5)] rather than Array(5).

src/encoded/static/inline.js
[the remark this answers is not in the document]
forresttanaka: Probably want to get rid of the extra blank line so it won’t show up as a changed file when it wasn‘t really changed.

src/encoded/static/components/matrix_chip_seq.js
```diff
@@ -0,0 +1,900 @@
...
+    context: PropTypes.object.isRequired,
+};
+
+ChIPSeqMatrix.contextTypes = {
+    location_href: PropTypes.string,
+    navigate: PropTypes.func,
+    biosampleTypeColors: PropTypes.object, // DataColor instance for experiment project
+};
+
+globals.contentViews.register(ChIPSeqMatrix, 'ChipSeqMatrix');
```
forresttanaka: I think this needs a blank line at the end.

zoldello: There is no blank link here but one in the file.

zoldello: In matrix_chip_seq.js, there was already a blank line in the end.

zoldello:
@forresttanaka When I made the Array change, I did it in two places rather than just the one I needed. I fixed that.

src/encoded/static/components/app.js
```diff
@@ -1332,4 +1333,4 @@ module.exports.getRenderedProps = function getRenderedProps(document) {
         props[elem.getAttribute('data-prop-name')] = value;
     }
     return props;
-};
+};
```
forresttanaka: Seems like this one really is missing the final carriage return.

zoldello: Fixed. I do not recall modifying this.

PR #3073, opened 2019-12-06
ENCD-4994-fix-broke-img-tag-appearance

zoldello:


src/encoded/static/components/objectutils.js
```diff
@@ -622,7 +622,7 @@ export class ImageWithFallback extends React.Component {
         // IE11 has an issue where it frequently throws a "Permission denied" exception, when the image
         // exist. This workaround makes IE11 show either the image if it exist or the browser's
         // inbuilt no-image display
-        if (document && document.all !== undefined) {
+        if (!!window.MSInputMethodContext && !!document.documentMode) {
```
zoldello: @emro My bad. document.all is not a good IE11 detector. What I used above detects IE11 but not lower browsers like IE10. Since only IE11 is supported in the IE family, I am not worried. Let me know if you are.

emro: Did you think about using the function in `browserfeat.js` to detect IE11? In `genome_browser.js` I check for IE11 to see if I should disable the genome browser:
> if (BrowserFeat.getBrowserCaps('uaTrident')) {

and it works well I think.

zoldello: @emro Thanks. I did not know we had a method that handles IE11 detection. I updated the codebase and am waiting for the build.

PR #3081, opened 2019-12-18
ENCD-4986-fix-summary-page-filter-miscount

zoldello:


zoldello:
@forresttanaka For an example of overlapping x-axis, see- https://encd-4986-4af630636-phil.demo.encodedcc.org/summary/?type=Experiment&advancedQuery=date_released:[2009-01-01%20TO%202018-12-31]

In chart.js, you can control the number of ticks in the x-axis. However, you cannot control the step size (like you can in y-axis). I could not think of a formula to set x-tick to ensure the last date always appears (needed to ensure users the graph reflected date range.) So, what I did was add the last date always. This sometimes makes the last date very close to the second to last.

src/encoded/static/components/award.js
```diff
@@ -1789,18 +1802,36 @@ export const ExperimentDate = (props) => {
...
+
+    if (dateRangeString) {
+        dateRange = dateRangeString
+            .find(d => !!d)
+            .replace(':', '')
+            .replace('[', '')
+            .replace(']', '')
+            .split('TO');
+    }
+
```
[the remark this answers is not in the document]
zoldello: Good catch. I was thinking step-by-step process, rather then using inbuilt power of JavaScript like regex.

src/encoded/static/components/award.js
```diff
@@ -1989,6 +2010,23 @@ class CumulativeGraph extends React.Component {
...
             const Chart = require('chart.js');
             const ctx = document.getElementById('myGraph').getContext('2d');
+            const xticksBallParkCount = 25; // ballpark number desired x-axis ticks
+            const monthReleasedLength = monthReleased.length;
+            const lastxAxisIndex = monthReleasedLength - 1;
+            const xticks = monthReleasedLength < xticksBallParkCount ? 1 : Math.ceil(monthReleasedLength / xticksBallParkCount);
+            const dateValues = [];
+            const totals = [];
+
+            // eslint-disable-next-line no-plusplus
```
forresttanaka: I’d like to not have any new exceptions to ESLint rules, and gradually remove the exceptions we have from code written before we used ESLint. This one seems like an easy rule to follow without disabling it.

zoldello: @forresttanaka Corrected.

src/encoded/static/components/award.js
```diff
@@ -2052,6 +2052,8 @@ class CumulativeGraph extends React.Component {
                             ticks: {
                                 autoSkip: false,
                                 maxTicksLimit: 25,
+                                maxRotation: 45,
```
zoldello: @forresttanaka chart.js stopped tilting the numbers probably because of the change in sampling rate. This forces the tilt again.

forresttanaka: OK, that sounds likely.

PR #3083, opened 2019-12-20
ENCD-4936-add-location-to-gene

bonitalam:


src/encoded/schemas/gene.json
```diff
@@ -122,6 +204,56 @@
...
             "permission": "import_items"
         },
+        "locations": {
+            "title": "Gene location coordinates",
+            "description": "Genomic locations for different assembles of this gene.",
+            "type": "array",
+            "default": [],
+            "uniqueItems": true,
+            "items": {
+                "title": "Gene Location Coordinate",
```
gabdank: I think to be consistent it should be "Gene location coordinate" and not "Gene Location Coordinate"

src/encoded/schemas/gene.json
```diff
@@ -122,6 +204,56 @@
...
+                            "hg19",
+                            "mm10",
+                            "mm9"
+                        ]
+                    },
+                    "chromosome": {
+                        "title": "Chromosome",
+                        "description": "The number (or letter) designation for the chromosome, e.g. chr1 or chrX",
+                        "type": "string",
+                        "pattern": "^(chr[0-9XYMT]+)$"
```
gabdank: I think this regex is not going to work with C.elegans and Drosophila chromosome names. See: https://github.com/ENCODE-DCC/encValData/blob/master/dm3/chrom.sizes and https://github.com/ENCODE-DCC/encValData/blob/master/ce11/chrom.sizes

bonitalam: Ok, thanks for point that out. Will change it.

src/encoded/schemas/gene.json
```diff
@@ -77,7 +159,57 @@
...
+        "locations": {
+            "title": "Gene location coordinates",
+            "description": "Genomic locations using 1-based, closed coordinates for different assembles of this gene.",
+            "type": "array",
+            "default": [],
+            "uniqueItems": true,
+            "items": {
+                "title": "Gene location coordinate",
+                "description": "Genome position information for one assembly of this gene.",
+                "type": "object",
```
gabdank: I would go with title "Gene location", otherwise you need to say "coordinate" but each location has two coordinates?
For description I would go with: "Genomic position information of the gene for a specific genome assembly"

gabdank: MAybe: "Gene location specified using 1-based, closed coordinates for a specific version of the reference genome assembly."

PR #3089, opened 2020-01-07
ENCD-4956-add-content-to-schema

emro:
We want to display "anyOf" and "oneOf" schemas on the formatted schema pages.

### DIFF
--- a/src/encoded/static/components/schema.js
+++ b/src/encoded/static/components/schema.js
@@ -35,8 +35,6 @@ const excludedTerms = [
     'output_type_output_category',
     'file_format_file_extension',
     'sort_by',
-    'anyOf',
-    'oneOf',
 ];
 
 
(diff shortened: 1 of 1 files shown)


src/encoded/static/components/schema.js
```diff
@@ -398,6 +396,16 @@ class DisplayObjectSection extends React.PureComponent {
     render() {
         const { term, schema, schemaName, profilesMap } = this.props;
 
+        // No simple term types in the schema value, so display the schema object
+        const schemaTerm = schema[term];
+        const schemaIsObject = typeof schemaTerm === 'object';
+        const schemaIsArray = Array.isArray(schemaTerm);
+        let simpleTermValuesExist = true;
+        if (schemaIsObject && schemaIsArray) {
```
forresttanaka: Technically, if schemaIsArray is true, then we know schemaIsObject must also be true.

emro:
Rebased in order to build demo.

PR #3090, opened 2020-01-08
ENCD-4987 Add audits for queried_RNP_size_range

jenjou:


src/encoded/audit/experiment.py
```diff
@@ -3920,6 +3920,59 @@ def audit_experiment_nih_institutional_certification(value, system, excluded_typ
...
+            lib = rep.get('library', {})
+            if 'queried_RNP_size_range' in lib:
+                experiment_size_range.add(lib['queried_RNP_size_range'])
+            else:
+                detail = ('Libraries of experiment {} are missing specification'
+                    ' of queried_RNP_size_range'.format(
+                        audit_link(path_to_text(value['@id']), value['@id'])
+                    )
+                )
+                yield AuditFailure('missing queried_RNP_size_range', detail, level='ERROR')
```
gabdank: I think that in the case of one or more replicates missing size range you should not continue to comparing with control, add "return" after yielding

src/encoded/audit/experiment.py
```diff
@@ -3920,6 +3920,59 @@ def audit_experiment_nih_institutional_certification(value, system, excluded_typ
...
+
+    if 'possible_controls' in value:
+        for control in value.get('possible_controls'):
+            control_accessions.append(audit_link(path_to_text(control['@id']), control['@id']))
+            for rep in control['replicates']:
+                if rep.get('status') not in excluded_types and 'library' in rep:
+                    lib = rep.get('library', {})
+                    if 'queried_RNP_size_range' in lib:
+                        control_size_range.add(lib['queried_RNP_size_range'])
```
gabdank: I think here you can also throw "missing queried_RNAP_size_range", in the details saying it is a control that is missing it. Again here you can "return" after yield

src/encoded/audit/experiment.py
```diff
@@ -3920,6 +3920,59 @@ def audit_experiment_nih_institutional_certification(value, system, excluded_typ
...
+    if not experiment_size_range:
+        experiment_size_range = 'none'
+    else:
+        experiment_size_range = ', '.join(experiment_size_range)
+    if not control_size_range:
+        control_size_range = 'none'
+    else:
+        control_size_range = ', '.join(control_size_range)
+
+    if experiment_size_range != control_size_range:
```
gabdank: Instead of comparing strings generated from sets, you can compare sets, you don't need the "none" because you already made sure any empty ranges stopped the function and "returned" earlier

src/encoded/audit/experiment.py
```diff
@@ -3920,6 +3920,59 @@ def audit_experiment_nih_institutional_certification(value, system, excluded_typ
...
 
+def audit_experiment_eclip_queried_RNP_size_range(value, system, excluded_types):
+    '''
+    Check if libraries of eCLIP experiment and its control have matching queried_RNP_size_range.
+    '''
+    if value.get('assay_term_name') != 'eCLIP':
+        return
+
+    experiment_size_range = set()
+    control_size_range = set()
```
gabdank: Instead of trying to capture potential differences in ranges of different replicates - you can throw an audit that experiment has mixed size ranges - and "return" after that. Then you can be sure you have only one range you are dealing with

src/encoded/audit/experiment.py
```diff
@@ -3932,6 +3932,75 @@ def audit_experiment_nih_institutional_certification(value, system, excluded_typ
...
+        yield AuditFailure('mixed queried_RNP_size_range', detail, level='ERROR')
+
+    if len(control_size_range) > 1:
+        detail = 'Libraries of control experiment {} have mixed queried_RNP_size_range values of {}.'.format(
+            ', '.join(control_accessions),
+            ', '.join(control_size_range)
+        )
+        yield AuditFailure('mixed queried_RNP_size_range', detail, level='ERROR')
+
+    if experiment_size_range != control_size_range and value['possible_controls']:
```
gabdank: I think only if the size ranges in experiment and in control are "OK" you should continue to the comparison between them, otherwise, you are potentially comparing multiple ranges in control to a range in the experiment. In addition to that, the Detail message in the "inconsistent queried .." audit should be not sets but the string values of the size_ranges.

src/encoded/audit/experiment.py
```diff
@@ -3932,6 +3932,75 @@ def audit_experiment_nih_institutional_certification(value, system, excluded_typ
...
+        detail = ('Libraries of experiment {} have queried_RNP_size_range of {},'
+            ' but the libraries of its control experiment {} have queried_RNP_size_range of {}.'.format(
+                audit_link(path_to_text(value['@id']), value['@id']),
+                experiment_size_range,
+                ', '.join(control_accessions),
+                control_size_range
+            )
+        )
+        yield AuditFailure('inconsistent queried_RNP_size_range', detail, level='ERROR')
+    return
```
gabdank: You don't need the return statement in the end of the function

src/encoded/audit/experiment.py
```diff
@@ -3932,6 +3932,75 @@ def audit_experiment_nih_institutional_certification(value, system, excluded_typ
...
+    if len(experiment_size_range) > 1:
+        detail = 'Libraries of experiment {} have mixed queried_RNP_size_range values of {}.'.format(
+            audit_link(path_to_text(value['@id']), value['@id']),
+            ', '.join(experiment_size_range)
+        )
+        yield AuditFailure('mixed queried_RNP_size_range', detail, level='ERROR')
+
+    if len(control_size_range) > 1:
+        detail = 'Libraries of control experiment {} have mixed queried_RNP_size_range values of {}.'.format(
+            ', '.join(control_accessions),
```
gabdank: if there were more than 1 control the sentence should be "of control experiments {}", I thin you can do "Libraries of control experiment(s) {} have..."

PR #3092, opened 2020-01-08
ENCD-5010-Reduce-min-read-length-miRNA

jnau:
added audit check for insufficient read length under the miRNA standards

### DIFF
--- a/src/encoded/audit/experiment.py
+++ b/src/encoded/audit/experiment.py
@@ -964,9 +964,11 @@ def check_experiment_rna_seq_standards(value,
             value,
             fastq_files,
             alignment_files,
+            pipeline_title,
             gene_quantifications,
             desired_assembly,
             desired_annotation,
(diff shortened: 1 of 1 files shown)


src/encoded/audit/experiment.py
```diff
@@ -1329,6 +1331,13 @@ def check_experiment_micro_rna_standards(
     upper_limit_expressed_mirnas,
     lower_limit_expressed_mirnas,
 ):
+    # Audit read length
+    minimum_threshold = 30 
+    for f in fastq_files:
+        yield from check_file_read_length_rna(f, minimum_threshold,
+                                                  'microRNA-seq pipeline',
+                                                  experiment['assay_term_name'],
+                                                  '/microrna/microrna-seq/')
```
yunhailuo: The last argument here should following other function as being passed over here from `standards_links[pipeline_title]`.

PR #3094, opened 2020-01-08
ENCD-4948 Facet registry

forresttanaka:
* search.js has all facet rendering cut out, moving all to the new src/encoded/static/components/facet directory that contains the external interface (index.js) the default facet rendering code (defaults.js) as well as individual files for each kind of custom facet. I put instructions on writing custom facet components in index.js.
* You’ll find inline functions in my new React code quite a bit even though they cause new instances of those functions on each render — something I’ve said to avoid in past code reviews because I used to get that feedback from a past dev here all the time. The React team’s view is to not worry about that — new function instances have little to no lasting memory impact and virtually no performance impact, and make the code clearer. I often (and the React team often) memoizes functions, but that’s simply to avoid sending a new function instance to components and cause them to rerender needlessly. The function still gets a new instance created that gets ignored on each render, but the point is memoization allows us to pass the same instance to components, which allows them to skip rerendering. I had found a great citation for this, but I didn’t save it and now I can’t find it.
* The Date Selector facet is nearly unchanged from dev, but it has been split out from search.js to its own module and registered. date_selector.js also demonstrates how to unconditionally suppress a facet.
* I had to add features to my experimental query_string module to allow for negated query string elements, and add/modify a couple convenience methods. I’m targeting us getting rid of the npm query-string module eventually which hasn’t served us very well. I also removed the `getNotKeyElements` method which I ended up never using, and its API is pretty confusing.
* I didn’t like how we used the existence or absence of `context` to determine whether to draw the search results title and Clear Filters link. I now always pass `context` and made a new boolean flag for rendering the title and Clear Filters link.
* I added a new CSS class to browserfeat.js so the “hiddenscroll” class appears in the `<html>` tag if the user has configured vanishing scroll bars in macOS, and “no-hiddenscroll” if the user has configured always-visible scroll bars, or if the user uses Windows. Unlike all the other browser features, you have to append the test `<div>` to the DOM to perform the test.
* I’m still learning best practices with hooks, so I’m guessing I’ll look at this code in the future and groan — or you might groan now, in which case I’d really like the feedback.
* I removed the unused tabbed browser display from search.js.
* The `restrictions` flag from forms.js appears not to have had any effect in about four years, and without any useful documentation, I feel unclear what `restrictions` was for. Though technically unrelated, its last reference was in the now-removed browser-display code in search.js, causing an ESLint warning.
* I didn’t implement the facet.appended special case from SNO-85 because it doesn’t appear to be needed anymore, maybe because of the new search code.


forresttanaka:
I recommend looking at src/encoded/static/components/facets/index.js and defaults.js in the same directory first, to get a feel for how you work with the facet registry. The other files in the new facets directory implement the actual special-cased facets using the registry.

src/encoded/static/components/facets/date_selector.js
```diff
@@ -0,0 +1,353 @@
...
+    /** Complete search-results object */
+    results: PropTypes.object.isRequired,
+    /** Query-string portion of current URL without initial ? */
+    queryString: PropTypes.string.isRequired,
+};
+
+DateSelectorFacet.contextTypes = {
+    navigate: PropTypes.func,
+};
+
```
emro: I was thinking about the architecture of the facet registry and I feel like maybe most of this date selector code could make more sense in `defaults.js`. Because the date selector facet type is a facet type, and then `date_released` and `date_submitted` are instances of that facet type.

You have a file `datasets_accession.js` which sets parameters for the datasets.accession facet which is an instance of an "exists" facet. And an `nih_institutional_certification.js` file which sets parameters for another instance of an "exists" facet. So to have the same sort of paradigm here, you would have a `date_released.js` and `date_submitted.js` which would set parameters for a "date selector" facet. At least, that is what would make more sense to me. I recognize that there aren't actually any parameters to set for a date selector facet so it's not exactly the same.

Overall, I wonder if this set-up could get a little unwieldy. I really like how all of the default facet types can be all stored in one file and then their parameters can be set elsewhere. But having one file for NIH certification and each specific instance of a default facet type seems like it could get to be a lot of very specific files. Another way to structure it could be to have the `defaults.js` file and then a file for all the instances of "exists" facets and another file for all the instances of the "boolean" facets and "date selector" and "typeahead" facets etc. (Although, here "typeahead" isn't really a default facet type, it's more like an option for the default facet.)

src/encoded/static/components/facets/defaults.js
```diff
@@ -0,0 +1,572 @@
...
+                            <div className="facet__radio-label">{term.key_as_string}</div>
+                            <div className="facet__radio-count">{term.doc_count}</div>
+                        </label>
+                    </div>
+                ))}
+                <div className="facet__radio">
+                    <input type="radio" name={facet.field} value="either" id={`${facet.field}-either`} checked={currentOption === 'either'} onChange={handleRadioClick} />
+                    <label htmlFor={`${facet.field}-either`}>
+                        <div className="facet__radio-label">either</div>
+                        <div className="facet__radio-count">{facet.total}</div>
```
emro: I would change this to "Either" and convert `term.key_as_string` to be uppercase, but this is a style preference and I'm not sure what the Encode style should be.

src/encoded/static/components/search.js
```diff
@@ -1436,108 +664,75 @@ TextFilter.propTypes = {
...
-    if (searchQuery) {
-        // Convert search query string to a query object for easy parsing.
-        const terms = queryString.parse(searchQuery);
+    const parsedUrl = context && context['@id'] && url.parse(context['@id']);
 
-        // See if there are terms in the query string aside from `searchTerm` or `advancedQuery`. We have a Clear
-        // Filters button if we do.
-        let nonPersistentTerms = _(Object.keys(terms)).any(term => term !== 'searchTerm' && term !== 'advancedQuery');
-        clearButton = nonPersistentTerms && (terms.searchTerm || terms.advancedQuery);
+    // See if we need the Clear Filters link or not. context.clear_filters.
```
emro: Could this be clarified? Maybe "see if we need the Clear filters link by checking context.clear_filters" or something?

src/encoded/static/components/search.js
```diff
@@ -1436,108 +664,75 @@ TextFilter.propTypes = {
...
- * @param {object} sessionProperties From <App> React context
- *
- * @return True if given facet should be hidden
- */
-const isFacetHidden = (facet, session, sessionProperties) => (
-    facet.field === 'internal_status' && !sessionProperties.admin
-);
-
-
 // Displays the entire list of facets. It contains a number of <Facet> cmoponents.
```
emro: "components"

src/encoded/static/components/facets/date_selector.js
forresttanaka: I was going to say I don’t mind lots of small files, and I don’t, except with AirBnB who insist on one component per .jsx file, and Redux that goes even finer than that, so I guess I don’t mind — up to a point. But I think you bring up a good point with the datasets_accession and the nih one, and that could (and likely will) happen again. So I’ll rearrange it by type rather than field. Hopefully done after lunch and appointment.

src/encoded/static/components/facets/defaults.js
forresttanaka: I’ll leave it lowercase for now, only because "exists" facets have always been lowercase, and I haven’t heard otherwise from the wrangling side yet.

forresttanaka:
Along with the other changes, I moved the DateSelectorFacet to DefaultDateSelectorFacet in defaults.js. I also combined both exists facet and added a note about how to handle that going forward in index.js. I also updated the demo with these changes.

forresttanaka:
I don’t know why “Show changes since last review” is grayed out for me. Maybe because I’m not a reviewer? Hopefully it’s enabled for you.

forresttanaka:
I made a small update that needs review. I had to move a div to always render even with a custom term name component because IE11 was formatting it wrong without that div.

forresttanaka:
I had ignored the "appended" flag on facets as the test cases in SNO-85 seemed to work for me even ignoring that flag, and without comments I didn’t feel like reverse-engineering code just to find out what it was for. But selected facets not included in the schema still need a special case using the "appended" flag, where we need to keep going even without a doc_count.

PR #3095, opened 2020-01-09
ENCD-4995 adding index reads

gabdank:
adding index of

adding index reads output type

adding dependencies

updating insert

fixing loadxl and inserts

### DIFF
--- a/src/encoded/schemas/file.json
+++ b/src/encoded/schemas/file.json
@@ -210,6 +210,17 @@
                 }
             }
         },
+        "index_of": {
+            "comment": "Only fastq files can have index_of specified.",
+            "properties": {
+                "file_format": {
+                    "enum": ["fastq"]
+                },
+                "output_type": {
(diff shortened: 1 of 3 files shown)


src/encoded/schemas/file.json
```diff
@@ -364,11 +375,19 @@
...
+                            "enum": ["index reads"]
+                        }
+                    }
+                },
                 {
                     "not": {
                         "properties": {
                             "output_type": {
-                                "enum": ["reads", "restriction enzyme site locations"]
+                                "enum": ["reads", "index reads", "restriction enzyme site locations"]
```
yunhailuo: In general, I consider "index reads" as similar to "reads". So at line 397 below here for "assembly" dependency. Should "index read" be not able to have assembly?

gabdank: you are right, fixed

PR #3096, opened 2020-01-09
ENCD-4997-adjust-GM-schema-for-FCC-expts

iyoungworth:


src/encoded/schemas/genetic_modification.json
```diff
@@ -70,14 +70,27 @@
                     ],
                     "properties": {
                         "purpose": {
-                                "enum": ["repression", "activation", "overexpression", "characterization", "screening"]
+                                "enum": ["repression", "characterization"]
```
yunhailuo: Is it expression here?

iyoungworth:
After fixing the code to make sure the upgrade happened (and included the edge case), errors cropped up with previous tests on the GM object. It looked like some GM objects were then valid for multiple sections in a 'oneOf' clause, so I've tried to make my additions in a different way.

PR #3097, opened 2020-01-09
ENCD-5014-new-FCE-property-target-expression-percentile

iyoungworth:
2nd commit is adjustments to code in experiment.js to get the property to render, 3rd commit is reverting those changes. Visualization won't be included in this branch/ticket.

### DIFF
--- a/src/encoded/tests/test_schema_functional_characterization_experiment.py
+++ b/src/encoded/tests/test_schema_functional_characterization_experiment.py
@@ -40,6 +40,18 @@ def functional_characterization_experiment(testapp, lab, award, cell_free):
     return testapp.post_json('/functional_characterization_experiment', item).json['@graph'][0]
 
 
+@pytest.fixture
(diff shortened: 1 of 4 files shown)


src/encoded/tests/test_schema_functional_characterization_experiment.py
```diff
@@ -40,6 +40,18 @@ def functional_characterization_experiment(testapp, lab, award, cell_free):
     return testapp.post_json('/functional_characterization_experiment', item).json['@graph'][0]
 
 
+@pytest.fixture
```
hitz: @gabdank for the record, I think we are going to start moving ALL fixtures to datafixtures.py.

hitz: (I am going to merge this anyway... but in the future let's put fixtures in datafixtures.py

PR #3098, opened 2020-01-11
ENCD-5005 replace GO properties with GO links

yunhailuo:


src/encoded/static/components/gene.js
```diff
@@ -10,96 +10,142 @@ import { RelatedItems } from './item';
...
+        }).then((annotation) => {
+            if (annotation.numFound) {
+                validGOs.push('GOGene:'.concat(goID));
+            }
+        }))).then(() => {
+            this.setState({
+                goIDs: validGOs,
+            });
+        });
+    }
```
forresttanaka: Promise.all passes an array of all your fetch results after they all resolve (or reject), so I think you can make this simpler to understand by nesting the fetch `.then()` chain inside the `.map()` loop. Then in the outer `.then()` receives an array of annotations that it can pick through and put into the goIDs state. You can see the `Promise.all()` use in experiment_series.js for a pretty simple example.

src/encoded/static/components/app.js
```diff
@@ -1164,6 +1164,7 @@ class App extends React.Component {
                     <meta charSet="utf-8" />
                     <meta httpEquiv="X-UA-Compatible" content="IE=edge" />
                     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+                    {this.props.context['@type'].includes('Gene') ? <meta http-equiv="Content-Security-Policy" content="upgrade-insecure-requests" /> : null}
```
forresttanaka: The reason for this line will be a mystery in a month, so I would add a comment on the line before. Since this part is JSX and not Javascript, you’ll need to do it in a weird way. On the line before, you’d need something like:

```
{/* Here is a JSX comment */}
```

yunhailuo: Thanks. Added 3 comments.

src/encoded/static/components/gene.js
yunhailuo: Tried but not sure if it's right.

src/encoded/static/components/gene.js
```diff
@@ -10,96 +10,142 @@ import { RelatedItems } from './item';
...
+            } else if (dbxref.startsWith('FlyBase:FBgn')) {
+                nadbIDs.push(dbxref.replace('FlyBase:', 'FB:'));
+            } else if (dbxref.startsWith('MGI:')) {
+                nadbIDs.push('MGI:'.concat(dbxref));
+            } else if (dbxref.startsWith('UniProtKB:')) {
+                uniprotIDs.push(dbxref);
+            }
+        });
+        const validGOs = [];
+        Promise.all((nadbIDs.length > 0 ? nadbIDs : uniprotIDs).map(goID => fetch(
```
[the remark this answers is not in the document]
yunhailuo: Used. Thank you for the example. I did miss `Gene.contextTypes` for `fetch` without it!

src/encoded/static/components/gene.js
```diff
@@ -10,96 +10,144 @@ import { RelatedItems } from './item';
...
+                this.baseGOUrl.replace(/\{0\}/g, goID), {
+                    method: 'GET',
+                }
+            ).then((response) => {
+                if (response.ok) {
+                    return response.json();
+                }
+                return { numFound: 0 };
+            }).then((annotation) => {
+                if (annotation.numFound) {
```
forresttanaka: Just one little thing: numbers should always be compared with a number (our ESLint configuration). ESLint doesn’t seem to detect it, but that’s rule we have. So `numFound > 0`.

yunhailuo: Thank you. You've mentioned this before; but I forgot. Sorry. Fixed.

PR #3101, opened 2020-01-14
ENCD-4999 Add PRO-cap assay term

jenjou:


hitz:
@gabdank It's a little dicey to combine the ontology update with the schema updates... but I'll let it slide

PR #3105, opened 2020-01-15
ENCD-5022 ENTEx updates

forresttanaka:
* The issue with the Mars/Venus symbols was that Helvetica didn’t have those symbols defined so it fell back to a default font (not sure which) that has them shaped wrong. I now use Menlo which should appear on macOS, and Courier for Windows.
* matrixAssaySortOrder determines the sort order of the assays across the x-axis and was shared between the ENTEx and Reference Epignome matrices, but Idan wanted them to have their own sorting orders, so this array moved out of matrix.js and each matrix now has its own copy.
* The `legendText` properties of `entexDonors` might look odd, but I was surprised it still separates the digit from the unicode digits. I use letter-spacing to space them as the space I used to have between them had gotten very large with the new font.
* We have a new concept of “dividers” — the dark lines defining the edges of a group of targets combined into an assay.

### DIFF
--- a/src/encoded/static/components/matrix.js
+++ b/src/encoded/static/components/matrix.js
@@ -14,28 +14,6 @@ export const MATRIX_VISUALIZE_LIMIT = 500;
 
 
 /**
- * Order in which assay_titles should appear along the horizontal axis of the matrix. Anything not
- * included gets sorted after these.
- */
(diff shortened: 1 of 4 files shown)


forresttanaka:
Closing this PR and will reopen for v96.

PR #3106, opened 2020-01-15
ENCD-5032-add-reset-to-browser

emro:
Adds a reset button to the Valis genome browser so that users can return to the initial browser position.

### DIFF
--- a/src/encoded/static/scss/encoded/modules/_genome_browser.scss
+++ b/src/encoded/static/scss/encoded/modules/_genome_browser.scss
@@ -222,3 +222,30 @@
         }
     }
 }
+
+.browser-container {
+    position: relative;
+}
+
+.reset-browser-button {
+    position: absolute;
+    z-index: 9;
+    line-height: 1;
+    left: 5px;
+    text-align: left;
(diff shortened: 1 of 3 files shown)


src/encoded/static/scss/encoded/modules/_genome_browser.scss
```diff
@@ -222,3 +222,30 @@
...
+    text-align: left;
+    background-color: #f2f2f2;
+    border: 1px solid #BFBFBF;
+    padding: 7px 20px;
+    height: 50px;
+    width: 176px;
+    font-size: 14px;
+    border-radius: 4px;
+    display: flex;
+    .icon-undo {
```
zoldello: Does there supposed to be a space here? I frequently see a space on classes at this position. This is not a deal breaker and it is fine if it is ignored, I am just curious.

emro: I don't have a particularly strong opinion about this. If it will help with readability, I can add the space.

emro:
I pushed an update 5 days ago because I was rebasing. No code changes.

PR #3107, opened 2020-01-17
ENCD-5025-create-account-created-messagebox

zoldello:


src/encoded/static/components/app.js
```diff
@@ -213,6 +213,25 @@ EulaModal.propTypes = {
 };
 
 
+const AccountCreatedModal = ({ closeModal }) => (
+    <Modal>
+        <ModalHeader title="Account Created" closeModal={closeModal} />
+        <ModalBody>
+            <p>
+                The account was created and you were automatically logged-in.
```
forresttanaka: I won’t request changes for this, but in case you feel like it, I _think_ “logged-in” in this case should be “logged in.” You’d only use the hyphen when used as an adjective phrase, like “logged-in user.”

zoldello: Fixed.

zoldello:
Note, there is a system-wide issue with CI, the code passed test before and the change was minor and had no impact on automated test. So, I did not focus on getting the automated test working.

zoldello:
I rebased this to fix the broken automated test.

PR #3108, opened 2020-01-17
ENCD-4975-three-library-properties-restricted-to-enum

iyoungworth:


src/encoded/schemas/library.json
```diff
@@ -242,29 +259,60 @@
...
-            ],
-            "format": "semi-controlled"
+            "oneOf": [
+                {
+                    "enum": [
+                        "ATAC buffer",
+                        "Clontech UltraLow for Illumina sequencing",
+                        "RIPA",
+                        "SDS",
+                        "[NPB (5% BSA (Sigma), 0.2% IGEPAL-CA630 (Sigma), cOmplete (Roche), 1mM DTT in PBS)]",
```
gabdank: I am wondering why not to move the long and super specimport_only" enums?
like: "[NPB (5% BSA (Sigma), 0.2% IGEPAL-CA630 (Sigma), cOmplete (Roche), 1mM DTT in PBS)]",
                        "72 degrees for 3 minutes in the presence of Triton"
"0.01% digitonin"

instead to put Digitonin, Triton ?

PR #3109, opened 2020-01-22
ENCD-3983-SNO-132-multiticket-do-not-squash-merge

caseylitton:
Look at the commits, there are many different features included.

This goes with SNO pr 

SNO-132 Upgrade python related deps #249
https://github.com/ENCODE-DCC/snovault/pull/249


cloud-config/deploy-run-scripts/es-install.sh
```diff
@@ -4,51 +4,79 @@
...
-# Setup/Restart
-update-rc.d elasticsearch defaults
-sudo /usr/share/elasticsearch/bin/elasticsearch-plugin install discovery-ec2
-service elasticsearch restart
+
+if [ -z "$CLUSTER_NAME" ]; then
+    # Only for single demos
+    # start es and wait for yellow or green
+    while true; do
+        es_status="$(curl -fsSL 'localhost:9201/_cat/health?h=status')"
```
hitz: I don't think port 9201 should be hard coded here.

requirements.txt
```diff
@@ -0,0 +1,10 @@
+boto3>=1.10.49
```
hitz: why are we adding these here instead of setup.py?  Are they deploy-only requirements?

src/encoded/commands/import_data.py
```diff
@@ -22,7 +22,7 @@
 """
 from webtest import TestApp
 from urllib.parse import urlparse
-from .. import loadxl
+from encoded import loadxl
```
hitz: seems like an odd time clean this up; is this a python6+ issue?

src/encoded/region_search.py
```diff
@@ -265,7 +264,7 @@ def region_search(context, request):
         'notification': '',
         'filters': []
     }
-    principals = effective_principals(request)
+    principals = request.effective_principals
```
hitz: this , I assume is a pyramid upgrade issue?

src/encoded/tests/features/matrix_reference_epigenome.featurematrix.feature
```diff
@@ -0,0 +1,29 @@
+@matrix @usefixtures(workbook)
```
hitz: very confused by the presence of this file in the PR...

src/encoded/tests/features/test_audit.py
```diff
@@ -8,6 +8,13 @@
     pytest.mark.usefixtures('workbook'),
 ]
 
+
+def test_auditview(testapp, workbook):
```
hitz: confused by this as well

src/encoded/tests/test_access_key.py
```diff
@@ -160,4 +160,4 @@ def test_access_key_uses_edw_hash(app, access_key):
     root = app.registry[COLLECTIONS]
     obj = root.by_item_type['access_key'][access_key['access_key_id']]
     pwhash = obj.properties['secret_access_key_hash']
-    assert EDWHash.encrypt(access_key['secret_access_key']) == pwhash
```
hitz: oh this must get rid of one of the annoying dependencies.

src/encoded/commands/deploy.py
```diff
@@ -4,22 +4,19 @@
...
 
-from base64 import b64encode
 from difflib import Differ
 from os.path import expanduser
 
 import boto3
 
 
-def nameify(in_str):
+def _nameify(in_str):
```
hitz: I can't tell if anyone of the changes in deploy.py are functional at all, or just style/lint changes.

cloud-config/deploy-run-scripts/es-install.sh
caseylitton: that's a good idea.  We can pass in the port.  That parts isn't quite completed too.

requirements.txt
caseylitton: I use it install python deps locally so pylint can pick them up.  It should be in the package anyway, but it is not used by app install.

src/encoded/commands/import_data.py
caseylitton: It got rid of some of the pytest warning.

src/encoded/region_search.py
caseylitton: yea, That is from lrowe.  Did you look at the commit history?  That explains some of the changes and there is a specific one for lrowes updates.

src/encoded/tests/features/test_audit.py
caseylitton: A space?  You should look at the commit history.

src/encoded/tests/test_access_key.py
caseylitton: Yea, cryptography was removed.  There were a few changes for that in lrowe's pr.

src/encoded/commands/deploy.py
caseylitton: Commit history.

src/encoded/tests/features/matrix_reference_epigenome.featurematrix.feature
caseylitton: This was a weird pytest issue where the file as 
src/encoded/tests/features/matrix_reference_epigenome.feature 

was being looked for at 

src/encoded/tests/features/matrix_reference_epigenome.featurematrix.feature

Also, if you look at the commit history you'll see this as flag with WARNING.

cloud-config/deploy-run-scripts/ami-java.sh
```diff
@@ -0,0 +1,26 @@
...
+# Check previous failure flag
+if [ -f "$encd_failed_flag" ]; then
+    echo -e "\n\t$ENCD_INSTALL_TAG $(basename $0) Skipping: encd_failed_flag exits"
+    exit 1
+fi
+echo -e "\n\t$ENCD_INSTALL_TAG $(basename $0) Running"
+
+# Script Below
+# Manually install java
+sudo -u ubuntu aws s3 cp --region=us-west-2 --recursive s3://encoded-conf-prod/encd-tars ~ubuntu/encd-tars
```
hitz: this is just java?

cloud-config/deploy-run-scripts/ami-elasticsearch.sh
```diff
@@ -0,0 +1,77 @@
...
+        # For es builds we set the do install flag here since this is the main app
+        # We could have alternativily created an app-wrapper.sh for elasticsearch
+        touch "$encd_do_install_flag"
+        exit 0
+    fi
+fi
+echo -e "\n\t$ENCD_INSTALL_TAG $(basename $0) Running"
+
+# Script Below
+opts_src="$ENCD_CC_DIR/conf-es"
```
hitz: where are the files in conf-es?    I feel like we need to be sure the indices.query.bool.max_clause_count: 8192 gets set.

.circleci/config.yml
```diff
@@ -38,16 +38,17 @@ commands:
               graphviz \
               nodejs \
               openjdk-11-jdk \
-              postgresql-9.3
+              postgresql-10
```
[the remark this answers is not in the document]
caseylitton: nope, not 11.  It's been like this too.

If we integrate amis or docker images we can lock this down.

.circleci/config.yml
```diff
@@ -57,10 +58,15 @@ jobs:
...
       - run:
-          name: NPM and non ES indexing tests
+          name: NPM, non indexing, and indexing
           command: |
             npm run circleci-test
-            bin/test -s -vv -m "not bdd"
+            # collected 2189 items / 70 deselected / 2119 selected
+            # bin/test -s -vv -m "not bdd"
+            # collected 2189 items / 75 deselected / 2114 selected
+            bin/test -s -vv -m "not bdd" -k 'not test_indexing'
```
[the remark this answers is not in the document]
caseylitton: This is a requirement.  The indexing test would not pass when run with the other tests.  Separating them out is...better resource management?

src/encoded/commands/deploy.py
```diff
@@ -849,7 +849,7 @@ def hostname(value):
         'default': 'ami-0d1cd67c26f5fca19',
         # Private AMIs: Add comments to each build
         # Es builds were on 2020-02-02 - testing
-        'demo': 'ami-0fdfbf6cc258c999e',
+        'demo': 'ami-0dce3204ee0ef9c8e',
```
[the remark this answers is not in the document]
caseylitton: These are our private amis so not everyone would have access to urls.  

This is the kind of stuff we should have in an encd-deployment repo.  And the encode repo should just have a basic deployment, maybe even just dev-servers readme for osx or linux.

cloud-config/prebuilt-config-yamls/20191112-pg11-u18-demo.yml
```diff
@@ -0,0 +1,112 @@
+#cloud-config
```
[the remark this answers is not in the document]
caseylitton: They are important for development of deployment, and it's nice to have a backup in case the config build fails.  

We need the ones used to create the current amis, which could also be used as backups.
And it's nice to keep a copy of one before major updates.    

Ansible and a deployment repo would clean this up.

cloud-config/deploy-run-scripts/app-pg.sh
```diff
@@ -0,0 +1,165 @@
+#!/bin/bash
+# Setup postgres 9.3 or 11,
```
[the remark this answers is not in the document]
caseylitton: I think I removed it...I'll check.

cloud-config/deploy-run-scripts/ami-preapp.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
caseylitton: Yea, that was a late addition.  I made the demo first, then es node, then frontend.  The separation became important after a first organization. 

I can fix up the names, but this will be cleaned up with ansible.    And changing them would require retesting...which I may have to do anyway.

cloud-config/deploy-run-scripts/ami-java.sh
caseylitton: Yea, we download the tar from s3...At least now it only downloads when building the base ami instead of ever time we create a demo.

cloud-config/deploy-run-scripts/ami-elasticsearch.sh
caseylitton: It is in there.  See /cloud-config/deploy-scripts/conf-es

cloud-config/deploy-run-scripts/es-install.sh
caseylitton: This was removed.  There was a bug where es wasn't starting but now he don't need to check if its's on.

PR #3110, opened 2020-01-23
ENCD-5065-account-creation-error-msg

zoldello:


src/encoded/static/components/app.js
```diff
@@ -1204,6 +1230,12 @@ class App extends React.Component {
                                     signup={this.signup}
                                 /> :
                             null}
+                            {this.state.accountCreationFailedVisibility ?
+                                <AccountCreationFailedModal
+                                    closeModal={this.closeAccountCreationErrorModal}
+                                    date={(new Date()).toUTCString()}
+                                /> :
+                                null}
```
emro: This is an optional style comment, but @forresttanaka and I generally format conditionals like this differently. We generally put `: null}` together on one line which is aligned with the opening `{`. Whether or not you want to put the `:` with `null`, I think `null}` should be aligned with the opening `{`, in particular because line 1232 is formatted that way.

src/encoded/static/components/app.js
```diff
@@ -212,6 +212,26 @@ EulaModal.propTypes = {
     signup: PropTypes.func.isRequired,
 };
 
+const AccountCreationFailedModal = ({ closeModal, date }) => (
+    <Modal>
+        <ModalHeader title="Failed to create a new account." closeModal={closeModal} />
+        <ModalBody>
+            <p>
+                Creating a new acount failed. Please, contact <a href={`mailto:encode-help@lists.stanford.edu?subject=Creating e-mail account failed&body=Creating an account failed at time: ${date}`}>support</a>.
```
emro: "acount" should be spelled "account". Also I think "Please" doesn't need a comma after it. It could just be "Please contact support."

src/encoded/static/components/app.js
zoldello: Do you mean like this:
                            {this.state.accountCreationFailedVisibility ?
                                <AccountCreationFailedModal
                                    closeModal={this.closeAccountCreationErrorModal}
                                    date={(new Date()).toUTCString()}
                                />
                            : null}

emro: Yes. Where to put the `:` is completely up to you, though, I have no strong preference. I do think the indentation should be fixed though.

zoldello: Done.

src/encoded/static/components/app.js
zoldello: Done.

PR #3116, opened 2020-01-24
ENCD-5055-update-audit-detail-insufficient-read-length-rna

jnau:


src/encoded/audit/experiment.py
```diff
@@ -2348,7 +2348,7 @@ def check_file_read_length_chip(file_to_check,
     return
 
 
-def check_file_read_length_rna(file_to_check, threshold_length, assay_term_name, pipeline_title, standard_link):
+def check_file_read_length_rna(file_to_check, threshold_length, pipeline_title, assay_term_name, standard_link):
```
yunhailuo: It seems `assay_term_name` is not used in the function anymore? If so, deleting it seems cleaner.

Also, this function is called at line 675. Please check and make sure the arguments passed in match the new order here.

PR #3117, opened 2020-01-24
ENCD-5066-calculate-s3-uri-for-all-files

jnau:


src/encoded/types/file.py
```diff
@@ -49,9 +49,9 @@ def show_upload_credentials(request=None, context=None, status=None):
 
 
 def show_cloud_metadata(status=None, md5sum=None, file_size=None, restricted=None, no_file_available=None):
-    if restricted or not md5sum or not file_size or no_file_available:
+    if restricted or not md5sum or not file_size or no_file_available or status in ['upload failed']:
```
keenangraham: Code looks good but I don't see benefit of special casing this for upload failed.

PR #3118, opened 2020-01-24
ENCD-5054-revise-biosample-search-result-link-age-display

jenjou:


src/encoded/static/components/search.js
```diff
@@ -213,9 +212,7 @@ class BiosampleComponent extends React.Component {
         let synchText;
         if (result.synchronization) {
             synchText = result.synchronization +
-                (result.post_synchronization_time ?
-                    ` + ${result.post_synchronization_time}${result.post_synchronization_time_units ? ` ${result.post_synchronization_time_units}` : ''}`
-                : '');
+                (result.post_synchronization_time ? ` +${ageDisplay}` : '');
```
forresttanaka: This isn’t really your problem, but since you’re touching this code, the existing code combines string arithmetic with template strings, when really we should only use template strings.

This could then be:

```
synchText = `${result.synchronization}${result.post_synchronization_time ? ` +${ageDisplay}` : ''}`;
```

String templates can be nested like this without issue.

jenjou: Ok, will change. I just noticed the same thing exists in ExperimentComponent (also with the synchronization time), so I'll update that similarly.

PR #3119, opened 2020-01-27
ENCD-5037-valis-library-update

emro:
We've added an update to the Valis library which allows it to display a highlight location. This helps the user not lose track of the initial position of the browser. Currently this feature is not useful for Encode. However, we want Encode to point to the Valis update which enables this configuration.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4682,7 +4682,7 @@
       "requires": {
         "d3": "^3.3.8",
         "dagre": "^0.8.4",
-        "graphlib": "git+https://github.com/ENCODE-DCC/graphlib.git#v1.0.8",
+        "graphlib": "git+https://github.com/ENCODE-DCC/graphlib.git#df17f0dc51178d5a49f4a0cc638dcebcf9746b2d",
         "lodash": "^4.17.15"
       },
       "dependencies": {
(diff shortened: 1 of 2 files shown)


zoldello:
@emro The change looks good. Two questions- (1)Did a different ticket review changes in valis (REG-161)? (2) I read your comments. However, since ENCODE does not benefit from the new Yellow line-change, is there anything gained by making ENCODE point to the branch (seems to violate YAGNI)?

emro:
Addressing your comments:
(1) Forrest reviewed changes in REG-161 and Ben started reviewing the changes to Valis.
(2) I think that Encode may benefit from the highlight change in the future, and this way we've verified that the configuration is flexible enough to allow us to display highlights if we want or not display highlights if we don't want. I think the plan is to merge the change into our fork of Valis, or possibly submit a PR to Valis, in which case it will be important for the changes which we originally made for the benefit of Regulome to be compatible with Encode as well. Overall, adding a new configuration option that Encode doesn't require doesn't seem bad to me - Encode will either use it or not and then we don't have two different versions of Valis that we need to maintain separately.

forresttanaka:
I just updated Valis in my cart visualization branch and found what looks like the same changes to package-lock.json. Apparently they are Valis dependencies.

zoldello:
I spoke offline with Emma about changes to package-lock.json. It looks good not.

PR #3120, opened 2020-01-27
ENCD-4948 Facet registry

forresttanaka:
This PR includes all changes from last time, plus the performance fix that centers around `initialState`. I also clarified some comments.


forresttanaka:
Current demo: https://encd-4948-facet-registry-pr1-fytanaka.demo.encodedcc.org/

src/encoded/static/components/facets/exists.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { DefaultExistsFacet } from './defaults';
+import FacetRegistry from './registry';
+
+
+/**
+ * The Publication datasets.accession facet field renders an "exists" facet.
```
emro: It seems a little odd to only mention the `datasets.accession` facet field here when this file registers two exists type facets (also the `nih_institutional_certifications` facet).

src/encoded/static/components/facets/defaults.js
```diff
@@ -0,0 +1,928 @@
...
+                                return term;
+                            }
+                            return null;
+                        }
+                        return null;
+                    }
+                )
+            );
+        }
+        return processedTerms;
```
emro: It looks like you're doing operations in a different order than the original typeahead code. The original typeahead code filtered to make sure that terms matched searches and then took the first 50 terms if the component hadn't mounted yet. I think by doing the operation in this order, you could end up with an odd situation. Hypothetically, suppose that the first 50 terms for the facet begin with "a" and the user searches for "z". Then, `truncatedProcessedTerms` will get filtered to no terms even though you're trying to get the first 50 terms that match a given search. Is there a reason that you switched the order of operations here?

Also, I'm just curious, the original facet implementation had typeahead facets as a default facet type. Does it work better to have typeahead facets be a type of regular facets? Just wondering if having them separate would make things like this easier, or if this is simpler.

src/encoded/static/components/facets/exists.js
forresttanaka: Agreed.

src/encoded/static/components/facets/defaults.js
forresttanaka: I’ll look at the order.

If I understand right, you’re suggesting splitting the typeahead facet renderer out of `DefaultFacet`? Since whether a facet is typeahead or not is determined not by facet.field but by the schema, it made sense to me that typeahead should be part of `DefaultFacet`. I feel like I could be misunderstanding you though.

emro: I think I wasn't being clear. I was just sort of wondering why you decided not to have a `DefaultTypeaheadFacet` the way that there's a `DefaultExistsFacet` and a `DefaultDateSelectorFacet`. Maybe it's not too different from a regular facet so it's easier to fold it into `DefaultFacet`?

forresttanaka: Oh, I see. Yeah, this way if wranglers/associates change a facet from non-typeahead to typeahead, we don’t have to change the corresponding Javascript — DefaultFacet adapts, and their code is largely the same anyway.

forresttanaka: Regarding the order, I thought the way I had it meant the same thing, but it clearly doesn't. I corrected the order.

forresttanaka: Regarding the order, I thought the way I had it meant the same thing, but it clearly doesn't. I corrected the order.

_Odd, when I typed this the first time, it showed as a pending review comment so I deleted it. I’ll try it again._

forresttanaka:
Updated demo with region-search enabled: https://encd-4948-facet-registry-pr2-fytanaka.demo.encodedcc.org/

src/encoded/static/components/facets/defaults.js
emro: Looks good now!

PR #3122, opened 2020-01-29
ENCD-5022-ENTEx-updates

forresttanaka:
* The issue with the Mars/Venus symbols was that Helvetica didn’t have those symbols defined so it fell back to a default font (not sure which) that has them shaped wrong. I now use Menlo which should appear on macOS, and Courier for Windows.
* matrixAssaySortOrder determines the sort order of the assays across the x-axis and was shared between the ENTEx and Reference Epignome matrices, but Idan wanted them to have their own sorting orders, so this array moved out of matrix.js and each matrix now has its own copy.
* The `legendText` properties of `entexDonors` might look odd, but I was surprised it still separates the digit from the unicode digits. I use letter-spacing to space them as the space I used to have between them had gotten very large with the new font.
* We have a new concept of “dividers” — the dark lines defining the edges of a group of targets combined into an assay.

### DIFF
--- a/src/encoded/static/components/matrix_entex.js
+++ b/src/encoded/static/components/matrix_entex.js
@@ -286,9 +286,8 @@ const convertContextToDataTable = (context) => {
 
     // Generate the matrix header row labels for the assays with targets. Need a max-width inline
     // style so that wide labels don't make the target columns expand.
-    let targetAssayHeader;
     if (targetAssays.length > 0) {
-        targetAssayHeader = [{ css: 'matrix__col-category-targetassay-corner' }].concat(targetAssays.map(((targetAssayElement) => {
+        const targetAssayHeader = [{ css: 'matrix__col-category-targetassay-corner' }].concat(targetAssays.map(((targetAssayElement) => {
             if (targetAssayElement) {
                 // Add cell with assay title and span for the number of targets it has.
                 const categoryQuery = `${colCategory}=${encoding.encodedURIComponent(targetAssayElement.category)}`;
--- a/src/encoded/static/components/matrix_entex.js
+++ b/src/encoded/static/components/matrix_entex.js
@@ -6,7 +6,7 @@ import { svgIcon } from '../libs/svg-icons';
 import { Panel, PanelBody } from '../libs/ui/panel';
 import DataTable from './datatable';
 import * as globals from './globals';
-import { matrixAssaySortOrder, MATRIX_VISUALIZE_LIMIT, SearchFilter } from './matrix';
+import { MATRIX_VISUALIZE_LIMIT, SearchFilter } from './matrix';
 import { MatrixInternalTags } from './objectutils';
 import { SearchControls } from './search';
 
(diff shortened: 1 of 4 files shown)


forresttanaka:
Fully indexed demo: https://encd-5053-entex-update-pr1-fytanaka.demo.encodedcc.org/

src/encoded/static/components/matrix_entex.js
```diff
@@ -164,15 +208,14 @@ const generateColMap = (context) => {
             }
         }
     });
-    return colMap;
-};
 
+    // If targetAssays only contains nulls, then just empty it so we can skip the targetAssay row.
+    if (!targetAssays.some(assay => assay !== null)) {
```
zoldello: It seems like .every is a better option than .some in this case. The inverse of !== gets confusing sometimes.

src/encoded/static/components/matrix_entex.js
```diff
@@ -226,59 +273,114 @@ const convertContextToDataTable = (context) => {
...
     // Convert column map to an array of column map values sorted by column number for displaying
     // in the matrix header.
     const sortedCols = Object.keys(colMap).map(assayColKey => colMap[assayColKey]).sort((colInfoA, colInfoB) => colInfoA.col - colInfoB.col);
 
-    // Generate array of names of assays that have targets and don't collapse their targets, for
-    // rendering those columns as disabled.
-    const colCategoriesWithSubcategories = Object.keys(colMap).filter(colCategoryName => colMap[colCategoryName].hasSubcategories && !collapsedAssays.includes(colCategoryName));
+    // Generate the matrix header row labels for the assays with targets. Need a max-width inline
+    // style so that wide labels don't make the target columns expand.
+    let targetAssayHeader;
```
zoldello: Should targetAssayHeader be in the if-statement as it seems not to be used anywhere else. If yes, then it can be const and be just a tad faster than using let in evergreen browsers.

src/encoded/static/components/matrix_entex.js
forresttanaka: I agree it’s less confusing, but this way the loop will bail early if it finds a non-null assay. With .every(), it always goes through the entire targetAssays array no matter what.

zoldello:
@forresttanaka  Not related to code-review, just a UX observation. It would be nice if placeholders were used. They could be used to tell the user what the circles and other icons means. While most people should know the male/female venus/mars symbols; it is not always safe to assume so. I for one keep forgetting which is which.

zoldello:
@forresttanaka  Just to be thorough, did you do IE11 smoke test? I am not sure how good it is with handling symbols.

src/encoded/static/components/matrix_entex.js
```diff
@@ -41,25 +64,25 @@ const entexDonors = [
         accession: 'ENCDO845WKR',
         cssSuffix: 'male1',
         voice: 'Male 1',
```
zoldello: voice?

src/encoded/static/components/matrix_entex.js
```diff
@@ -106,24 +129,36 @@ const DonorLegend = () => (
  * in the matrix. Any determination of column order or inclusion/exclusion happens in this
  * function. Do not rely on the order of the keys. The resulting object has the form:
  * {
- *   'cat1': {col: 0, category: 'cat1', hasSubcategories: true},
+ *   'cat1': {col: 0, category: 'cat1'},
  *   'cat1|subcat1': {col: 1, category: 'cat1', subcategory: 'subcat1'},
  *   'cat1|subcat2': {col: 2, category: 'cat1', subcategory: 'subcat2'},
```
zoldello: I have no seems anyone use pipe in a variable name

src/encoded/static/components/matrix_entex.js
forresttanaka: Text for screen readers to say.

src/encoded/static/components/matrix_entex.js
forresttanaka: It’s a key, not a variable name.

forresttanaka:
I chose Courier for Windows because it was one that worked in IE11.

forresttanaka:
Since the male/female symbols were chosen and approved by Idan and Ben, I think it’s fairly safe that the target audience knows what they mean. It was also approved by the ENTEx committee.

src/encoded/static/components/matrix_entex.js
forresttanaka: I’d like to address this one. I think it might need to be a const, but I need to see it in more context after I’m done with one thing first.

forresttanaka: Yes, you’re right. I pushed a change to make it a const, and updated the demo: https://encd-5053-entex-update-pr2-fytanaka.demo.encodedcc.org/entex-matrix/?type=Experiment&status=released&internal_tags=ENTEx

PR #3123, opened 2020-01-29
ENCD-5068-update-regex-pattern-for-IHEC-dbxref

yunhailuo:
…umber

### DIFF
--- a/src/encoded/upgrade/dataset.py
+++ b/src/encoded/upgrade/dataset.py
@@ -515,3 +515,31 @@ def annotation_25_26(value, system):
     elif enc_ver == 'Blacklists':
         value.pop('encyclopedia_version', None)
     return
+
+
+@upgrade_step('aggregate_series', '1', '2')
+@upgrade_step('annotation', '26', '27')
+@upgrade_step('experiment_series', '1', '2')
+@upgrade_step('functional_characterization_experiment', '1', '2')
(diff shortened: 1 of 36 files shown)


src/encoded/upgrade/dataset.py
```diff
@@ -514,4 +514,32 @@ def annotation_25_26(value, system):
...
+@upgrade_step('matched_set', '15', '16')
+@upgrade_step('organism_development_series', '15', '16')
+@upgrade_step('publication_data', '15', '16')
+@upgrade_step('reference_epigenome', '16', '17')
+@upgrade_step('reference', '16', '17')
+@upgrade_step('replication_timing_series', '15', '16')
+@upgrade_step('single_cell_rna_series', '1', '2')
+@upgrade_step('treatment_concentration_series', '15', '16')
+@upgrade_step('treatment_time_series', '16', '17')
+@upgrade_step('ucsc_browser_composite', '15', '16')
```
gabdank: I think you missed Project and Experiment here? perhaps other dataset types?

yunhailuo: Add project. Experiment has its own dbxrefs which doesn't have IHEC: https://github.com/ENCODE-DCC/encoded/blob/dev/src/encoded/schemas/experiment.json#L109-L123

PR #3126, opened 2020-01-29
ENCD-5057-add-experiment-audit-for-no-processed-data

jnau:


src/encoded/audit/experiment.py
```diff
@@ -3990,6 +3990,23 @@ def audit_experiment_eclip_queried_RNP_size_range(value, system, excluded_types)
...
+
+    if not raw_data:
+        return
+
+    if not processed_data:
+        detail = 'Experiment {} only has raw data and does not contain any processed data'.format(audit_link(path_to_text(value['@id']), value['@id']))
+        yield AuditFailure('lacking processed data', detail, level='WARNING')
+    else:
+        return
+
```
gabdank: You don't need else return, it is the end of the function anyway

src/encoded/audit/functional_characterization_experiment.py
```diff
@@ -0,0 +1,42 @@
...
+        value.get('original_files'), excluded_files)
+
+    for function_name in function_dispatcher_with_files.keys():
+        yield from function_dispatcher_with_files[function_name](value, system, files_structure)
+
+    excluded_types = excluded_files + ['deleted', 'replaced']
+    for function_name in function_dispatcher_without_files.keys():
+        yield from function_dispatcher_without_files[function_name](value, system, excluded_types)
+
+    return
```
gabdank: I don't think we need this return statement

src/encoded/tests/test_audit_experiment.py
```diff
@@ -1,5 +1,6 @@
 import pytest, re
 
+
```
gabdank: I don't think you need to add a new line here

src/encoded/audit/functional_characterization_experiment.py
jnau: src/encoded/audit/experiment.py also has this return statement. should i remove it there or ignore it?

hitz:
branch has conflicts.

hitz:
Merging it without new demo, only changes were to test code; tests pass.

PR #3127, opened 2020-01-30
ENCD-5053-display-target-expression-percentile

forresttanaka:
The ticket also asked to display the minimum and maximum range, but the code to display them already exists. This just adds the expression percentile. Its value can potentially be zero, and should be displayed if it is.

### DIFF
--- a/src/encoded/static/components/experiment.js
+++ b/src/encoded/static/components/experiment.js
@@ -507,6 +507,12 @@ const ExperimentComponent = ({ context, auditIndicators, auditDetail }, reactCon
                                             <dd>{context.target_expression_range_minimum}% &ndash; {context.target_expression_range_maximum}%</dd>
                                         </div>
                                     : null}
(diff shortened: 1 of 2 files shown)


forresttanaka:
This demo shows an FCC with a target expression percentile: https://encd-5053-fcc-target-pr1-fytanaka.demo.encodedcc.org/functional-characterization-experiments/ENCSR652MCH/

Not part of the ticket, but no production data has target expression range minimums or maximums, so here's one I modified that demonstrates these: https://encd-5053-fcc-target-pr1-fytanaka.demo.encodedcc.org/functional-characterization-experiments/ENCSR024WBS/

The schema enforces that you can have data for the target expression range minimum if and only if you have a maximum. The schema also enforces that you can have target expression ranges OR a target expression percentile, but not both. And of course you can set an FCC to have none of these, as is the case for all production data right now.

PR #3128, opened 2020-01-30
ENCD-5056-computation-model-object

khzlin:


forresttanaka:
@khzlin Could you make one more change in this PR? For some reason your file table from the test inserts is scrolling vertically a couple pixels even though I can find no other FileSet tables that do that. But yours should be fixed by going to src/encoded/static/scss/encoded/modules/_tables.scss and changing this section:

```
    @at-root #{&}__scrollarea {
        overflow-x: auto;
    }
```

to

```
    @at-root #{&}__scrollarea {
        overflow-x: auto;
        overflow-y: hidden;
    }
```

That should stop the vertical scrolling without affecting the horizontal scrolling that's needed on occasion.

src/encoded/static/components/dataset.js
```diff
@@ -407,6 +407,149 @@ const PublicationData = auditDecor(PublicationDataComponent);
...
+                            {context.description ?
+                                <div data-test="description">
+                                    <dt>Description</dt>
+                                    <dd>{context.description}</dd>
+                                </div>
+                            : null}
+
+                            {context.computational_model_type ?
+                                <div data-test="type">
+                                    <dt>Annotation type</dt>
```
forresttanaka: Is this the right title? In the schema it’s called “Computational model type.”

khzlin: changed title

PR #3132, opened 2020-02-01
ENCD-5049-mint-mixture-identifier

meenakshikagda:


src/encoded/schemas/library.json
```diff
@@ -112,6 +112,12 @@
         "lot_id":{
             "description": "The lot identifier provided by the vendor, for nucleic acids or proteins purchased directly from a vendor (e.g. total RNA)."
         },
+        "mint_mixture_identifier": {
+            "title": "Chromatin preps mixture identifier",
+            "description": "Specifies the identifier to allow normalization for MINT-ChIP experiments.",
+            "type": "string",
```
gabdank: Please see my comment in the changelog. We don't know if and when this value going to be used. Update accordingly.

hitz:
@gabdank please approve

PR #3133, opened 2020-02-05
ENCD-5026 toolbar updates

emro:
Updates the navigation menus!
- Organizes the "Data" menu into two sub-categories: "Functional Genomics data" and "Functional Characterization data"
- Adds link to ChIP data matrix (see Phil's branch)
- Makes the menus hoverable on desktop and retains click interaction for mobile / touch devices


src/encoded/static/components/navigation.js
```diff
@@ -86,11 +86,56 @@ export default class Navigation extends React.Component {
...
+        // Check if device has touch events
+        function isMobile() {
+            try {
+                document.createEvent('TouchEvent');
+                return true;
+            } catch (evt) {
+                return false;
+            }
+        }
+        const isMobileDevice = isMobile();
```
forresttanaka: I think this should be added to browserfeat.js so that any code can tell touch events are supported, and CSS can target it. But not necessarily in this ticket. [Modernizr](https://modernizr.com/docs/) has a "touchevents"/"no-touchevents" CSS class.

PR #3137, opened 2020-02-09
ENCD-5088 graphlib crash fix

forresttanaka:
This changes package-lock.json for ENCD-5037 _might_ have caused this crash. I haven't confirmed this yet, but I think updating Valis in package-lock.json and updating package-lock.json brought in — maybe — the wrong version of graphlib.

I’ll describe what I did for this Valis update to package-lock.json in the ticket.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1143,9 +1143,9 @@
       }
     },
     "@google-cloud/common": {
-      "version": "2.2.2",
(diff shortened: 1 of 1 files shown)


forresttanaka:
Including both @hitz and @emro in case one or the other can code review this first.

PR #3138, opened 2020-02-10
ENCD-5051-generating-QC-table-for-experimental-series

yunhailuo:


src/encoded/static/components/experiment_series.js
```diff
@@ -2,53 +2,259 @@ import React from 'react';
...
 import { Breadcrumbs } from './navigation';
-import { ItemAccessories, InternalTags } from './objectutils';
+import { requestObjects, ItemAccessories, InternalTags } from './objectutils';
 import { PickerActions, resultItemClass } from './search';
-import { SortTablePanel, SortTable } from './sorttable';
+import { SortTablePanel } from './sorttable';
 import Status, { getObjectStatuses, sessionToAccessLevel } from './status';
 
 
+const selectedAssembly = ['GRCh38'];
```
forresttanaka: If you could comment this, describing what situations could make a future dev need to change it, that would help maintenance.

src/encoded/static/components/experiment_series.js
```diff
@@ -2,53 +2,259 @@ import React from 'react';
...
+        display: (experiment, meta) => {
+            const low = 5000000;
+            let minimal = 10000000;
+            let recommended = 20000000;
+            if (experiment.target.investigated_as.includes('broad histone mark')) {
+                minimal = 35000000;
+                recommended = 45000000;
+            }
+            const filteredFiles = experiment.files.filter(f => f.biological_replicates.includes(meta.bioRepNum));
+            const qualityMetrics = filteredFiles.map(f => f.quality_metrics).flat();
```
forresttanaka: Array.prototype.flat() doesn’t work in IE11 and probably not Legacy Edge either, and babel doesn’t polyfill it. You could use your own inline polyfill here like https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat#reduce_and_concat

src/encoded/static/components/experiment_series.js
```diff
@@ -2,53 +2,259 @@ import React from 'react';
...
+                    if (qualityMetricsByReplicate[num]) {
+                        qualityMetricsByReplicate[num].push(qm[field]);
+                    } else {
+                        qualityMetricsByReplicate[num] = [qm[field]];
+                    }
+                });
+            }
+        });
+    });
+    return Object.entries(qualityMetricsByReplicate).reduce(
```
forresttanaka: `Object.entries` will cause IE11 to crash because that browser doesn’t implement this function, and babel doesn’t polyfill it. There is a defined alternative, but I think I would just implement this in a different way, probably using Object.keys instead (which is polyfilled by babel) and then getting the corresponding `qms` values with `qualityMetricsByReplicate[rep]` within the reduce loop. So the parameters to `reduce()` would be `(qmByRep, rep)`.

src/encoded/static/components/experiment_series.js
```diff
@@ -2,53 +2,259 @@ import React from 'react';
...
+        (qmByRep, [rep, qms]) => {
+            qmByRep[rep] = _.uniq(qms);
+            return qmByRep;
+        },
+        {}
+    );
```
forresttanaka: This doesn’t look to me like a reduce() operation. You’re starting out with something like {1: [...], 3: [...]} and ending up with the same thing, but with duplicates in each array removed. You’re also mutating your accuulator (`qmByRep`) which I don’t think is a defined operation — it’s not the point of reduce() anyway. You could do `Object.keys(qualityMetricsByReplicate).forEach()` and just build up `qmByRep` in that loop in pretty much the same way.

You could also check for duplicates in line 37 above, but that would be four nested loops, which I think would be difficult to justify.

src/encoded/static/components/experiment_series.js
```diff
@@ -85,85 +291,71 @@ class ExperimentSeriesComponent extends React.Component {
...
-                if (datasetAudits && datasetAudits.length > 0) {
-                    datasetAudits.forEach((datasetAudit) => {
-                        auditsByDataset[datasetAudit['@id']] = datasetAudit.audit || {};
-                    });
-                }
-                this.setState({ auditsByDataset });
-            });
-        }
+        this.getViewableDatasets();
     }
```
forresttanaka, line 395: Is it a problem to not have `componentDidUpdate` here? If someone logs in while viewing the page, could the page become out of date?

src/encoded/static/components/experiment_series.js
```diff
@@ -85,85 +291,71 @@ class ExperimentSeriesComponent extends React.Component {
...
+        const viewableStatuses = getObjectStatuses('Dataset', accessLevel, viewableDatasetStatuses);
+        const viewableDatasets = this.props.context.related_datasets.reduce(
+            (datasetObjects, dataset) => {
+                if (viewableStatuses.includes(dataset.status)) {
+                    datasetObjects[dataset['@id']] = dataset;
+                }
+                return datasetObjects;
+            },
+            {}
+        );
```
forresttanaka: Simiarly to above, this doesn’t look like a reduce() operation. This is generating a filtered object keyed by dataset.@id which could be done in a `forEach`.

src/encoded/static/components/experiment_series.js
```diff
@@ -192,21 +384,82 @@ class ExperimentSeriesComponent extends React.Component {
...
+                    if (seriesAssayTermName.filter(term => column.assay_term_name.includes(term)).length > 0) {
+                        colObjs[columnId] = column;
+                    }
+                } else {
+                    colObjs[columnId] = column;
+                }
+                return colObjs;
+            },
+            {}
+        );
```
forresttanaka, line 517: Similar to above. I would use a forEach instead of a reduce().

src/encoded/static/components/experiment_series.js
```diff
@@ -192,21 +384,82 @@ class ExperimentSeriesComponent extends React.Component {
...
+                // for extracting quality metrics
+                const filesByDesiredAssembly = experiment.files.reduce(
+                    (fileAtIds, f) => {
+                        if (selectedAssembly.includes(f.assembly)) {
+                            fileAtIds.push(f['@id']);
+                        }
+                        return fileAtIds;
+                    },
+                    []
+                );
```
forresttanaka: Similar comment to above.

src/encoded/static/components/experiment_series.js
```diff
@@ -192,21 +384,82 @@ class ExperimentSeriesComponent extends React.Component {
...
             );
 
             // Collect unique internal_tags from all relevant experiments.
-            internalTags = _.uniq(viewableDatasets.reduce((allInternalTags, experiment) => (
+            internalTags = _.uniq(Object.values(viewableDatasets).reduce((allInternalTags, experiment) => (
                 experiment.internal_tags && experiment.internal_tags.length > 0 ? allInternalTags.concat(experiment.internal_tags) : allInternalTags
             ), []));
+
+            // Experiment table data
+            Object.values(viewableDatasets).forEach((experiment) => {
```
forresttanaka: `Object.values` will crash on IE11. You can instead use `Object.keys(viewableDAtasets)` and then witin the `forEach` loop use `viewableDAtasets[whatever your loop variable is]` to get the corresponding value.

src/encoded/static/components/experiment_series.js
```diff
@@ -192,21 +384,82 @@ class ExperimentSeriesComponent extends React.Component {
...
+                const analyses = experiment.analyses || [];
+                const selectedAnalysis = analyses.filter(
+                    analysis => analysis.files.every(f => filesByDesiredAssembly.includes(f))
+                ).sort((a, b) => a.length - b.length)[0] || { files: [] };
+                experiment.files = experiment.files.filter(
+                    f => selectedAnalysis.files.includes(f['@id'])
+                );
+                // Number of subrows to be shown corresponds to the number of biological replicates. Show at least one by default.
+                let bioRepNums = _.uniq(
+                    experiment.files.map(f => f.biological_replicates).flat()
```
forresttanaka: Polyfill `flat()`.

src/encoded/static/components/experiment_series.js
```diff
@@ -192,21 +384,82 @@ class ExperimentSeriesComponent extends React.Component {
...
+                if (bioRepNums.length === 0) {
+                    bioRepNums = [''];
+                }
+                // Extract experiment metadata
+                const rowCount = bioRepNums.length;
+                const datasetRows = [];
+                bioRepNums.sort((a, b) => a - b).forEach((num, i) => {
+                    const cellMeta = { bioRepNum: num, isAuthorized, rowCount };
+                    datasetRows.push(
+                        <tr key={experiment['@id'].concat('_', i)}>
```
forresttanaka: I think `num` will be unique for each row, right? It would be better to use `num` instead of `i` because `i` always starts at 0 even if you change the order of the rows, and React might not notice you changed the order. If you use `num` instead, then React will notice if you change the order because `num` will move with the data.

src/encoded/static/components/experiment_series.js
```diff
@@ -192,21 +384,82 @@ class ExperimentSeriesComponent extends React.Component {
...
+                            {Object.entries(columns).map(([columnId, expCol]) => (columns[columnId].replicateSpecific || i === 0 ?
```
forresttanaka: Use `Object.keys` instead of `Object.entries`.

src/encoded/static/components/experiment_series.js
```diff
@@ -192,21 +384,82 @@ class ExperimentSeriesComponent extends React.Component {
...
+                                (expCol.display ? expCol.display(experiment, cellMeta) : <td key={columnId} rowSpan={columns[columnId].replicateSpecific ? 1 : rowCount}>{expCol.getValue ? expCol.getValue(experiment, cellMeta) : experiment[columnId]}</td>)
```
forresttanaka: Nested ?: aren’t allowed in AirBnB, but ESLint rarely detects it. Better to break it up (might have to use a temporary variable to hold one value).

src/encoded/static/components/experiment_series.js
```diff
@@ -243,6 +496,15 @@ class ExperimentSeriesComponent extends React.Component {
                                     </div>
                                 : null}
 
+                                {context.target && context.target.length > 0 ?
+                                    <div data-test="target">
+                                        <dt>Target</dt>
+                                        <dd>
+                                            {context.target.map((targetObj, i) => <span key={i}>{i !== 0 ? ', ' : ''}<a href={targetObj['@id']}>{targetObj.label}{' ('}{targetObj.organism ? <i>{targetObj.organism.scientific_name}</i> : <span>{targetObj.investigated_as[0]}</span>}{')'}</a></span>)}
```
forresttanaka: Can you use `targetObj['@id']` as the `<span>` key instead of `i`? That way the key moves with the data as I mentioned.

forresttanaka (changes requested):
Really impressive job! That’s really complex. Hopefully we’ll have a SortTable that can be configured to use rowSpan someday.

src/encoded/static/components/experiment_series.js
yunhailuo: That's a nice doc and example. Thank you!

src/encoded/static/components/experiment_series.js
yunhailuo: Yes

src/encoded/static/components/experiment_series.js
yunhailuo: Is using if here OK?

forresttanaka: That’s often the best solution.

src/encoded/static/components/experiment_series.js
```diff
@@ -2,53 +2,260 @@ import React from 'react';
...
+const selectedAssembly = ['GRCh38'];
+
+/**
+ * Giving an experiment, extract one specific quality metric from its files.
+ * Extracted quality metrics are grouped by corresponding biological replicate number.
+ * For a specific biological replicate, there could be more than one quality metric
+ * values extracted. They are all stored uniquely in an array.
+ *
+ * @return {object} Quality metrics grouped by biological replicate number
+ *                 none.
```
forresttanaka: Is this a stray “none”? it’s in an odd place and doesn’t seem to fit the sentence.

forresttanaka (changes requested):
OK, just one extremely small comment change only because it looks a little strange, and I think it's ready. It works in IE11 fine.

src/encoded/static/components/experiment_series.js
yunhailuo: Missed. Sorry. Fixed.

forresttanaka (dismissed):
Nice!

src/encoded/static/components/antibody.js
```diff
@@ -647,7 +647,7 @@ StatusIndicator.defaultProps = {
 
 
 // Display the status indicators for one target
-const StatusIndicators = (props) => {
+export const StatusIndicators = (props) => {
```
forresttanaka: Was this a change you needed temporarily? Looks like this isn’t used outside of antibody.js.

src/encoded/static/components/experiment_series.js
```diff
@@ -2,53 +2,296 @@ import React from 'react';
...
 import { Breadcrumbs } from './navigation';
-import { ItemAccessories, InternalTags } from './objectutils';
+import { requestObjects, ItemAccessories, InternalTags } from './objectutils';
 import { PickerActions, resultItemClass } from './search';
-import { SortTablePanel, SortTable } from './sorttable';
+import { SortTablePanel } from './sorttable';
 import Status, { getObjectStatuses, sessionToAccessLevel } from './status';
 
 
+// Only GRCh38 analysis will be used for QC reporting
```
forresttanaka: “…and mm10,” or you might not want to mention the specific assemblies and just say what `selectedAssembly` is for generally.

src/encoded/static/components/experiment_series.js
```diff
@@ -85,85 +328,93 @@ class ExperimentSeriesComponent extends React.Component {
...
-                this.getDatasetAudits();
-            }
+        const datasetAtIds = Object.keys(this.statusFilteredDatasets());
+        const updatedDatasetAtIds = Object.keys(this.state.viewableDatasets);
+        const symmDiffAtIds = datasetAtIds.filter(
+            i => !updatedDatasetAtIds.includes(i)
+        ).concat(
+            updatedDatasetAtIds.filter(i => !datasetAtIds.includes(i))
+        );
+        if (symmDiffAtIds && symmDiffAtIds.length > 0) {
```
forresttanaka: `Array.filter` is guaranteed to give you _something_ — a zero-length array at minimum — so you don’t need to check for `symmDiffAtIds`’s existence, though it really doesn’t make a lot of difference performance-wise of course.

forresttanaka (changes requested):
Just a couple minor and optional things, though antibody.js should be reverted it looks like.

src/encoded/static/components/antibody.js
yunhailuo: You are right. This is a mistake. I changed my strategy and won't use this component.

src/encoded/static/components/experiment_series.js
yunhailuo: I'll try to make it general.

src/encoded/static/components/experiment_series.js
yunhailuo: If people take efforts to guarantee it, I respect that and should make my code simpler 😊.

src/encoded/static/components/experiment_series.js
```diff
@@ -16,7 +16,7 @@ import { SortTablePanel } from './sorttable';
 import Status, { getObjectStatuses, sessionToAccessLevel } from './status';
 
 
-// Only GRCh38 analysis will be used for QC reporting
+// Only analysis on selcted assembly will be used for QC reporting
```
forresttanaka: Since this will be quick, might as well correct “selcted” to “selected.”

forresttanaka (changes requested):

src/encoded/static/components/experiment_series.js
yunhailuo: Sorry. That's stupid. Fixed

forresttanaka (dismissed):

yunhailuo:
@forresttanaka if this is helpful, your previous review has covered all commits except the last one: 4625068f092b40febc26bdeaff2a0d150596e3b2

src/encoded/static/components/experiment_series.js
```diff
@@ -1,58 +1,339 @@
...
 
+// Display the color legend for quality metrics.
+const QualityMetricLegend = () => (
+    <div className="qc-legend">
+        <div className="qc-legend-item"><b>Quality metric status:</b></div>
+        <div className="qc-legend-item qc-report--ideal">Recommended</div>
+        <div className="qc-legend-item qc-report--warning">Sufficient</div>
+        <div className="qc-legend-item qc-report--not-compliant">Insufficient</div>
+        <div className="qc-legend-item qc-report--error">Error</div>
+    </div>
```
forresttanaka, line 352: I won’t gate this review on this because it’s not a technical problem, but a BEM convention (and others have done it wrong too, probably including me). But if you use a Modifier (the “M” in BEM), then it should come with the Element (the “E” in BEM). So each of the secondary items should have `className="qc-report qc-report--ideal"` and so on. So “qc-report” would have the styles common to all of them, and ”qc-report--warning“ would have the additional styles needed for that one thing.

forresttanaka (changes requested):
I think you’re putting in some changes, so here are the comments I have. Nothing big.

src/encoded/static/components/experiment_series.js
yunhailuo: Oh, right. You mentioned this before but I forgot. I would follow that and add here.

forresttanaka (dismissed):
Nice! Really like it. Thanks for implementing all my hand-waving philosophical things. Nice commenting too.

yunhailuo:
Added one more change to exclude unreplicated (experiment.replicates.length <= 1) experiments for peakCount per Idan's request @forresttanaka .

forresttanaka (approved):

PR #3139, opened 2020-02-10
ENCD-5089-add-account-creation-msg

zoldello:


src/encoded/static/components/app.js
```diff
@@ -577,6 +578,10 @@ class App extends React.Component {
         this.setState({ eulaModalVisibility: false });
     }
 
+    closeAccountCreationNotification() {
+        this.setState({ accountCreatedModalVisibility: false });
```
emro: I think you might want to intialize 'accountCreatedModalVisibility' the way that you initialize 'accountCreationFailedVisibility' on line 291.

zoldello: Done

PR #3147, opened 2020-02-18
ENCD-5083-restrict-IHEC-dbxref-to-RefEpi-and-Annotation

iyoungworth:


src/encoded/upgrade/dataset.py
```diff
@@ -544,3 +544,32 @@ def dataset_27_28(value, system):
...
+@upgrade_step('ucsc_browser_composite', '16', '17')
+def dataset_28_29(value, system):
+    # https://encodedcc.atlassian.net/browse/ENCD-5083
+    if not value.get('dbxrefs'):
+        return
+    new_dbxrefs = set()
+    for dbxref in value['dbxrefs']:
+        if not dbxref.startswith('IHEC:IHECRE'):
+            new_dbxrefs.add(dbxref)
+            continue
```
gabdank: I think we need to record this dbxref in notes, instead of discarding it.

iyoungworth: Makes sense, I've adjusted the upgrade to do that

PR #3151, opened 2020-02-19
ENCD-5106-add-control-link-FCE

bonitalam:


src/encoded/static/components/typeutils.js
```diff
@@ -425,19 +425,24 @@ ExperimentTable.defaultProps = {
...
+const tableContentMap = {
+    Experiment: 'Experiments',
+    FunctionalCharacterizationExperiment: 'Functional Characterization Experiments',
+};
 /**
  * Display a table of experiments with the dataset in `context` as a possible_controls.
  */
 export const ControllingExperiments = ({ context, items, total, url }) => {
     if (items.length > 0) {
+        const tableContent = context['@type'].map(typ => tableContentMap[typ]).filter(s => s)[0] || 'Experiments';
```
forresttanaka: I see what you’re doing here, but I think it’s too complicated for the task. You can achieve _nearly_ the same thing with:

```
const tableContent = tableContentMap[context['@type'][0]] || 'Experiments';
```

This assumes “Experiment” or “FunctionalCharacterizationExperiment” will always be at context['@type'][0], similar to assumptions we make throughout the code. But if you can’t make that assumption in this case, then what you have seems reasonable. Lots of ways to do that, none of them really tight that I’m aware of.

yunhailuo: Thank you. @bonitalam please do that.

PR #3155, opened 2020-02-21
ENCD-4953-calculated-tags-property-experiment

bonitalam:


src/encoded/types/experiment.py
```diff
@@ -177,6 +176,102 @@ def related_series(self, request, related_series):
...
+                replicateObject = request.embed(rep, '@@object?skip_calculated=true')
+                if replicateObject['status'] == 'deleted':
+                    continue
+                if 'library' in replicateObject:
+                    libraryObject = request.embed(replicateObject['library'], '@@object?skip_calculated=true')
+                    if libraryObject['status'] == 'deleted':
+                        continue
+                    if 'biosample' in libraryObject:
+                        biosampleObject = request.embed(libraryObject['biosample'], '@@object')
+                        if biosampleObject['status'] == 'deleted':
```
yunhailuo: Not sure if other status like "revoked" should be excluded here?

src/encoded/types/experiment.py
```diff
@@ -177,6 +176,102 @@ def related_series(self, request, related_series):
...
+                                        tag_dict = {'location': tag['location'], 'name': tag['name']}
+                                        if gm_object.get('modified_site_by_target_id'):
+                                            tag_dict.update({'target': gm_object.get('modified_site_by_target_id')})
+                                        else:
+                                            tag_dict.update({'target': 'none'})
+                                        modification_tags.append(tag_dict)
+                                tag_list.append(modification_tags)
+                                one_biosample_tags = [item for sublist in tag_list for item in sublist]
+                            protein_tags_lists.append(one_biosample_tags)
+                            protein_tags = [item for sublist in protein_tags_lists for item in sublist]
```
yunhailuo: For one thing, it looks like you are trying to grow `protein_tags` list with `tag_dict` here. Why don't we just append `tag_dict` to `protein_tags` directly?
For another thing, is `{'target': 'none'}` valid to be in `protein_tags` list?

src/encoded/types/experiment.py
```diff
@@ -177,6 +176,102 @@ def related_series(self, request, related_series):
     def superseded_by(self, request, superseded_by):
         return paths_filtered_by_status(request, superseded_by)
 
+    @calculated_property(schema={
+        "title": "Protein tags",
+        "description": "The protein tags introduced through the genetic modifications of biosamples investigated in the experiment.",
+        "comment": "Do not submit. This field is calculated through applied_modifications.",
+        "type": "array",
+        "notSubmittable": True,
+        "items": {
```
yunhailuo: If you don't return empty list below, you have `"minItems": 1,` here.

src/encoded/types/experiment.py
```diff
@@ -177,6 +176,102 @@ def related_series(self, request, related_series):
...
+                    if 'biosample' in libraryObject:
+                        biosampleObject = request.embed(libraryObject['biosample'], '@@object')
+                        if biosampleObject['status'] == 'deleted':
+                            continue
+                        genetic_modifications = biosampleObject.get('applied_modifications')
+                        tag_list = None
+                        if genetic_modifications:
+                            tag_list = []
+                            for gm in genetic_modifications:
+                                gm_object = request.embed(gm, '@@object')
```
yunhailuo: Can't skip calculation here?

src/encoded/types/experiment.py
bonitalam: Agreed, will add.

src/encoded/types/experiment.py
bonitalam: Will skip calculation here.

src/encoded/types/experiment.py
bonitalam: Not sure. For calculated biosample summary, only "deleted" objects were excluded. But I guess it wouldn't hurt to also consider revoked?

src/encoded/types/experiment.py
bonitalam: I think the reason why is because each introduced_tag might actually have multiple tags within one gm introduced_tag. In addition, each experiment could have several gms per biosample. I can try to clean it up a bit, so that I can minimize the amount of lists/flattening of the final protein_tags list. Some introduced_tags are not connected with a target but could be with "modified_site_by_gene_id" for example, so I guess adding the none option was to handle the cases where no target was specified. Do you suggest it saying something else?

PR #3158, opened 2020-02-21
ENCD-5033-index-user-made-accounts

zoldello:


src/encoded/auth0.py
```diff
@@ -127,7 +127,7 @@ def signup(context, request):
     result = collection_add(context, request, user_info)
     if not result or result['status'] != 'success':
         raise HTTPInternalServerError(explanation='attempt to create account was not successful')
-    return HTTPCreated(explanation='User created')
```
caseylitton: HTTPCreated looks like it might be an import.  As a developer when you remove an imported object and as code reviewer 
 when you see something removed that may be an imported object, it is important to check if we still need it in the file.  If the import is no longer used, then it should be removed from the file.  Please remove the HTTPCreated  import in this file and the test file if it is no longer used.

zoldello: Done.

PR #3159, opened 2020-02-21
ENCD-5119 IGV visualization

lrowe:
Ideally the IGV url would be configurable, but it exists in both JS and Python code so doing so would require a bigger refactoring.

We use IGV for visualizing vcf files, but that requires enabling tabix indexing to be performant, so have left that configuration commented out.

### DIFF
--- a/src/encoded/region_search.py
+++ b/src/encoded/region_search.py
@@ -797,6 +797,14 @@ def search_result_actions(request, doc_types, es_results, position=None):
                     ensembl_url = vis_format_url("ensembl", hub_url, assembly, pos)
                     if ensembl_url is not None:
                         browser_urls['Ensembl'] = ensembl_url
+                    igv_search_params = request.params.dict_of_lists()
+                    igv_search_params['assembly'] = [assembly]
+                    igv_params = {
+                        "sessionURL": request.route_url('batch_igv', _query=igv_search_params),
+                    }
+                    if pos is not None:
+                        igv_params['locus'] = pos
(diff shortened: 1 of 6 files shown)


hitz:
@lrowe @cricketsloan we implemented this but it appears to be somewhat buggy.   Your UI is different think?  Also some of our datasets don't show any data, are their strange defaults sent?

Do you know how to set a default (zoomed in) location?

src/encoded/region_search.py
```diff
@@ -797,6 +797,14 @@ def search_result_actions(request, doc_types, es_results, position=None):
                     ensembl_url = vis_format_url("ensembl", hub_url, assembly, pos)
                     if ensembl_url is not None:
                         browser_urls['Ensembl'] = ensembl_url
+                    igv_search_params = request.params.dict_of_lists()
+                    igv_search_params['assembly'] = [assembly]
+                    igv_params = {
+                        "sessionURL": request.route_url('batch_igv', _query=igv_search_params),
+                    }
+                    if pos is not None:
+                        igv_params['locus'] = pos
```
lrowe, line 806: @hitz This sets the position when you browse from region search.

lrowe:
Using @gabdank's test instance it looks like there are errors with both the UI (name not appearing in the dropdown list on experiment page) and tracks are not appearing for bigWigs in the igv.json. I should get to fixing this in the next month or so as I get to rebasing my local changes on the current enocoded version.

PR #3163, opened 2020-02-24
ENCD-4540 Cart visualizations DRAFT PR

forresttanaka:
@keenangraham Could you look over batch_download.py? I added a couple columns to metadata_tsv and included those same couple for batch_download. I use this data to determine whether the “visualize=true” query string parameter was included. If it is, then I only include visualizable files in the files.txt and metadata.tsv files.


src/encoded/batch_download.py
```diff
@@ -175,6 +177,21 @@ def get_peak_metadata_links(request):
...
 
+
+def is_file_visualizable(file):
+    """Determines whether a file can be visualized in a genome browser.
+    Needs to be kept in sync with isFileVisualizable in objectutils.js.
+
+    Keyword arguments:
+    file -- file object including props to test for visualizability
+    """
+    return (
```
keenangraham: Formatting nit:

```
def is_file_visualizable(file):
    """Determines whether a file can be visualized in a genome browser.
    Needs to be kept in sync with isFileVisualizable in objectutils.js.
    Keyword arguments:
    file -- file object including props to test for visualizability
    """
    conditions = [
        file.get('file_format') in [
            'bigWig',
            'bigBed',
        ],
        file.get('file_format_type') not in [
            'bedMethyl',
            'bedLogR',
            'idr_peak',
            'tss_peak',
            'pepMap',
            'modPepMap',
        ],
        file.get('status') in [
            'released',
            'in progress',
            'archived',
        ],
    ]
    return all(conditions)
```

src/encoded/batch_download.py
```diff
@@ -395,6 +414,11 @@ def metadata_tsv(context, request):
         for p in param_list.get('@id', [])
     ]
     qs.drop('limit')
+
+    # Determine if "visualizable=true" in query string to select only visualizable files.
+    visualizable_only = 'true' in param_list.get('visualizable', [])
```
keenangraham: I think you can also do:

```
visualizable_only = qs.is_param('visualizable', 'true')
```

https://github.com/ENCODE-DCC/snovault/blob/02f7a9cbb70fa21964fbc445ab41d88fa91580a5/src/snovault/elasticsearch/searches/parsers.py#L31

src/encoded/batch_download.py
```diff
@@ -462,9 +487,13 @@ def batch_download(context, request):
...
         ('limit', 'all'),
         ('field', 'files.href'),
-        ('field', 'files.restricted')
+        ('field', 'files.restricted'),
+        ('field', 'files.file_format'),
+        ('field', 'files.file_format_type'),
+        ('field', 'files.status'),
     ]
     qs = QueryString(request)
+    param_list = qs.group_values_by_key()
```
keenangraham: Looks like you could avoid this call if you did `visualizable_only = qs.is_param('visualizable', 'true')` instead.

src/encoded/batch_download.py
forresttanaka: Oh wow. `qs.is_param` makes that way cleaner. Thanks!

src/encoded/batch_download.py
hitz: Isn't this code supposed to be in visualization.py?  Or some analog?  Not that I would recommend it unless you were really into "italian food"?

forresttanaka: Huh. I never thought of visualization.py as a kind of library you import utility functions from, but I guess that makes sense.

hitz: probably visualization and batch_download and cart etc. should all import a vis module.  Perhaps this is vis_defines.py?

PR #3165, opened 2020-02-24
ENCD-5101-add-target-to-annotation-summary-page

jnau:


src/encoded/static/components/dataset.js
```diff
@@ -121,6 +121,27 @@ const AnnotationComponent = (props, reactContext) => {
...
+                            {context.targets && context.targets.length > 0 ?
+                                <div data-test="targets">
+                                    <dt>Target</dt>
+                                    <dd>
+                                        {context.targets.map((target, i) =>
+                                        <span>
+                                            {i > 0 ? <span>, </span> : null}
+                                            <a href={target['@id']}>{target.label}</a>
+                                        </span>
+                                        )}
```
forresttanaka: You’ll need to indent this `<span>` section one to the right. It shows up as a formatting error in my editor.

Also, and optionally, you can replace the `<span>`/`</span>` tags (the ones within the `context.target` loop; not the ones around the comma) with `<React.Fragment>`/`</React.Fragment>` tags. The advantage is the generated HTML with React.Fragment won’t include `<span>` tags; it generates no tags at all so the HTML is a little more efficient.

PR #3166, opened 2020-02-25
ENCD-5069 fix error in gene search

emro:
Gene searches were failing for annotations because urls were configured incorrectly (urls were only configured for experiments). This update configures urls properly for either case.

### DIFF
--- a/src/encoded/static/components/genome_browser.js
+++ b/src/encoded/static/components/genome_browser.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import _ from 'underscore';
+import url from 'url';
 import { FetchedData, Param } from './fetched';
 import { BrowserFeat } from './browserfeat';
 import AutocompleteBox from './region_search';
@@ -445,8 +446,8 @@ class GenomeBrowser extends React.Component {
 
     handleOnFocus() {
         this.setState({ showAutoSuggest: false });
-        const pageType = this.context.location_href.indexOf('/experiments') > -1 ? 'experiments' : 'annotations';
-        const coordinateHref = `${this.context.location_href.split(`/${pageType}/`)[0]}/suggest/?genome=${this.state.genome}&q=${this.state.searchTerm}`;
+        const urlComponents = url.parse(this.context.location_href);
+        const coordinateHref = `${urlComponents.protocol}//${urlComponents.host}/suggest/?genome=${this.state.genome}&q=${this.state.searchTerm}`;
         getCoordinateData(coordinateHref, this.context.fetch).then((response) => {
             // Find the response line that matches the search
             const responseIndex = response['@graph'].findIndex(responseLine => responseLine.text === this.state.searchTerm);
--- a/src/encoded/static/components/genome_browser.js
+++ b/src/encoded/static/components/genome_browser.js
@@ -445,7 +445,13 @@ class GenomeBrowser extends React.Component {
 
     handleOnFocus() {
         this.setState({ showAutoSuggest: false });
-        getCoordinateData(`${this.context.location_href.split('/experiments/')[0]}/suggest/?genome=${this.state.genome}&q=${this.state.searchTerm}`, this.context.fetch).then((response) => {
+        let coordinateHref = '';
+        if (this.context.location_href.indexOf('/experiments') > -1) {
+            coordinateHref = `${this.context.location_href.split('/experiments/')[0]}/suggest/?genome=${this.state.genome}&q=${this.state.searchTerm}`;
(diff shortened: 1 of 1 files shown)


src/encoded/static/components/genome_browser.js
```diff
@@ -445,7 +445,13 @@ class GenomeBrowser extends React.Component {
 
     handleOnFocus() {
         this.setState({ showAutoSuggest: false });
-        getCoordinateData(`${this.context.location_href.split('/experiments/')[0]}/suggest/?genome=${this.state.genome}&q=${this.state.searchTerm}`, this.context.fetch).then((response) => {
+        let coordinateHref = '';
```
zoldello: This is fine. But I think it can be simplified to be: 

```
const type = this.context.location_href.indexOf('/experiments') > -1 ? ‘experiment : ‘annotations’;

const coordinateHref = `${this.context.location_href.split(‘/${type}/‘)[0]}/suggest/?genome=${this.state.genome}&q=${this.state.searchTerm}`;
```

src/encoded/static/components/genome_browser.js
```diff
@@ -445,7 +445,13 @@ class GenomeBrowser extends React.Component {
...
+        if (this.context.location_href.indexOf('/experiments') > -1) {
```
zoldello: If you look at the traffic, the search is performed twice. I am not sure why and it may be outside the scope of this ticket.

Is there ever a case where the type is not an experiment nor annotation?

zoldello (dismissed):
The code looks fine. My comments are not significant.

src/encoded/static/components/genome_browser.js
emro: I like this way of writing the code better, it is more succinct!

src/encoded/static/components/genome_browser.js
emro: I think that looking into the search being performed twice might be a new ticket.

I thought there wasn't a case where the type is not an experiment nor annotation but I checked with @hitz and he said this was not a valid assumption, that the browser is already on the cart pages and could be on any number of search related pages. So I updated again to make the query more general so that it should work for any page.

src/encoded/static/components/genome_browser.js
```diff
@@ -445,7 +446,9 @@ class GenomeBrowser extends React.Component {
 
     handleOnFocus() {
         this.setState({ showAutoSuggest: false });
-        getCoordinateData(`${this.context.location_href.split('/experiments/')[0]}/suggest/?genome=${this.state.genome}&q=${this.state.searchTerm}`, this.context.fetch).then((response) => {
+        const urlComponents = url.parse(this.context.location_href);
+        const coordinateHref = `${urlComponents.protocol}//${urlComponents.host}/suggest/?genome=${this.state.genome}&q=${this.state.searchTerm}`;
```
zoldello: Is there a reason you are making an absolute URL rather than a relative one?

emro: Yes. If we're making sure this is generic, then we can't be certain what the url is, and need to account for how there could be more parameters than just /experiments/ENCDXXXX or /annotations/ENCSRXXXX. For instance, I'm not sure what the cart page browser url will be, but I do know that we want to use the root of that page to construct this query. So if we were to try to construct a relative url in that case, we would need to know how far up the directory structure to travel. It's simpler, and more robust, in my opinion, to just specify the full url.

emro: Ok, tried this and it seems to work! Hopefully it also works on Forrest's cart page. He says he hasn't tried that yet.

emro:
Rebased, tests are passing

src/encoded/static/components/genome_browser.js
```diff
@@ -445,7 +446,8 @@ class GenomeBrowser extends React.Component {
 
     handleOnFocus() {
         this.setState({ showAutoSuggest: false });
-        getCoordinateData(`${this.context.location_href.split('/experiments/')[0]}/suggest/?genome=${this.state.genome}&q=${this.state.searchTerm}`, this.context.fetch).then((response) => {
+        const coordinateHref = `/suggest/?genome=${this.state.genome}&q=${this.state.searchTerm}`;
```
zoldello, line 449: When do you use `/experiment/`?

zoldello (approved):
The code looks okay. I found wrong.

src/encoded/static/components/genome_browser.js
zoldello: My question can be ignored. "Experiment" is not needed in the URL. It was just unnecessarily used to split a URL to make a new one and not part of the new one.

PR #3168, opened 2020-02-25
ENCD-5072 Library construction platform display

forresttanaka:


src/encoded/static/components/experiment.js
```diff
@@ -302,6 +302,46 @@ LibrarySubmitterComments.propTypes = {
...
+        return platforms;
+    }, []);
+
+    if (constructionPlatforms.length > 0) {
+        return (
+            <div data-test="constructionplatform">
+                <dt>Library construction platform</dt>
+                <dd>
+                    {constructionPlatforms.map((platform, i) =>
+                        <React.Fragment key={platform['@id']}>
```
zoldello: Most likely I am wrong; but that has never deterred me from asking... My understanding of React fragments is that it basically tell react now to complain that there no container-tag. That is why you can use <></>. If that is true, why is there a key in the fragment?

src/encoded/static/components/experiment.js
```diff
@@ -302,6 +302,46 @@ LibrarySubmitterComments.propTypes = {
...
+                            {i > 0 ? <span>, </span> : null}
```
zoldello: This reminds me of life under Ecmascript 5. Map creates an array. Would using .join() on the resultant array be sufficient in creating the comma separated list and clearer?

src/encoded/static/components/experiment.js
```diff
@@ -302,6 +302,46 @@ LibrarySubmitterComments.propTypes = {
...
+ * Display, as one element of a "key-value" definition list, all the `construction_platform`
+ * properties in a dataset's replicates' library.
+ */
+const DatasetConstructionPlatform = ({ context }) => {
+    // Collect the library construction platforms from all the dataset replicates.
+    const constructionPlatforms = context.replicates.reduce((platforms, replicate) => {
+        // We easily have duplicates, so only add the replicate library platform if we haven't
+        // already seen it.
+        if ((replicate.library && replicate.library.construction_platform) &&
+            (platforms.findIndex(platform => platform['@id'] === replicate.library.construction_platform['@id']) === -1)) {
```
zoldello: Is it a good idea to put the "platforms.findIndex(platform => platform['@id'] === replicate.library.construction_platform['@id'])" onto its own line. It is the last parenthesis can easily be missed and it easy to think you are doing something like x===y===-1.

src/encoded/static/components/experiment.js
forresttanaka: You need the key here also, or else you get (and I just tried it removing that key to be sure):

image

src/encoded/static/components/experiment.js
forresttanaka: You cannot join JSX elements. You can only join primitive elements. Other people have other ways to do this, also not involving join, and all of them very difficult to read IMO.

PR #3177, opened 2020-02-28
VALIS-2 Allow additional file properties in track headers

forresttanaka:
This is the ENCODE side of VALIS-2 which temporarily refers to the VALIS-2 branch in package.json, and which has a small CSS change as well as the encoded-side memory fixes that relied on VALIS-1. This also has a fix to the graphlib crash (yet again) in package.json.

In this branch, this currently only affects the file gallery “Genome browser” tab display, where the track headers contain the existing content, but with the newer style including the disclosure arrow being at the bottom of the track header instead of to the left.

The demo for this broke after rebasing to dev, so I’ve just started a new demo.


forresttanaka:
Demo, but just started indexing at 2020-02-28 at noon:
~https://valis-2-file-props-pr1-fytanaka.demo.encodedcc.org/~

emro (approved):
I was confused about the demo and why the browser labels didn't look the same as the cart visualization branch browser labels, but I now note that it's because this branch makes no use of the new classes `gb-accession` and `gb-info`. It might be helpful for future users to document somewhere how to use the new styles and how the new long name property works overall.

I also think that I noted, looking at these demos, that the track file height is taller than the signal track height, and I'm not sure why that is, but I think it might make sense to keep those shorter (except in the case of track files with two tracks). However, that is a QA-type comment and this code all looks good, so this is approved!

forresttanaka:
All the changes in this branch are also in ENCD-4540 which Ben merged yesterday, so closing this PR.

PR #3178, opened 2020-02-29
ENCD-5137-Fix-ENTEx-text-issue

zoldello:


forresttanaka (approved):
not-bdd-non-indexing fails both on CircleCI and locally on this branch as well as on dev, a problem I expect was introduced on Friday, as early CircleCI tests that day passed completely. This branch is good to go.

PR #3182, opened 2020-03-04
SNO-144-redirect-downloads-through-nginx-or-apache

lrowe:
This is the encoded change to accompany encode-dcc/snovault#192.

### DIFF
--- a/ini-templates/production-template.ini
+++ b/ini-templates/production-template.ini
@@ -5,6 +5,7 @@ accession_factory = ${accession_factory}
 blob_bucket = ${blob_bucket}
 blob_store_profile_name = ${file_upload_profile_name}
 file_upload_bucket = ${file_upload_bucket}
+accel_redirect_header = ${accel_redirect_header}
 
 indexer.chunk_size = 1024
 indexer.processes = 16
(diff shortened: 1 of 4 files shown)


ini-templates/production-template.ini
```diff
@@ -5,6 +5,7 @@ accession_factory = ${accession_factory}
 blob_bucket = ${blob_bucket}
 blob_store_profile_name = ${file_upload_profile_name}
 file_upload_bucket = ${file_upload_bucket}
+accel_redirect_header = ${accel_redirect_header}
```
keenangraham: Probably want this in base.ini so it applies to demos as well. @caseylitton can confirm.

keenangraham (changes requested):
We'll have to point this to the sno branch so tests run and we can spin up demo.

keenangraham:
Putting param in base.ini and pointing to sno branch.

keenangraham:
Rebased.

ini-templates/production-template.ini
lrowe: I thought demo instances ran production.ini? Removing the necessity to run nginx in development seems useful.

keenangraham: Yeah, I think you are right. Concerned the `if proxy and accel_redirect_header` logic would change current behavior (local and demo) if the `accel_redirect_header` param wasn't available to both INIs.

keenangraham: If that makes sense to you should we make symmetric change to snovault PR?

keenangraham (approved):

PR #3183, opened 2020-03-04
ENCD-5145-remove-gRNA-tiled-loci-update-examined-loci

iyoungworth:


src/encoded/schemas/functional_characterization_experiment.json
```diff
@@ -236,36 +236,25 @@
...
-                "description": "Gene locus targeted by gRNAs.",
-                "type": "string",
-                "linkTo": "Gene"
-            }
-        },
         "examined_loci": {
             "title": "Examined loci",
             "description": "The loci whose expression or product is under investigation for the functional characterization experiment.",
             "type": "array",
             "uniqueItems": true,
```
gabdank, line 243: I think it should be minItem=1. Otherwise, this could be an empty array - that is not useful and confusing.

gabdank (changes requested):

src/encoded/schemas/functional_characterization_experiment.json
iyoungworth: I added minItems: 1 back in.

gabdank (approved):

src/encoded/schemas/functional_characterization_experiment.json
```diff
@@ -236,36 +236,26 @@
...
-            "minItems": 2,
+            "minItems": 1,
             "items": {
                 "title": "Locus",
                 "description": "The gene whose expression or product is under investigation for the functional characterization experiment.",
                 "type": "object",
                 "additionalProperties": false,
                 "anyOf": [
```
gabdank: probably should be oneOf and not anyOf

gabdank (changes requested):

src/encoded/schemas/functional_characterization_experiment.json
iyoungworth: Addressed; see JIRA ticket for detailed comment. Needed to add another restriction to the gene-only case

gabdank (dismissed):

src/encoded/schemas/functional_characterization_experiment.json
```diff
@@ -236,67 +236,60 @@
...
+                        "maxProperties": 1
+                    },
+                    {
+                        "required": ["gene", "expression_percentile"],
+                        "maxProperties": 2
+                    },
+                    {
+                        "required": ["gene", "expression_range_minimum", "expression_range_maximum"],
+                        "maxProperties": 3
+                    }
```
gabdank: will that work if in the future we add another property (that is not strictly required for each locus) - will that mess up the maxProperties?

iyoungworth: I think that probably would break it, yes. I think part of the issue with the three cases in oneOf is that "gene" may be specified by itself. Once we add survival properties, I think we would no longer allow "gene" to be submitted alone. This should mean the min/max requirements are no longer needed either (as when Bonita originally wrote this section)

gabdank: I think we need to change the dependency. Let's talk about some other way to specify the logic in this dependency?

gabdank (approved):

PR #3189, opened 2020-03-07
SNO-145-handle-missing-obj-on-request

caseylitton:
Going in v97rc2 but not being merged yet.


hitz:
This is irelevant as written, will point to snovault 1.0.42

PR #3191, opened 2020-03-07
ENCD- 5125-add-factorbook

meenakshikagda:


src/encoded/schemas/namespaces.json
```diff
@@ -6,6 +6,7 @@
     "GEO": "http://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=",
     "Caltech": "http://jumpgate.caltech.edu/library/",
     "FlyBase": "http://flybase.org/cgi-bin/quicksearch_solr.cgi?caller=quicksearch&tab=basic_tab&data_class=FBgn&species=Dmel&search_type=all&context=",
+    "FactorBook": "http://factorbook.beta.wenglab.org/experiment/",
     "WormBase": "http://www.wormbase.org/species/c_elegans/gene/",
```
gabdank: would like to double-check with DAC is the URL not going to change to "non-beta" soon

src/encoded/static/components/dbxref.js
```diff
@@ -107,6 +107,9 @@ export const dbxrefPrefixMap = {
     Cellosaurus: {
         pattern: 'https://web.expasy.org/cellosaurus/{0}',
     },
+    FactorBook: {
+        pattern: 'http://factorbook.beta.wenglab.org/experiment/{0}',
+    },
```
gabdank: would like to double-check with DAC is the URL not going to change to "non-beta" soon

gabdank (changes requested):
not approved until we hear back from DAC

src/encoded/static/components/dbxref.js
hitz: this page is SUPER slow

gabdank (approved):

PR #3192, opened 2020-03-09
ENCD-5123-expt-audit-processed-files-analyses

bonitalam:


src/encoded/audit/experiment.py
```diff
@@ -4018,6 +4018,24 @@ def audit_experiment_no_processed_data(value, system, files_structure):
...
+                    return
+                if file_object not in analysis:
+                    detail = ('Experiment {} '
+                        'contains a processed file {} '
+                        'that is not in an analysis'.format(
+                            audit_link(path_to_text(value['@id']), value['@id']),
+                            audit_link(path_to_text(file_object['@id']), file_object['@id']),
+                        )
+                    )
+                    yield AuditFailure('inconsistent analyses files', detail, level='INTERNAL_ACTION')
```
gabdank: I think this code is incorrect. Should have been something like:
`
for analysis in value['analyses']:
...
`
and I think that it should not be file_object in analysis, but file_object['@id'] in analysis

gabdank (changes requested):

gabdank (changes requested):
please add test for a situation where the audit is not triggered because all the processed files are in analyses.

src/encoded/audit/experiment.py
```diff
@@ -4018,6 +4018,45 @@ def audit_experiment_no_processed_data(value, system, files_structure):
...
+                            audit_link(path_to_text(value['@id']), value['@id']),
+                            ', '.join(files_not_in_analysis_links)
+                        )
+                    )
+                yield AuditFailure('inconsistent analyses files', detail, level='INTERNAL_ACTION')
+        files_not_in_processed_data = []
+        for analysis_file in analysis_outputs:
+            if analysis_file in processed_data:
+                continue
+            if analysis_file not in processed_data:
```
gabdank: @bonitalam please update the variable names. processed_data is a dictionary, so every time you do something like "for processed_file in processed_data" or "if analysis_file in processed_data" - you actually iterating over or checking for the existence of a file_id in processed_data dictionary keys. I propose to make it more clear by renaming in the proper places "processed_file" into "processed_file_id" and similarly "analysis_file" into "analysis_file_id", I would also make it clear that you are traversing keys of a dictionary by using "processed_data.keys()"

gabdank (changes requested):

bonitalam:
I had to move out the links and audit detail lines, so that only one audit should be raised (listing all the files) rather than multiple.

gabdank (approved):

gabdank (dismissed):

gabdank (approved):

PR #3198, opened 2020-03-10
ENCD-2666 mouse dev matrix

emro:
Adds a mouse development matrix with the following key features:
- Buttons for filtering by mouse stage/age
- Links to other pages (search, report, summary) which respect selected filters
- Row/column links which respect selected filters
- Link to matrix on navigation menu
- Two-column format on wide screens


src/encoded/static/components/app.js
```diff
@@ -56,7 +56,7 @@ const portal = {
                 { id: 'sescc', title: 'SE Stem Cell Consortium', url: '/matrix/?type=Experiment&status=released&internal_tags=SESCC', tag: 'collection' },
                 { id: 'reference-epigenomes-human', title: 'Human reference epigenomes', url: '/reference-epigenome-matrix/?type=Experiment&related_series.@type=ReferenceEpigenome&replicates.library.biosample.donor.organism.scientific_name=Homo+sapiens', tag: 'collection' },
                 { id: 'reference-epigenomes-mouse', title: 'Mouse reference epigenomes', url: '/reference-epigenome-matrix/?type=Experiment&related_series.@type=ReferenceEpigenome&replicates.library.biosample.donor.organism.scientific_name=Mus+musculus', tag: 'collection' },
-                { id: 'mouse-dev-series', title: 'Mouse development series', url: '/search/?type=OrganismDevelopmentSeries&internal_tags=MouseDevSeries&status=released', tag: 'collection' },
+                { id: 'mouse_development_matrix', title: 'Mouse development matrix', url: '/mouse-development-matrix/?type=Experiment&status=released&related_series.@type=OrganismDevelopmentSeries&replicates.library.biosample.organism.scientific_name=Mus+musculus', tag: 'collection' },
```
forresttanaka: Doesn’t _actually_ matter, but all other multi-word ids use hyphens.

src/encoded/static/components/matrix_mouse_development.js
```diff
@@ -0,0 +1,1165 @@
...
+ * @param {number} colCount Number of columns in matrix
+ *
+ * @return {object} Summary information about given matrix data:
+ *     {
+ *         {array} subCategorySums: Column sums for all subcategory rows
+ *         {number} maxSubCategoryValue: Maximum value in all cells in all subcategory rows
+ *         {number} minSubCategoryValue: Minimum value in all cells in all subcategory rows
+ *     }
+ */
+const analyzeSubCategoryData = (subCategoryData, columnCategory, columnSubcategory, colMap, colCount, stageFilter) => {
```
forresttanaka: Looks like you can leave out the `columnSubcategory` parameter; you’re not using it.

src/encoded/static/components/matrix_mouse_development.js
```diff
@@ -0,0 +1,1165 @@
...
+};
+
+/**
+ * Given one subcategory of matrix data (all the subcategory rows within a category), collect
+ * summary information about that data.
+ * @param {array}  subCategoryData Array of subcategory objects, each containing an array of data
+ * @param {string} columnCategoryType `subCategoryData` property that contains data array
+ * @param {array}  colMap Maps column titles to the column indices they correspond to
+ * @param {number} colCount Number of columns in matrix
+ *
```
forresttanaka: The parameter comments are for a different `analyzeSubCategoryData`.

src/encoded/static/components/matrix_mouse_development.js
```diff
@@ -0,0 +1,1165 @@
...
+    subCategoryData.forEach((rowData) => {
+        // `rowData` has all the data for one row. Collect sums of all data for each column.
+        rowData[columnCategory].buckets.forEach((value) => {
+            if (stageFilter) {
+                stageFilter.forEach((singleFilter) => {
+                    const filterString = singleFilter.replace(/[()]/g, '');
+                    const keyString = rowData.key.replace(/[()]/g, '');
+                    if (!filterString || keyString.includes(filterString)) {
+                        let colIndex = 0;
+                        if (value && value['target.label'] && value['target.label'].buckets.length > 1) {
```
forresttanaka: I think you can use `COL_SUBCATEGORY` everywhere you’re using `'target.label'`, in the highly unlikely chance you eventually use something other than `'target.label'`, though that has happened before.

src/encoded/static/components/matrix_mouse_development.js
```diff
@@ -0,0 +1,1165 @@
...
+ * @param {func}   getRowSubCategories Returns subCategory desired display order
+ * @param {func}   mapRowCategoryQueries Callback to map row category query values
+ * @param {func}   mapSubCategoryQueries Callback to map subcategory query values
+ * @param {array}  expandedRowCategories Names of rowCategories the user has expanded
+ * @param {func}   expanderClickHandler Called when the user expands/collapses a row category
+ *
+ * @return {object} Generated object suitable for passing to <DataTable>
+ */
+
+const convertExperimentToDataTable = (context, getRowCategories, getRowSubCategories, mapRowCategoryQueries, mapSubCategoryQueries, expandedRowCategories, expanderClickHandler, clearClassifications, stageFilter) => {
```
forresttanaka: Looks like `getRowSubCategories` and `mapSubCategoryQueries` aren’t used in this function, and `clearClassifications` and `stageFilter` need to be added to the @param comments.

src/encoded/static/components/matrix_mouse_development.js
```diff
@@ -0,0 +1,1165 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
```
forresttanaka: ESLint marking this as unused.

src/encoded/static/components/matrix_mouse_development.js
```diff
@@ -0,0 +1,1165 @@
...
+                        let colIndex = 0;
+                        if (value && value['target.label'] && value['target.label'].buckets.length > 1) {
+                            value['target.label'].buckets.forEach((bucket) => {
+                                const newKey = `${value.key}|${bucket.key}`;
+                                if (bucket.key && colMap[newKey]) {
+                                    colIndex = colMap[newKey].col || 0;
+                                    subCategorySums[colIndex] = (subCategorySums[colIndex] || 0) + 1;
+                                } else {
+                                    colIndex = colMap[value.key].col || 0;
+                                    subCategorySums[colIndex] = (subCategorySums[colIndex] || 0) + 1;
```
forresttanaka: I think this line and the same one in the `if` part can be put below the if? Since they’re executed in both cases.

src/encoded/static/components/matrix_mouse_development.js
```diff
@@ -0,0 +1,1165 @@
...
+                    }
+                });
+            } else {
+                let colIndex = 0;
+                if (value && value['target.label'] && value['target.label'].buckets.length > 1) {
+                    value['target.label'].buckets.forEach((bucket) => {
+                        const newKey = `${value.key}|${bucket.key}`;
+                        if (bucket.key && colMap[newKey]) {
+                            colIndex = colMap[newKey].col || 0;
+                            subCategorySums[colIndex] = (subCategorySums[colIndex] || 0) + 1;
```
forresttanaka: Same here. Makes nearly no difference in execution, but if I maintained this code later I would have changed it to that.

src/encoded/static/components/matrix_mouse_development.js
```diff
@@ -0,0 +1,1165 @@
...
+                        } else {
+                            colIndex = colMap[value.key].col || 0;
+                            subCategorySums[colIndex] = (subCategorySums[colIndex] || 0) + 1;
+                        }
+                    });
+                } else {
+                    colIndex = colMap[value.key].col || 0;
+                    subCategorySums[colIndex] = (subCategorySums[colIndex] || 0) + 1;
+                }
+            }
```
forresttanaka, line 114: Looks like this entire if/else block is identical to the one inside the forEach() above. Could this block be thrown into a utility function and return `subCategorySums`? Looks like a utility function would only need `value` as an input parameter, and `subCategorySums` as a mutated parameter.

src/encoded/static/components/matrix_mouse_development.js
```diff
@@ -0,0 +1,1165 @@
...
+    const sortedColCategoryBuckets = _(colCategoryBuckets).sortBy((colCategoryBucket) => {
+        const sortIndex = matrixAssaySortOrder.indexOf(colCategoryBucket.key);
+        return sortIndex >= 0 ? sortIndex : colCategoryBuckets.length;
+    });
+
+    // Generate the column map based on the sorted category buckets.
+    sortedColCategoryBuckets.forEach((colCategoryBucket) => {
+        if (!excludedAssays.includes(colCategoryBucket.key)) {
+            const colSubcategoryBuckets = colCategoryBucket[colSubcategory].buckets;
+            if (colSubcategoryBuckets.length <= 0) {
```
forresttanaka: “=== 0”? Since this is an array length.

src/encoded/static/components/matrix_mouse_development.js
```diff
@@ -0,0 +1,1165 @@
...
+        <div className="matrix-header">
+            <div className="matrix-header__title">
+                <h1>{context.title}</h1>
+                <div className="matrix-tags">
+                    <MatrixInternalTags context={context} />
+                </div>
+            </div>
+        </div>
+    );
+};
```
forresttanaka: ESLint error on this function. It wants an implied return.

src/encoded/static/components/matrix_mouse_development.js
```diff
@@ -0,0 +1,1165 @@
...
+            this.handleScrollIndicator(this.scrollElement);
+            this.setState({ expandedRowCategories: [] });
+        }
+    }
+
+    updateWindowWidth() {
+        this.setState({
+          windowWidth: window.innerWidth,
+        });
+    }
```
forresttanaka: “updateWindowWidth should be placed after getClearClassificationsLink” ESLint error. I have no idea how ESLint wants methods ordered, but it wants some order.

src/encoded/static/components/matrix_mouse_development.js
```diff
@@ -0,0 +1,1165 @@
...
+        }
+
+        // Convert matrix data to a DataTable object.
+        const { dataTable, rowKeys } = convertExperimentToDataTable(context, rowCategoryGetter, rowSubCategoryGetter, mapRowCategoryQueries, mapSubCategoryQueries, this.state.expandedRowCategories, this.expanderClickHandler, clearClassifications, stageFilter);
+        // If we have a wide window, split the table in two
+        let matrixConfig;
+        let matrixConfig2;
+        if (this.state.windowWidth > 1160) {
+            // Determining best index for splitting the matrix
+            const spacerRowIndices = dataTable.map((d, i) => d.css === "matrix__row-spacer" ? i : '').filter(String);
```
forresttanaka: Need parentheses around the conditional expression inside the .map() to avoid an ESLint error.

src/encoded/static/components/matrix_mouse_development.js
```diff
@@ -0,0 +1,1165 @@
...
+        const postnatalWidth = (mouseAgeObject.postnatal.length * (defaultWidth + 2)) - 2;
+        const adultWidth = (mouseAgeObject.adult.length * (defaultWidth + 2)) - 2;
+
+        // check if embryo width is greater than window width and if so, split in two
+        let embryoSplitRows = false;
+        let embryoFirstRowWidth = 0;
+        let embryoSecondRowWidth = 0;
+        let countEmbryoFirstRow;
+        let countEmbryoSecondRow;
+        let embryoFirstRow
```
forresttanaka: Semicolon to avoid the ESLint error.

src/encoded/static/components/matrix_mouse_development.js
```diff
@@ -0,0 +1,1165 @@
...
+
+    // Generate array of names of assays that have targets and don't collapse their targets, for
+    // rendering those columns as disabled.
+    const colCategoriesWithSubcategories = Object.keys(colMap).filter(colCategoryName => colMap[colCategoryName].hasSubcategories && !collapsedAssays.includes(colCategoryName));
+
+    const { rowCategoryData, rowCategoryNames } = getRowCategories();
+    const rowKeys = ['column-categories'];
+
+    // loop through row data to combine gendered mouse stage/ages into single row
+    const combinedRowCategoryData = JSON.parse(JSON.stringify(rowCategoryData));
```
forresttanaka, line 258: I’ve avoided using JSON.parse(JSON.stringify()) since they can make a lossy clone (though probably not in your specific case), and in nearly every benchmark this method comes out slowest. There’s a `clone()` function [here](https://jsperf.com/cloning-an-object/79) here that might work better, and certainly faster, though I’ve not tested it. I’ve seen other JS benchmarks that have other clone functions.

But I also wonder if you really need to clone rowCategoryData here? You’re doing little mutation here, and the bit you do looks harmless? I’m not sure; it’s hard to tell.

src/encoded/static/components/matrix_mouse_development.js
```diff
@@ -0,0 +1,1165 @@
...
+                        success = bucket;
+                    }
+                });
+                return success;
+            });
+        } else {
+            filteredRowSubcategoryBuckets = visibleRowSubcategoryBuckets;
+        }
+
+        const categoryNameQuery = encoding.encodedURIComponentOLD(rowCategoryBucket.key);
```
forresttanaka: Does `encodedURIComponentOLD` do something `encodedURIComponent` doesn‘t? I’m trying to eventually delete `encodedURIComponentOLD`, and the new one needs to be fixed if it can’t do something the old one did (aside from  the “space” option that we never used).

src/encoded/static/components/matrix_mouse_development.js
```diff
@@ -0,0 +1,1165 @@
...
+        this.selectMouseDevelopmentStage = this.selectMouseDevelopmentStage.bind(this);
+        this.hasRequestedClassifications = filteredClassifications.length > 0;
+        this.clearFilters = this.clearFilters.bind(this);
+        this.updateWindowWidth = this.updateWindowWidth.bind(this);
+    }
+
+    componentDidMount() {
+        this.handleScrollIndicator(this.scrollElement);
+        this.updateWindowWidth();
+        window.addEventListener("resize", this.updateWindowWidth);
```
forresttanaka: Single-quote string to avoid ESLint error.

forresttanaka (changes requested):
This looks like way more work than any other matrix, so that’s really impressive. I think the only changes I’d _insist_ on are the ESLint errors and `encodedURIComponentOLD`, just so we can easily get rid of that function through attrition.

src/encoded/static/components/matrix_mouse_development.js
emro: When I set up my new laptop I had to reinstall Atom and forgot to install the linting that I'd had on my previous setup, so I missed a bunch of these. I've reconfigured Atom now, though, and fixed all these errors.

src/encoded/static/components/matrix_mouse_development.js
emro: I think I really do need to clone rowCategoryData. This code was very tricky actually. The problem is that I'm using the original rowCategoryData in order to generate the new rowCategoryData and if you don't deep clone the original, you end up with infinite rows and the browser crashes, and it's no good. I agree that this generally isn't optimal but couldn't see a much better way to do this. I think the cloning is kind of tricky because the object has a lot of depth. I looked at the function you suggested and am not sure if it will help, I'll look again on Monday.

src/encoded/static/components/matrix_mouse_development.js
emro: For some reason, this does not work, and I'm not sure why not, because I think it should.

forresttanaka (dismissed):
Very cool.

src/encoded/static/components/app.js
hitz: I think we should fix this especially since it needs a rebase anyway.

hitz (changes requested):
fix L59 in app.js and rebase.

forresttanaka (approved):

hitz (approved):

PR #3201, opened 2020-03-12
ENCD-5082 File table pagination

forresttanaka:
* The main addition in this branch is a new component, `FileTablePaged`, that shows a non-sortable table of files which. If more than 50 files exist in the table, a pager appears which lets the user page between the pages of files. The table displays either an array of file objects or an array of file `@id`s — `FileTablePaged` does a fetch of the file objects displayed on the current page in the latter case. The most local 10 pages get cached in memory, kicking out farther pages from the current if the cache overflows.
* The change to `DerivedFiles` in file.js is because it contained all the functionality I needed for this ticket, though not generically enough. `DerivedFiles` now calls the new `FileTablePaged`.


zoldello (approved):
Code-wise, things look fine. I have no comments worth mentioning. I like the use of hooks. I glanced at the UI to make sure I am not missing something and it too looked fine.

PR #3202, opened 2020-03-13
Examined loci alternative dependencies

iyoungworth:
Please see commits 4a73e5e, 2383234, 0a61cb0, 5813de3, and 6726a6c for a few different tweaks to the dependencies in my attempts to understand the validation behavior better.


iyoungworth:
**decreased-nesting comment**: Just in case, I removed the "properties": section from the "not" designation, but it doesn't appear to make a difference, I think it's failing the tests the same as commit "**explicit-property-dependency-nesting-CORRECTED**" did.

PR #3203, opened 2020-03-13
ENCD-5105-SECC-Stem-cell-matrix

zoldello:


src/encoded/static/components/matrix_sescc_stem_cell.js
```diff
@@ -0,0 +1,676 @@
+import React from 'react';
+import { Panel, PanelBody } from '../libs/ui/panel';
+import PropTypes from 'prop-types';
+import { svgIcon } from '../libs/svg-icons';
+import DataTable from './datatable';
+import _ from 'underscore';
```
forresttanaka: For both underscore and prop-types I’m getting an ESLint error “Absolute imports should come before relative imports.eslintimport/first” so they both have to move up above Panel and PanelBOdy.

src/encoded/static/components/matrix_sescc_stem_cell.js
```diff
@@ -0,0 +1,676 @@
...
+        'RRBS',
+        'MeDIP-seq',
+        'MRE-seq',
+        'Repli-chip',
+        'DNAme array',
+        'genotyping array',
+        'RAMPAGE',
+        'TF ChIP-seq',
+        'Histone ChIP-seq',
+    ];
```
forresttanaka: Could move `headerDataOrder` initialization to the root level so it doesn’t get initialized on every matrix render, and `rowDataOrder` below.

src/encoded/static/components/matrix_sescc_stem_cell.js
```diff
@@ -0,0 +1,676 @@
...
+                    // Update the links.
+                    link.enter().insert('path', 'g')
+                        .attr('class', 'sescc_matrix__link')
+                        .style('stroke', d => d.target.linkColor)
+                        .attr('d', diagonal);
+                };
+                update(treeData[0]);
+            });
+        }
+    }
```
forresttanaka, line 528: Oh wow; I thought this was a static SVG. I didn’t know you were rendering it in D3. Do you think this is better than using a static image? I just worry this is a lot more difficult to maintain than having an SVG you can modify in Illustrator. I don’t _think_ the shape of this SVG changes except for CSS classes.

forresttanaka (changes requested):
The only required change is the ordering of the imports. The mechanisms look good.

src/encoded/static/components/matrix_sescc_stem_cell.js
zoldello: Thanks. Fixed.

src/encoded/static/components/matrix_sescc_stem_cell.js
zoldello: Fixed. I thought they were not large and specific to the method, so I did not put them class-wide, initially. Now, I more them class-wide

src/encoded/static/components/matrix_sescc_stem_cell.js
zoldello: Are you okay with me leaving it as-is? The issue is that the SVG code can get messy wiring up the events and handing the styles if embedded; and can be a sore to look at if inline. Plus, the SVG has some dynamic changes like the height and width is adjusted at runtime if the screen changes. If you feel strongly against D3.js, I will look into it.

forresttanaka: No, I would recommend keeping it this way; I was just surprised. And maybe others would prefer the way you did it. An embedded or file-based SVG should be able to handle height and width changes on window-size changes though, generally at least. I don’t know if you have a complicated case.

forresttanaka (approved):

PR #3205, opened 2020-03-13
ENCD-4540 Cart visualizations

forresttanaka:
* Not only did I move `filterForVisualizableFiles` so it could be shared, I also refactored it so that someone could check an individual file for visualizability, which is useful for the “Visualizable” indicators on the facets. I moved `computeAssemblyAnnotationValue` for the same reason.
* Added a new `mode` for search results, and I pass it through to the components that actually render search results. Previously `mode` had only been used for forms with the “picker” value. Now I added a “cart-view” mode so we can show a truncated display of experiment search results specifically for carts.
* I’m guessing I had `addFileTermToFacet` as a static method of `FileFacets` because `FileFacets` used it  exclusively at some point. Now `addFileTermToFacet` only gets called from other components, so I made it a top-level function. Many other methods also became top-level functions to avoid the confusion of mixing use of state variables and parameters
* A lot of functionality has moved from `FileFacets` to the page-level `CartComponent` so that it can keep track of states and props among all facets.
* I removed the `retrieveFileFacetTemplate` method that had done a limit=0 file search to start building a realistic-looking dummy file-search facets object. Now no dummy file-search object exists — I instead use the files within searched datasets to build a simplified facet object. This eliminates a bit of network traffic and a lot of complication, but as I also now have facets with data from files’ datasets instead of the files themselves, `retrieveFileFacetTemplate` couldn’t to do a complete job any more.
* This branch is our first use of the `useReducer` hook. I’m using it just like `useState` but with multiple independent values in an object instead of a single value like other states, and a reducer function to manipulate those values within the object.
* The datacolors.js change adjusts the tint level for the black/white text transition point for background colors across the site, notably on the experiment matrix. I found the existing level made the new facet term count bubbles difficult to read at the existing transition point.
* Added properties to `dummyFiles` to allow the display of complete cart facets locally.
* I was finding the graphlib package crash happening sometimes again because npm simply doesn’t install the graphlib package at all sometimes. I found the change to package.json fixes it. I don’t know why it still happened sometimes. I had two demos, one which had the crash and the other didn’t — both demos were _identical_.
* Ben said I should remove the Beta tag from the Genome Browser tab. Related, I modified TabPanel to accept React components as tab labels. Not sure if this is flexible enough long term, but it’s enough for the tab counts for now.


forresttanaka:
~Current demo~ updated from earlier versions.

src/encoded/vis_defines.py
```diff
@@ -1631,6 +1631,33 @@ def visualizable_assemblies(
...
+            'pepMap',
+            'modPepMap',
+        ],
+        file.get('status') in [
+            'released',
+            'in progress',
+            'archived',
+        ],
+    ]
+    return all(conditions)
```
emro, line 1660: This looks good to me and seems pretty straightforward, but should someone else review the Python?

src/encoded/batch_download.py
```diff
@@ -555,6 +577,12 @@ def batch_download(context, request):
 
 
 def files_prop_param_list(exp_file, param_list):
+    """Does a file in experiment search results match query-string parms?
+
+    Keyword arguments:
+    exp_file -- file object from experiment search results
+    param_list -- grouped query-string parameters for experiment search
+    """
```
emro, line 585: This all looks good to me, although I'm not sure I'm the most reliable Python code reviewer.

src/encoded/static/components/genome_browser.js
```diff
@@ -349,15 +453,15 @@ class GenomeBrowser extends React.Component {
         const tracks = files.map((file) => {
             if (file.name) {
                 const trackObj = {};
-                trackObj.name = file.name;
+                trackObj.name = <i>{file.name}</i>;
```
emro, line 457: I'm trying to remember why this case exists. When do we get here?

src/encoded/static/components/genome_browser.js
```diff
@@ -92,6 +143,40 @@ function mapGenome(inputAssembly) {
...
+                    <React.Fragment>
+                        <li>{file.output_type}</li>
+                        {(biologicalReplicates !== '') ? <li>{`rep ${biologicalReplicates}`}</li> : null}
+                    </React.Fragment>
+                : null}
+                {file.assay_term_name ? <li>{file.assay_term_name}</li> : null}
+                {file.biosample_ontology && file.biosample_ontology.term_name ? <li>{file.biosample_ontology.term_name}</li> : null}
+                {file.target ? <li>{file.target.label}</li> : null}
+            </ul>
+        </React.Fragment>
```
emro, line 166: Probably this was already discussed and there's a reason for having it this way, but I would swap the order of the elements in the `gb-info` list, so that the `file.output_type` and biological replicates were at the end. I don't really like how the order of the list changes when you click on the expander (I don't like things moving around if it's avoidable, in general), and I think it might be more intuitive to have them at the end because then it's like you've expanded the list to show the new information.

src/encoded/static/components/genome_browser.js
```diff
@@ -378,8 +482,7 @@ class GenomeBrowser extends React.Component {
                 return trackObj;
             }
             const trackObj = {};
-            const biologicalReplicates = file.biological_replicates.join(',');
-            trackObj.name = `${file.accession} ${file.output_type} ${(biologicalReplicates !== '') ? `rep ${biologicalReplicates}` : ''}`;
+            trackObj.name = <TrackLabel file={file} />;
```
emro, line 486: Do we not want the annotation tracks to be expandable?

src/encoded/static/scss/encoded/modules/_cart.scss
```diff
@@ -328,33 +321,67 @@
...
+$cart-facet-check-box-size: 18px;
+%facet-term-control {
+    display: flex;
+    flex: 0 0 $cart-facet-check-box-size;
+    width: $cart-facet-check-box-size;
+    height: $cart-facet-check-box-size;
+    border: 1px solid #c0c0c0;
+    justify-content: center;
+    align-items: center;
+}
```
emro: Do you have this twice on purpose? These two definitions of `%facet-term-control` look identical.
Also, this is fancy CSS syntax! I'll have to remember placeholder selectors.

src/encoded/static/scss/encoded/modules/_cart.scss
```diff
@@ -705,6 +750,350 @@ $deleted-bg: #ffe0e0;
...
+    (rcc, #acccbe, #000000),
+    (sam, #8eb2b2, #000000),
+    (tagAlign, #b0ff8c, #000000),
+    (tar, #e8ce95, #000000),
+    (tsv, #ff9a99, #000000),
+    (vcf, #a08ce8, #ffffff),
+    (wig, #a3fff2, #000000),
+    (idx, #ffd6ba, #000000),
+    (txt, #e897dc, #000000),
+);
```
emro, line 844: I almost always put color look-ups in JS, but I think this works very well!

src/encoded/static/components/search.js
```diff
@@ -371,12 +379,14 @@ const ExperimentComponent = (props, reactContext) => {
...
     context: PropTypes.object.isRequired, // Experiment search results
     cartControls: PropTypes.bool, // True if displayed in active cart
+    mode: PropTypes.string, // Special search-result modes, e.g. "picker"
     auditIndicators: PropTypes.func.isRequired, // Audit decorator function
     auditDetail: PropTypes.func.isRequired,
 };
 
 ExperimentComponent.defaultProps = {
     cartControls: false,
+    mode: '',
```
emro, line 389: Do we need "mode" as a property or could we use `mode === 'cart-view'` instead?

src/encoded/static/components/cart/cart.js
```diff
@@ -979,256 +888,646 @@ PagerArea.propTypes = {
...
+                    datasetFiles: addToAccumulatingFiles(accumulatingResults.datasetFiles, currentResults),
+                    datasets: accumulatingResults.datasets.concat(currentResults['@graph'].map(experiment => experiment['@id'])),
+                };
+            })
+        ))
+    ), Promise.resolve({ datasetFiles: [], datasets: [] })).then(({ datasetFiles, datasets }) => {
+        facetProgressHandler(-1);
+        return { datasetFiles, datasets };
+    });
+};
```
emro, line 1254: I think this isn't useful, but `requestFiles` and `requestObjects` in objectutils.js do a similar operation: https://github.com/ENCODE-DCC/encoded/blob/dev/src/encoded/static/components/objectutils.js#L209. It doesn't handle the progress bar the way you might need, and it also handles filtering and duplicate @ids which seem not relevant here. But, it is chunking data and fetching it!

emro (commented):
It looks great! I don't have any big comments, just a few small ones, and some questions that I was curious about.

One thing I was trying to think about was if we could try to share more code between the filegallery.js facets and the cart.js facets, because they work in somewhat similar ways and it would be nice to reduce code / simplify these things. I couldn't think of a straightforward or easy way to do that, though, so I think it's great as is.

The other thing I was looking at was the backend changes. It looks like these changes maybe allow you to execute a search for visualizable files only so the filtering isn't just on the front end? This wasn't completely clear to me. But if that's true, that could be very helpful.

src/encoded/vis_defines.py
forresttanaka: This particular part is Keenan’s code actually. I put this up for a draft PR a while back and he reviewed the Python.

src/encoded/static/components/genome_browser.js
forresttanaka: I can’t figure that out either, but I know I’ve seen it or I wouldn’t have made it italic. I thought it was for the pinned files, but they clearly don’t use this case, at least not that I’ve seen so far, and “name” isn’t a file property. Would have been a good place for a comment I guess.

src/encoded/static/components/genome_browser.js
forresttanaka: I honestly don’t know. I was only aware of bigWigs needing that, but maybe that was a bad assumption. What do you think, @hitz?

src/encoded/static/components/search.js
forresttanaka: I’m not sure what you mean. "mode" now gets passed through to the type-specific search result rendering so that they can check for modes just like the existing top-level facet code has always done.

src/encoded/static/components/genome_browser.js
forresttanaka: I _think_ @hitz and I discussed the order, and we definitely discussed what should be in the non-expanded and expanded versions. It’s been so long I don’t remember clearly on the order though.

src/encoded/static/scss/encoded/modules/_cart.scss
forresttanaka: That looks like a total mistake.

src/encoded/static/components/cart/cart.js
forresttanaka: They are similar and I’m sure could be combined. But this needed just that little bit more and I didn’t want to touch functions used elsewhere yet.

src/encoded/static/components/genome_browser.js
hitz: Order should be: Biosample (Target) Assay.  And it should not flip around when you expand contract.

src/encoded/static/components/genome_browser.js
hitz: I might have forgotten about annotation tracks, but yes - they should be expandable.  Probably the default should be expandable and BigBed or similar "boring" tracks should have this option off.

src/encoded/static/components/search.js
emro: It just didn't seem to me like the "picker" mode was used for anything in this component, so maybe the only thing that was relevant was whether the mode was "cart-view". I could be wrong about this though.

forresttanaka: No that’s right. "picker" mode is only used at the level covering all search results so that they display differently in the object-editor form. But that’s where `mode` stopped so I now pass `mode` all the way down to the individual search result level, and that’s where 'cart-view' mode is used. So the edit forms can affect the entire search results using mode "picker," and the cart view can affect individual search results using mode "cart-view." The individual results don’t care about "picker" mode. I’m just overloading this `mode` property since their functionalities currently don’t overlap yet they have a similar role.

If we ever have entire search results displayed differently for cart view, or individual search results care about picker mode, then this mechanism won’t work anymore.

src/encoded/static/components/genome_browser.js
forresttanaka: I think we have to leave this as is. George has some special-case code specifically for annotations, which is why annotation files on production in the file gallery Valis view don’t have arrows on them. I figured out how to force an arrow onto annotation tracks on the encoded side, but I can’t figure out how to make the tracks as short as bigWig files to begin with, so bigBed files start tall, and you expand them to make them _really_ tall.

I think we leave this as is for now. Do you agree @emro and @hitz?

forresttanaka:
Updated demo with the repeated CSS template removed, and the track file info items reordered: ~https://encd-4540-cart-visualization-pr2-fytanaka.demo.encodedcc.org/~

src/encoded/static/components/genome_browser.js
emro: I forgot about this, I agree we should leave it as is for now.

src/encoded/static/components/genome_browser.js
emro: I wish I'd made a comment originally! Maybe we'll figure it out later.

src/encoded/static/components/genome_browser.js
forresttanaka: OK, thanks! I’ll move this back to in code review.

src/encoded/static/components/search.js
emro: Ok, this all makes sense. I guess I was suggesting that you just pass a flag saying if the mode was "cart-view" since you don't care if the mode is "picker" for this component. I think also I found the comment `mode: PropTypes.string, // Special search-result modes, e.g. "picker"` a little confusing because the example is not relevant for the component. This isn't very important though, just explaining why I mentioned it.

forresttanaka: What might be nice is a `mode` _object_, so you can have multiple flags at the same time. That crossed my mind, as did a new property for cart view, but I felt like mode could be more useful than just for one thing, but I didn’t know if making an object was worth how little we use it. Something to think about if we need to expand on this someday.

emro (dismissed):
Looks great!

hitz (approved):
after rebase

PR #3209, opened 2020-03-20
ENCD-5150 Fix rendering error when Type facet negated

forresttanaka:
* This involves a new property to the facet renderers — allowNegation — that’s to false currently only for the "type" facet so that we no longer allow the user to negate it with the icons to the right of each facet term.
* This fix _might_ be temporary. If search someday makes negated "type" facets work like the others, we can set allowNegation back to the default True for the "type" facet.

### DIFF
--- a/src/encoded/static/components/facets/defaults.js
+++ b/src/encoded/static/components/facets/defaults.js
@@ -569,7 +569,7 @@ DefaultTermName.propTypes = {
 /**
  * Default component to render a single term within the default facet.
  */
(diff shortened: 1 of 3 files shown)


forresttanaka:
Here’s the demo: https://encd-5150-rendering-error-pr1-fytanaka.demo.encodedcc.org/

zoldello (approved):
The changes look fine. There is a strong correlation to the new feature and the code change- Make negation optional and dataset should not allow negation.. There is nothing in the commit that makes my eyebrow rise.

PR #3211, opened 2020-03-20
ENCD-5163 deemphasize hg19 screen link

emro:
We want to de-emphasize the hg19 link to SCREEN.
We'll do this by turning it into a link rather than a button.


zoldello (approved):
The PR looks fine. It does what it need to do and removes what it needs to. Functionally, UX-wise, this PR prevents a user from searching hg19 on the homepage. Plus, it is a bit awkward to me to see buttons (technically button-like link) and under that a link. Would these changes be okay to people outside Weng's lab (they are not the only users)?

PR #3216, opened 2020-03-24
ENCD-5180-re-matrix-left-align-term-names

zoldello:


src/encoded/static/scss/encoded/modules/_matrix.scss
```diff
@@ -156,7 +156,6 @@ $row-data-header-width: 200px;
 }
 
 .subcategory-row-text {
```
emro: These styles are all specific to the mouse matrix, so I would probably change line 158 to be the new lines you added (174 - 176). The regular "subcategory-row-text" does not need the "label-stage" or "label-length" classes because those only correspond to the mouse matrix and `display:flex` so might as well just make it specific to the mouse matrix.

src/encoded/static/scss/encoded/modules/_matrix.scss
```diff
@@ -916,6 +920,7 @@ $divider-color: #000; // Color of dividers between assays with targets
             width: $donor-cell-width;
             padding: 0;
             border-left: none;
+            text-align: right;
```
emro, line 921: I can see that this line works and that the labels aren't aligned properly without it, but I'm confused because the matrix pages used to not require it, so I think I must have added some line that changes the default, but I can't figure out where. I just think this shouldn't be necessary, but it clearly is. Do you see what the change is that I made that necessitates this update?

emro (commented):
Overall, looks great, and this definitely fixes the problem! Just a question and one comment.

src/encoded/static/scss/encoded/modules/_matrix.scss
zoldello: Done.

src/encoded/static/scss/encoded/modules/_matrix.scss
zoldello: ~I am in the middle of something important so will update this answer later. It had to do with display:flex, overriding the style~ 

There was text-align: right; under .matrix__row-data > th in the code before (like in prod currently). I am not sure how or why it was removed, but can find if you feel it is important. That said, having it makes no difference if combined with display: flex, because display: flex alters the text alignment to still pulled to the right but in a way that made it not seem to not be so (like left align on small text but right align on long text string [something weird]). So, I moved display: flex to its own class, so text-align:right can be allowed to work without hindrance. I had to throw a little mud at the wall before I discerned this.

src/encoded/static/scss/encoded/modules/_matrix.scss
emro: Looks good! I probably would have also put `align-items: center` under the subcategory-row-text-mouse class because it's a property that only applies to flexbox display, and I also would have put `line-height: 1.5` under the mouse subcategory because otherwise we're over-writing the line-height on the reference epigenome matrix as well (it has a line-height of 1.15 on production). I was looking at the reference epigenome matrix more and I think there's another change needed there also, it seems like a mouse matrix change might have also eliminated some row height. Maybe lines 184 - 185 need to be moved so they aren't impacting the other matrices? I'm not sure if that's the fix needed, but hopefully.

src/encoded/static/scss/encoded/modules/_matrix.scss
emro: Thank you for looking into this! This seems good!

src/encoded/static/scss/encoded/modules/_matrix.scss
zoldello: I did the first part (moved line-height and align-item.) For the padding:0 (line 184-185), I moved it unto a class specific for mouse. Prior to the last commit (shah #: 84cd42204), it was causing a very small padding change in other matrices. I moved it so prevent that while not affect mouse matrix itself.

emro (approved):
Looks great!

PR #3217, opened 2020-03-25
ENCD-5182 respect filters for viz

emro:
We want browser links to apply user-selected filters, the way the other buttons/links apply user-selected filters.

### DIFF
--- a/src/encoded/static/components/vis_defines.js
+++ b/src/encoded/static/components/vis_defines.js
@@ -388,12 +388,21 @@ export class BrowserSelector extends React.Component {
             activeFilters.forEach((f) => {
                 if (['adult', 'postnatal', 'embryo'].includes(f)) {
                     const stageTerm = f === 'embryo' ? 'embryonic' : f;
-                    resultsId = `${resultsId}&replicates.library.biosample.life_stage=${stageTerm}`;
+                    const stageFilter = `&replicates.library.biosample.life_stage=${stageTerm}`;
+                    if (!resultsId.includes(stageFilter)) {
+                        resultsId = `${resultsId}${stageFilter}`;
+                    }
                 } else {
                     const stageTerm = f.split(' ')[0] === 'embryo' ? 'embryonic' : f.split(' ')[0];
                     const ageTerm = f.split(' ').slice(1).join(' ');
-                    resultsId = `${resultsId}&replicates.library.biosample.life_stage=${stageTerm}`;
-                    resultsId = `${resultsId}&replicates.library.biosample.age_display=${ageTerm}`;
+                    const stageFilter = `&replicates.library.biosample.life_stage=${stageTerm}`;
+                    if (!resultsId.includes(stageFilter)) {
+                        resultsId = `${resultsId}${stageFilter}`;
+                    }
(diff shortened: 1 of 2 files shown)


src/encoded/static/components/vis_defines.js
```diff
@@ -235,8 +235,8 @@ export const visMapBrowserName = browser => browserNameMap[browser];
  *
  * @return {string} hub URL used in visualization URLs
  */
-const generateBatchHubUrl = (results, hostName) => {
-    const parsedUrl = url.parse(results['@id']).search;
+const generateBatchHubUrl = (resultsId, hostName) => {
```
forresttanaka, line 238: Be nice to change the comment for `results` but let’s not worry about it since we need to get this into rc2.

src/encoded/static/components/vis_defines.js
```diff
@@ -382,7 +382,31 @@ export class BrowserSelector extends React.Component {
...
+                    const stageTerm = f.split(' ')[0] === 'embryo' ? 'embryonic' : f.split(' ')[0];
+                    const ageTerm = f.split(' ').slice(1).join(' ');
+                    const stageFilter = `&replicates.library.biosample.life_stage=${stageTerm}`;
+                    if (!resultsId.includes(stageFilter)) {
+                        resultsId = `${resultsId}${stageFilter}`;
+                    }
+                    const ageFilter = `&replicates.library.biosample.age_display=${ageTerm}`;
+                    if (!resultsId.includes(ageFilter)) {
+                        resultsId = `${resultsId}${ageFilter}`;
+                    }
```
forresttanaka, line 405: There’s one potentially outlandish case where I think this could fail, but I’m not sure if it can practically happen. If you go to https://encd-2666-f761e000e-emma.demo.encodedcc.org/mouse-development-matrix/?type=Experiment&status=released&related_series.@type=OrganismDevelopmentSeries&replicates.library.biosample.organism.scientific_name=Mus+musculus&replicates.library.biosample.life_stage=adult&replicates.library.biosample.age_display=8 weeks&replicates.library.biosample.life_stage=postnatal the server replaces the “8 weeks” with “8%20weeks” which arrives in this code in `resultsId`. If `ageTerm` gets the value “8 weeks” then this `includes` test fails, and `resultsId` gets _another_ copy of “&replicates.library.biosample.age_display=” — the existing one with “8%20weeks” and a new one with “8 weeks.” Again, I’m not sure if this is too unrealistic or not, as this was the only way I could trigger that. Might need to use encodeURIComponent which will replace the space with a plus, and hope that people use a plus in the URL. Ben sometimes doesn’t care what happens when people manually enter the URL, so this might not matter. I’ll leave this as a question.

forresttanaka:
One long comment, and I’m not sure it’s a real concern or not. If it is, then you might need to change a couple lines to fix it.

forresttanaka:
If it’s not a concern, I’ll just approve this. The rest looks good!

forresttanaka (approved):
Emma and I discussed this, and we feel there’s likely no problem — with the possible exception of sending a query string to UCSC with the same term repeated twice. If there is, we can address that later.

PR #3221, opened 2020-03-26
SNO-143-switch-update-readmes

caseylitton:


README.md
```diff
@@ -1,338 +1,58 @@
...
-  
-  
-- **JavaScript**: Make a syntax error, or style error, in a JS file. Warnings and errors should show on the left side of the line.  
-
+ENCODE Metadata Database
+========================
+
+
+## System Installation (OSX Catlina 10.15.2)
+See [Snovault OSX System Installation][].  Snovault is a dependecy of the Encoded application.
```
zoldello: type: "dependency" is incorrectly spelt "dependecy"

zoldello: Suggested wording: Encode installs Snovault as it is a dependency. However, you do not need to set up a running instance yourself.

README.md
```diff
@@ -1,338 +1,58 @@
...
+Installing Snovault is not a requirement to run the this application, however, the
+system installation is the same.
+
+
+## Application Installation
+For issues see [Snovault OSX App Installation][] first.
+
+1. Create a virtual env in your work directory.
+    Here we use python3 venv module.  Use venv, like conda, if you please
```
zoldello: Suggested word: This example uses venv. Other options would also work like conda

README.md
```diff
@@ -1,338 +1,58 @@
...
+    ```
+    $ cd your-work-dir
+    $ python3 -m venv encoded-venv
+    $ source encoded-venv/bin/activate
+    ```
+
+1. Clone the repo and install requirements
+    ```
+    $ cd encoded
+    # Make sure you are in the encoded-venv
```
zoldello, line 24: I think "# Make sure you are in the encoded-venv" should be mentioned once early. I believe saying it multiple times is noise.

README.md
```diff
@@ -1,338 +1,58 @@
...
+
+
+## Application Installation
+For issues see [Snovault OSX App Installation][] first.
+
+1. Create a virtual env in your work directory.
+    Here we use python3 venv module.  Use venv, like conda, if you please
+    ```
+    $ cd your-work-dir
+    $ python3 -m venv encoded-venv
```
zoldello, line 18: In theory, no one should be using Python2 after January 1st, 2020... So, it "Python3" still needed?

README.md.pre-ubuntu18
```diff
@@ -0,0 +1,338 @@
+# ENCODE Metadata Database
```
zoldello, line 1: Are we still supporting ENCODE/SNOVAULT pre-ubuntu18? If no, do we need this page?

README.md.pre-ubuntu18
```diff
@@ -0,0 +1,338 @@
...
+-------------------------------------------------------
+
+
+[![CircleCI](https://circleci.com/gh/ENCODE-DCC/encoded/tree/dev.svg?style=svg)](https://circleci.com/gh/ENCODE-DCC/encoded/tree/dev)
+
+## Setting Up Your Environment
+
+These are the primary software versions used in production, and you should be able to use them locally:
```
zoldello, line 9: I do not see the point of mentioning this in that these Encode will not work without even one of the other dependencies.

README.osx.catalina.md
```diff
@@ -1,58 +0,0 @@
-ENCODE Metadata Database
```
zoldello, line 1: Why was this deleted?

zoldello: Are we still supporting mojave and other older versions of mac?

zoldello (dismissed):
Most of my comments are suggestions. There is one small typo that may need to be fixed (if you deem a wording suggestion irrelevant.)

README.osx.catalina.md
caseylitton: This is the README now.  Also having catalina in the file name was confusing for people on mojave.

README.md.pre-ubuntu18
caseylitton: These were the original instructions.  I didn't edit it, just moved the file.

README.md.pre-ubuntu18
caseylitton: I was going to keep it for a bit.  There are few external users of the repo who may still be on ubuntu14.  Although we could very well be the last people to use it!

README.md
caseylitton: One day maybe!  The default python command on osx and ubuntu 18 is still python2.  

When I see 'python' in readme I assume python3, but don't mind the clarity when people write the 3.

hitz (approved):

PR #3228, opened 2020-03-31
ENCD-5173-add-to-reference-summary-page

khzlin:


forresttanaka (dismissed):
Very nice! And thanks for fixing the comment.

forresttanaka (approved):
I think that should work. Interesting we haven’t needed that before. We use ExperimentTable in several places.

PR #3229, opened 2020-04-01
ENCD-5164 Create v98 demo and prod amis

caseylitton:


cloud-config/create-ami.py
```diff
@@ -70,9 +70,10 @@ def _find_similar_amis(ec2_client, ami_name):
                 'description': res_img['Description'],
                 'state': res_img['State'],
             }
-            for tag in res_img['Tags']:
-                ami_obj[tag['Key']] = tag['Value']
-            ami_objs.append(ami_obj)
+            if 'Tags' in res_img:
+                for tag in res_img['Tags']:
```
hitz, line 74: I think you can also use:
`for tag in reg_img.get('Tags',[]):`
I think that's the pythonic way of iterating over a list in an dict.

hitz: there is probably also a list comprehension way to do this.

hitz (approved):

cloud-config/create-ami.py
caseylitton: yea, I threw this script together...It will not be around long.  For some reason prod aws instances had some issue with the previous code.

PR #3231, opened 2020-04-02
ENCD-5197-return-paths-from-libraries-calculated-property

lrowe:
The uuid is escaping into the object frame of this replicate when you request it with frame=object.


lrowe:
Example here: https://www.encodeproject.org/replicates/51c85fe3-0993-4157-ad5e-bf0b803dc6d8/?format=json&frame=object

yunhailuo (approved):

PR #3233, opened 2020-04-02
ENCD-5199-add-ref-epig-bdd-and-polish

zoldello:


forresttanaka (approved):
That does look better. Wanted to get your opinion on the test output where you’ll see things like:

```
src/encoded/tests/features/test_matrix.py::test_matrix_reference_epigenome_2
Scenario: Matrix Reference Epigenome
Step: I visit "/" PASSED
Step: I wait for the content to load PASSED
Step: the title should contain the text "ENCODE" PASSED
Step: I press "Data" PASSED
Step: I click the link with text that contains "Human reference epigenomes" PASSED
```

I _think_ the `test_matrix_reference_epigenome_2` is probably generated by bdd testing itself, I’m guessing from the “Scenario” lines from the .feature files. Should we put more effort into making those scenario titles unique to make debugging BDD tests easier? Looking through the console output (and reference epigenome tests aren’t the only place this happens) it can be tough to figure out which test I’m looking at when they’re only differentiated with “_1” “_2” and so on.

PR #3234, opened 2020-04-02
ENCD-5194-add-grnas-file-output-type

jenjou:


gabdank (changes requested):
I think you need to add these also to analysis_steps inputs/outputs

gabdank (dismissed):

hitz (approved):

PR #3237, opened 2020-04-03
ENCD-4374-fix-restrict-file-edit-checkbx

zoldello:


src/encoded/static/components/form.js
```diff
@@ -688,7 +688,7 @@ export class Field extends UpdateChildMixin(React.Component) {
                 searchBase={`?mode=picker&type=${schema.linkTo}`}
             />);
         } else if (schema.type === 'boolean') {
-            input = <input type="checkbox" {...inputProps} />;
+            input = <input type="checkbox" {...inputProps} checked={!!inputProps.value} />;
```
emro: Could we use just 'value' instead of 'inputProps.value' here? I think 'value' is defined as 'inputProps.value' on line 617.

emro (dismissed):
Looks good! Just one minor suggestion.

src/encoded/static/components/form.js
zoldello: Done.

emro (approved):

PR #3239, opened 2020-04-03
ENCD-5140-add-clear-classification-of-related-datasets

yunhailuo:


src/encoded/static/components/experiment.js
```diff
@@ -456,10 +456,21 @@ const ExperimentComponent = ({ context, auditIndicators, auditDetail }, reactCon
...
+            (series) => {
+                if (loggedIn || series.status === 'released') {
+                    if (series['@type'][0] in seriesMap) {
+                        seriesMap[series['@type'][0]].push(series);
+                    } else {
+                        seriesMap[series['@type'][0]] = [series];
+                    }
+                }
+            }
+        );
```
forresttanaka, line 468: I’m not recommending you change this because this works, but for future reference, the underscore library has [one function](http://underscorejs.org/#groupBy) that does exactly this. Inside the callback you would, after the checks, would just return `series['@type'][0]`.

forresttanaka (dismissed):
Looks nice! The one suggestion wouldn’t make the code more efficient; it just saves on coding — coding you’ve already done.

src/encoded/static/components/experiment.js
yunhailuo: Oh. Yes, `groupBy` is what I'm looking for. I stopped when I didn't see useful things on MDN. Thank you for that!

forresttanaka (approved):
Cool!

PR #3245, opened 2020-04-07
ENCD-5151-fix-software-breadcrump

zoldello:


emro (dismissed):
Looks good! You checked that the BDD tests passed locally?

emro:
It sounds like there might be a fix for the BDD tests not passing, maybe it is worth waiting for that fix and re-running the BDD tests?

zoldello:
@emro Good point. Since this is not going in the current spring AFAIK, it should be okay to wait until Paul's bdd-fix is merged

emro (approved):

PR #3248, opened 2020-04-08
ENCD-4991-display-assembly-N-gene_annot-on-UI

zoldello:


src/encoded/static/components/file.js
```diff
@@ -430,7 +430,18 @@ class FileComponent extends React.Component {
...
-
+                                {context.assembly ?
+                                    <div data-test="assembly">
+                                        <dt>Assembly</dt>
+                                        <dd>{context.assembly}</dd>
+                                    </div>
+                                : null}
+                                {context.genome_annotation ?
+                                    <div data-test="genome_annotation">
+                                        <dt>Genome Annotation</dt>
```
emro: Copy-editing comment: all of the other properties with multiple words (aside from PI which is an acronym) have the convention of the first word being capitalized and the following being lowercase. For example: "Date added", "Output type", "File format" etc. I think we should follow this convention for this added property and change "Genome Annotation" to "Genome annotation".

emro (changes requested):
Code is great except for minor style convention.

src/encoded/static/components/file.js
emro: Another thing I just noticed that I think matters not at all is that none of the other "data-test" fields have underscores in them. I'm not sure why this is or if this matters in any way, I just noted it.

emro (approved):

PR #3252, opened 2020-04-09
ENCD-5171-download-fastq-files

forresttanaka:
* I noticed JSDoc doesn't really adequately handle React hooks, so I'm trying out an alternative style of my own, but hopefully some de facto documentation standard comes out.
* I completely rewrote the DownloadButton component to now offer two kinds of dropdown buttons
    * Immediate: The button includes the chevron and executes an action the moment you choose an option from the dropdown menu.
    * Selected: “Github style” — The button and chevron are separate buttons combined, and selecting an option from the dropdown changes the label of the button, and clicking the button then executes the action with the chosen option.
* We have two `<DropdownButton>` uses outside of carts that I converted to use the new `<DropdownButton.Immediate>` component but haven’t tested it because I’m fairly confident that code can never execute. One is in ExperimentSeries when visualizable files exist, but the `visualizable` property is not implemented on ExperimentSeries objects. The other is in edit forms but I could not find a situation that would make that button render.
* In cart/batch_download.js I moved the `batchDownload` function out of `CartBatchDownloadComponent` because I converted to React hooks and didn’t want such a large function allocated on every execution of `CartBatchDownloadComponent.`
* `<DropdownButton.Selected>` is part of a new reusable component to manage the new Download button.
* In view_controls.js, I split the component to display the Download button and modal for non-cart pages into two components — one for the button and one for the modal. This allows cart batch download to use just the modal without also including the button, as the Download button for the cart isn’t a simple button anymore, while non-cart pages do still use a simple button. This modal had to change from plug-and-play into a managed modal.
* For the back end, I removed the “visualizable=true” query string parameter and replaced it with an “option=” parameter. The possible values are “visualizable” and “raw.” The “visualizable” option is just like the old “visualizable=true” parameter, while “option=raw” only downloads files that have no assembly.
* I think this is our first-ever use of custom React hooks, with `useDropdownButton` in button.js. It’s quite a bit like the old mixins so that multiple components can embed the same functionality. The unusual thing I did with this one was to return an object so that I don’t have to return so many array items, which is what you see with most custom React hooks.

### DIFF
--- a/src/encoded/static/scss/encoded/modules/_cart.scss
+++ b/src/encoded/static/scss/encoded/modules/_cart.scss
@@ -250,10 +250,6 @@
     @at-root #{&}__tools {
         display: flex;
         flex: 0 1 auto;
-
-        button {
-            margin-left: 5px;
-        }
     }
 
     // Header of cart panel.
@@ -262,7 +258,7 @@
         justify-content: space-between;
     }
 
-    @at-root #{&}__facet-progress-overlay {
+    @at-root #{&}__facet-disabled-overlay {
         position: absolute;
         top: 0;
         right: 0;
@@ -1096,4 +1092,42 @@ $file-type-colors: (
         flex: 1 1 auto;
     }
 
+}
+
+
+// Control to batch download cart files.
+.cart-download {
+    margin-right: 5px;
+
+    button {
+        @extend .btn;
+        @extend .btn-info;
+        @extend .btn-sm;
+    }
+
+    @at-root #{&}__option-title {
+        margin: 5px 0;
+        font-size: 1rem;
+        font-weight: bold;
+        pointer-events: none;
+    }
+
+    @at-root #{&}__option-description {
+        margin: 5px 0;
+        font-size: 0.9rem;
+        font-weight: normal;
+        pointer-events: none;
+    }
+
+    .dropdown-button__content {
+        button {
+            color: #000;
+
+            &:hover {
+                color: #fff;
+                
+                background-color: #4f689e;
+            }
+        }
+    }
 }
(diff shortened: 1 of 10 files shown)


forresttanaka:
Updated demo: ~https://encd-5171-dl-cart-fastq-pr1-fytanaka.demo.encodedcc.org/~

keenangraham (dismissed):

src/encoded/static/scss/encoded/modules/_cart.scss
```diff
@@ -1097,3 +1093,41 @@ $file-type-colors: (
...
+        pointer-events: none;
+    }
+
+    .dropdown-button__content {
+        button {
+            color: #000;
+
+            &:hover {
+                color: #fff;
+
```
emro, line 1128: extra space?

emro (dismissed):
I looked through everything carefully and it all looks terrific! I like the implementation too, the new drop-down button type, the way that the facets are grayed out on the raw data tab, etc. I wish I'd had one more day to come back to this because I sometimes notice new things that way but it's unlikely!

forresttanaka:
New demo: https://encd-5171-dl-cart-fastq-pr2-fytanaka.demo.encodedcc.org/

forresttanaka:
For the updates since the last code review:

* To disable the Download button if no files can be downloaded, I now pass three different counts in `fileCounts` to be used depending on which dropdown option the user selected.
* Changed CartBatchDownloadComponent to have downloadType be a state instead of a ref because I now use it within the returned JSX.
* I added a red banner to the batch download modal explaining why the user can't download files.

Updated demo:
https://encd-5171-dl-cart-fastq-pr2-fytanaka.demo.encodedcc.org/

emro (approved):

PR #3255, opened 2020-04-13
ENCD-5212-deploy-demo-pointing-at-es-cluster

caseylitton:


cloud-config/deploy-run-scripts/app-wrapper.sh
```diff
@@ -75,7 +75,7 @@ $ENCD_CC_DIR/app-es-status.sh
 $ENCD_CC_DIR/app-pg.sh "$standby_mode"
 $ENCD_CC_DIR/app-encd.sh $ENCD_ROLE $ES_IP $ES_PORT $REGION_INDEX $APP_WORKERS
 sudo -u root $ENCD_CC_DIR/app-a2en.sh
-if [ "$ENCD_ROLE" == "demo" ]; then
+if [ "$ENCD_ROLE" == "demo" ] && [ ! "$ENCD_BUILD_TYPE" == 'encd-demo-no-es-build' ]; then
     $ENCD_CC_DIR/app-batchupgrade.sh production.ini $BUP_VAR_1 $BUP_VAR_2 $BUP_VAR_3 $BUP_VAR_4
```
hitz, line 79: I think we said otherwise, but technically this will work with only local postgres.  It just won't match what's the in the external ES, so probably it's best to leave out.

hitz (approved):

cloud-config/deploy-run-scripts/app-wrapper.sh
caseylitton: I think batch upgrade crashed during deployment then I added this.  I thought it was due to missing elasticsearch but I didn't look into it.

hitz (changes requested):
I know I just approved this but I have to reject it now because the change to regular clusters is mixed in (L796)

hitz (approved):

caseylitton:
Ticket was included https://github.com/ENCODE-DCC/encoded/pull/3273

PR #3256, opened 2020-04-13
ENCD-5099-add-index-read-on-file-page

zoldello:


src/encoded/static/components/file.js
```diff
@@ -324,6 +324,13 @@ class FileComponent extends React.Component {
                                     <dd>{context.output_type}</dd>
                                 </div>
 
+                                {context.index_of ?
+                                <div data-test="outputtype">
+                                    <dt>Index of</dt>
+                                    <dd>{context.index_of.map(file => file.replace(/\/files\/|\//g, '')).join(', ')}</dd>
+                                </div>
+                                : null }
+
```
emro: A style comment: in order to match the other blocks of code here, I would indent lines 328 - 331 and I would get rid of the space after "null" in line 332.

emro (commented):
Just a minor style comment, but I am not approving at this time because of a couple other questions discussed with @zoldello not relating to code.

emro:
Just as another note, I wanted to be sure that the code change accomplished the goal of the ticket, and after reading the Jira ticket and comments I was still slightly confused about how to confirm that and what the data format would be, so it could be super helpful to add a really short description to the PR sometimes, just saying "a property is being added to certain file types (but is not present in any existing data), the property will have this format, the goal of this ticket is to display a list of the property entries" or something along those lines.

src/encoded/static/components/file.js
zoldello: Done. Thanks.

zoldello:
Devs
There are no index_of files in the system yet. That is expected as labs have not started using it. However, the necessary structure supporting it should exist before they start. This ticket mades "index of" available on file pages.

Adding an "Index_of"
You can search for a file by doing <demo url>/search/?type=File&flowcell_details=*

After you find a file, look for "index of" and add a file or two. 

Now, go to the file (<demo url>/files/<fileId>) you modified and you should see the "Index of" column. If not, you know where to find Phil.

src/encoded/static/components/file.js
```diff
@@ -579,6 +579,18 @@ class SequenceFileInfo extends React.Component {
                             </div>
                         : null}
 
+                        {file.index_of ?
+                            <div data-test="outputtype">
+                                <dt>Index of</dt>
+                                <dd>
+                                    {file.index_of.map((f) => {
```
zoldello: react refused to work properly when I used map and join. After searching online, some recommended using map and reduce. See- https://stackoverflow.com/a/40788571/178550

forresttanaka: Wow! I’ve never thought of that. Kind of a mind bender, and that it keeps nesting arrays on each iteration is kind of a shame, though as that post says, React 16 should have no issue with that. I think I would probably use this myself going forward though.

zoldello: @forresttanaka Addressed, based on- https://stackoverflow.com/a/47459785/178550

emro (approved):
Looks great and works well! I like your solution for not being allowed to combine a .map and a .join.

PR #3258, opened 2020-04-15
ENCD-5223-metadata-shift

forresttanaka:
Mostly adding a column to the header row of metadata.tsv so that it lines up with the columns of data.


forresttanaka:
Oh, this is probably going to break a test. Heading to a meeting shortly, but I’ll check that after.

forresttanaka:
Recycled a failed BDD test on report that I see a lot.

keenangraham (approved):
Looks good assuming BDD passes.

PR #3260, opened 2020-04-17
ENCD-5070-missing-img

forresttanaka:
The problem was the resize event mount wasn't getting called after the DOM was rendered, so it based the size of the image on a zero height of the parent div which hadn’t been rendered yet. I decided to rewrite this component with React hooks so that I can use useEffect which first gets called _after_ render, so the height of the image gets calculated after we know how big the parent div is.

### DIFF
--- a/src/encoded/static/components/image.js
+++ b/src/encoded/static/components/image.js
@@ -5,55 +5,52 @@ import * as globals from './globals';
 
 
 // Fixed-position lightbox background and image
-class Lightbox extends React.Component {
-    constructor() {
-        super();
+const Lightbox = ({ lightboxVisible, clearLightbox, lightboxImg }) => {
+    // Height of the image within the light box.
+    const [imgHeight, setImgHeight] = React.useState(0);
+    const lightbox = React.useRef(null);
 
-        // Set initial React state.
-        this.state = { imgHeight: 0 };
+    // Called when the window is resized to resize the image within it.
+    const handleResize = () => {
+        setImgHeight(lightbox.current.offsetHeight - 40);
+    };
 
-        // Bind this to non-React methods.
-        this.handleResize = this.handleResize.bind(this);
-    }
+    // Called when a key is pressed to close the lightbox with ESC.
+    const handleKeyDown = (e) => {
+        if (e.keyCode === 27) {
+            clearLightbox();
+        }
+    };
 
-    componentDidMount() {
-        globals.bindEvent(window, 'resize', this.handleResize);
-        this.setState({ imgHeight: this.lightbox.offsetHeight - 40 });
-    }
+    React.useEffect(() => {
+        // Need to resize and attach event handlers at mount.
+        handleResize();
+        window.addEventListener('resize', handleResize);
+        return (() => {
+            window.removeEventListener('resize', handleResize);
+        });
+    }, []);
 
-    componentWillUnmount() {
-        globals.unbindEvent(window, 'resize', this.handleResize);
-    }
-
-    // Window resized; set max-height of image
-    handleResize() {
-        this.setState({ imgHeight: this.lightbox.offsetHeight - 40 });
-    }
-
-    render() {
-        const lightboxVisible = this.props.lightboxVisible;
-        const lightboxClass = `lightbox${lightboxVisible ? ' active' : ''}`;
-        const imgStyle = { maxHeight: this.state.imgHeight };
-
-        /* eslint-disable jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions */
-        return (
-            <div className={lightboxClass} onClick={this.props.clearLightbox} aria-label="Close" ref={(div) => { this.lightbox = div; }}>
-                <div className="lightbox-img">
-                    <a aria-label="Open image" data-bypass="true" href={this.props.lightboxImg}>
-                        <img src={this.props.lightboxImg} alt="Attachment from submitters" style={imgStyle} />
-                    </a>
-                    <button className="lightbox-close" aria-label="Close" onClick={this.clearLightbox} />
-                </div>
+    const lightboxClass = `lightbox${lightboxVisible ? ' active' : ''}`;
+    return (
+        <div tabIndex="-1" className={lightboxClass} onClick={clearLightbox} onKeyDown={handleKeyDown} role="button" aria-label="Close" ref={lightbox}>
+            <div className="lightbox-img">
+                <a aria-label="Open image" data-bypass="true" href={lightboxImg}>
+                    <img src={lightboxImg} alt="Attachment from submitters" style={{ maxHeight: imgHeight }} />
+                </a>
+                <button className="lightbox-close" aria-label="Close" onClick={clearLightbox} />
             </div>
-        );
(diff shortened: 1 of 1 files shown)


forresttanaka:
Indexed demo at https://encd-5070-missing-img-dv1-fytanaka.demo.encodedcc.org/. A good example of the issue on test is /files/ENCFF284ZEP/ where the quality metric image doesn't appear when you click on it. With this demo, it does appear.

src/encoded/static/components/image.js
```diff
@@ -5,177 +5,149 @@ import * as globals from './globals';
...
-                    <button className="lightbox-close" aria-label="Close" onClick={this.clearLightbox} />
-                </div>
+const Lightbox = ({ lightboxVisible, clearLightbox, lightboxImg }) => {
+    // Height of the image within the light box.
+    const [imgHeight, setImgHeight] = React.useState(0);
+    const lightbox = React.useRef(null);
+
+    // Called when the window is resized to resize the image within it.
+    const handleResize = () => {
+        setImgHeight(lightbox.current.offsetHeight - 40);
```
zoldello, line 15: How did you determine 40?

forresttanaka: It’s a margin from the first implementation that looked reasonable.

src/encoded/static/components/image.js
```diff
@@ -5,177 +5,149 @@ import * as globals from './globals';
...
+    };
+
+    // Called when a key is pressed to close the lightbox with ESC.
+    const handleKeyDown = (e) => {
+        if (e.keyCode === 27) {
```
zoldello, line 20: Recommend adding key code meaning in comment above it. Some may not infer from the comment on top of method name.

src/encoded/static/components/image.js
```diff
@@ -5,177 +5,149 @@ import * as globals from './globals';
...
+            clearLightbox();
+        }
+    };
+
+    React.useEffect(() => {
+        // Need to resize and attach event handlers at mount.
+        handleResize();
+        window.addEventListener('resize', handleResize);
```
zoldello, line 28: If not mistaken, React.useEffect is use tasking needed after mounting or unmounting like clean-up and unsubscring. If true, should event be wired up here?

src/encoded/static/components/image.js
```diff
@@ -5,177 +5,149 @@ import * as globals from './globals';
...
+        return (() => {
+            window.removeEventListener('resize', handleResize);
+        });
+    }, []);
+
+    const lightboxClass = `lightbox${lightboxVisible ? ' active' : ''}`;
+    return (
+        <div tabIndex="-1" className={lightboxClass} onClick={clearLightbox} onKeyDown={handleKeyDown} role="button" aria-label="Close" ref={lightbox}>
```
zoldello, line 36: Is the role and aria-label appropriate for this "div" tag? Is the "figure" tag better to use than "div" (if used then the role and aria would make sense.)

src/encoded/static/components/image.js
```diff
@@ -5,177 +5,149 @@ import * as globals from './globals';
...
+
+    // Close the lightbox and image.
+    const clearLightbox = () => {
+        setLightboxVisible(false);
+    };
+
+    let attachmentHref;
+    let src;
+    let alt;
+    let height = '100';
```
zoldello, line 79: can some of these variables like "height" be in the if-statement? If yes, then they can made const, as it seems they are only used within the if-statement

src/encoded/static/components/image.js
```diff
@@ -5,177 +5,149 @@ import * as globals from './globals';
...
-    // Register for keyup events for ESC key
-    componentDidMount() {
-        globals.bindEvent(window, 'keyup', this.handleEscKey);
-    }
-
-    // Unregister keyup events when component closes
-    componentWillUnmount() {
-        globals.unbindEvent(window, 'keyup', this.handleEscKey);
-    }
+export const Attachment = ({ context, attachment, className, showLink }) => {
```
zoldello, line 61: "showLink"- show what link?
"className"- what affect does the css class have?

src/encoded/static/components/image.js
```diff
@@ -5,177 +5,149 @@ import * as globals from './globals';
...
-                    </div>
-                );
-            } else if (attachment.type === 'application/pdf') {
-                // Attachment is a PDF. Show the PDF icon, and clicks in it show the PDF in a new
-                // window using the browser's PDF viewer.
-                return (
+            // Display the attachment image in a light box.
+            return (
+                <div>
                     <div className="attachment">
```
zoldello, line 98: I have seen people use something like "js-attachment" for css classes that are used for javascript markers rather than styling. Is that a good approach to adopt.

src/encoded/static/components/image.js
```diff
@@ -5,177 +5,149 @@ import * as globals from './globals';
...
 
-        // No attachment given; display a broken file icon.
+        // Non image, non PDF attachment (likely a text file). Show the generic document icon
+        // and open it in a new window.
         return (
             <div className="attachment">
-                <div className="attachment__button">
-                    <div className="file-missing">Attachment file broken icon</div>
-                </div>
+                <a data-bypass="true" href={attachmentHref} className="attachment__button" target="_blank" rel="noopener noreferrer" title="Open attachment in a new window">
```
zoldello, line 124: is data-bypass still needed?

zoldello (approved):
I saw no security problems or code that must be changed. None of my comments are suggestions that can be ignored.

src/encoded/static/components/image.js
forresttanaka: Actually, I wasn't sure which one to use. Since it's clickable, I figured “button” was a good one.

src/encoded/static/components/image.js
forresttanaka: showLink: “False to just display image preview without link or lightbox” from the comment in propTypes.

className: “CSS class name to add to image element; '-img' added to it”.

src/encoded/static/components/image.js
forresttanaka: This is always the trade off. I rejiggered this function to use hooks, but I didn’t touch the guts of it to avoid having to specifically test that. But yes, you’re right. It probably was used outside the `if` at some point in the distant past.

src/encoded/static/components/image.js
forresttanaka: Haven’t thought about that specifically. `.attachment` has actual styles associated with it. Is it used as a Javascript maker? I don’t see where that happens, but I haven‘t looked at this code in a long time.

src/encoded/static/components/image.js
forresttanaka: I really have no idea. We use it all over the place but I don’t remember what it’s for. It might be worth a ticket to investigate, and remove all uses of it if we don't need it.

src/encoded/static/components/image.js
zoldello: It was rhetorical. :)  The name are fine. I was just suggesting that maybe the names need another noun as they required me to look at the code to get a glimpse of their purpose. But it could just be me.

src/encoded/static/components/image.js
zoldello: Acknowledged.

PR #3266, opened 2020-04-21
ENCD-5200-add-FCC-ExperimentTable

khzlin:


src/encoded/static/components/gene.js
```diff
@@ -140,10 +140,16 @@ class Gene extends React.Component {
...
                 <RelatedItems
-                    title={`Experiments targeting gene ${context.symbol}`}
+                    title={`Functional genomics experiments targeting gene ${context.symbol}`}
                     url={`/search/?type=Experiment&target.genes.uuid=${context.uuid}`}
                     Component={ExperimentTable}
                 />
+
+                <RelatedItems
+                    title={`Functional characterization experiments targeting gene ${context.symbol}`}
+                    url={`/search/?type=FunctionalCharacterizationExperiment&target.genes.uuid=$context.uuid}`}
```
forresttanaka: Missing an open brace before context.uuid, so this puts the literal string “...genes.uuid=$context.uuid}” as the query string.

forresttanaka (changes requested):
Looks like this should work (I haven’t tried the demo) except for one small mistake.

src/encoded/static/components/gene.js
khzlin: Thanks for catching this! I'll fix this asap.

forresttanaka (dismissed):

forresttanaka (approved):

PR #3268, opened 2020-04-22
ENCD-5230-change-tech-replicate-field-of-metadata-tsv

jenjou:


gabdank (approved):

src/encoded/batch_download.py
```diff
@@ -72,7 +72,7 @@ def includeme(config):
     ('Library fragmentation method', ['files.replicate.library.fragmentation_method']),
     ('Library size range', ['files.replicate.library.size_range']),
     ('Biological replicate(s)', ['files.biological_replicates']),
-    ('Technical replicate', ['files.replicate.technical_replicate_number']),
+    ('Technical replicate', ['files.technical_replicates']),
```
gabdank: I think it should be changed into "Technical replicate(s)" to be similar to the behavior of biological replicates, while files.replicate.tech_rep was ONLY one, after the change to the calculated property it could be a list

src/encoded/tests/test_batch_download.py
```diff
@@ -80,7 +80,7 @@ def test__tsv_mapping_value():
         ('Library fragmentation method', ['files.replicate.library.fragmentation_method']),
         ('Library size range', ['files.replicate.library.size_range']),
         ('Biological replicate(s)', ['files.biological_replicates']),
-        ('Technical replicate', ['files.replicate.technical_replicate_number']),
+        ('Technical replicate', ['files.technical_replicates']),
```
gabdank: Technical replicate -> Technical replicate(s)

gabdank (changes requested):

gabdank (approved):

PR #3270, opened 2020-04-22
ENCD-5229-fix-run-type-dependency

jenjou:


gabdank (changes requested):
@jenjou can you please add schema tests that make sure the dependency behaves properly, like posting FASTQ not PacBio and making sure run_type is required, and few more fixtures to have dependencies in check for future schema updates?

jenjou:
There is an existing test_no_runtype_readlength_dependency so I just added test_no_runtype_dependency as I'm not sure what else would be useful.

gabdank (dismissed):

hitz:
@jenjou has conflicts with tests/fixtures/schemas/file.py  please rebase and force-push

gabdank (approved):

PR #3271, opened 2020-04-22
ENCD-4378-remove-restrict-target-dbxref

bonitalam:


yunhailuo (changes requested):
Find a few more places:
https://github.com/ENCODE-DCC/encoded/blob/ENCD-4378-remove-restrict-target-dbxref/src/encoded/static/components/search.js#L554-L555
https://github.com/ENCODE-DCC/encoded/blob/ENCD-4378-remove-restrict-target-dbxref/src/encoded/static/components/testdata/lot_review.js#L20-L47
https://github.com/ENCODE-DCC/encoded/blob/ENCD-4378-remove-restrict-target-dbxref/src/encoded/static/components/testdata/target/HNRNPA1-human.js#L12-L21

yunhailuo (approved):

PR #3274, opened 2020-04-23
ENCD-5203-biosample-audit-inconsistent-GMs

iyoungworth:


src/encoded/audit/biosample.py
```diff
@@ -61,6 +61,31 @@ def audit_biosample_modifications(value, system):
...
+            purpose = 0
+            GM_ids = set()
+            for GM in value['applied_modifications']:
+                if GM['method'] == 'CRISPR':
+                    method += 1
+                    if GM['purpose'] == 'characterization':
+                        purpose += 1
+                        GM_ids.add(GM['@id'])
+            GM_ids_links = [audit_link(path_to_text(m), m) for m in GM_ids]
+            if purpose == method & purpose >1:
```
gabdank: I think this is not going to work in cases where you have (CRISPR-characterization, CRISPR-characterization, CRISPR-not_characterization) method will be 3 and purpose will be 2 so despite 2 CRISPR characterization the audit will not trigger?

gabdank (changes requested):

src/encoded/audit/biosample.py
iyoungworth: Somehow it does seem to work correctly (I checked on the demo), but rereading the code I'm not sure how that is possible. I think I see now how to write this in a way that's more clean and robust, thanks!

gabdank (approved):

PR #3278, opened 2020-04-24
ENCD-5116-pipeline-graph

forresttanaka:
* Most of this change revolves around the “compiled analyses” object which takes the array of analyses objects from an experiment and reassembles them to be more convenient for generating the analyses dropdown that replaces the Assembly facet if conditions are right.
* To be included in compiled analyses, analyses objects have to satisfy several criteria. `compileAnalyses` handles all these checks as well as the conversion to compiled analyses.

An often-updated demo here:
~https://encd-5116-pipeline-graph-pr1-fytanaka.demo.encodedcc.org/search/?type=Experiment&analyses=*~

### DIFF
--- a/src/encoded/static/components/filegallery.js
+++ b/src/encoded/static/components/filegallery.js
@@ -1986,8 +1986,154 @@ function createFacetObject(propertyKey, fileList, filters) {
     return facetObject;
 }
 
+
+/**
+ * Compile the usable experiment analyses objects into a form for rendering a dropdown of pipeline
+ * labs. Export for Jest test.
+ * @param {object} experiment Contains the analyses to convert into an pipeline labs dropdown
+ * @param {array} files Array of all files from search that gets included in file gallery
+ *
+ * @return {array} Compiled analyses information, each element with the form:
+ * {
+ *      title: Analyses dropdown title -- `pipelineLab`+`assembly`
+ *      pipelineLab: Pipeline processing lab title, e.g. "ENCODE4 uniform"
+ *      assembly: Assembly and annotation string matching assembly facet terms
+ *      assemblyAnnotationValue: Value used to sort the compiled analyses
+ *      files: Array of files selected with the pipline lab and assembly
+ * }
+ */
+const UNIFORM_PIPELINE_LAB = '/labs/encode-processing-pipeline/';
+export const compileAnalyses = (experiment, files) => {
+    let compiledAnalyses = [];
+    if (experiment.analyses && experiment.analyses.length > 0) {
+        // Get all the analyses objects that qualify for inclusion in the Pipeline facet.
+        const qualifyingAnalyses = experiment.analyses.filter((analyses) => {
+            const rfas = _.uniq(analyses.pipeline_award_rfas);
+
+            // More than one lab OK, as long as none of them is `UNIFORM_PIPELINE_LAB` --
+            // `UNIFORM_PIPELINE_LAB` is only valid if alone.
+            return (
+                analyses.assemblies.length === 1
+                && analyses.genome_annotations.length <= 1
+                && analyses.pipeline_award_rfas.length === 1
+                && analyses.pipeline_labs.length > 0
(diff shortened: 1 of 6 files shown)


src/encoded/static/components/filegallery.js
```diff
@@ -1986,8 +1986,154 @@ function createFacetObject(propertyKey, fileList, filters) {
...
+
+
+/**
+ * Find the compiled analyses object matching the given assembly and annotation and return its
+ * index in the compiled analyses array.
+ * match found.
+ * @param {string} assembly Assembly and annotation to map to an analyses title
+ * @param {array} compiledAnalyses Compiled analyses for the experiment
+ *
+ * @return {object} Array index of compiled analyses object matching given assembly, or -1
```
emro, line 2077: I think that this is just the formatting from a plug-in you use for defining functions and components, but I would find it more readable if there was a colon after the variable name, like "assembly: Assembly and annotation" etc. I don't think you should change this, it's just a comment about the plug-in.

src/encoded/static/components/filegallery.js
```diff
@@ -2542,7 +2784,7 @@ class FileGalleryRendererComponent extends React.Component {
                         >
                             <TabPanelPane key="browser">
                                 <GenomeBrowser
-                                    files={includedFiles}
+                                    files={graphIncludedFiles}
```
emro: I'm not sure that this change quite works the way it should. Trying to remember why there are so many different arrays of files and what they are all for took a little while though - really, when I first wrote all this code I should have documented this better. It seems like you want to make sure that you're allowing the user to choose an analysis, just like for the graph, which is very important. However, I think that the difference between "includedFiles" and "graphIncludedFiles" lies in whether or not the array of files is filtered based on the user-selected filters (see lines 2444 and 2465) and it seems like it might be important to keep this filtering on the genome browser. After looking at the code, I compared the demo and production and if you select filters on the graph tab and then switch to the genome browser tab, the filters are not applied to the displayed files, and I think that is because of this line change.

Screen Shot 2020-04-28 at 11 28 06 AM

emro (changes requested):
Overall this looks great! Everything is well-commented and clear and it's well integrated. There is a lot of complexity to collecting all the analyses into a selector! I just have one comment about how the change is being implemented on the genome browser.

src/encoded/static/components/filegallery.js
forresttanaka: Yeah, JSDoc likes only a space after the variable name, but I’m not sure what I think about JSDoc anymore. I complied with it so it’ll pass JSDoc tools, but we never run those tools because we have so much code, even new code, that doesn’t comply, and it works poorly for React functional components. It might be best just to do whatever seems clearest, going forward.

src/encoded/static/components/filegallery.js
forresttanaka: Switching between tabs was a nightmare in this ticket. I’ll try to figure out what’s happening here.

forresttanaka: @emro I patched the demo directly to change `graphIncludedFiles` back to `includedFiles`. Does it look better to you? I’m not sure why I changed that.

~https://encd-5116-pipeline-graph-pr1-fytanaka.demo.encodedcc.org/experiments/ENCSR000AFA/~

forresttanaka:
Updated demo here: ~https://encd-5116-pipeline-graph-pr2-fytanaka.demo.encodedcc.org/~

src/encoded/static/components/filegallery.js
emro: The update looks good in the code but for some reason I'm getting a "bad gateway" error when I try to visit your demo. The demo says it's running on AWS, though.

forresttanaka: I changed the domain name this morning so you’ll have to go to https://encd-5116-pipeline-graph-pr2-fytanaka.demo.encodedcc.org/ now.

emro (approved):
Looks good!

PR #3280, opened 2020-04-27
ENCD-5157-add-to-biosample-summary-and-display

khzlin:


src/encoded/schemas/changelogs/biosample.md
```diff
@@ -4,7 +4,7 @@
 
 * Added *post_differentiation_time* and *post_differentiation_time_units* properties.
 * Added *perturbed* calculated property to indicate if the biosample has been perturbed using a treatment or a genetic modification for non-tagging purposes.
-* Added *post_nucleic_acid_delivery_time* and *post_nucleic_acid_delivery_time_units* properties.
+* Added *post_nucleic_acid_delivery_time* and *post_nucleic_acid_delivery_time_units* properties and to the calculated property of summary.
```
zoldello, line 7: Shouldn't ENCD-5139 had done this?

khzlin: ENCD-5139 doesn't add the properties to the biosample_summary. It only created the properties. ENCD-5157 ticket does 2 things. 1) display post_nucleic_acid_delivery_time and post_nucleic_acid_delivery_time_units on the biosample page, 2) add post_nucleic_acid_delivery_time and post_nucleic_acid_delivery_time_units to calculated biosample_summary. Let me know if this makes sense.

src/encoded/tests/test_types_biosample.py
```diff
@@ -116,10 +116,13 @@ def test_biosample_summary(testapp,
     testapp.patch_json(biosample_1['@id'], {'donor': donor_1['@id'],
                                             "biosample_ontology": liver['uuid'],
                                             "preservation_method": "cryopreservation",
+                                            "post_nucleic_acid_delivery_time": 3,
+                                            "post_nucleic_acid_delivery_time_units": "week",
                                             'treatments': [treatment_5['@id']]})
     res = testapp.get(biosample_1['@id']+'@@index-data')
     assert res.json['object']['summary'] == (
-        'Homo sapiens male child (10 days) liver tissue treated with ethanol, preserved by cryopreservation')
+        'Homo sapiens male child (10 days) liver tissue treated with ethanol,'
```
zoldello, line 124: This is a non-code review comment. I thought living tissue was preserved in formaldehyde?

src/encoded/types/biosample.py
```diff
@@ -709,6 +717,13 @@ def generate_summary_dictionary(
             pluralize(post_treatment_time, post_treatment_time_units)
             )
 
+    if post_nucleic_acid_delivery_time is not None and \
```
zoldello, line 720: As-is this is correct and is a good way to do it. If interested, since we updated to Python 3.6+, you can use f-string (https://cito.github.io/blog/f-strings/). I feel it is an even better way.

src/encoded/types/biosample.py
```diff
@@ -709,6 +717,13 @@ def generate_summary_dictionary(
...
+        post_nucleic_acid_delivery_time_units is not None:
+        dict_of_phrases['post_nucleic_acid_delivery_time'] = '{} post-nucleic acid delivery time'.format(
```
zoldello, line 722: As-is this is correct and is a good way to do it. If interested, since we updated to Python 3.6+, you can use f-string (https://cito.github.io/blog/f-strings/). I feel it is an even better way.

src/encoded/types/biosample.py
```diff
@@ -832,6 +847,9 @@ def generate_sentence(phrases_dict, values_list):
             if 'preservation_method' in key:
                 sentence = sentence.strip() + ', ' + \
                                     phrases_dict[key].strip() + ' '
+            elif 'post_nucleic_acid_delivery_time' in key:
```
zoldello, line 850: As-is this works so it is fine. But I would recommend f-string (https://cito.github.io/blog/f-strings/) for the future.

src/encoded/tests/test_types_biosample.py
khzlin: The properties are arbitrary. The summary printout just depends on the properties patched just a few lines prior (L115-121).

src/encoded/types/shared_calculated_properties.py
```diff
@@ -313,6 +313,8 @@ def biosample_summary(self,
                                 biosampleObject.get('post_synchronization_time_units'),
                                 biosampleObject.get('post_treatment_time'),
                                 biosampleObject.get('post_treatment_time_units'),
+                                biosampleObject.get('post_nucleic_acid_delivery_time'),
```
zoldello, line 316: Do you need to add a default value? Something like biosampleObject.get('post_nucleic_acid_delivery_time', '1')

zoldello (dismissed):
Looks good. None of my comments are required changes. I saw no security issue or anything else that could raise an alarm

src/encoded/schemas/changelogs/biosample.md
gabdank: Instead of expanding the line with "and to the .." please add a separate line, saying you have updated the summary property to include post nucleic acid delivery time info.

gabdank (changes requested):

src/encoded/schemas/changelogs/biosample.md
khzlin: Ok

gabdank (approved):

src/encoded/tests/test_types_biosample.py
hitz: @zoldello there are many ways to preserve tissue - and in this case it's no longer "living".  Formaldehyde specifically interferes with many downstream biology experiments.

PR #3281, opened 2020-04-27
ENCD-5100-index-reads-grouping

forresttanaka:
Nearly all the relevant change is in filegallery.js. I had to add a lot of local test data to be able to develop this. I have a demo with this current build, with a search that shows the paired-ended case. We have no single-ended cases in production data, but I have one in local test data.


src/encoded/static/components/filegallery.js
```diff
@@ -151,7 +151,13 @@ export class FileTable extends React.Component {
             // Extract four kinds of file arrays
             const files = _(datasetFiles).groupBy((file) => {
                 if (file.output_category === 'raw data') {
-                    return file.output_type === 'reads' ? 'raw' : 'rawArray';
+                    if (file.output_type === 'reads') {
```
zoldello, line 154: Would an if- else if be a tad better? Isolated if-statements imply the if-statements are not related nor operating on the same variable.

src/encoded/static/components/filegallery.js
```diff
@@ -482,16 +489,30 @@ class RawSequencingTable extends React.Component {
         };
 
         // Bind `this` to non-React methods.
+        this.findIndexFile = this.findIndexFile.bind(this);
```
zoldello, line 492: findIndexFile or GetIndexFile/FetchIndexFile?

src/encoded/static/components/filegallery.js
```diff
@@ -482,16 +489,30 @@ class RawSequencingTable extends React.Component {
...
+    /**
+     * Find the index reads file associated with the one or two given files. For paired reads, both
+     * files have to match to return a matching index reads file. For singled-ended reads, only
+     * `file0` needs to match.
+     * @param {object} file0 - First reads file of pair or only file of single
+     * @param {object} file1 - Second reads file of pair
+     *
+     * @return {object} Index reads file with index_of matching given files; undefined if none
+     */
+    findIndexFile(file0, file1) {
```
zoldello, line 505: This is a find. However, from the POV of the consumer of the method, it is a GET

src/encoded/static/components/filegallery.js
```diff
@@ -528,24 +549,61 @@ class RawSequencingTable extends React.Component {
...
-                            // Both the file and its partner qualify as good pairs of each other. Let
-                            // them pass the filter, and record set their sort keys to the lower of
-                            // the two accessions -- that's how pairs will sort within a biological
-                            // replicate.
-                            partner.pairSortKey = file.title < partner.title ? file.title : partner.title;
-                            file.pairSortKey = partner.pairSortKey;
-                            file.pairSortKey += file.paired_end;
-                            partner.pairSortKey += partner.paired_end;
+                            // Both the file and its partner qualify as good pairs of each other.
+                            // Let hem pass the filter, and set their sort keys to the lower of the
```
zoldello: "them" and not "hem"

src/encoded/static/components/filegallery.js
```diff
@@ -505,8 +526,8 @@ class RawSequencingTable extends React.Component {
             // Make lists of files that are and aren't paired. Paired files with missing partners
             // count as not paired. Files with more than one biological replicate don't count as
             // paired.
-            const nonpairedFiles = [];
-            const pairedFiles = _(files).filter((file) => {
+            let nonpairedFiles = [];
```
zoldello, line 529: Why did you change this from "const" to "let"? ESLINT did not bark at me when pulled down this code and changed these to const locally.

src/encoded/static/components/filegallery.js
```diff
@@ -528,24 +549,61 @@ class RawSequencingTable extends React.Component {
...
                             return true;
                         }
                     }
                 }
 
-                // File not part of a pair; add to non-paired list and filter it out
-                nonpairedFiles.push(file);
+                // File not part of a pair; add it to the non-paired list as a copy so we can
+                // mutate them later.
+                nonpairedFiles.push(Object.assign({}, file));
```
zoldello, line 570: Is there a reason why you created a new object via Object.assign. rather than use file-object itself?

It would be cool if we upgrade babel to support object spread (https://stackoverflow.com/q/32925460/178550)

src/encoded/static/components/filegallery.js
```diff
@@ -528,24 +549,61 @@ class RawSequencingTable extends React.Component {
...
+            // we only look for the one file of the pair in `pairedFiles` to get its pairSortKey
+            // root (without appended `paired_end` string) -- not both files.
+            const pairedFileIndexReads = [];
+            if (indexFiles) {
+                indexFiles.forEach((indexFile) => {
+                    const indexFileCopy = Object.assign({}, indexFile);
+                    const matchingPairFile = pairedFiles.find(pairedFile => indexFile.index_of.includes(pairedFile['@id']));
+                    if (matchingPairFile) {
+                        indexFileCopy.pairSortId = matchingPairFile.pairSortId;
+                        indexFileCopy.pairSortKey = `${matchingPairFile.pairSortId}-I`;
```
zoldello, line 603: Is the 'I' at the end intentional?

src/encoded/tests/test_batch_download.py
```diff
@@ -328,7 +328,7 @@ def test_batch_download_report_download(testapp, workbook):
         b'Post-synchronization time', b'Post-synchronization time units',
         b'Replicates',
     ]
-    assert len(lines) == 64
+    assert len(lines) == 65
```
zoldello, line 331: This increased by 1 because you added a new experiments. Now, what is special about the one you added? The data looks like the others. Was it "alias"?

src/encoded/static/components/filegallery.js
```diff
@@ -554,12 +612,12 @@ class RawSequencingTable extends React.Component {
...
             let pairedRepKeys = [];
             if (pairedFiles.length > 0) {
-                pairedRepGroups = _(pairedFiles).groupBy(file => (
-                    (file.biological_replicates && file.biological_replicates.length === 1) ?
-                        globals.zeroFill(file.biological_replicates[0]) + ((file.replicate && file.replicate.library) ? file.replicate.library.accession : '')
-                    :
-                        'Z'
-                ));
+                pairedRepGroups = _(pairedFiles).groupBy((file) => {
+                    if (file.biological_replicates && file.biological_replicates.length === 1) {
```
zoldello, line 619: How is this an improvement from the previous code?

zoldello (dismissed):
The code looks fine. My comments are nothing that must be changed. I am not sure if it is worth even changing things for them small typo

src/encoded/static/components/filegallery.js
forresttanaka: With embedded returns, that wouldn’t be allowed under [AirBnB](https://github.com/airbnb/javascript#arrays--callback-return).

src/encoded/static/components/filegallery.js
forresttanaka: I’m going to change that..

src/encoded/static/components/filegallery.js
forresttanaka: nonpairedFiles gets reassigned on line 573. ESLint gives me an error.

src/encoded/static/components/filegallery.js
forresttanaka: I’m mutating the file objects, and the file objects come in as a prop.

src/encoded/static/components/filegallery.js
forresttanaka: Very much, so that it always sorts behind {accession}-1 and {accession}-2. I never explain that though, and I should.

src/encoded/tests/test_batch_download.py
forresttanaka: Because I didn’t want to use another experiment to test specifically one with index reads files, in case that removed a test of something else.

src/encoded/static/components/filegallery.js
forresttanaka: I had initially modified this to do additional things and had to remove the ternary expression. I removed that change, but felt the ternary condition was so long and unwieldy, I preferred using the if.

src/encoded/static/components/filegallery.js
zoldello: Thanks. I missed that.

src/encoded/static/components/filegallery.js
zoldello: did you change this?

forresttanaka: Yes. You’re looking at the outdated version.

zoldello (dismissed):
Looks good.

zoldello (approved):
Updated (rawIsIndex) looks fine.

PR #3284, opened 2020-04-27
ENCD-5217-expt-fcc-audit-inconsistent-gms

bonitalam:


src/encoded/audit/experiment.py
```diff
@@ -4042,6 +4042,41 @@ def audit_experiment_inconsistent_analyses_files(value, system, files_structure)
...
+            if (rep['status'] not in excluded_types and 'library' in rep and rep['library']['status'] not in excluded_types and 'biosample' in rep['library'] and rep['library']['biosample']['status'] not in excluded_types):
+                biosampleObject = rep['library']['biosample']
+                modifications = biosampleObject.get('applied_modifications')
+                if not modifications:
+                    genetic_modifications['no genetic modifications'].add(biosampleObject['@id'])
+                else:
+                    for gm in modifications:
+                        gm_ids = set()
+                        gm_ids.add(gm['@id'])
+                        gm_combined = tuple(sorted(gm_ids))
```
yunhailuo: Not sure if things are right here. For what you have here, doesn't that end up just `gm_combined = (modifications[-1]['@id],)`?

Do you actually want `gm_combined = tuple(sorted(gm['@id'] for gm in modifications))`?

yunhailuo (changes requested):

src/encoded/audit/experiment.py
bonitalam: Ah yes you're right, my mistake. Comprehension is better too, will push a commit to change it.

yunhailuo (approved):

PR #3285, opened 2020-04-27
ENCD-5178 add body map to summary page

emro:
The main goal of the ticket is to create a body facet component which is integrated into the Summary page. There are a number of updates associated with this change:
- Selection of organism is now a radiobutton-type functionality.
- There is a new body map component, which consists of a diagram, inset images, and lists of organs, tissue types, and systems. You can select organs or systems either by clicking on the svg image or by clicking on the lists of terms.
- Facets on the summary page are reduced.
- The summary page has been extended for a new, wider design.
- New parameters have been added to the ViewControls and SummaryData components to allow for more display flexibility.

Two notes about the code:
(1) I'm a little concerned that if there are changes in the ontology, that will necessitate code changes in `body_map.js`, because that is where I'm storing all of the relevant mappings. Ideally, there might be a file somewhere from which we are loading these mappings. It's a little complicated, though, particularly with the dependencies between the organs and systems, so this may be the best solution for now.
(2) I needed to combine organ slims and system slims at times and had to write some verbose code to make spread syntax work, please let me know if you're aware of a more concise way to write that code.


src/encoded/static/components/body_map.js
```diff
@@ -0,0 +1,766 @@
...
+    'skeletal system': ['bone element', 'skeleton'],
+};
+
+// Unhighlight all highlighted organ / inset image / systems terms and all highlighted svg paths / shapes and all highlighted inset images
+const unHighlightOrgan = () => {
+    const matchingElems = document.querySelectorAll('.highlight');
+    [].forEach.call(matchingElems, (el) => {
+        el.classList.remove('highlight');
+    });
+};
```
forresttanaka, line 113: Could this be done with a simple forEach, or did you try that and it didn’t work? Since the result of `querySelectorAll` isn’t live, it seems like this would work, but I’ve never done this so I’m not sure. https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelectorAll#Accessing_the_matches

src/encoded/static/components/body_map.js
```diff
@@ -0,0 +1,766 @@
...
+// Highlight all of the svg paths / shapes corresponding to a hovered-over svg path / shape and highlight corresponding term(s)
+// Most organs are comprised of multiple svg paths and we want all of the corresponding svg components to highlight together
+// For example, when the user hovers over one kidney, we want both kidneys to highlight because both will be selected upon click
+// As another example, "musculature of body" is comprised of 7 paths right next to each other and it would be confusing for just one line or section to highlight on hover
+const svgHighlight = (e) => {
+    const highlightElems = document.querySelectorAll('.highlight');
+    // Remove existing highlights
+    [].forEach.call(highlightElems, (el) => {
+        el.classList.remove('highlight');
+    });
```
forresttanaka: Could call `unHighlightOrgan` here? These lines look identical to that function.

src/encoded/static/components/body_map.js
```diff
@@ -0,0 +1,766 @@
...
+                                [].forEach.call(matchingElems, (el) => {
+                                    el.classList.add('highlight');
+                                });
+                            });
+                        }
+                    } else {
+                        const matchingElems = document.querySelectorAll(`.${bodyClass}`);
+                        [].forEach.call(matchingElems, (el) => {
+                            el.classList.add('highlight');
+                        });
```
forresttanaka: This same operation is done in so many places, I feel like it could be a function that just takes the selector string and CSS class as a parameter. You could even make a utility class that has add and remove operations, if it’s worth it (you’d have to make an instance of that class, which I’m not sure makes things easier or harder).

src/encoded/static/components/body_map.js
```diff
@@ -0,0 +1,766 @@
...
+        this.chooseOrgan = this.chooseOrgan.bind(this);
+        this.clearOrgans = this.clearOrgans.bind(this);
+    }
+
+    // Once the page has loaded, we want to check to see if any terms are selected
+    // and highlight the body map elements which correspond to those terms
+    componentDidMount() {
+        const searchQuery = url.parse(this.props.context['@id']).search;
+        const terms = queryString.parse(searchQuery);
+        if (terms[organField]) {
```
forresttanaka: You can use our own query-string library to do this.

```
const query = new QueryString(searchQuery);
const terms = query.getKeyValues(organField);
if (terms.length > 0) {
...
```

I’m trying to one day remove our reliance on the queryString npm module as it does several things wrong, like formatting query strings with elements in a different order than the query string you intially gave it, and it returns an array as a value or an individual value whereas ours aways returns an array, so you don’t need the two cases you in many places below. And that would let you remove the typeof checks on the `selectedOrgan` state. Ours also handles negative query strings (key!=value) which the npm package doesn’t.

src/encoded/static/components/body_map.js
```diff
@@ -0,0 +1,766 @@
...
+                    this.setState({
+                        selectedOrgan: [...new Set(newState)],
+                    });
+                }
+
+                // Append or remove organ(s) to / from the search query
+                const parsedUrl = url.parse(this.props.context['@id']);
+                const query = new QueryString(parsedUrl.query);
+                if (active) {
+                    query.addKeyValue(organField, newOrgan, '');
```
forresttanaka: For a tiny bit of simplicity, you can just leave off the third parameter to `addKeyValue`. It’s just looking for something truthy.

src/encoded/static/components/summary.js
```diff
@@ -70,6 +40,21 @@ function generateStatusData(buckets, labels) {
     return statusData;
 }
 
+// Data field for organism
+// We will display different facets depending on the selected organism
+const organismField = 'replicates.library.biosample.donor.organism.scientific_name';
+// Query for systems slim
+const systemsField = 'biosample_ontology.system_slims';
+// Query for organ slim
+const organField = 'biosample_ontology.organ_slims';
```
forresttanaka: Could you import `systemsField` and `organField` from body_map.js so both places don’t need to change if they ever need changing? You could still make `BodyMap` the default export.

src/encoded/static/components/summary.js
```diff
@@ -306,59 +304,131 @@ class SummaryData extends React.Component {
...
+    }
+    render() {
+        let clearButton;
+        const searchQuery = url.parse(this.props.context['@id']).search;
+        if (searchQuery) {
+            // If we have a 'type' query string term along with others terms, we need a Clear Filters
+            // button.
+            const terms = queryString.parse(searchQuery);
+            const nonPersistentTerms = _(Object.keys(terms)).any(term => term !== 'type');
+            clearButton = nonPersistentTerms && terms.type;
```
forresttanaka: For this one, you can use:

```
const query = new QueryString(searchQuery);
const nonPersistentQuery = query.clone();
nonPersistentQuery.deleteKeyValue('type');
clearButton = nonPersistentQuery.getCount() > 0 && query.getCount('type') > 0;
```

forresttanaka (changes requested):
That is a lot of functionality in surprisingly little code! I think there are places you can save even more code.

src/encoded/static/components/body_map.js
emro: I just made one function, with an optional flag parameter for removing a class name (default is adding a class name). Is this slower than a utility function?

There's only one case that I can't use the function for, on line 196 or so:
const matchingElems = document.querySelectorAll(\`.${bodyClass}, .${bodyClass.replace(' ', '-')}\`);
and I could modify the input to the function so it takes the full \`.${bodyClass}\` for every case, but it seems cleaner to just have the one special case where I can't use the function and keep the simpler "bodyClass" as the input.

src/encoded/static/components/body_map.js
emro: I'm really not sure why I didn't just use "forEach" to start with.

emro:
Changes since the last code review:
(1) Addressed code review. I tried to be very careful adding the new "addingClass" function because it is so many places with slight variations!
(2) Updates to ontology. Bonita provided me with updates to some of the system slims, which I incorporated.
(3) Addressed feedback from Ben. I eliminated the three regular facets on for the Homo sapiens page, and added back the icons next to the links to the search / report / matrix links.

Another note:
@hitz and I discussed how I've hard-coded the ontology information, and we agree that this is not a very ideal way to reference this information. However, I'm not going to address it at this time; it will be updated in future tickets.

src/encoded/static/components/summary.js
```diff
@@ -353,16 +353,11 @@ class SummaryBody extends React.Component {
...
-        if (query) {
-            // If we have a 'type' query string term along with others terms, we need a Clear Filters
-            // button.
-            const nonPersistentQuery = query.clone();
-            nonPersistentQuery.deleteKeyValue('type');
-            clearButton = nonPersistentQuery.queryCount() > 0 && query.queryCount('type') > 0;
-        }
+        const nonPersistentQuery = query.clone();
+        nonPersistentQuery.deleteKeyValue('?type');
+        const clearButton = nonPersistentQuery.queryCount() > 0 && query.queryCount('?type') > 0;
```
forresttanaka, line 360: Oh! Yeah that works too. I forgot the way I usually do this is to use url.parse(this.props.context['@id']).query instead of .search. They give you the same thing, but .query doesn’t have the question mark. I wonder if QueryString should check for a leading question mark and ignore it, so you can use either?

forresttanaka (dismissed):
Cool — like it!

src/encoded/static/components/summary.js
emro: That could be handy! Or, I could remember to use `.query` instead.

emro:
Latest updates:
(1) Fix hover on inset images
(2) Fix weird hover behavior due to unused svg paths
(3) Add colon to intestine

forresttanaka (approved):

PR #3287, opened 2020-04-28
ENCD-5225-display-cropped-read-length-and-tolerance

khzlin:


src/encoded/schemas/file.json
```diff
@@ -271,6 +271,15 @@
                 }
             }
         },
+        "cropped_read_length_tolerance": {
+            "comment": "Only bam files with cropped_read_length specified can have cropped_read_length_tolerance.",
```
zoldello: does "bam" supposed to be capitalized ("BAM")?

src/encoded/static/components/filegallery.js
```diff
@@ -2815,6 +2815,20 @@ const FileDetailView = function FileDetailView(node, qcClick, auditIndicators, a
                         </div>
                     : null}
 
+                    {selectedFile.cropped_read_length !== undefined ?
```
zoldello, line 3129: Why are you checking specifically for undefined?  Why not:

{selectedFile.cropped_read_length > 0 ?

src/encoded/static/components/filegallery.js
```diff
@@ -2815,6 +2815,20 @@ const FileDetailView = function FileDetailView(node, qcClick, auditIndicators, a
...
+                        <div data-test="croppedreadlength">
+                            <dt>Cropped read length</dt>
+                            <dd>{selectedFile.cropped_read_length}</dd>
+                        </div>
+                    : null}
+
+                    {selectedFile.cropped_read_length_tolerance !== undefined ?
```
zoldello, line 3136: Why are you checking specifically for undefined? Why not:

{selectedFile.ropped_read_length_tolerance > 0 ?

zoldello (changes requested):
I made two comments and I suspect that they will result in changes.

src/encoded/static/components/filegallery.js
khzlin: I am checking to see if cropped_read_length_tolerance has a value to be displayed.

src/encoded/schemas/file.json
khzlin: I don't think it has to be.

zoldello (dismissed):
Changes look good. My comments are nothing significant.

zoldello (approved):
This was a rebase on dev. I see the rebased commit entry in the history. A new demo was built.

PR #3289, opened 2020-04-29
ENCD-5159-show-donor-on-ref-page

zoldello:


zoldello:
Development Notes
There are no references with donors (https://test.encodedcc.org/search/?type=Reference&donor=*). So I added one. Here is an example: https://encd-5159-e47596719-phil.demo.encodedcc.org/references/ENCSR792RVV/

emro (approved):
Looks great!

PR #3290, opened 2020-05-01
ENCD-5136-hide-nih-in-facets-on-0-results

zoldello:


forresttanaka:
This has given me something to think about. I think there’s a little weakness in your fix because of a weakness in getKeyValues. You can check for a=b with `getKeyValue('a')` or `getKeyValue('a', false)`, and you can check for a!=b with `getKeyValue('a', true)`. But what if you wanted to check for both cases? Right now you’d have to call getKeyValue twice, with the `negative` parameter true in one case.

I wonder, to keep consistency with existing code using getKeyValue, if we should make false or undefined mean =, true means !=, and null means either case? Or should we just call it twice, in which case I think you need to call it twice here.

zoldello:
@forresttanaka 
Can we consider a 3rd option? How about a method called "getPresentKeyValues" where internally it checked for both a=b and a!=b?

forresttanaka:
I like that idea actually. While keeping the number of methods down is nice, your idea has the advantage of being less confusing I think, and it’d be easy to implement since _it_ can call getKeyValues twice.

zoldello:
@forresttanaka 
Thanks for the catch. I will implement the change.

forresttanaka (approved):
Nice! Thanks for adding that query-string method.

PR #3291, opened 2020-05-02
ENCD-5259-fix-validation-fail-reference-objects

iyoungworth:


gabdank (changes requested):
I think that you are theoretically invalidating references by this change. You probably need to write schema upgrade and test it.

gabdank (approved):

PR #3292, opened 2020-05-04
ENCD-5258-add-file-upgrade

jenjou:


gabdank (changes requested):
I think you should add to the "pre-existing" notes. If there was something in "notes" you are going to overwrite if you do it the way it is written right now.

src/encoded/schemas/changelogs/file.md
```diff
@@ -1,5 +1,8 @@
 ## Changelog for file.json
 
+### Schema version 20
+* Reimplemented *run_type* dependency on fastq files.
```
gabdank: I would probably say "added upgrade required due to re-implementation of the run_type dependency

gabdank (changes requested):
the change log re-wording and the upgrade "notes" update, see my comments

gabdank (approved):

PR #3293, opened 2020-05-04
ENCD-5264 fix error on pages missing type

emro:
Some pages do not have an @type, so when I added a class name based on @type (so that I could make some pages, like the summary page with the body map facet, wider), these pages returned an error. Adding a check for @type eliminates this error.

### DIFF
--- a/src/encoded/static/components/app.js
+++ b/src/encoded/static/components/app.js
@@ -1247,7 +1247,7 @@ class App extends React.Component {
                             <Provider store={cartStore}>
                                 <div id="layout">
                                     <Navigation isHomePage={isHomePage} />
-                                    <div id="content" className={`container ${context['@type'].join(' ')}`} key={key}>
+                                    <div id="content" className={context['@type'] ? `container ${context['@type'].join(' ')}` : 'container'} key={key}>
                                         {content}
                                     </div>
                                     {errors}


src/encoded/static/components/app.js
```diff
@@ -1247,7 +1247,7 @@ class App extends React.Component {
                             <Provider store={cartStore}>
                                 <div id="layout">
                                     <Navigation isHomePage={isHomePage} />
-                                    <div id="content" className={`container ${context['@type'].join(' ')}`} key={key}>
+                                    <div id="content" className={context['@type'] ? `container ${context['@type'].join(' ')}` : 'container'} key={key}>
```
zoldello, line 1250: className={`container  ${context['@type'] ? context['@type'].join(' ') : ''`} }

Here is an alternate way

zoldello (dismissed):
The code looks good. My alternate way is just me thinking aloud in a sense and it not significantly different from your approach.

emro:
I thought about writing it that way and then I switched partway through, and didn't have a good reason for doing that. I think it shouldn't make much of a difference, but if the way you suggested is more conventional, I'm happy to push an update.

zoldello:
@emro Your call. I would recommend leaving it as-is. "Tomato" and "Tommatooo" are still the same fruit. The code looks fine as-is.

zoldello (dismissed):
I am approving it assuming you want to leave it as-is. If you do decide to change it, I will know from the new commit and return to this PR.

zoldello (approved):
Looks fine.

zoldello (approved):
Looks good.

zoldello:
@emro  Please ignore the re-request for code review.

PR #3294, opened 2020-05-05
ENCD-4913-fix-filtered-url-in-matrix

zoldello:


src/encoded/static/components/matrix_experiment.js
```diff
@@ -106,6 +106,7 @@ const convertExperimentToDataTable = (context, getRowCategories, getRowSubCatego
         colTitleMap[colCategoryBucket.key] = colIndex;
         return colCategoryBucket.key;
     });
+    const searchUrl = '/search/?type=Experiment&status=released';
```
forresttanaka: I feel uncomfortable hard-coding this. I’d prefer to do something like taking the search_base and removing any part of it that matches a column property before appending that column property.

forresttanaka (changes requested):
I tried a local build as well as your demo, but I still see the repeated properties. For example, https://encd-4913-e2926704d-phil.demo.encodedcc.org/search/?type=Experiment&status=released&assay_title=TF+ChIP-seq. If you click the TF ChIP-seq header column, I still get two of the same assay_title properties in the query string.

forresttanaka:
I’m also getting a key error in the Javascript console, at least when running locally (I didn’t check when I had your demo up).

”Encountered two children with the same key, `Histone ChIP-seq`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.“

forresttanaka (dismissed):
OK, though I thought you were going to use the QueryString module to do this, since that’s what it does. I’m still not clear what Ben wants with stripping other selected facets, because it seems strange that going from matrix to search is different than going from search to matrix, but Jennifer can figure that out I guess.

forresttanaka (approved):
You can also use nested string templates, which is another option to the .join('').

zoldello:
@forresttanaka <strike> Good point. I did not use string template because it seemed awkward to have a string template inside a string template. I will consider it for next time. </strike>  &nbsp; You are right. I will do that next time.

PR #3295, opened 2020-05-06
ENCD-5260-cart-auto-save

forresttanaka:
In ENCD-2091 I changed some property names from “*Elements” to “*Datasets” to be clearer. Unfortunately I didn’t take this far enough down, which prevented the “Add items to cart’ button from ever appearing. I’ve now taken that change down to the component that handles this merging of carts.


emro (approved):
Code looks good, and works well

PR #3298, opened 2020-05-07
ENCD-5265-correct-stem-cell-matrix-links

zoldello:


zoldello:
@emro PR updated.

src/encoded/static/components/matrix_sescc_stem_cell.js
```diff
@@ -466,18 +467,31 @@ class MatrixPresentation extends React.Component {
...
                             const elementClass = formatPebbleNameToCssClassFriendly(name);
                             const element = document.querySelector(`.${elementClass} th`);
+                            const text = document.querySelector(`.${elementClass} th .subcategory-row-text`);
 
                             if (element) {
                                 element.style.backgroundColor = 'white';
+
+                                if (text) {
+                                    text.style.color = 'black';
+                                }
```
emro, line 494: This works so well!

Did you tint the color so that it would be right format? I think you could have tinted with 0 if you didn't want to change the color, but this is also great.

Just as a comment, and it's barely even worth mentioning, I would have used '#000' for 'black' and '#fff' for 'white' here on lines 490 and 493 just because that's what you have on 474, but this super doesn't matter.

emro (dismissed):
Looks good and works well!

Just to note here as a record, we had a meeting and a couple conversations on slack about when the links should be updated to account for filtering of the matrix, and I think everyone is in agreement that this update should be in a future ticket.

src/encoded/static/components/matrix_sescc_stem_cell.js
zoldello: @emro 
I added a small tint but nothing significant. Are you okay with the small tint? Are you okay if I leave the "black" and "white" instead of "000" and "fff"?

emro: Yes! That is all good.

emro (approved):

PR #3299, opened 2020-05-07
ENCD-4614-no-assembly-graph

forresttanaka:
I now bypass the assembly checks for displaying the graph specifically for RNA Bind-n-seq experiments. This might get adjusted in the future, but this works for now.

### DIFF
--- a/src/encoded/static/components/filegallery.js
+++ b/src/encoded/static/components/filegallery.js
@@ -1390,6 +1390,7 @@ export function assembleGraph(files, highlightedFiles, dataset, options, loggedI
     let matchingFiles = {}; // All files that match the current assembly/annotation, keyed by file @id
     const fileQcMetrics = {}; // List of all file QC metrics indexed by file @id
     const allPipelines = {}; // List of all pipelines indexed by step @id
+    const allowNoAssembly = dataset.assay_term_name === 'RNA Bind-n-Seq';
     files.forEach((file) => {
         // allFiles gets all files from search regardless of filtering.
         allFiles[file['@id']] = file;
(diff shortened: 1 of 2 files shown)


forresttanaka:
Demo: ~https://encd-4614-no-assembly-graph-pr1-fytanaka.demo.encodedcc.org/~

src/encoded/tests/data/inserts/file.json
```diff
@@ -3407,5 +3407,105 @@
         "file_format_type": "bed3+",
         "accession": "ENCFF880CDH",
         "file_size": 13224
+    },
```
zoldello, line 3410: Good idea to add data locally to see the changes. I think this approach should be made standard in code reviews.

src/encoded/static/components/filegallery.js
```diff
@@ -1390,14 +1390,15 @@ export function assembleGraph(files, highlightedFiles, dataset, options, loggedI
     let matchingFiles = {}; // All files that match the current assembly/annotation, keyed by file @id
     const fileQcMetrics = {}; // List of all file QC metrics indexed by file @id
     const allPipelines = {}; // List of all pipelines indexed by step @id
+    const allowNoAssembly = dataset.assay_term_name === 'RNA Bind-n-Seq';
```
zoldello, line 1393: It is there any chance someone will use a different case like "rNA bind-n-seq"? I tend to make both hands of comparators lower case to take this into account.

zoldello (approved):
The change looks fine. All you did was allow a specific assembly-less type to be displayed. My comments are not significant.

src/encoded/static/components/filegallery.js
forresttanaka: There’s no danger of that because assay_term_name has a controlled vocabulary, so people can only choose from a set of pre-defined options.

PR #3300, opened 2020-05-08
ENCD-5243 fix crash on date submitted

emro:
Two main changes:
(1) create date selector facet for both date_released and date_submitted and if data for both exists, only display one
(2) if data does not exist for either date_released or date_submitted, disable that radio button on the date selector facet

### DIFF
--- a/src/encoded/static/components/facets/defaults.js
+++ b/src/encoded/static/components/facets/defaults.js
@@ -413,26 +413,85 @@ export class DefaultDateSelectorFacet extends React.Component {
 
     render() {
         const { facet, results, queryString } = this.props;
-        const searchBase = `?${queryString}&`;
+        let searchBase = `?${queryString}&`;
         const field = this.state.activeFacet;
         const activeFacet = results.facets.filter(f => f.field === this.state.activeFacet)[0];
+        let disableDateReleased = false;
+        let disableDateSubmitted = false;
+        // filterFlag is true to indicate that we need to display filters
+        let filterFlag = false;
+        // missingField indicates which field to disable (date_released or date_submitted) if one field is not disabled
+        let missingField = null;
 
-        // We have two facets registered for the date selector facet, one for date_released and one for date_submitted, but we only want to display one
-        // "displayField", when false, hides the duplicate facet when needed (most of the time)
-        // If we have date_released or date_submitted data only, then only one facet will be displayed and we don't need to hide one
-        // If we have both, we will just display the facet for date_released
-        if (facet.field === 'date_submitted' && results.facets.filter(f => f.field === 'date_released').length > 0) {
+        // Check which of date released and date submitted might be disabled, either for lack of data or because of !=* filter
+        if ((queryString.indexOf('date_released!=*') > -1) || !(results.facets.filter(f => f.field === 'date_released').length > 0)) {
+            disableDateReleased = true;
+            missingField = 'date_released';
+        }
+        if ((queryString.indexOf('date_submitted!=*') > -1) || !(results.facets.filter(f => f.field === 'date_submitted').length > 0)) {
+            disableDateSubmitted = true;
+            missingField = 'date_submitted';
+        }
+
+        // If date released and/or date submitted has a !=* filter, determine link(s) to delete the filter(s)
+        let deleteSubmittedFilter = '';
+        let deleteReleasedFilter = '';
+        let missingFilter = '';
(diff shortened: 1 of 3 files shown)


src/encoded/static/components/facets/defaults.js
```diff
@@ -441,9 +462,9 @@ export class DefaultDateSelectorFacet extends React.Component {
             }
         }
 
-        if (((activeFacet.terms.length > 0) && activeFacet.terms.some(term => term.doc_count)) || (field.charAt(field.length - 1) === '!')) {
+        if ((activeFacet && (activeFacet.terms.length > 0) && activeFacet.terms.some(term => term.doc_count)) || (field.charAt(field.length - 1) === '!')) {
             return (
-                <div className={`facet date-selector-facet ${facet.field === 'date_released' ? 'display-date-selector' : ''}`}>
+                <div className={`facet date-selector-facet ${facet.field === 'date_released' ? 'display-date-selector' : ''} ${!displayField ? 'hide-facet' : ''}`}>
```
forresttanaka: I think once you find out `displayField` is false, you can immediately bail at the if statement above, instead of doing all the calculations and rendering the date_submitted facet, and then hiding it.

src/encoded/static/components/facets/defaults.js
```diff
@@ -416,6 +417,26 @@ export class DefaultDateSelectorFacet extends React.Component {
...
+        let displayField = true;
+        if (facet.field === 'date_submitted' && results.facets.filter(f => f.field === 'date_released').length > 0) {
+            displayField = false;
+        }
+
+        // If we are missing data for "date_submitted" or "date_released", we want to disable that radio button
+        // "missingField" indicates if there is a field to disable, and if so, which one
+        let missingField = null;
+        // Check if we are displaying date_submitted facet and there is no date_released data
+        if (facet.field === 'date_submitted' && !(results.facets.filter(f => f.field === 'date_released')).length > 0) {
```
forresttanaka: Since you changed the `if` statement above to bail at that stage (re: my comment below), then at this stage if facet.field is date_submitted, you’ll already know no date_released facet exists, so I think you don’t need the filter loop unless I’m not seeing a case where that filtering is still needed. I’ll leave it up to you, as I’m sure this works, but might not be necessary.

forresttanaka (changes requested):
Looks like a good way to do it. I think with my second suggestion you can make it more efficient.

src/encoded/static/components/facets/defaults.js
emro: This is much simpler!

src/encoded/static/components/facets/defaults.js
```diff
@@ -416,6 +417,24 @@ export class DefaultDateSelectorFacet extends React.Component {
...
+        // If we have date_released or date_submitted data only, then only one facet will be displayed and we don't need to hide one
+        // If we have both, we will just display the facet for date_released
+        if (facet.field === 'date_submitted' && results.facets.filter(f => f.field === 'date_released').length > 0) {
+            return null;
+        }
+        // If we are missing data for "date_submitted" or "date_released", we want to disable that radio button
+        // "missingField" indicates if there is a field to disable, and if so, which one
+        let missingField = null;
+        // Check if we are displaying date_submitted facet and there is no date_released data
+        if (facet.field === 'date_submitted' && !(results.facets.filter(f => f.field === 'date_released')).length > 0) {
```
forresttanaka: Technically, since you already tested for both date_submitted and date_released existing and then returned null in that case, then if you get to this line and `facet.field` is date_submitted, you’ll know that no date_released exists, guaranteed.

forresttanaka (dismissed):
One optional change, but I’m not 100% sure I’m analyzing that situation right.

src/encoded/static/components/facets/defaults.js
emro: I noticed this could be simplified at some point and then forgot to go back and take care of it. Thank you for reminding me!

forresttanaka (dismissed):
Cool!

emro:
QA illuminated a problematic oversight on the date selection facet: when the !=* filter is applied to date_released or date_submitted, we need to disable the corresponding radio button and display the filter as selected (so that the user can delete it easily if desired).

forresttanaka (approved):
Wow, this got really complicated with all the different cases you need to take care of. Looks good to me.

PR #3302, opened 2020-05-11
ENCD-5269-audit-annotation-file-derived-from-revoked

jenjou:


src/encoded/audit/annotation.py
```diff
@@ -19,3 +20,20 @@ def audit_annotation_organism(value, system):
...
+def audit_annotation_derived_from_revoked(value, system):
+    '''
+    Annotations with files derived from a revoked file should be flagged with an audit.
+    '''
+    request = system.get('request')
+    for file in value.get('files'):
+        if 'derived_from' in file:
+            for f in file['derived_from']:
+                parent = request.embed(f + '@@object')
+                if parent['status'] == 'revoked':
```
gabdank: I thikn you should use 
`parent = request.embed(f + '@@object?skip_calculated=true')`
instead.

gabdank (changes requested):

gabdank (approved):

PR #3305, opened 2020-05-12
ENCD-4927-internal-audit

forresttanaka:
I simply hadn’t passed session_properties to the GM-rendering component, so it never knew people had signed in.

### DIFF
--- a/src/encoded/static/components/genetic_modification.js
+++ b/src/encoded/static/components/genetic_modification.js
@@ -647,7 +647,7 @@ GeneticModificationComponent.defaultProps = {
 };
 
 const GeneticModificationInternal = (props, reactContext) => (
-    <GeneticModificationComponent {...props} session={reactContext.session} />
+    <GeneticModificationComponent {...props} session={reactContext.session} sessionProperties={reactContext.session_properties} />
 );
 
 GeneticModificationInternal.propTypes = {


src/encoded/static/components/genetic_modification.js
```diff
@@ -647,7 +647,7 @@ GeneticModificationComponent.defaultProps = {
 };
 
 const GeneticModificationInternal = (props, reactContext) => (
-    <GeneticModificationComponent {...props} session={reactContext.session} />
+    <GeneticModificationComponent {...props} session={reactContext.session} sessionProperties={reactContext.session_properties} />
```
zoldello, line 650: @forresttanaka  Let's talk about this briefly.

zoldello (commented):
Self-note:
The point of this PR is to give sessionProperties a value at: https://github.com/ENCODE-DCC/encoded/blob/ENCD-4927-internal-audit/src/encoded/static/components/genetic_modification.js#L544

Do not confused line like https://github.com/ENCODE-DCC/encoded/blob/ENCD-4927-internal-audit/src/encoded/static/components/genetic_modification.js#L686 with this change.

zoldello (approved):
Looks good. I had some question that I got found answers to after reading the code again.

PR #3306, opened 2020-05-12
ENCD-5262 add skinny tracks

emro:
This ticket has several goals pertaining to the display of tracks on the Valis browser:
(1) Labels are customized for cart and file gallery views:
     On the cart, the labels for the collapsed tracks should be: target, assay title, biosample
     On the file gallery, the labels for the collapsed tracks should be: accession (replicate number)
(2) bigBed files are now expandable / collapsible ...
(3) ... except for bigBed files with two rows of data ('bedrnaelements', 'peptidemapping', 'bedexonscore', 'bed12', 'bed9') which are automatically expanded and are not collapsible
(4) We are using an updated Valis library which allows for specifying whether or not labels are displayed, and we do not display bigBed track labels except for the gene track
(5) Default track height is significantly smaller and expanded height is quite tall and we've eliminated any medium height
(6) Some of the cart labels are somewhat long (they can be up to 3 lines) so for those cases, the skinny track view is a little taller - we use an estimate of how many characters fit on a line in the label, and calculate from that how many lines we need (it mostly works)

### DIFF
--- a/src/encoded/static/components/genome_browser.js
+++ b/src/encoded/static/components/genome_browser.js
@@ -157,7 +157,7 @@ const TrackLabel = ({ file, label, long }) => {
                     {file.biosample_ontology && file.biosample_ontology.term_name ? <span>{file.biosample_ontology.term_name}</span> : null}
                     {long ?
                         <React.Fragment>
-                            <li><a href={file['@id']} className="gb-accession">{file.accession}<span className="sr-only">{`Details for file ${file.accession}`}</span></a></li>
+                            <li><a href={file['@id']} className="gb-accession">{file.title}<span className="sr-only">{`Details for file ${file.title}`}</span></a></li>
                             <li>{file.output_type}</li>
                             <li>{`rep ${biologicalReplicates}`}</li>
                         </React.Fragment>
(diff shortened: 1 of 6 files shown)


src/encoded/static/components/genome_browser.js
```diff
@@ -147,29 +146,48 @@ function mapGenome(inputAssembly) {
...
-                : null}
-            </ul>
+            {(label === 'cart') ?
+                <ul className="gb-info">
+                    {file.target ? <span>{file.target.label}, </span> : null}
+                    {file.assay_term_name ? <span>{file.assay_term_name}, </span> : null}
+                    {file.biosample_ontology && file.biosample_ontology.term_name ? <span>{file.biosample_ontology.term_name}</span> : null}
+                    {long ?
+                        <React.Fragment>
+                            <li><a href={file['@id']} className="gb-accession">{file.accession}<span className="sr-only">{`Details for file ${file.accession}`}</span></a></li>
```
forresttanaka: Oh, sorry, that was my mistake in the original code. We’re not supposed to ever use file.accession. We’re supposed to use file.title. Usually they’re the same, but on occasion file.accession is undefined, while file.title has the real identifier. I keep forgetting about that. I think file is the only accessioned object where you can’t use the accession property.

forresttanaka (changes requested):
That looks really good. The one change is because of something I did wrong.

src/encoded/static/components/genome_browser.js
emro: Just for the cart? Or we're not supposed to use file.accession anywhere?

forresttanaka: Not supposed to use file.accession anywhere. We’ve had several bugs in the past about this, though this happens only in unusual cases where we use lab IDs instead of accessions.

forresttanaka (dismissed):
It’ll be nice to see this in dev.

emro:
Yeah! Although I'm a little afraid of finding more cases where my approximate math for determining track height for multi-line labels will not be accurate....

forresttanaka (approved):
Just the valis version update.

PR #3307, opened 2020-05-12
ENCD-5277 Update amis v100 and typos

caseylitton:


hitz (dismissed):

src/encoded/commands/deploy.py
```diff
@@ -812,6 +812,7 @@ def main():
         # helps vars for release and building amis
         for helper_var in helper_vars:
             print(helper_var)
+    print('Done')
```
hitz, line 813: Got tired of AWS leaving you hanging?

hitz (dismissed):

src/encoded/commands/deploy.py
caseylitton: The previous print line would not print...I think it has to do with my terminal config but printing Done seems okay.

hitz (approved):

PR #3308, opened 2020-05-13
ENCD-5252-add-index-reads-audit

iyoungworth:
minor-adjustments

### DIFF
--- a/src/encoded/audit/file.py
+++ b/src/encoded/audit/file.py
@@ -524,7 +524,8 @@ def audit_file_index_of(value, system):
         return
     if 'index_of' in value:
         index_exp = value['dataset']['@id']
-        fastq_with_expt = {}
+        files = []
+        expts = []
         indexed_files = 0
         error_files = 0
         run_type = set()
@@ -533,11 +534,10 @@ def audit_file_index_of(value, system):
             indexed_files += 1
             run_type.add(each['run_type'])
             if index_exp != fastq_exp:
-                fastq_with_expt[each['@id']] = each['dataset']
+                files.append(each['@id'])
+                expts.append(each['dataset'])
                 error_files += 1
         if error_files >= 1:
-            files = tuple(sorted(fastq_with_expt.keys()))
-            expts = tuple(sorted(fastq_with_expt.values()))
             files_links = ', '.join(audit_link(path_to_text(m), m) for m in files)
             expts_links = ', '.join(audit_link(path_to_text(n), n) for n in expts)
             detail = (
--- a/src/encoded/audit/file.py
+++ b/src/encoded/audit/file.py
@@ -513,6 +513,53 @@ def audit_file_matching_md5sum(value, system):
     yield AuditFailure('matching md5 sums', detail, level='WARNING')
 
 
+def audit_file_index_of(value, system):
+    '''
+    Files with output_type: index reads may specify that they are index_of
+    either 1 SE or 2 PE files. The SE/PE files must be from the same
+    experiment as index, other combinations are flagged as ERROR.
+    https://encodedcc.atlassian.net/browse/ENCD-5252
+    '''
+    if value['output_type'] != 'index reads':
+        return
+    if 'index_of' in value:
+        index_exp = value['dataset']['@id']
+        fastq_ids = set()
+        exp_ids = set()
+        indexed_files = 0
+        error_files = 0
+        run_type = set()
+        for each in value['index_of']:
+            fastq_exp = each['dataset']
+            indexed_files += 1
+            run_type.add(each['run_type'])
+            if index_exp != fastq_exp:
+                fastq_ids.add(each['@id'])
+                exp_ids.add(each['dataset'])
+                error_files += 1
+        if error_files >= 1:
+            fastq_ids_links = [audit_link(path_to_text(m), m) for m in fastq_ids]
+            exp_ids_links = [audit_link(path_to_text(n), n) for n in exp_ids]
+            detail = ('Index file {} is from experiment {}, but is used as an index for file(s) {} from experiment(s) {}'.format(
+                                audit_link(path_to_text(value['@id']), value['@id']),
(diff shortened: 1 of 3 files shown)


src/encoded/tests/test_audit_file.py
```diff
@@ -912,3 +912,41 @@ def test_audit_self_matching_md5sum(testapp, file7):
         errors_list.extend(errors[error_type])
     assert any(error['category'] == 'inconsistent matching_md5sum'
                for error in errors_list)
+
+
+def test_audit_incorrect_index(testapp, fastq_index, single_fastq_indexed):
```
gabdank: I think you should also include a case where index_of is associated correctly and make sure the audit is not triggered.

src/encoded/audit/file.py
```diff
@@ -513,6 +513,53 @@ def audit_file_matching_md5sum(value, system):
...
+            fastq_ids_links = [audit_link(path_to_text(m), m) for m in fastq_ids]
+            exp_ids_links = [audit_link(path_to_text(n), n) for n in exp_ids]
+            detail = ('Index file {} is from experiment {}, but is used as an index for file(s) {} from experiment(s) {}'.format(
+                                audit_link(path_to_text(value['@id']), value['@id']),
+                                audit_link(path_to_text(value['dataset']['@id']), value['dataset']['@id']),
+                                ', '.join(fastq_ids_links),
+                                ', '.join(exp_ids_links)
+                                )
+                    )
+            yield AuditFailure('incorrect index file', detail, level='ERROR')
```
gabdank: for all three audit failures you created here you gave different audit categories (incorrect index file, mismatched index file, fastq missing index) while it seems more clear - it also adds 3 entries to already very long list of error audits facet - I think you should make one category ("inconsistent index reads file") or something like that and then the details could be different under it depending on the specific case, but all of them will fall under the same Audit category in the facet.

src/encoded/audit/file.py
```diff
@@ -513,6 +513,53 @@ def audit_file_matching_md5sum(value, system):
...
+
+        if indexed_files > 1 and 'single-ended' in run_type:
+            detail = ('Index file {} is incorrectly specified for both single- and paired-end fastq files'.format(
+                audit_link(path_to_text(value['@id']), value['@id'])))
+            yield AuditFailure('mismatched index file', detail, level='ERROR')
+
+        elif indexed_files == 1 and 'single-ended' not in run_type:
```
gabdank: I would change the condition to be "and "paired-ended" in run_type" instead of negating single-ended

src/encoded/audit/file.py
```diff
@@ -513,6 +513,53 @@ def audit_file_matching_md5sum(value, system):
...
+                fastq_ids.add(each['@id'])
+                exp_ids.add(each['dataset'])
+                error_files += 1
+        if error_files >= 1:
+            fastq_ids_links = [audit_link(path_to_text(m), m) for m in fastq_ids]
+            exp_ids_links = [audit_link(path_to_text(n), n) for n in exp_ids]
+            detail = ('Index file {} is from experiment {}, but is used as an index for file(s) {} from experiment(s) {}'.format(
+                                audit_link(path_to_text(value['@id']), value['@id']),
+                                audit_link(path_to_text(value['dataset']['@id']), value['dataset']['@id']),
+                                ', '.join(fastq_ids_links),
```
gabdank: I think we agreed to use f-strings for audit details and in general for strings in the code base. And I think that if you have multiple files that lead to multiple experiments, your details will be incorrect because file IDs in set will not necessarily be sorted like experiment IDs in the set.

gabdank (changes requested):
comments outline the chnages

src/encoded/tests/test_audit_file.py
iyoungworth: I've added a test for this, thank you

src/encoded/audit/file.py
iyoungworth: All are now under 'inconsistent index file'

src/encoded/audit/file.py
iyoungworth: Made this change

src/encoded/audit/file.py
iyoungworth: I reworked each detail into an f-string. I tried creating sorted tuples from a dict with the file/experiment pairs, and I think this is working to keep the order the same for each (I'm not sure how else to test it, but I triggered the audit with pairs of files several times in a local demo and they appeared in the correct order)

iyoungworth: I think because lists in python are ordered, if I append the new items to the end, the order should match in separate lists for files and fastqs- I returned to using two lists rather than dict used to then generate sorted lists.

src/encoded/audit/file.py
```diff
@@ -513,6 +513,55 @@ def audit_file_matching_md5sum(value, system):
...
+            fastq_exp = each['dataset']
+            indexed_files += 1
+            run_type.add(each['run_type'])
+            if index_exp != fastq_exp:
+                files.append(each['@id'])
+                expts.append(each['dataset'])
+                error_files += 1
+        if error_files >= 1:
+            files_links = ', '.join(audit_link(path_to_text(m), m) for m in files)
+            expts_links = ', '.join(audit_link(path_to_text(n), n) for n in expts)
```
gabdank: item[0] for item in file-expt-list
item[1] for item in file-expt-list

src/encoded/audit/file.py
```diff
@@ -513,6 +513,55 @@ def audit_file_matching_md5sum(value, system):
...
+    if value['output_type'] != 'index reads':
+        return
+    if 'index_of' in value:
+        index_exp = value['dataset']['@id']
+        files = []
+        expts = []
+        indexed_files = 0
+        error_files = 0
+        run_type = set()
+        for each in value['index_of']:
```
gabdank: each -> indexed_file

gabdank (approved):

PR #3311, opened 2020-05-15
ENCD-5261-copy-schema

forresttanaka:
* I used to use the brace editor that we use for the object-editing forms in read-only mode for these colorized schema displays in the Raw tab. But brace intercepts selections, making selection difficult and copying unreliable. This change removes brace  for the profiles page and instead just uses `<pre>` and JSON.stringify to display the schemas without colorization, meaning these schemas are now server rendered. brace is still used for the object-editing forms, so we need to keep that npm package.
* I hadn’t heard before that people like to copy the schema off this page, so I added an absolutely positioned button to copy the entire schema to the clipboard, or the selected portion if the user selects a portion of the schema. The title of the button changes depending on which case applies.
* Handling the two cases involves event listeners, and that handling takes quite a bit of this branch’s changes.
* The Add button that appears when you have logged in used to be absolutely positioned in the panel header, but I changed that to use flex box here. I had initially put the new Copy button next to it so I needed this change. Now the Copy button is elsewhere but this flexbox change still seemed better so I kept it.
* Parts of schemas get displayed on the Formatted tab, so the same mechanism gets used there, including the copy button.

https://encd-5261-copy-schema-pr1-fytanaka.demo.encodedcc.org/

### DIFF
--- a/src/encoded/static/components/schema.js
+++ b/src/encoded/static/components/schema.js
@@ -51,47 +51,6 @@ const linkedTerms = ['properties'];
 
 
 /**
- * Create a brace read-only editor instance that goes into the HTML element with the given ID and
- * displaying the given schema term.
- *
- * @param {*} elementId - HTML element ID to draw the editor into.
- * @param {*} term - Schema term object to display.
- */
-function createJsonDisplay(elementId, term) {
-    return new Promise((resolve) => {
-        require.ensure([
-            'brace',
-            'brace/mode/json',
-            'brace/theme/katzenmilch',
-        ], (require) => {
-            const ace = require('brace');
-            require('brace/mode/json');
-            require('brace/theme/katzenmilch');
-            const value = JSON.stringify(term, null, 4);
-            const editor = ace.edit(elementId);
-            const session = editor.getSession();
-            session.setMode('ace/mode/json');
-            editor.setTheme('ace/theme/katzenmilch');
-            editor.setValue(value);
-            editor.setOptions({
-                maxLines: 500,
-                minLines: 1,
-                readOnly: true,
-                highlightActiveLine: false,
-                highlightGutterLine: false,
-                showGutter: false,
-                showPrintMargin: false,
-            });
-            editor.clearSelection();
-            editor.textInput.getElement().disabled = true;
-            editor.renderer.$cursorLayer.element.style.opacity = 0;
-            resolve(editor);
-        }, 'brace');
-    });
-}
-
-
-/**
  * Convert the contents of the `id` field of a schema to a schema name.
  *
  * @param {*} schemaId - `id` property from schema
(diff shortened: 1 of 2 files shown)


src/encoded/static/scss/encoded/modules/_schema.scss
```diff
@@ -95,23 +95,46 @@
...
 
     @at-root #{&}__title {
         font-weight: bold;
     }
-}
\ No newline at end of file
+}
+
+// Styles for the raw JSON schema display.
+$raw-schema-height: 8000px;
```
emro, line 107: Just curious, is this an arbitrarily large number? A height which we don't expect any JSON schema to exceed? Or was it derived from something?

src/encoded/static/components/schema.js
```diff
@@ -471,38 +386,117 @@ DisplayObject.defaultProps = {
...
+    };
+
+    return (
+        <div className="raw-schema__controls">
+            <button className="btn btn-info btn-xs" onClick={copyHandler}>
+                {textHasSelection ? <>Copy selected JSON</> : <>Copy all JSON</>}
+            </button>
+        </div>
+    );
+};
```
emro, line 427: This is a little more of a usability comment, but since you're not highlighting all the JSON text when the user clicks on "Copy all JSON", maybe you could make it so that the button style changes when you've clicked on it to show the user that the text is in the clipboard?

src/encoded/static/components/schema.js
```diff
@@ -471,38 +386,117 @@ DisplayObject.defaultProps = {
...
+        if (!textHasSelection) {
+            const selection = window.getSelection();
+            selection.removeAllRanges();
+        }
+    };
+
+    return (
+        <div className="raw-schema__controls">
+            <button className="btn btn-info btn-xs" onClick={copyHandler}>
+                {textHasSelection ? <>Copy selected JSON</> : <>Copy all JSON</>}
```
emro, line 423: I've been using the full <React.Fragment> but maybe I should switch to using the <> shorthand.

src/encoded/static/components/schema.js
```diff
@@ -471,38 +386,117 @@ DisplayObject.defaultProps = {
...
+    // that RawObjectControls can rerender at mount.
+    const rawSchemaRef = (element) => {
+        if (element) {
+            setRawSchemaElement(element);
         }
-    }
+    };
+
+    React.useEffect(() => {
+        // React to text-selection changes anywhere on the page, partly because text selections
```
emro, line 474: Using "React" here is mildly confusing because in this context I think you mean "respond to" but this isn't worth changing, just a comment.

src/encoded/static/components/schema.js
```diff
@@ -471,38 +386,117 @@ DisplayObject.defaultProps = {
...
+
+/**
+ * Display a complete schema object as text. At mount rawSchemaRef.current gets mutated with
+ * <pre>'s DOM node, which we need to handle text selections within it.
+ */
+const DisplayRawObject = ({ schema, term }) => {
+    // True if user has selected raw schema text.
+    const [textHasSelection, setTextHasSelection] = React.useState(false);
+    // Raw schema DOM element; `rawSchemaRef.current` doesn't cause the rerender we need.
+    const [rawSchemaElement, setRawSchemaElement] = React.useState(null);
```
emro, line 450: Using React hooks here seems very helpful.

emro (commented):
Looks good and works well! Just a few optional comments.

src/encoded/static/scss/encoded/modules/_schema.scss
forresttanaka: That was an arbitrary limit set with this page initially two years ago or whenever that was. Lots (most?) schemas end up taller than this, and they scroll vertically. This was just to not have these pages end up stupendously long.

src/encoded/static/components/schema.js
forresttanaka: I thought about highlighting the text when you copied (partly because I accidentally had the text selected when you clicked Copy at first) but thought that might be more confusing (now it’s selected, does that mean I should copy it with command-C, or did it already copy?). But I notice when you use the copy button on Github to clone a repo, it doesn’t do nor say anything, so I decided to go with that. Not that Github is necessarily right, but I figure they might do user testing. I don’t know.

src/encoded/static/components/schema.js
forresttanaka: I use `<React.Fragment>` mostly, but for inline JSX I use `<>` since `<React.Fragment>` would take up so much space in that case. I think `<>` on its own line looks distasteful — just a weird personal thing.

src/encoded/static/components/schema.js
forresttanaka: Oh yeah, that’s true.

src/encoded/static/components/schema.js
forresttanaka: I can’t go back to class components. Hooks generally has made code simpler — sometimes a lot, sometimes slightly.

emro (approved):

src/encoded/static/components/schema.js
emro: I guess you could add a little line of text under the button saying "JSON object copied" or "JSON selection copied" but I'm not sure this is very important.

PR #3314, opened 2020-05-18
ENCD-5214-in-vivo-enhancer-screen

bonitalam:


src/encoded/schemas/transgenic_enhancer_experiment.json
```diff
@@ -0,0 +1,269 @@
...
+        "tissue_with_enhancer_activity": {
+            "title": "Tissues with enhancer activity",
+            "description": "The tissue(s) in which enhancer activity was positively expressed in.",
+            "type": "array",
+            "uniqueItems": true,
+            "minItems": 1,
+            "items": {
+                "title": "Tissue",
+                "description": "Tissue where positive expression of the enhancer was observed.",
+                "type": "string",
```
gabdank: I would re-word to be something like "The tissue in which enhancer activity was detected"

src/encoded/types/shared_biosample.py
```diff
@@ -0,0 +1,86 @@
+from .biosample import generate_summary_dictionary
+
+def biosample_summary_dictionary(request, biosampleObject=None):
```
gabdank: since this function is no longer returns only a dictionary I think it should be renamed, otherwise, it is misleading. Perhaps "biosample_summary_information" or something similar to that. I would also probably define variables with explicit names in dataset.py tp clearly distinguish between the [0] that is a dictionary and [1] that is a boolean flag.

src/encoded/types/dataset.py
```diff
@@ -181,6 +190,98 @@ def hub(self, request):
...
+    def biosample_summary(self, request, biosamples=None):
+        drop_age_sex_flag = False
+        dictionaries_of_phrases = []
+        if biosamples is not None:
+            for bs in biosamples:
+                biosampleObject = request.embed(bs, '@@object')
+                biosample_dict = biosample_summary_dictionary(request, biosampleObject)
+                dictionaries_of_phrases.append(biosample_dict[0])
+                if biosample_dict[1] is True:
+                    drop_age_sex_flag is True
```
gabdank: I would explicitly define 2 variables here - make the code more readable:
"biosample_summary_dictionary = item[0]" and "biosample_drop_age_sex_flag = item[1]", see my next comment for more details.

src/encoded/types/shared_calculated_properties.py
```diff
@@ -232,97 +229,10 @@ def biosample_summary(self,
...
-                                biosampleObject.get('post_nucleic_acid_delivery_time_units'),
-                                treatment_objects_list,
-                                preservation_method,
-                                part_of_object,
-                                originated_from_object,
-                                modifications_list,
-                                True)
-
-                            dictionaries_of_phrases.append(dictionary_to_add)
+                        biosample_dict = biosample_summary_dictionary(request, biosampleObject)
```
gabdank: update the function name if you change it according to comments above

gabdank (changes requested):
In addition to comments on the code, I would also add a test that actually checks the biosample summary calculation correctness for the new Object you are defining.

src/encoded/schemas/transgenic_enhancer_experiment.json
bonitalam: Changed wording accordingly

src/encoded/types/shared_biosample.py
bonitalam: Used biosample_summary_information instead

src/encoded/types/dataset.py
bonitalam: Defined

src/encoded/types/shared_calculated_properties.py
bonitalam: Updated

gabdank (approved):

PR #3318, opened 2020-05-19
ENCD-5239-fix-css-problem-on-examine-loci

zoldello:


emro (dismissed):
Looks good!

zoldello:
@emro Kline requested an additional change.

zoldello:
@forresttanaka Do you have any issue with this commit:

https://github.com/ENCODE-DCC/encoded/pull/3318/commits/27be320fd87c1dc1881762fca0d9f8a6f3169cac

It makes text wrap in many add-pages like award and donor.

zoldello:
> Would `word-break: break-word` do what you need? It at least works in Safari. Might have to just sacrifice IE.

@forresttanaka  Thanks for the catch. I  went with "word-break: break-all" despite the way it wraps words. "word-break: break-word" is deprecated and some sites say its non-standard.

emro (dismissed):

zoldello:
I rebased on origin/dev.

hitz (approved):
Appproved based on rebase/force push

PR #3323, opened 2020-05-20
ENCD-5268-analysis-object

yunhailuo:


src/encoded/schemas/analysis.json
```diff
@@ -0,0 +1,96 @@
+{
+    "title": "Analysis",
+    "description": "Schema for grouping files into one analysis.",
+    "comment": "An analysis is typically performed on one experiments and is initially designed to group only files from one same experiment.",
```
hitz: "one same" => "a single"

src/encoded/schemas/analysis.json
```diff
@@ -0,0 +1,96 @@
...
+    "id": "/profiles/analysis.json",
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "type": "object",
+    "required": ["dataset", "files"],
```
hitz: I am not 100% dataset should be required.  I am concerned that if/when we create analyses that span multiple datasets this will be hard to port.

src/encoded/schemas/analysis.json
```diff
@@ -0,0 +1,96 @@
...
+            "type": "string",
+            "linkTo": "Dataset"
+        },
+        "files": {
+            "title": "Analysis files",
+            "description": "A list of processed files belonging to this analsis",
+            "type": "array",
+            "minItems": 1,
+            "uniqueItems": true,
+            "permission": "import_items",
```
hitz, line 34: Not sure why this is restricted to admins.

src/encoded/types/analysis.py
```diff
@@ -0,0 +1,141 @@
...
+        "title": "Assemblies",
+        "description": "A list of genome assemblies on which this analysis is performed.",
+        "comment": "Do not submit. This field is calculated from files in this analysis.",
+        "type": "array",
+        "notSubmittable": True,
+        "items": {
+            "type": "string"
+        }
+    })
+    def assemblies(self, request, files):
```
hitz: isn't this guaranteed to be singular? Unless something has gone horribly awry?  I feel like we should do something like file.assay_term_name which uses take_one_or_return_none (could return some error code or "mixed" or something that audits could trap.

Same for genome_annotations

src/encoded/types/analysis.py
```diff
@@ -0,0 +1,141 @@
...
+        "description": "A list of award bioproject phases to which pipelines "
+                       "that used to generate this analysis belong to.",
+        "comment": "Do not submit. This field is calculated from files in this analysis.",
+        "type": "array",
+        "notSubmittable": True,
+        "items": {
+            "type": "string"
+        }
+    })
+    def pipeline_award_rfas(self, request, pipelines=[]):
```
hitz, line 154: I suspect there may be a "valid" corner case where awards and labs have multiple values but I am not sure.  If there isn't one -- see comment on "assemblies"

hitz (commented):
So this code looks like it works fine but I have several conceptual questions; possibly they were already brought up in the ticket.

hitz (dismissed):

src/encoded/loadxl.py
```diff
@@ -64,6 +64,7 @@
     'pipeline',
     'analysis_step_run',
     'file',
+    'analysis',
```
hitz: I wonder if this should be after QC metrics for future considerations?

src/encoded/schemas/analysis.json
```diff
@@ -0,0 +1,110 @@
+{
+    "title": "Analysis",
+    "description": "Schema for grouping files into one analysis.",
+    "comment": "An analysis is typically performed on one experiments and is initially designed to group only files from a single experiment.",
+    "id": "/profiles/analysis.json",
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "type": "object",
+    "required": ["files"],
```
hitz, line 8: also, future considerations, I suspect this will not be required.  If we post ENCAN at pipeline start time.

src/encoded/types/analysis.py
```diff
@@ -0,0 +1,190 @@
...
+            ]
+            if dataset is not None
+        })
+
+    file_schema = load_schema(
+        'encoded:schemas/file.json'
+    )
+
+    @calculated_property(schema={
+        "title": "Assemblies",
```
hitz: title should be assembly

src/encoded/types/analysis.py
```diff
@@ -0,0 +1,190 @@
...
+        }
+        if len(assemblies) > 1:
+            return 'mixed'
+        if len(assemblies) == 1:
+            return assemblies.pop()
+        if len(assemblies) < 1:
+            return
+
+    @calculated_property(schema={
+        "title": "Genome Annotations",
```
hitz: also title should be singular

src/encoded/types/analysis.py
```diff
@@ -0,0 +1,190 @@
...
+            "comment": "Do not submit. This field is calculated from files in this analysis.",
+            "type": "array",
+            "notSubmittable": True,
+            "items": {
+                "type": "string",
+                "linkTo": "Pipeline"
+            }
+        }
+    )
+    def pipelines(self, request, files):
```
hitz, line 123: this method looks very expensive.

src/encoded/types/experiment.py
```diff
@@ -336,6 +337,28 @@ def perturbed(self, request, replicates=None):
             return any(bio_perturbed)
         return False
 
+    @calculated_property(schema={
+        "title": "[Deprecated] Analysis",
```
hitz: it's not critical, but the language here is strange to me.  "Deprecated" to me means "currently bad".   It seems like you mean it here as 'we will change the name of this property in the future'

hitz (dismissed):
approve final changes

hitz (approved):
resolving some conflicts.

PR #3326, opened 2020-05-21
ENCD-5267 body map updates

emro:
There are several updates in this ticket:
(1) a link to the body map is added to the "Functional genomics" drop-down menu
(2) several schema updates are reflected
      -- exocrine system with mammary gland and liver is added
      -- the system slim "musculature of body" is renamed to be "musculature"
      -- "limb" is added to skeletal system and musculature
(3) I discovered a weird bug where "musculature" was being highlighted when "musculature of body" was the only term selected and that has to do with JS treating arrays of one element as strings and arrays of multiple elements as arrays, so I tried to account for that everywhere, though I think this was a very special and unusual case.
(4) I updated some comments where I had used "musculature of body" as an example and it was no longer relevant.

### DIFF
--- a/src/encoded/static/components/app.js
+++ b/src/encoded/static/components/app.js
@@ -41,9 +41,10 @@ const portal = {
             title: 'Data',
             children: [
                 { id: 'functional-genomics', title: 'Functional Genomics data' },
+                { id: 'assaysearch', title: 'Experiment search', url: '/search/?type=Experiment&status=released', tag: 'collection' },
                 { id: 'assaymatrix', title: 'Experiment matrix', url: '/matrix/?type=Experiment&status=released', tag: 'collection' },
                 { id: 'chip', title: 'ChIP-seq matrix', url: '/chip-seq-matrix/?type=Experiment&replicates.library.biosample.donor.organism.scientific_name=Homo%20sapiens&assay_title=Histone%20ChIP-seq&status=released', tag: 'collection' },
-                { id: 'assaysearch', title: 'Experiment search', url: '/search/?type=Experiment&status=released', tag: 'collection' },
+                { id: 'bodymap', title: 'Human body map', url: '/summary/?type=Experiment&replicates.library.biosample.donor.organism.scientific_name=Homo+sapiens', tag: 'collection' },
                 { id: 'sep-mm-0' },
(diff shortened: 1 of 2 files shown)


forresttanaka:
On (3) in your description, when does JS treat a single-element string array as a string? Is there a specific condition? I just tried this in node:

```
> typeof ['hello']
'object'
>
```

emro:
When there is just one selection in `this.state.selectedOrgan`, it gets interpreted as a string. So if you're on the summary page with the human body map and you select "musculature of body", this gets set in state: `this.state.selectedOrgan = ["musculature of body"]` and on line 608, one of the elements of `SystemsList` is "musculature", so then `this.state.selectedOrgan.includes("musculature")` evaluates to be true, and an "active" class is set on "musculature" when it shouldn't be. The real issue here might be in the constructor() when the state is set. I tried solutions like adding extra brackets on line 255 around "organTerms" and some other solutions to force the code to recognize an array of one element as an array but none of the solutions I tried worked. Was I missing something straightforward? It would make the code a lot simpler if I didn't need to check for `typeof` anywhere.

emro:
I actually sort of mentioned this on my original PR for the body map because there are some additional coding oddities due to this issue:

> (2) I needed to combine organ slims and system slims at times and had to write some verbose code to make spread syntax work, please let me know if you're aware of a more concise way to write that code.

Spread syntax converts a string into an array of letters and because single element states are interpreted as strings, this caused me to write extra code to check for type.

forresttanaka:
OK, I’ll let it go and keep going, but I feel like something else must be happening; I’m not sure what. I don’t see anything that could cause that.

```
> let so = ['musculature of body']
undefined
> so
[ 'musculature of body' ]
> so.includes('musculature')
false
> so.includes('musculature of body')
true
>
```

forresttanaka (approved):

emro:
I tried looking at this again, and added the brackets on line 255, and doing that makes the classes work as expected, but then appending classes doesn't work as expected - you can get the array of `this.state.selectedOrgan` into a very strange and unhelpful state: 
Screen Shot 2020-05-21 at 1 00 44 PM
I think this isn't as trivial as it seems like it should be, but it's probably all fixable / able to be simplified.

PR #3328, opened 2020-05-21
ENCD-5302-cart-link

forresttanaka:
Ben hasn’t approved of how this works yet, but given the timing, I figured might as well get the PR in at least.

Nothing uses the .btn-inline CSS class, so I felt safe modifying it to be more useful to situations where buttons sit side-by-side.

Updated demo: https://encd-5302-cart-link-dv1-fytanaka.demo.encodedcc.org/

A QA demo is coming up as well.

### DIFF
--- a/src/encoded/static/components/cart/cart.js
+++ b/src/encoded/static/components/cart/cart.js
@@ -5,7 +5,9 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import _ from 'underscore';
+import { encodedURIComponent } from '../../libs/query_encoding';
 import { svgIcon } from '../../libs/svg-icons';
+import { Modal, ModalHeader, ModalBody, ModalFooter } from '../../libs/ui/modal';
 import Pager from '../../libs/ui/pager';
 import { Panel, PanelBody, PanelHeading, TabPanel, TabPanelPane } from '../../libs/ui/panel';
 import { tintColor, isLight } from '../datacolors';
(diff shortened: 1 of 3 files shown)


src/encoded/static/components/cart/cart.js
```diff
@@ -822,6 +824,66 @@ FileFacets.defaultProps = {
...
+/**
+ * Display a button that links to experiment search results. This *only* works for Experiments, not
+ * FCCs for now.
+ */
+const MAX_SEARCHABLE_DATASETS = 125; // Maximum number of datasets for search-results links.
+const CartDatasetSearch = ({ elements }, reactContext) => {
+    const [isWarningVisible, setIsWarningVisible] = React.useState(false);
+
+    // * For now only include experiments. Might need to expand this in the future.
+    const experimentElements = elements.filter(element => element.includes('/experiments/'));
```
zoldello, line 836: Are the leading and trailing '/' needed?

forresttanaka: This eliminates the possibility of confusion if some other type with “experiments” in it comes up.

src/encoded/static/components/cart/cart.js
```diff
@@ -822,6 +824,66 @@ FileFacets.defaultProps = {
...
+
+    // Called when the user clicks the Dataset Search button.
+    const handleButtonClick = () => {
+        if (experimentElements.length > MAX_SEARCHABLE_DATASETS) {
```
zoldello, line 840: Does this supposed to be '>' or '>=='?

forresttanaka: “Greater than.” The maximum value is allowable.

src/encoded/static/components/cart/cart.js
```diff
@@ -822,6 +824,66 @@ FileFacets.defaultProps = {
...
+    if (experimentElements.length > 0) {
+        return (
+            <React.Fragment>
+                <button onClick={handleButtonClick} className="btn btn-sm btn-info btn-inline">Dataset search</button>
+                {isWarningVisible ?
+                    <Modal closeModal={handleCloseClick}>
+                        <ModalHeader title={<h4>Cart dataset search results</h4>} closeModal={handleCloseClick} />
+                        <ModalBody>
+                            Viewing cart dataset search results requires {MAX_SEARCHABLE_DATASETS} datasets
+                            or fewer. This cart contains {elements.length} datasets.
```
zoldello, line 864: if element.length === 1, it reads- "This cart contains 1 dataset**s**". That is a grammatical error. A similar principle applies to the line above this.

forresttanaka: The only way to have that situation is if the maximum number of allowable elements is zero.

src/encoded/static/components/cart/cart.js
```diff
@@ -822,6 +824,66 @@ FileFacets.defaultProps = {
 };
 
 
+/**
+ * Display a button that links to experiment search results. This *only* works for Experiments, not
+ * FCCs for now.
+ */
+const MAX_SEARCHABLE_DATASETS = 125; // Maximum number of datasets for search-results links.
```
zoldello, line 831: Where did you get 125 from? The ticket does not mention it. I assume you obtained it empirically.

forresttanaka: I determined it empirically. I mentioned it in the ticket comments.

zoldello (approved):
The code looks good. I could not find any code smell worth mentioning. All my comments are more like nitpicking rather than anything significant.

src/encoded/static/components/cart/cart.js
zoldello: @forresttanaka Thanks. I only looked at the description. I saw it after I looked at the comments.

PR #3329, opened 2020-05-22
ENCD-5296-fix-matrix-header-link

zoldello:


src/encoded/static/components/matrix_experiment.js
```diff
@@ -106,7 +106,12 @@ const convertExperimentToDataTable = (context, getRowCategories, getRowSubCatego
         colTitleMap[colCategoryBucket.key] = colIndex;
         return colCategoryBucket.key;
     });
-    const searchUrl = [context.search_base.split('&')[0], '&status=released'].join(''); // fancy way of removing query string parameters from URL
+
+    // This gets the statues and the base url and combines them
+    const urlComponents = context.search_base.split('&');
+    const statues = urlComponents.filter(component => component.includes('status')).join('&');
+    const searchUrl = `${urlComponents[0]}${statues ? ['&', statues].join('') : ''}`;
```
zoldello: I joined .join because string templating gave me an error

forresttanaka: I think you meant the const to be `status` and not `statues`?

What error did you get? This works for me:
```
const searchUrl = `${urlComponents[0]}${statues ? `&${statues}` : ''}`;
```

forresttanaka (changes requested):

src/encoded/static/components/matrix_experiment.js
zoldello: @forresttanaka  There can be multiples statues as a user can select "status=released" AND "status=archived". So I used the variable names "statuses".

You care correct on the string template. I probably improperly used a tick.

zoldello: @forresttanaka This code was removed. I also had to rebase on origin/dev

src/encoded/static/components/matrix_experiment.js
```diff
@@ -106,9 +106,13 @@ const convertExperimentToDataTable = (context, getRowCategories, getRowSubCatego
         colTitleMap[colCategoryBucket.key] = colIndex;
         return colCategoryBucket.key;
     });
-    const searchUrl = [context.search_base.split('&')[0], '&status=released'].join(''); // fancy way of removing query string parameters from URL
+
+    // Breaks up url so unneeded parts can be discarded
+    const searchBaseUrlComponents = context.search_base.split('&').filter(s => s.trim() !== '');
+
+    const headerUrl = searchBaseUrlComponents.filter(s => !s.includes('assay_title')).join('&');
```
forresttanaka: The query_string module can do this too in a way that might be easier to understand and maintain, though you do have to extract just the query string portion to use it. Should think about if query_string should accept entire URLs and extract the query string itself.

src/encoded/static/components/matrix_experiment.js
```diff
@@ -140,6 +144,9 @@ const convertExperimentToDataTable = (context, getRowCategories, getRowSubCatego
         const renderedData = categoryExpanded ? subCategoryData : subCategoryData.slice(0, SUB_CATEGORY_SHORT_SIZE);
         matrixRowKeys[matrixRow] = rowCategoryBucket.key;
         matrixRow += 1;
+        const cellDataUrl = searchBaseUrlComponents
+            .filter(s => !(s.includes('biosample_ontology.term_name') || s.includes('assay_title')))
+            .join('&');
```
forresttanaka: I think hardcoding these strings is risky in case they change in the page JSON. “biosample_ontology.term_name” is already in the `subCategory` variable, and “assay_title” is already in the `columnCategoryType` variable. The same applies to the first change above for “assay_title” and the changes below for “biosample_ontology.term_name” and “assay_title.”

forresttanaka (changes requested):

src/encoded/static/components/matrix_experiment.js
zoldello: I think better in terms of filters, but I can see how query_string is easier to understand. I updated the code.

src/encoded/static/components/matrix_experiment.js
zoldello: Good catch. I am tend to use magic strings so that thought did cross my mind. I fixed this.

src/encoded/static/components/matrix_experiment.js
forresttanaka: OK, then you meant “statuses” instead of “statues” — perfectly good word, but not the one you wanted.

forresttanaka (approved):

src/encoded/static/components/matrix_experiment.js
```diff
@@ -106,9 +107,25 @@ const convertExperimentToDataTable = (context, getRowCategories, getRowSubCatego
...
+
+    query.deleteKeyValue(columnCategoryType);
+    const baseUrlWithoutSubNorColCategoriesType = query.format();
+
+    query = new QueryString(context.search_base);
+    query.deleteKeyValue(columnCategoryType);
+    const baseUrlWithoutColCategoryType = query.format();
+
+
+    // query.deleteKeyValue('biosample_ontology.term_name');
```
forresttanaka: After approving, I just noticed commented-out code. I think you need to not have that there.

forresttanaka (dismissed):
Sorry, I noticed commented-out code just as I clicked Approve. I don't think we should have commented-out code generally, at least not without an explanation of why it needs to be there.

src/encoded/static/components/matrix_experiment.js
zoldello: I was testing out something before using it. Done.

src/encoded/static/components/matrix_experiment.js
hitz: this code is gone now.

hitz (approved):
LGTM

PR #3330, opened 2020-05-22
ENCD-5285-table-lines

forresttanaka:
Added a supremely complicated condition for drawing the heavy bar below certain rows. I also fixed one outdated colSpan.

### DIFF
--- a/src/encoded/static/components/filegallery.js
+++ b/src/encoded/static/components/filegallery.js
@@ -658,7 +658,10 @@ class RawSequencingTable extends React.Component {
                             {pairedRepKeys.map((pairedRepKey, j) => {
                                 // groupFiles is an array of files under a bioreplicate/library
                                 const groupFiles = pairedRepGroups[pairedRepKey].sort((a, b) => (a.pairSortKey < b.pairSortKey ? -1 : 1));
-                                const bottomClass = j < (pairedRepKeys.length - 1) ? 'merge-bottom' : '';
+
+                                // Draw bottom border if this isn't the last paired replicate, or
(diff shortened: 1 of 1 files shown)


forresttanaka:
Demo: ~https://encd-5285-table-lines-pr1-fytanaka.demo.encodedcc.org/~

forresttanaka:
Demo for PR: https://encd-5285-table-lines-pr1-fytanaka.demo.encodedcc.org/experiments/ENCSR141MUD/

src/encoded/static/components/filegallery.js
```diff
@@ -670,7 +673,15 @@ class RawSequencingTable extends React.Component {
                                         // Must be an index reads file if it's not paired-ended.
                                         pairClass = 'index-reads';
                                     }
-                                    if (groupFiles[i + 1] && groupFiles[i + 1].pairSortId !== file.pairSortId) {
+
+                                    // Draw bottom border if second file of pair and no index reads
+                                    // follow, or file *is* index reads, but not if last row of
+                                    // table.
+                                    const isNextInGroupIndexReads = groupFiles[i + 1] && groupFiles[i + 1].output_type === 'index reads';
+                                    if (
```
zoldello, line 681: Would it be a good idea to place each condition in the if-statement in a variable and use the if-statement on those variables? That has the advantage of keeping the readers' attention focus and may help reduce the length. That is something I picked up from Python.

zoldello (approved):
Code looks fine. I saw nothing posing a security threat or that looks improper. My comment on the PR is nothing significant and can be ignored if so desired.

PR #3335, opened 2020-05-27
ENCD-5319 fix date released link

emro:
A bug introduced by modifying "searchBase" when an unmodified "searchBase" was required later.

### DIFF
--- a/src/encoded/static/components/facets/defaults.js
+++ b/src/encoded/static/components/facets/defaults.js
@@ -443,19 +443,20 @@ export class DefaultDateSelectorFacet extends React.Component {
         let deleteSubmittedFilter = '';
         let deleteReleasedFilter = '';
         let missingFilter = '';
+        let searchBaseCopy = searchBase;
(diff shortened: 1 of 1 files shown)


src/encoded/static/components/facets/defaults.js
```diff
@@ -443,19 +443,20 @@ export class DefaultDateSelectorFacet extends React.Component {
         let deleteSubmittedFilter = '';
         let deleteReleasedFilter = '';
         let missingFilter = '';
+        let searchBaseCopy = searchBase;
```
forresttanaka, line 446: I’m getting an ESLint error on line 422 because `searchBase` doesn’t get modified anymore, so it wants a const.

forresttanaka (changes requested):
ESLint error only.

src/encoded/static/components/facets/defaults.js
emro: Oops, thanks for catching that!

forresttanaka (approved):

PR #3336, opened 2020-05-27
ENCD-5191 add ENCSR id to cart

emro:
Adding links to datasets on the cart genome browser labels.

### DIFF
--- a/src/encoded/static/components/genome_browser.js
+++ b/src/encoded/static/components/genome_browser.js
@@ -148,6 +148,8 @@ function mapGenome(inputAssembly) {
  */
 const TrackLabel = ({ file, label, long }) => {
     const biologicalReplicates = file.biological_replicates && file.biological_replicates.join(', ');
+    const splitDataset = file.dataset.split('/');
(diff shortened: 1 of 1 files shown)


src/encoded/static/components/genome_browser.js
```diff
@@ -148,6 +148,8 @@ function mapGenome(inputAssembly) {
  */
 const TrackLabel = ({ file, label, long }) => {
     const biologicalReplicates = file.biological_replicates && file.biological_replicates.join(', ');
+    const splitDataset = file.dataset.split('/');
+    const datasetName = splitDataset[splitDataset.length - 2];
```
forresttanaka, line 152: globals.js has the function `atIdToAccession` that converts an `@id` to an accession. Probably easier to read, as I had to work out what these two lines did at first.

forresttanaka (approved):

src/encoded/static/components/genome_browser.js
emro: Ah, that is very good to know! I would have used that function if I'd realized it existed.

PR #3338, opened 2020-05-28
ENCD-5324-experiment-perturbed

forresttanaka:
An update to ENCD-5181. This implements a new `perturbed` property and facet in Experiments. It also now uses everything it can to use “true” and “false” instead of “1” and “0.”


forresttanaka:
I have @yunhailuo for Python and @emro for Javascript on this one.

yunhailuo (approved):

emro (approved):
Looks good!

PR #3340, opened 2020-05-30
ENCD-5310-choose-analysis

forresttanaka:
* Before, I based the selected pipeline menu item by calculating it from the currently selected assembly — causing an issue if two menu items used the same assembly. I now store the selected menu index in the `selectedAnalysesIndex` state, and recalculate it when switching tabs.
* To help with this calculation, I keep the last-selected lab and assembly as properties in `this.selectedAnalysisProps`.
* I added a couple fields to all the dummy files in genome_browser.js as some recent code changes couldn’t handle those fields not being there. That issue doesn’t affect demos and production; just local.
* `findCompiledAnalysesIndex` which used to find an appropriate pipeline menu item based only on an assembly, now exists as a method and uses both a given assembly and the last-selected pipeline menu to figure out the index of the best-matching menu item.

### DIFF
--- a/src/encoded/static/components/filegallery.js
+++ b/src/encoded/static/components/filegallery.js
@@ -2138,27 +2138,10 @@ export const compileAnalyses = (experiment, files) => {
 
 
 /**
- * Find the compiled analyses object matching the given assembly and annotation and return its
- * index in the compiled analyses array.
- * match found.
- * @param {string} assembly Assembly and annotation to map to an analyses title
- * @param {array} compiledAnalyses Compiled analyses for the experiment
- *
- * @return {object} Array index of compiled analyses object matching given assembly, or -1
- */
-const findCompiledAnalysisIndex = (assembly, compiledAnalyses) => (
-    // Find the matching assembly in compiledAnalyses, or the first otherwise.
-    compiledAnalyses.findIndex(analyses => analyses.assembly === assembly)
-);
-
-
-/**
  * Display the analyses selector, a dropdown menu to choose which pipeline lab's files to view in
  * the file association graph.
  */
-const AnalysesSelector = ({ assembly, analyses, handleAnalysesSelection }) => {
-    const selectedAnalysesIndex = findCompiledAnalysisIndex(assembly, analyses);
-
+const AnalysesSelector = ({ analyses, selectedAnalysesIndex, handleAnalysesSelection }) => {
     React.useEffect(() => {
         if (selectedAnalysesIndex === -1 && analyses.length > 0) {
             // No selected pipeline lab analyses, but if we have at least one qualifying one,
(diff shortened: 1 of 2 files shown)


src/encoded/static/components/filegallery.js
```diff
@@ -2452,6 +2443,52 @@ class FileGalleryRendererComponent extends React.Component {
...
+    saveSelectedAnalysesProps(selectedAnalysesIndex, compiledAnalyses) {
+        const localCompiledAnalyses = compiledAnalyses || this.state.compiledAnalyses;
+        const selectedAnalyses = localCompiledAnalyses[selectedAnalysesIndex];
+        this.selectedAnalysisProps = { assembly: selectedAnalyses.assembly, pipelineLab: selectedAnalyses.pipelineLab };
+    }
+
+    /**
+     * Find the compiled analyses object best matching the given assembly/annotation and the last
+     * selected pipeline lab and return its index into the compiled analyses array. Only pass in
+     * `compiledAnalyses` if it has been newly determined and might not yet have propogated to
```
emro: "propagated" not "propogated"

src/encoded/static/components/filegallery.js
```diff
@@ -2452,6 +2443,52 @@ class FileGalleryRendererComponent extends React.Component {
         return [];
     }
 
+    /**
+     * Sets the `selectedAnalysisProps` property to track the corresponding selected assembly/annotation
+     * and pipeline lab. Only pass in `compiledAnalyses` if it has been newly determined and might
+     * not yet have propogated to `this.state.compiledAnalyses`.
```
emro: "propagated" not "propogated"

src/encoded/static/components/filegallery.js
```diff
@@ -2452,6 +2443,52 @@ class FileGalleryRendererComponent extends React.Component {
...
+     */
+    findCompiledAnalysesIndex(assembly, compiledAnalyses) {
+        let compiledAnalysisIndex = 0;
+        const localCompiledAnalyses = compiledAnalyses || this.state.compiledAnalyses;
+
+        // Find all analyses matching the given assembly, then of those find one that also matches
+        // the pipeline lab.
+        const analysesWithMatchingAssembly = localCompiledAnalyses.filter(analysis => analysis.assembly === assembly);
+        if (analysesWithMatchingAssembly.length > 0) {
+            const matchingAnalysis = analysesWithMatchingAssembly.find(analysis => analysis.pipelineLab === this.selectedAnalysisProps.pipelineLab);
```
emro, line 2478: Just curious, could there be multiple analyses with a matching assembly and pipeline lab?

forresttanaka: Not at this level because any entries with duplicate assembly and pipeline labs get their files combined into one entry. So one or the other must be different for there to be separate entries.

forresttanaka:
I’ve _long_ had problems spelling “propagated” because the proper spelling always looks weird to me.

emro (approved):

PR #3341, opened 2020-06-01
ENCD-5031-make-facets-collapsible

zoldello:


src/encoded/static/components/facets/date_selector.js
```diff
@@ -31,11 +33,17 @@ DateReleasedFacet.propTypes = {
     pathname: PropTypes.string.isRequired,
     /** Query-string portion of current URL without initial ? */
     queryString: PropTypes.string,
+    /** List of expand facet */
+    expandedFacets: PropTypes.instanceOf(Set),
+    /** toogles facets */
```
forresttanaka, line 38: “Toggles”

src/encoded/static/components/facets/defaults.js
```diff
@@ -20,6 +20,14 @@ const sanitizedString = inputString => inputString.toLowerCase()
...
     .replace(/[*?()+[\]\\/]/g, ''); // remove certain special characters (these cause console errors)
 
+// session storage used to preserve opened/closed facets
+const FACET_STORAGE = 'FACET_STORAGE';
+
+// marker for determining user just opened the page
+const MARKER_FOR_NEWLY_LOADED_FACET_PREFIX = 'MARKER_FOR_NEWLY_LOADED_FACETS_';
+
+// List of facets deemed to open when the page first starts
+const preopenedFacetList = ['status', 'perturbed', 'assay_slims', 'biosample_ontology.term_name', 'target.label'];
```
forresttanaka, line 30: On Slack I think, Ben said the facets opened by default would be defined in the schema. Was this decided against? No problem defining them here as well, but the schema seemed like the obvious place so that if you delete or add a facet to the schema that should be expanded by default, you wouldn’t also have to modify this array.

src/encoded/static/components/facets/defaults.js
```diff
@@ -1015,15 +1066,59 @@ export const DefaultFacet = ({ facet, results, mode, relevantFilters, pathname,
...
 
+    /**
+     * Handlers opening or closing a tab
+     *
+     * @param {event} e React synthetic event
+     * @param {bool} status True for open, false for closed
+     * @param {string} field Tab name
+     */
+    const handleExpanderClick = (e, status, field) => {
+        if (toggleFacets) {
```
forresttanaka, line 1077: Why wouldn’t `toggleFacets` exist?

src/encoded/static/components/facets/defaults.js
```diff
@@ -940,11 +948,54 @@ export const DefaultFacet = ({ facet, results, mode, relevantFilters, pathname,
...
+
+            // auto-open facets based on (1)With selected terms OR facet pre-determined to be opened
+            if ((relevantFilters && relevantFilters.length > 0) || preopenedFacetList.includes(field)) {
+                facetList.add(field);
+            }
+        }
+
+        sessionStorage.setItem(FACET_STORAGE, facetList.size !== 0 ? [...facetList].join(',') : []);
+        setExpandFacets(facetList); // initalize facet collapse-state
+    }, []); // eslint-disable-line react-hooks/exhaustive-deps
```
forresttanaka, line 981: I feel like this code should be handled by search.js, not the default facet. If we wanted to make the date-selector facet or the exists facet also collapse (and I felt surprised they don’t), then we’ll have to copy this code to the other kinds of default and custom facets. I think each facet component should be told through props if they appear expanded or not. We’ll still need a trigger callback, but search.js should be the one and only place that stores the expanded/collapsed states. The facet and title components should only know whether their particular facet appears expanded or not.

src/encoded/static/components/facets/defaults.js
```diff
@@ -940,11 +948,54 @@ export const DefaultFacet = ({ facet, results, mode, relevantFilters, pathname,
...
+            sessionStorage.removeItem(FACET_STORAGE);
+            facetList = new Set(status ? [] : results.facets.map(f => f.field));
+        } else {
+            facetList = new Set(expandedFacets);
+            facetList[status ? 'delete' : 'add'](field);
+        }
+
+        sessionStorage.setItem(FACET_STORAGE, [...facetList].join(',')); // replace rather than update memory
+        setExpandFacets(facetList); // controls open/closed facets
+    };
```
forresttanaka, line 997: Like here, `toggleFacets` should just tell `<Search>` that the user clicked the expand/collapsed trigger. It shouldn’t care how that gets recorded. Let search.js worry about the current states and storing them.

src/encoded/static/components/facets/defaults.js
```diff
@@ -1015,15 +1066,59 @@ export const DefaultFacet = ({ facet, results, mode, relevantFilters, pathname,
...
+    * @param {bool} status True for open, false for closed
+    * @param {string} field Tab name
+    */
+    const handleKeyDown = (e, status, field) => {
+        if (e.keyCode === 13 || e.keyCode === 32) {
+            // keyCode: 13 = enter-key. 32 = spacebar
+            e.preventDefault();
+            handleExpanderClick(e, status, field);
+        }
+    };
```
forresttanaka, line 1097: I think this and `handleExpanderClick` should either be a service provided by search.js, or should be in a custom hook that all facet types can use. Or maybe just be implemented once in TitleComponent.

src/encoded/static/components/facets/exists.js
```diff
@@ -7,14 +7,16 @@ import FacetRegistry from './registry';
...
     <DefaultExistsFacet
         facet={facet}
         results={results}
         mode={mode}
         relevantFilters={relevantFilters}
         pathname={pathname}
         queryString={queryString}
+        expandedFacets={expandedFacets}
+        setExpandFacets={setExpandFacets}
     />
```
forresttanaka, line 20: `DefaultExistsFacet` doesn’t seem to use these new props, nor does `DefaultDateSelector`.

forresttanaka (changes requested):
The main feedback: I think the localstorage code should be close to the code handling the expand/collapse state so that all facet types can use it without needing to worry much about it.

zoldello:
closed in lieu of SNO-157

PR #3343, opened 2020-06-01
ENCD-5113-make-curl-use-server-filename

keenangraham:


keenangraham:
-J, --remote-header-name
(HTTP) This option tells the -O, --remote-name option to use the server-specified Content-Disposition filename instead of extracting a filename from the URL.
If the server specifies a file name and a file with that name already exists in the current working directory it will not be overwritten and an error will occur. If the server doesn't specify a file name then this option has no effect.

yunhailuo (approved):

hitz (approved):

PR #3346, opened 2020-06-02
ENCD-5328 fix released start date

emro:
The default start year for the "released" radio button needs to be 2009. I'm not sure why this worked originally but since setting state is asynchronous, the function for setting parameters needs to be a callback.


forresttanaka (approved):
That _is_ strange this problem didn’t happen before.

PR #3348, opened 2020-06-02
ENCD-5308-fix-react-error-on-summary-page

zoldello:


src/encoded/static/components/file.js
```diff
@@ -596,7 +596,7 @@ class SequenceFileInfo extends React.Component {
                                 <dd>
                                     {file.index_of.reduce((fUrls, f) => {
                                         const fileId = f.replace(/\/files\/|\//g, '');
-                                        const fUrl = <a href={`/files/${fileId}`} title={fileId}>{fileId}</a>;
+                                        const fUrl = <a key={f} href={`/files/${fileId}`} title={fileId}>{fileId}</a>;
```
forresttanaka: This isn’t part of this ticket really, but I just noticed that you don’t really need the string template for the href — you can just use `href={f}` since you have `f` already. Did you want to change that for this ticket, or leave it?

zoldello: @forresttanaka I will change it. I will look at it tomorrow (Wednesday)

forresttanaka (approved):

PR #3355, opened 2020-06-05
ENCD-5309-cart-locks

forresttanaka:
A new property in the cart object, `locked`, keeps track of the cart lock. This property isn’t used on the back end, though a couple back-end functions set it to false as part of their initialization.

https://encd-5309-cart-locks-dv1-fytanaka.demo.encodedcc.org/

### DIFF
--- a/src/encoded/static/components/cart/lock.js
+++ b/src/encoded/static/components/cart/lock.js
@@ -1,0 +1,74 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import { svgIcon } from '../../libs/svg-icons';
+import { setCartLockAndSave } from './actions';
+
+
+/**
+ * Renders a lock icon which, on a user click, initiates the locking of the given cart. That
+ * involves setting the `locked` flag in the cart object, though all other processing of this flag
+ * happens in the front end.
+ */
+const CartLockTriggerComponent = ({ savedCartObj, inProgress, onLock }) => {
+    // Called when the user clicks the lock/unlock button.
+    const handleLockClick = () => {
+        onLock(!savedCartObj.locked);
+    };
+
+    // Determine the tooltip text.
+    let disabledTooltip = '';
+    if (savedCartObj.status === 'deleted') {
+        disabledTooltip = 'Cannot lock a deleted cart';
+    } else if (savedCartObj.status === 'disabled') {
+        disabledTooltip = 'Cannot share the auto-save cart';
+    } else if (inProgress) {
+        disabledTooltip = 'Cart operation in progress';
+    }
+
+    return (
+        <div className="cart-manager-table__tooltip-group">
+            {disabledTooltip ?
+                <div
+                    className="cart-manager-table__button-overlay"
+                    title={disabledTooltip}
+                />
+            : null}
+            <button
+                onClick={handleLockClick}
+                className="btn btn-sm btn-warning btn-inline"
+                disabled={inProgress || savedCartObj.status === 'deleted' || savedCartObj.status === 'disabled'}
+                aria-label={`${savedCartObj.locked ? 'Unlock' : 'Lock'} cart`}
+            >
+                {svgIcon(savedCartObj.locked ? 'lockClosed' : 'lockOpen', { width: 13, marginRight: 4 })}
+                {savedCartObj.locked ? <span>Unlock</span> : <span>Lock</span>}
+            </button>
+        </div>
+    );
+};
+
+CartLockTriggerComponent.propTypes = {
+    /** Cart as it exists in the database */
+    savedCartObj: PropTypes.object.isRequired,
+    /** True if cart operation in progress */
+    inProgress: PropTypes.bool.isRequired,
+    /** Redux function called to lock/unlock cart */
+    onLock: PropTypes.func.isRequired,
+};
+
+const mapDispatchToProps = (dispatch, ownProps) => ({
+    onLock: locked => dispatch(setCartLockAndSave(locked, ownProps.savedCartObj, ownProps.sessionProperties && ownProps.sessionProperties.user, ownProps.fetch)),
+});
+
+const CartLockTriggerInternal = connect(null, mapDispatchToProps)(CartLockTriggerComponent);
+
+const CartLockTrigger = (props, reactContext) => (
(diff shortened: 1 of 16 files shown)


src/encoded/static/libs/svg-icons.js
```diff
@@ -46,6 +64,8 @@ const icons = {
     spinner,
     largeArrow,
     genomeBrowser,
+    lockOpen,
+    lockClosed,
```
emro, line 68: I really like these icons.

src/encoded/static/components/cart/toggle.js
```diff
@@ -33,13 +33,14 @@ class CartToggleComponent extends React.Component {
...
         const cartAtLimitToolTip = cartAtLimit ? `Cart can contain a maximum of ${CART_MAXIMUM_ELEMENTS_LOGGEDOUT} items` : '';
+        const locked = savedCartObj && Object.keys(savedCartObj).length > 0 ? savedCartObj.locked : false;
 
         // "name" attribute needed for BDD test targeting.
         return (
             <button
                 className={`cart__toggle${inCart ? ' cart__toggle--in-cart' : ''}${css ? ` ${css}` : ''}`}
                 onClick={this.handleClick}
-                disabled={inProgress || (!loggedIn && !inCart && cartAtLimit)}
+                disabled={inProgress || locked || (!loggedIn && !inCart && cartAtLimit)}
```
emro, line 43: I would consider adding / removing a couple styles to these disabled `cart__toggle` buttons. Currently, when the buttons are disabled, the cursor is a pointer and there's a light gray background on hover. I would expect the cursor to be the default and no change on hover for a disabled button.

src/encoded/static/components/cart/status.js
```diff
@@ -117,10 +126,10 @@ class CartStatusComponent extends React.Component {
             return (
                 <NavItem
                     dropdownId="cart-control"
-                    dropdownTitle={<CartNavTitle elements={elements} inProgress={inProgress} />}
+                    dropdownTitle={<CartNavTitle elements={elements} locked={locked} inProgress={inProgress} />}
                     openDropdown={openDropdown}
                     dropdownClick={dropdownClick}
-                    label={`Cart containing ${elements.length} ${elements.length > 1 ? 'items' : 'item'}`}
+                    label={`${locked ? 'locked' : ''} cart containing ${elements.length} ${elements.length > 1 ? 'items' : 'item'}`}
```
emro, line 132: Did you think about having the label read "unlocked cart" for carts that are not locked? Just a thought, it probably depends on how much emphasis you want to put on the locking capability. Also, this is me being overly detail-oriented, but I'd put the extra space with "locked" so that the label doesn't start with a space for unlocked carts.

src/encoded/static/components/cart/status.js
```diff
@@ -92,15 +94,22 @@ class CartStatusComponent extends React.Component {
...
             const cartName = (loggedIn && savedCartObj && savedCartObj.name) ? truncateString(savedCartObj.name, 22) : '';
             const menuItems = [];
             const viewCartItem = <a key="view" href="/cart-view/">View cart</a>;
             const clearCartItem = <button key="clear" onClick={this.clearCartClick}>Clear cart</button>;
+            const lockIcon = cartName ? <div className="cart-nav-lock">{svgIcon(locked ? 'lockClosed' : 'lockOpen')}</div> : null;
             if (loggedIn) {
-                menuItems.push(<span key="name" className="disabled-menu-item">{`Current: ${cartName}`}</span>, <DropdownMenuSep key="sep-1" />);
+                menuItems.push(
+                    <span key="name" className="disabled-menu-item">
+                        {`Current: ${cartName}`}{lockIcon}
```
emro, line 109: All the visual cues about the carts being locked or not are great.

src/encoded/static/components/cart/lock.js
```diff
@@ -0,0 +1,74 @@
...
+            : null}
+            <button
+                onClick={handleLockClick}
+                className="btn btn-sm btn-warning btn-inline"
+                disabled={inProgress || savedCartObj.status === 'deleted' || savedCartObj.status === 'disabled'}
+                aria-label={`${savedCartObj.locked ? 'Unlock' : 'Lock'} cart`}
+            >
+                {svgIcon(savedCartObj.locked ? 'lockClosed' : 'lockOpen', { width: 13, marginRight: 4 })}
+                {savedCartObj.locked ? <span>Unlock</span> : <span>Lock</span>}
+            </button>
```
emro, line 45: Another really minor comment - did you think about making the Lock and Unlock buttons the same size? The other buttons have the same names so their buttons are automatically the same width, and you could do that if you wanted with the Lock/Unlock buttons. Really not a big deal, just a thought as I was looking at these buttons.
Screen Shot 2020-06-08 at 5 25 16 PM

emro (commented):
All the code looks great, and I really like this new functionality and the way that it is visualized on the cart pages / navigation. All of my comments are just minor things, feel free to ignore them all.

forresttanaka:
Thanks! I'll probably make a couple of these changes tomorrow (Tuesday).

src/encoded/static/components/cart/status.js
forresttanaka: I was OK with the extra space because you never see this text — you only hear it with a screen reader. That’s why I use the lower-case letters to start the line as well.

src/encoded/static/components/cart/toggle.js
forresttanaka: Fixed

src/encoded/static/components/cart/lock.js
forresttanaka: Fixed. I tried to do this in a way that the button is only as wide as the widest possible label, and wrote a mostly-working mechanism to do it, but it was so complicated that it hardly seemed worth it when I can just hard code one line of CSS to do what’s probably good enough, so that’s what I went with.

forresttanaka:
I changed the lock button text and icon to black because I found out that white on orange fails the WCAG contrast test.

emro (dismissed):

forresttanaka:
Forgot to remove the “Clear Cart” menu item when your cart is locked. “Clear Cart” in the menu now only appears while your cart is unlocked.

emro (approved):

PR #3357, opened 2020-06-05
ENCD-3470-update-facet-item-blocked-out

zoldello:


src/encoded/static/components/facets/defaults.js
```diff
@@ -943,6 +943,8 @@ export const DefaultFacet = ({ facet, results, mode, relevantFilters, pathname,
     // For straightforward numerical facets, just sort by value
     const processedTerms = significantTerms.every(numericalTest) ? _.sortBy(significantTerms, obj => obj.key) : significantTerms;
 
+    const disabledCss = filters => filters.some(f => f.field.indexOf('!') !== -1 && f.term.trim() === '*');
```
emro, line 946: I would probably write a comment here, mostly because I'm forgetful and sometimes forget the purpose of code that I write.

emro (approved):

PR #3360, opened 2020-06-08
ENCD-5331-target-factorbook-dbxref

bonitalam:


gabdank (dismissed):
the dbxref should probably specify a regex that limits it to be only FactorBook for Target objects.
It should not be a free text field

hitz (approved):
We need to make sure to test the regex against all gene names in the database... sometimes people slip odd characters in there.

PR #3366, opened 2020-06-09
ENCD-5237 add body map to reference epigenome

emro:
- Adds a clickable version of the body map to the human reference epigenome page
- Minor updates to the tab component allow for additional tab features (disabled tabs, individually styles tabs)


src/encoded/static/components/body_map.js
```diff
@@ -126,6 +126,34 @@ const addingClass = (changedClass, matchingString, removeFlag = false) => {
     }
 };
 
+// Set body map diagram colors based on url, when component mounts
```
zoldello, line 129: I recommend using JSDocs rather than a vanilla comment as JSDocs are used by IDEs to provide information to consumers this function.

src/encoded/static/components/body_map.js
```diff
@@ -126,6 +126,34 @@ const addingClass = (changedClass, matchingString, removeFlag = false) => {
...
+            }
+            if (SystemsList[term]) {
+                document.getElementById(term).classList.add('active');
+            }
+        });
+    }
+};
+
+// Clearing organ and systems key value selections from a url
+export const clearBodyMapSelectionsFromUrl = (originalUrl) => {
```
zoldello, line 148: "clearBodyMapSelectionsFromUrl " or "removeSystemsFromUrl"?

src/encoded/static/components/body_map.js
```diff
@@ -126,6 +126,34 @@ const addingClass = (changedClass, matchingString, removeFlag = false) => {
...
+    const parsedUrl = url.parse(originalUrl);
+    const query = new QueryString(parsedUrl.query);
+    query.deleteKeyValue(organField);
```
zoldello, line 151: Can organField or SystemField appear multiple times in the URL? If so, I believe this functions removes them just once.

src/encoded/static/components/body_map.js
```diff
@@ -263,20 +291,7 @@ class BodyMap extends React.Component {
...
-                    BodyList[term].forEach((bodyClass) => {
-                        addingClass('active', bodyClass);
-                    });
-                }
-                if (SystemsList[term]) {
-                    document.getElementById(term).classList.add('active');
-                }
-            });
-        }
+        initializeBodyMap(searchQuery);
```
zoldello, line 298: Good idea IMOH. I too prefer tertiary functionality outsourced.

src/encoded/static/components/body_map.js
```diff
@@ -602,24 +613,25 @@ class BodyMap extends React.Component {
...
-                                key={b}
-                            >
-                                {b}
-                            </button>
+                            <li key={b}>
+                                <button
+                                    className={`body-list-element ${((typeof this.state.selectedOrgan === 'string' && this.state.selectedOrgan === b) || (typeof this.state.selectedOrgan !== 'string' && this.state.selectedOrgan.includes(b))) ? 'active' : ''}`}
+                                    id={b}
+                                    onClick={e => this.chooseOrgan(e)}
+                                    onMouseEnter={e => highlightOrgan(e)}
```
zoldello, line 627: Why don't highlightOrgan and unHighlighOrgan have "this." in front? They both are within a React Class component.

src/encoded/static/components/body_map.js
```diff
@@ -629,20 +641,21 @@ class BodyMap extends React.Component {
...
-                                    onClick={e => this.chooseOrgan(e)}
-                                    onMouseEnter={e => highlightOrgan(e)}
-                                    onMouseLeave={unHighlightOrgan}
-                                    key={b}
-                                >
-                                    {b}
-                                </button>
+                                <li key={b}>
+                                    <button
+                                        className={`body-list-element ${((typeof this.state.selectedOrgan === 'string' && this.state.selectedOrgan === b) || (typeof this.state.selectedOrgan !== 'string' && this.state.selectedOrgan.includes(b))) ? 'active' : ''}`}
```
zoldello: This seems duplicated in line 620. Can this be moved into a method, so changes can take place just once?

src/encoded/static/components/body_map.js
```diff
@@ -602,24 +613,25 @@ class BodyMap extends React.Component {
...
-                                onClick={e => this.chooseOrgan(e)}
-                                onMouseEnter={e => highlightOrgan(e)}
-                                onMouseLeave={unHighlightOrgan}
-                                key={b}
-                            >
-                                {b}
-                            </button>
+                            <li key={b}>
+                                <button
+                                    className={`body-list-element ${((typeof this.state.selectedOrgan === 'string' && this.state.selectedOrgan === b) || (typeof this.state.selectedOrgan !== 'string' && this.state.selectedOrgan.includes(b))) ? 'active' : ''}`}
```
zoldello: typeof to me is a code smell in dynamic languages. selectOrgan can be a string or array. Why not always have it be a string ?

src/encoded/static/components/matrix.js
```diff
@@ -92,12 +92,14 @@ export class SearchFilter extends React.Component {
...
-                    <div className="searchform">
-                        <TextFilter filters={context.filters} searchBase={matrixSearch} onChange={this.onChange} />
-                    </div>
+                    <p>
+                        <i className="icon icon-search" />
+                        Filter the {type} included in the matrix:
+                    </p>
+                </div>
+                <div className="searchform">
+                    <TextFilter filters={context.filters} searchBase={matrixSearch} onChange={this.onChange} />
```
zoldello, line 102: why isn't matrixsearch binded?

src/encoded/static/components/matrix_reference_epigenome.js
```diff
@@ -21,6 +23,13 @@ import { SearchControls } from './search';
  */
 const SUB_CATEGORY_SHORT_SIZE = 10;
 
+// Organisms for which we display tabs
```
zoldello, line 26: Similar comment to my JSDoc earlier. JSDocs helps IDE make const more discoverable.

src/encoded/static/components/matrix_reference_epigenome.js
```diff
@@ -21,6 +23,13 @@ import { SearchControls } from './search';
...
+// Names for the tabs are hard-coded because we want to display disabled tabs for tabs for which there are no results
+// Mapping of shortened organism name and full scientific organism name is used for human and mouse thumbnail images
+const organismTerms = {
```
zoldello: The syntax is correct. However, it seems like the key and values are reverse. I normally see the key without a space and something sort and recognizable. The value is what gets more complicated.

src/encoded/static/components/matrix_reference_epigenome.js
```diff
@@ -415,6 +411,92 @@ MatrixHeader.contextTypes = {
     navigate: PropTypes.func,
 };
 
+const ClickableThumbnail = (props) => {
```
zoldello, line 424: There is something with the name that I cannot point my finger at. I know what it means but at the same time I cannot describe what it means based solely on the function name.

src/encoded/static/components/matrix_reference_epigenome.js
```diff
@@ -415,6 +411,92 @@ MatrixHeader.contextTypes = {
...
+    const expandedThumbnail = props.expandedThumbnail;
```
zoldello: should this be called something like "isExpand"?

src/encoded/static/components/matrix_reference_epigenome.js
```diff
@@ -415,6 +411,92 @@ MatrixHeader.contextTypes = {
...
+    const expandThumbnail = props.expandThumbnail;
```
zoldello: should this be called something like "expand"? The function names gets reading in the mind-set of working with thumbnails.

src/encoded/static/components/matrix_reference_epigenome.js
```diff
@@ -415,6 +411,92 @@ MatrixHeader.contextTypes = {
...
+            <BodyDiagram />
+            <div className="body-list body-list-narrow">
+                <ul className="body-list-inner">
+                    {Object.keys(CellsList).map(image =>
+                        <div
+                            className={`body-inset ${image}`}
+                            id={image}
+                            key={image}
+                        >
+                            <img className="active-image" src={`/static/img/bodyMap/insetSVGs/${image.replace(' ', '_')}.svg`} alt={image} />
```
zoldello, line 445: hardcoding part of the url path seems like a code smell I think. I feel there should be some pre-existing constant or it should be made into one.

src/encoded/static/components/matrix_reference_epigenome.js
```diff
@@ -415,6 +411,92 @@ MatrixHeader.contextTypes = {
...
+};
+
+const SelectedFilters = (props) => {
+    const selectedFilters = props.filters;
+    const organTerms = selectedFilters.filter(f => f.field === organField);
+    const systemsTerms = selectedFilters.filter(f => f.field === systemsField);
+    const freeSearchTerms = selectedFilters.filter(f => f.field === 'searchTerm');
+    const selectedTerms = [...organTerms, ...systemsTerms, ...freeSearchTerms];
+    return (
+        <React.Fragment>
```
zoldello, line 468: <React.Fragment> or <> is fine.

src/encoded/static/components/matrix_reference_epigenome.js
```diff
@@ -415,6 +411,92 @@ MatrixHeader.contextTypes = {
...
+SelectedFilters.propTypes = {
+    filters: PropTypes.array.isRequired,
+};
+
+const BodyMapModal = (props) => {
+    const expandedThumbnail = props.expandedThumbnail;
+    const expandThumbnail = props.expandThumbnail;
+    const context = props.context;
+    return (
+        <div className="modal" style={{ display: 'block' }}>
```
zoldello, line 480: Isn't there a pre-defined CSS for "display: block"?

src/encoded/static/components/matrix_reference_epigenome.js
```diff
@@ -462,9 +547,24 @@ class MatrixPresentation extends React.Component {
...
+        terms.forEach((term) => {
+            if (CellsList[term] && document.getElementById(term)) {
+                document.getElementById(term).classList.add('active');
+            }
+        });
+
+        this.setState({
+            // Display the organism-chooser modal if the query string doesn't specify an organism to
+            // display.
+            organismChooserVisible: this.initialSelectedTab === null,
```
zoldello, line 576: Do you mean null or falsey?

src/encoded/static/components/matrix_reference_epigenome.js
```diff
@@ -647,43 +756,80 @@ class MatrixPresentation extends React.Component {
...
+        const rowCount = matrix => matrix.y['biosample_ontology.classification'].buckets.reduce((currentTotalCount, termName) => currentTotalCount + termName['biosample_ontology.term_name'].buckets.length, 0);
+
         return (
-            <div className="matrix__presentation">
-                <div className={`matrix__label matrix__label--horz${!scrolledRight ? ' horz-scroll' : ''}`}>
-                    <span>{context.matrix.x.label}</span>
-                    {svgIcon('largeArrow')}
+            <React.Fragment>
+                <div className="view-controls-container">
+                    <ViewControls results={this.props.context} alternativeNames={['Search list', 'Tabular report', 'Summary matrix']} />
```
zoldello, line 776: ViewControls seems like a very generic name for a component that is not generic.

src/encoded/static/components/matrix_reference_epigenome.js
```diff
@@ -647,43 +756,80 @@ class MatrixPresentation extends React.Component {
...
-                                    <DataTable tableData={matrixConfig} />
-                                : null}
+                <div className="summary-controls">
+                    <div className="matrix-header__filter-controls">
+                        <SearchFilter context={context} />
+                    </div>
+                    <div className="results-count">Showing <b className="bold-total">{rowCount(context.matrix)}</b> result{rowCount(context.matrix) > 1 ? 's' : ''}.</div>
+                </div>
+                <TabPanel tabs={organismTabs} selectedTab={this.initialSelectedTab} handleTabClick={this.handleTabClick} tabPanelCss="matrix__data-wrapper">
+                    {(this.initialSelectedTab === 'Homo sapiens') ?
```
zoldello, line 786: If you go with my idea of swapping they keys and values of organismTerms, you can use it here (organismTerms['human'])

src/encoded/static/components/matrix_reference_epigenome.js
```diff
@@ -647,43 +756,80 @@ class MatrixPresentation extends React.Component {
...
+                                        <Modal>
+                                            <ModalHeader closeModal={false} addCss="matrix__modal-header">
+                                                <h2>Reference epigenome &mdash; choose organism</h2>
+                                            </ModalHeader>
+                                            <ModalBody addCss="matrix-reference-epigenome__organism-selector">
+                                                <div>Organism to view in matrix:</div>
+                                                <div className="selectors">
+                                                    {Object.keys(organismTabs).map(organism => (
+                                                        // Encode the organism name into the <a> class for BDD testing.
+                                                        <a key={organism} className={`btn btn-info btn__selector--${organism.replace(/ /g, '-')}`} href={`${context['@id']}&replicates.library.biosample.donor.organism.scientific_name=${encoding.encodedURIComponent(organism)}`}>{organism}</a>
```
zoldello, line 833: The keys of organismTerms have a space. Is it a good idea to have keys DOM property that contain space?

src/encoded/static/components/summary.js
```diff
@@ -1,7 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import queryString from 'query-string';
-import _ from 'underscore';
```
zoldello, line 4: Is this relevant to solving the ticket? I am not being sarcastic. No. I ask because I did not see how it contributes to solving the ticket and I would like to know just in case I missed it.

src/encoded/static/libs/ui/panel.js
```diff
@@ -272,14 +272,16 @@ class TabItem extends React.Component {
     }
 
     clickHandler() {
-        this.props.handleClick(this.props.tab);
+        if (!(this.props.children && this.props.children.props && this.props.children.props.className && this.props.children.props.className.includes('disabled'))) {
```
zoldello: I recommend the condition be put in a method as seeing a long check twice is a code smell.

src/encoded/static/scss/encoded/modules/_matrix.scss
```diff
@@ -3,6 +3,9 @@ $col-category-header-height: 200px;
 $row-category-cell-height: 22px;
 $row-data-header-width: 200px;
 
+$Homo-sapiens: #63326E;
```
zoldello: why is the variable name capitalized? Do you something more descriptive like "human-body-color:#63326E"?

src/encoded/static/scss/encoded/modules/_matrix.scss
```diff
@@ -15,6 +18,390 @@ $row-data-header-width: 200px;
...
+.epigenome-body-map-container {
+    border: 2px solid $Homo-sapiens;
+    position: relative;
+    .body-list-top {
+        border-top: 0;
+    }
+    &.expanded {
+        max-width: 675px;
+        position: absolute;
+        z-index: 99999;
```
zoldello: Is there a reason why you are making the z-index to the max?

src/encoded/static/scss/encoded/modules/_matrix.scss
```diff
@@ -15,6 +18,390 @@ $row-data-header-width: 200px;
...
+        }
+    }
+    .body-list-inner {
+        li {
+            display: inline;
+            &:after {
+                content: " ";
+                word-spacing: 0em;
+                letter-spacing: 1em;
+                background:url(data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAUAAAAFCAYAAACNbyblAAAABGdBTUEAALGPC/xhBQAAAAlwSFlzAAAOwAAADsABataJCQAAABl0RVh0U29mdHdhcmUAcGFpbnQubmV0IDQuMC4xMkMEa+wAAAAnSURBVBhXY/Dz89MA4sNA/B9Ka4AEYQIwfBgkiCwAxjhVopnppwEApxQqhnyQ+VkAAAAASUVORK5CYII=) center center no-repeat;
```
zoldello, line 63: This looks like a code smell. Shouldn't webpack do this for you?

src/encoded/static/scss/encoded/modules/_matrix.scss
```diff
@@ -15,6 +18,390 @@ $row-data-header-width: 200px;
...
+        .body-list-inner {
+            text-align: center;
+        }
+    }
+    .body-list-inner {
+        li {
+            display: inline;
+            &:after {
+                content: " ";
+                word-spacing: 0em;
```
zoldello: REQUIRED CHANGE. this should be: "word-space: 0;" I know it minor, but it causes a warning and technically its incorrect.

src/encoded/static/scss/encoded/modules/_matrix.scss
```diff
@@ -15,6 +18,390 @@ $row-data-header-width: 200px;
...
+                width: 55px;
+                height: 55px;
+            }
+        }
+        .body-list-narrow {
+            flex: 0 1 40%;
+        }
+    }
+    .body-facet :nth-child(1) {
+        order: 1;
```
zoldello, line 103: REQUIRED CHANGE: Based on- https://developer.mozilla.org/en-US/docs/web/css/order, 
is seems you need a second order with webkit prefix. Please see the link. It seems to be needed for Microsoft Edge pre-chromium (trident) and firefox.

src/encoded/static/scss/encoded/modules/_matrix.scss
```diff
@@ -15,6 +18,390 @@ $row-data-header-width: 200px;
...
+}
+
+.expand-arrows, .collapse-arrows {
+    height: 30px;
+    width: 30px;
+    position: absolute;
+    top: 43px;
+    left: 5px;
+}
+.expand-arrows {
```
zoldello, line 135: space needed?

src/encoded/static/scss/encoded/modules/_matrix.scss
```diff
@@ -15,6 +18,390 @@ $row-data-header-width: 200px;
...
+        width: 20px;
+    }
+    @media screen and (max-width: $screen-xs-min) {
+        top: 3px;
+    }
+}
+.body-map-expander, .body-map-collapser {
+    background: $Homo-sapiens;
+    text-align: center;
+    font-family: Mada;
```
zoldello, line 165: This is just for my personal curiosity. You and Forrest seems to have used this font. What is special about it that makes it appealing?

src/encoded/static/scss/encoded/modules/_matrix.scss
```diff
@@ -15,6 +18,390 @@ $row-data-header-width: 200px;
...
+    }
+    .tab-nav-Homosapiens {
+        border-bottom: 2px solid $Homo-sapiens;
+    }
+    .tab-nav-Musmusculus {
+        border-bottom: 2px solid $Mus-musculus;
+    }
+    .nav-tabs {
+        li {
+            margin-bottom: -1px;
```
zoldello, line 220: Negative margin seem like a code smell. Is there a reason an alternative is not sufficient?

src/encoded/static/scss/encoded/modules/_matrix.scss
```diff
@@ -15,6 +18,390 @@ $row-data-header-width: 200px;
...
+        }
+        @media screen and (max-width: $screen-sm-min) {
+            font-size: 0.9rem;
+            margin-bottom: 0;
+        }
+        @media screen and (max-width: $screen-xs-min) {
+            font-size: 0.8rem;
+            margin-bottom: 20px;
+        }
+        i {
```
zoldello, line 295: This is for my personal curiosity. Have you ever seen anyone remove the italics via CSS from i-html tag?

zoldello (changes requested):
Most of my comments are minor and can be ignored. Two though are required changes. I denoted the required ones with "REQUIRED CHANGE"

src/encoded/static/components/summary.js
emro: No, it's not relevant, but I was getting a linting error and wanted to fix it.

src/encoded/static/components/body_map.js
emro: I think that keeping this as "clearBodyMapSelectionsFromUrl" is the most clear, because the function clears both systems and organ fields from the url and "body map selections" encapsulates both.

src/encoded/static/components/body_map.js
emro: I should definitely think about using JSDocs. There's one thing I don't like about the syntax - it can be difficult to distinguish between the variable name and the definition/explanation of the variable - but I think there are a number of other benefits - the enforcement of defining input parameters and returned values seems very useful.

src/encoded/static/components/body_map.js
emro: They can appear multiple times, but this function removes all matching parameters: https://github.com/ENCODE-DCC/encoded/blob/dev/src/encoded/static/libs/query_string.js#L87

src/encoded/static/components/body_map.js
emro: I think they're actually not in React Class component and don't use `this`, though I might be missing something.

src/encoded/static/components/body_map.js
emro: Excellent point

src/encoded/static/components/body_map.js
emro: This is a very good point, and I think this code should definitely be fixed up. `this.state.selectedOrgan` should probably always be an array. I ran into some coding issues when I tried to implement it this way - it's less trivial than it seems like it should be. I'd like to fix it, but I think it may require some work and would prefer to do it as part of another ticket, just because this already has so much going on.

src/encoded/static/components/matrix_reference_epigenome.js
emro: I think I talked to @forresttanaka about this at some point and I don't remember exactly what we discussed, but I think we're mostly sticking to <React.Fragment> except for specific cases?

forresttanaka: We don’t have a style guide for this, and I already discussed this with Phil a while back. I think `<>` on its own line is ugly so I prefer `<React.Fragment>` for those cases, and I prefer `<>` when inline. But I don’t comment about it in code reviews because we have no guidance on it — whatever people like.

src/encoded/static/components/matrix_reference_epigenome.js
emro: I think I can completely eliminate the need for this to be an object, by just renaming the image files. I think that's the simplest solution here.

src/encoded/static/components/matrix_reference_epigenome.js
emro: I think this is supposed to be null, based on the return here: https://github.com/ENCODE-DCC/encoded/blob/4534124558358e33fec5ed399141146ff3ffcc4d/src/encoded/static/components/matrix_reference_epigenome.js#L622. I think I didn't change this line, I just added another state change to the setState.

src/encoded/static/components/matrix_reference_epigenome.js
emro: It is definitely getting more specialized all the time! I think Forrest is working on a ticket that should change a lot about how ViewControls is implemented.

src/encoded/static/components/matrix_reference_epigenome.js
emro: I think this will work a lot better as an array.

src/encoded/static/scss/encoded/modules/_matrix.scss
emro: The capitalization is because the class name is generated from organismTerms. I could add toLowerCase() or append `-color` on all the class names but I'd rather keep them more or less as permutations of the organism terms names to avoid making things too much more complicated.

Your comment reminded me that these colors exist in both `_summary.scss` and `matrix.scss` and that I should definitely avoid this duplication so I moved all the hard-coded organism colors to `_base.scss`.

src/encoded/static/scss/encoded/modules/_matrix.scss
emro: I think I added the font when I was making the mouse development matrix and I'm slowly propagating it through the site on more visual types of elements. In the newspaper world, I'd say it's a "graphics" font. This probably merits more discussion with the design group, but I started to use it because I find it visually appealing and easy to read. It was also sort of exploratory, to see what users and other team members would have to say about it. Overall it's nice to have a couple fonts on a site, because they allow you to come up with more interesting designs (you can have the one font for one page element and a different font for another, and that creates a distinction between the two elements without using say color or border or background or the other usual visual distinctions). Of course you don't want too many fonts either, or fonts that clash. I thought this one worked reasonably well for these purposes. Do you like the font?

src/encoded/static/scss/encoded/modules/_matrix.scss
emro: This is the pop-up. It needs to be in front! I can make it less than the max though.

src/encoded/static/components/matrix_reference_epigenome.js
emro: Just simplifying this.

src/encoded/static/scss/encoded/modules/_matrix.scss
emro: Thank you for catching this!

src/encoded/static/libs/ui/panel.js
emro: Excellent point.

src/encoded/static/components/matrix_reference_epigenome.js
emro: I think there's no need for this to be an object at all - I can just use an array and that will be simpler. Just need to rename the pngs.

src/encoded/static/components/matrix_reference_epigenome.js
emro: I've renamed "expandedThumbnail" to be "isThumbnailExpanded", hopefully this is clearer.

src/encoded/static/components/matrix_reference_epigenome.js
emro: I've renamed "expandThumbnail" to be "toggleThumbnail", hopefully this is clearer.

src/encoded/static/scss/encoded/modules/_matrix.scss
emro: This was really unreasonably tricky, getting the dots to only show in-between list elements and not on line ends.

src/encoded/static/scss/encoded/modules/_matrix.scss
emro: I increased the border width and therefore added the negative margin to not have an extra line at the bottom of the tab.

src/encoded/static/scss/encoded/modules/_matrix.scss
emro: No, I'm not sure to what you're referring?

src/encoded/static/components/matrix_reference_epigenome.js
emro: Hm, let me know if you have an idea of something to change it to. It's a thumbnail (small image) which is clickable (expandable) so that's why I called it this.

src/encoded/static/components/body_map.js
```diff
@@ -237,6 +237,10 @@ const highlightOrgan = (e) => {
     }
 };
 
+// Checks to see if "b" (string) is included in "selectedOrgan" (string or list)
+// "checkClass" is used to set an active class on buttons or images based on whether or not they are a selected organ
+const checkClass = (selectedOrgan, b) => ((typeof selectedOrgan === 'string' && selectedOrgan === b) || (typeof selectedOrgan !== 'string' && selectedOrgan.includes(b)));
```
zoldello: You need not worry about this. Your way is also probably clearer. 

Pardon me if this appears patronizing. It is not my intent.

I assume by "list" you mean "array".  Anyway, a string is an array-like object  (like an array of characters) and you can do array-like operations of it. For example, both of these return true

'xyz'.includes('xyz') // string
['xyz'].includes('xyz') //array

So in theory, you need not typecheck. Pardon me, I missed that in my previous review. On the downside, I have not done this much with using strings as an array other than iterator over a string. So, I do not know the edge cases.

'b' is a good when you pass as argument (what you pass in when a function/method is called). But as parameters (what you use in a function/method definition), it may be unclear

src/encoded/static/components/body_map.js
```diff
@@ -617,7 +621,7 @@ class BodyMap extends React.Component {
                         {Object.keys(SystemsList).map(b =>
                             <li key={b}>
                                 <button
-                                    className={`body-list-element ${((typeof this.state.selectedOrgan === 'string' && this.state.selectedOrgan === b) || (typeof this.state.selectedOrgan !== 'string' && this.state.selectedOrgan.includes(b))) ? 'active' : ''}`}
+                                    className={`body-list-element ${checkClass(this.state.selectedOrgan, b) ? 'active' : ''}`}
```
zoldello, line 624: Awesome.

src/encoded/static/libs/ui/panel.js
```diff
@@ -264,6 +264,10 @@ TabPanel.defaultProps = {
 };
 
 
+// Check to see if child of tab has "disabled" class
+const checkIfDisabled = (input) => input.children && input.children.props && input.children.props.className && input.children.props.className.includes('disabled');
```
zoldello, line 268: Since this is an internal class, I assume you can at least trust that no one will give you a null "input" parameter to deal with.

zoldello (dismissed):
Code looks good. Required changes were implemented. My comments are opinions and need not be addressed.

src/encoded/static/components/body_map.js
emro: Oh yes, this can be simpler. I think I needed `typeof` somewhere else because spread syntax was spreading my arrays (when I combined them) into things like ["h", "e", "a", "r", "t", "lungs"] but here there is no need! Thank you!

emro: I almost feel like this doesn't need to be a method now that it's so short, but I think it's fine in either case.

src/encoded/static/scss/encoded/modules/_matrix.scss
zoldello: I missed this comment. i HTML used to be used for italic. That was before CSS took over that role. People latter highjacked it and started using it for icons. I was just curious if you see people remove the italic-functionality in it.

src/encoded/static/components/body_map.js
```diff
@@ -237,9 +237,9 @@ const highlightOrgan = (e) => {
     }
 };
 
-// Checks to see if "b" (string) is included in "selectedOrgan" (string or list)
-// "checkClass" is used to set an active class on buttons or images based on whether or not they are a selected organ
-const checkClass = (selectedOrgan, b) => ((typeof selectedOrgan === 'string' && selectedOrgan === b) || (typeof selectedOrgan !== 'string' && selectedOrgan.includes(b)));
+// Checks to see if "organ" is included in "selectedOrgan"
+// checkClass is used to set an active class on buttons or images based on whether or not they are selected
+const checkClass = (selectedOrgan, organ) => selectedOrgan.includes(organ);
```
zoldello, line 242: Great. This looks cleaner to me now.

zoldello (dismissed):
Looks good.

emro:
There were merge conflicts with ENCD-5336, now fixed and updated.

zoldello (approved):
Changes look good

PR #3367, opened 2020-06-09
ENCD-5241-search-scroll

forresttanaka:
You can attach a new `data-noscroll` attribute to `<a>` elements which, when the user clicks the link, navigates to the given href but does _not_ scroll to the top of the page:

```
<a href="#" data-noscroll>Click</a>
```

You can pass a corresponding new option to `context.navigate`:

```
this.context.navigate(href, { noscroll: true });
```

All the normal text facet terms now use the data-noscroll attribute, and the more unusual facets (e.g. “exists” facets) use the `noscroll` option for `navigate`. It would be interesting to see how this works for the body map.

### DIFF
--- a/src/encoded/static/components/app.js
+++ b/src/encoded/static/components/app.js
@@ -844,8 +844,7 @@ class App extends React.Component {
         }
 
         // data-noscroll attribute prevents scrolling to the top when clicking a link.
-        const dataNoScroll = target.getAttribute('data-noscroll');
-        if (dataNoScroll !== null) {
+        if (target.getAttribute('data-noscroll') !== null) {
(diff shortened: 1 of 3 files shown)


forresttanaka:
Demo for PR: https://encd-5241-search-scroll-dv2-fytanaka.demo.encodedcc.org/

emro (dismissed):
This is a really excellent solution. It's wonderfully straightforward and easy to use, and will make the site work much better. 🎉

forresttanaka:
New PR has the same changes as last time, minus the actual usage in the facets.

emro (approved):
Still all good!

PR #3368, opened 2020-06-09
ENCD-5255-add-encode4-atac-seq-qc-audits

iyoungworth:


src/encoded/audit/experiment.py
```diff
@@ -550,7 +551,8 @@ def audit_experiment_standards_dispatcher(value, system, files_structure):
                                             'whole-genome shotgun bisulfite sequencing',
                                             'genetic modification followed by DNase-seq',
                                             'microRNA-seq',
-                                            'long read RNA-seq', 'icSHAPE']:
+                                            'long read RNA-seq', 'icSHAPE',
+                                            'ATAC-seq']:
```
hitz, line 555: usually in a long list like this you put the ]: on the last line by itself.
So you can just append:
'thing',
(in python... it's illegal to have a dangling , in JS for example.
Don't need to fix this, just FYI.

src/encoded/audit/experiment.py
```diff
@@ -620,6 +622,12 @@ def audit_experiment_standards_dispatcher(value, system, files_structure):
             desired_assembly)
         return
 
+    if value['assay_term_name'] == 'ATAC-seq':
```
hitz, line 625: Is this going to check ENCODE3 ATAC-seq under ENCODE4 standards?

src/encoded/tests/fixtures/schemas/atac_alignment_quality_metric.py
```diff
@@ -0,0 +1,17 @@
...
+def atac_alignment_quality_metric_low(testapp, award, encode_lab,
+                                        analysis_step_run_atac_encode4_alignment,
+                                        ATAC_bam):
+    item = {
+        "step_run": analysis_step_run_atac_encode4_alignment['@id'],
+        "award": award["uuid"],
+        "lab": encode_lab["uuid"],
+        "assay_term_name": 'ATAC-seq',
+        "quality_metric_of": [ATAC_bam['@id']],
+        "pct_mapped_reads": 76,
```
hitz, line 14: shouldn't this be a float, 76.0?

hitz: I guess there are plenty of these in the chip fixtures.

hitz (dismissed):

src/encoded/audit/experiment.py
forresttanaka: Oh, the AirBnB style guide _requires_ a dangling comma in JS.

src/encoded/audit/experiment.py
iyoungworth: I think it should check any ATAC-seq, but only when the pipeline run is ours (either 'ATAC-seq (replicated)' or 'ATAC-seq (unreplicated)')

hitz (approved):

PR #3370, opened 2020-06-10
ENCD-5336-matrix-desc

forresttanaka:
It’s not pretty, but I think we need to get all the matrix view headers back in line in the future, and we can design these descriptions in from the start when we do that.


src/encoded/static/components/matrix_experiment.js
```diff
@@ -292,12 +292,21 @@ const MatrixHeader = ({ context }) => {
         }
     }
 
+    // If the user has requested an ENCORE matrix, generate a matrix description.
+    const query = new QueryString(context.search_base);
+    const matrixDescription = query.getKeyValues('internal_tags').includes('ENCORE') ?
```
zoldello, line 297: Will ENCORE ever be lower case?

src/encoded/static/components/matrix_reference_epigenome.js
```diff
@@ -368,12 +368,26 @@ const MatrixHeader = ({ context, showProjects, project }, reactContext) => {
 
     const rowCount = matrix => matrix.y['biosample_ontology.classification'].buckets.reduce((currentTotalCount, termName) => currentTotalCount + termName['biosample_ontology.term_name'].buckets.length, 0);
 
+    const query = new QueryString(context.search_base);
+    let organism = '';
+    if (query.getKeyValues('replicates.library.biosample.donor.organism.scientific_name').includes('Mus musculus')) {
```
zoldello, line 373: I assume the organisms will never be more than human and mouse. If true, can you use conditional operator (: ?)? Also, organism too could be the const.

src/encoded/static/components/matrix_reference_epigenome.js
```diff
@@ -368,12 +368,26 @@ const MatrixHeader = ({ context, showProjects, project }, reactContext) => {
...
+        organism = 'mouse';
+    } else if (query.getKeyValues('replicates.library.biosample.donor.organism.scientific_name').includes('Homo sapiens')) {
+        organism = 'human';
+    }
+    const matrixDescription = organism ?
+        <span>Project data from {organism} tissue, cell line, primary cell, and in vitro differentiated cell biosamples organized as reference epigenomes following guidelines set out by IHEC.</span>
```
zoldello, line 379: Look at the reference epigenome page. The radio button on the far left are displaced. A similar issue occurs on the ENCORE page. I do not think that is intended.

zoldello (dismissed):
Code-wise the code looks fine. If you go to the UI, the description seem to displace the text on the far left. Is that by design?

src/encoded/static/components/matrix_reference_epigenome.js
forresttanaka: That is intended. As I said this is interim, and I think it’s likely once we get all these matrix headers back to consistency, we’ll keep the title, internal tags badge, and the description all in one place at the top, and all controls by the matrix itself, instead of mixing controls and titles.

src/encoded/static/components/matrix_experiment.js
forresttanaka: That's a good point, but through our menu it could never be lower case, and Ben has said if the user modifies the URL, they’re on their own.

forresttanaka:
Modified the Data menu titles for ENCORE and ENTEx according to Ben’s request in the ticket.

zoldello (dismissed):
Update looks fine.

forresttanaka:
@zoldello Sorry, another text update that cascaded to the tests.

zoldello (approved):
Cool looks good.

PR #3371, opened 2020-06-11
ENCD-5341 update deployment tags

caseylitton:


keenangraham (dismissed):

keenangraham:
Could also have an auto_resize with specified instance?

caseylitton:
> Could also have an auto_resize with specified instance?

Added auto_resize = c5.4xlarge .  Will = 'na' when we don't want to resize.

src/encoded/commands/deploy.py
```diff
@@ -98,34 +98,98 @@
...
+        profile_name='default',
+        dry_run=False,
+        arm_arch=False,
+        cluster_master=False,
+    ):
+    # Defaults to demo development
+    tags_dict = {
+        # Org
+        'project': 'encoded',
+        'section': 'app',
```
hitz, line 123: what is "section" supposed to represent?

src/encoded/commands/deploy.py
```diff
@@ -848,6 +945,7 @@ def hostname(value):
         dest='role',
         help="Set role"
     )
+    parser.add_argument('--is-qa-demo', action='store_true', help="Flagged as qa demo")
```
hitz, line 948: does this actually do anything but tag the instance? (apparently true!)

hitz (approved):

PR #3372, opened 2020-06-11
ENCD-5343-adjusting-internal-status-property

iyoungworth:


src/encoded/upgrade/dataset.py
```diff
@@ -612,3 +612,11 @@ def annotation_28_29(value, system):
         else:
             value['notes'] = info
     return
+
+
+@upgrade_step('experiment', '28', '29')
+def experiment_28_29(value, system):
+    # https://encodedcc.atlassian.net/browse/ENCD-5343
+    if value.get('internal_status') in ['requires lab review', 'unrunnable']:
+        value['internal_status'] = 'unreviewed'
```
iyoungworth: This is slightly cheating- I tried doing value.pop to remove internal_status, which on the portal I think should automatically move it back to internal_status: unreviewed as the default. But I couldn't figure out a way in the test_upgrade_dataset to match this behavior (I think I'd need to post to the testapp)

hitz (approved):

PR #3374, opened 2020-06-13
ENCD-4350-make-antibody-label-clearer

zoldello:


src/encoded/static/components/antibody.js
```diff
@@ -347,17 +347,20 @@ const AntibodyStatus = (props) => {
                             {Object.keys(organisms).map((organism, i) => {
                                 const terms = Object.keys(organisms[organism]);
                                 return (
-                                    <div key={i} className="antibody-status__group">
-                                        <div className="antibody-status__organism">
-                                            {organism}
+                                    <>
+                                        {i !== 0 ? <div className="antibody-status__group-gutter" /> : null }
```
forresttanaka: I think this could be done more easily in CSS by assigning a bottom margin to `antibody-status__group`, and then using `antibody-status__group:last-child` to have a 0 bottom margin. Saves a (granted, not very expensive) extra div, and then the only Javascript change is the `<Status>` line.

forresttanaka (changes requested):

src/encoded/static/components/antibody.js
zoldello: Good catch.

src/encoded/static/components/antibody.js
```diff
@@ -347,17 +347,19 @@ const AntibodyStatus = (props) => {
...
+                                            </div>
                                         </div>
-                                        <div className="antibody-status__terms">
-                                            {terms.length === 1 && terms[0] === 'not specified' ? '' : terms.join(', ')}
-                                        </div>
-                                        <div className="antibody-status__status">
-                                            {i === 0 ? <Status item={status} inline /> : null}
-                                        </div>
-                                    </div>
+                                    </>
```
forresttanaka: Oh just one more and it’s good. Since the antibody-status__group div is the only component within the `<>` and `</>`, we don’t need the `<>` and `</>` at all.

forresttanaka (changes requested):

src/encoded/static/components/antibody.js
zoldello: Thanks. Maybe this was a hangout from a previous change. I removed the react fragment.

forresttanaka (approved):

PR #3381, opened 2020-06-19
ENCD-5363-Fix-tags-on-cluster-deployments

caseylitton:


src/encoded/commands/deploy.py
```diff
@@ -179,7 +183,7 @@ def _tag_ec2_instance(
         tags_dict['elastic_ip_swtich_datatime'] = 'na'
         if cluster_master:
             tags_dict['elasticsearch_head'] += 'yes'
-            tags_dict['Role'] += '-datahead'
+            tags_dict['Role'] += 'head'
```
hitz, line 187: is this supposed to be -head?  I guess it doesn't matter that much.

hitz (dismissed):

src/encoded/commands/deploy.py
```diff
@@ -179,7 +183,7 @@ def _tag_ec2_instance(
         tags_dict['elastic_ip_swtich_datatime'] = 'na'
         if cluster_master:
             tags_dict['elasticsearch_head'] += 'yes'
```
hitz: This line (185) should be = 'yes' not += 'yes'

src/encoded/commands/deploy.py
caseylitton: no, the -data already exists.  Previously the role came out as -datadatahead

src/encoded/commands/deploy.py
caseylitton: yes, this is fixed along with a couple other bugs in the current pr.  There is also a read me now.

Bek: Looks good!

Bek (dismissed):
Verified these tags and resolution of the 'noyes' issue.

PR #3385, opened 2020-06-22
SNO-157-make-page-facets-collapsable

zoldello:


src/encoded/static/components/facets/defaults.js
```diff
@@ -1010,41 +1010,62 @@ export const DefaultFacet = ({ facet, results, mode, relevantFilters, pathname,
         setInitialState(false);
     }, [handleScrollShading, facet, typeaheadTerm]);
 
+    React.useEffect(() => {
+        setFieldAsNotNewlyLoaded(facet, relevantFilters);
+
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
zoldello: Let me know if this is a concern. I need this to run only once and this was a way to do it and I had to silent ESLint.

src/encoded/tests/features/customsteps.py
```diff
@@ -55,6 +55,26 @@ def click_element(browser, css):
     elements.first.click()
 
 
+@when(parse('I click all elements with the css selector "{css}"'))
```
zoldello, line 58: Let me know if this is a concern. I based this code off of "I click element with the css selector {css}"; but I excluded the assert statement.

src/encoded/tests/features/search.feature
```diff
@@ -11,16 +11,23 @@ Feature: Search
...
     Scenario: Search Antibodies
         When I press "Materials & Methods"
+        And I wait for the content to load
         And I click the link to "/search/?type=AntibodyLot&status=released"
         And I wait for the content to load
         Then I should see at least 7 elements with the css selector "ul.result-table > li"
         And I should see at least 5 elements with the css selector "div.box.facets > div.orientation > div.facet"
 
-        When I click the link to "?type=AntibodyLot&status=released&targets.organism.scientific_name=Homo+sapiens"
+        When I click all elements with the css selector "[data-test='facetcontainer'] .facet__expander--header[aria-pressed='true']"
```
zoldello, line 20: This closes all opened facets. I want this not to fail if no facet was opened (the pre-opened facets could change at anytime). So, I made a new customstep and excluded the assert len(element) == 1.

src/encoded/tests/features/search.feature
```diff
@@ -60,41 +72,79 @@ Feature: Search
...
+        And I click all elements with the css selector "[data-test='facetcontainer'] .facet__expander--header[aria-pressed='true']"
+        And I wait for the content to load
+        And I click the element with the css selector ".facet__expander--header[aria-label='assay_title']"
+        And I wait for the content to load
+        And I click the link to "?type=Experiment&assay_title=TF+ChIP-seq"
         And I wait for the content to load
         Then I should see "Showing 9 of 9 results"
 
-        When I click the link to "?type=Experiment"
+        When I press "Data"
```
zoldello, line 126: I had a hard time gettin the old code to work. So, I re-wrote it to be more explicit with what its doing

src/encoded/static/components/search.js
```diff
@@ -700,12 +705,94 @@ TextFilter.propTypes = {
 // Displays the entire list of facets. It contains a number of <Facet> components.
 export const FacetList = (props) => {
     const { context, facets, filters, mode, orientation, hideTextFilter, addClasses, docTypeTitleSuffix, supressTitle, onFilter } = props;
+
+    const [expandedFacets, setExpandFacets] = React.useState(new Set());
```
zoldello, line 709: It seemed as if passing () => {} makes no difference.

Lines 275 and 329 of
https://github.com/facebook/react/blob/df14b5bcc163516fc0f1ad35e9b93732c66c1085/packages/react-dom/src/server/ReactPartialRendererHooks.js#L275

https://stackoverflow.com/questions/55621212/is-it-possible-to-react-usestate-in-react

zoldello (commented):
@forresttanaka It would of had been hard to make ENCD-5031 work with the required changes and fixing a merge conflict. So, I made a new branch. Since there was a snovault change, I had to name the branch according. I made a mistake with the name so there are closed branches with "SNO-157" in their name. Please, ignore those.

src/encoded/static/components/facets/defaults.js
```diff
@@ -1010,41 +1010,62 @@ export const DefaultFacet = ({ facet, results, mode, relevantFilters, pathname,
...
+    }, []);
+
     // Callback to handle typeahead input events.
     const handleTypeAhead = (event) => {
         setTypeaheadTerm(event.target.value);
     };
 
+    const isExpanded = expandedFacets && expandedFacets.has(facet.field);
```
forresttanaka: Would this be easier if all facet components receive an `isExpanded` boolean, instead of the entire set and they have to do this check?

src/encoded/static/components/facets/defaults.js
forresttanaka: It’s a concern, as these eslint escapes can hide future bugs. But I do see the use case here, and I saw this simple custom hook to get around the issue without escaping ESLint: https://stackoverflow.com/questions/53120972/how-to-call-loading-function-with-react-useeffect-only-once/56767883#56767883.

src/encoded/static/components/search.js
forresttanaka: I agree now that this is fine. I’m not clear it makes _no_ difference, but it doesn’t appear to make any significant difference.

zoldello:
@forresttanaka Thanks for reviewing the snovault ticket. Also, please, do not forget the ENCODE-part of it :)

src/encoded/static/components/search.js
```diff
@@ -700,12 +705,94 @@ TextFilter.propTypes = {
...
+
+    React.useEffect(() => {
+        // Get facets from storage that need to be expanded
+        const facetsStorage = sessionStorage.getItem(FACET_STORAGE);
+        const facetList = new Set(facetsStorage ? facetsStorage.split(',') : []);
+
+
+        sessionStorage.setItem(FACET_STORAGE, facetList.size !== 0 ? [...facetList].join(',') : []);
+        setExpandFacets(facetList); // initalize facet collapse-state
+    }, []);
```
forresttanaka, line 718: I think this initial loading of open facets can be done more simply, and without the default or custom facets needing to detect rendering themselves. Right now `facetList` is empty the first time you load the site, but we have the information from `facet.open_on_load` by this time so you can initialize `facetList` with that here Then each facet gets re-rendered knowing whether it’s open or not, and session storage gets updated properly too. Nice idea on session storage — I probably would have used local storage, but session storage makes more sense for this.

With that handled here, you won’t need each facet to tell you that they’ve rendered and need to re-render either open or not, because this first useEffect will tell them to re-render, open or not. That way you don’t need the `setFieldAsNotNewlyLoaded` function, nor its callback prop.

forresttanaka:
Very cool! I think the mechanism is good, but I have a suggestion for what I _think_ would constitute a significant simplification. To be honest, I even tried a very messy version of this simplification, and it worked in the 30 seconds of testing I subjected it to.

forresttanaka (changes requested):

src/encoded/static/components/search.js
zoldello: I think "setFieldAsNotNewlyLoaded" needs to be commented better and it not properly named. When the page first loads, any facets with a selected item should be opened. For example 

http://localhost:6543/search/?type=Experiment&award.project=ENCODE

should have project opened. However, it should not be opened on subsequent clicks. "setFieldAsNotNewlyLoaded" helps with that as it is located somewhere where it can be called once. I placed facet.open_on_load check in this method too because placing it on the consuming affect resulted in it being called multiple times. Please, let me know if my idea makes no sense.

forresttanaka: What do you mean by “should not be opened on subsequent clicks”?

src/encoded/static/components/facets/defaults.js
zoldello: I do not think I get what you mean. Each facets has a different facet and facet.field; so only consumers can set isExpanded.

forresttanaka: All each `DefaultFacet` instance needs to know is whether itself should render expanded or not. It doesn’t need to know whether all facets need to appear expanded, and this code doesn’t check whether they all appear expanded. This code only checks if this particular facet appears expanded, so it only needs that one boolean. `FacetList` can check within the loop if `expandedFacets.has(facet.field)` return true and pass the result to each facet.

src/encoded/static/components/facets/defaults.js
zoldello: This code has been removed.

src/encoded/static/components/facets/defaults.js
zoldello: This code has been deleted.

src/encoded/static/components/search.js
zoldello: Great tip. The code was refactored and now is simplified.

src/encoded/static/components/search.js
```diff
@@ -700,12 +705,89 @@ TextFilter.propTypes = {
...
     const { context, facets, filters, mode, orientation, hideTextFilter, addClasses, docTypeTitleSuffix, supressTitle, onFilter } = props;
+
+    const [expandedFacets, setExpandFacets] = React.useState(new Set());
+
+    // Get facets from storage that need to be expanded
+    React.useEffect(() => {
+        const facetsStorage = sessionStorage.getItem(FACET_STORAGE);
+        const facetList = new Set(facetsStorage ? facetsStorage.split(',') : []);
+
+        sessionStorage.setItem(FACET_STORAGE, facetList.size !== 0 ? [...facetList].join(',') : []);
```
forresttanaka, line 716: I think we talked about this before, but I don’t remember the reason. Why `[...facetList].join(',')` instead of `facetList.join(',')`?

src/encoded/tests/features/customsteps.py
forresttanaka: I think this is pretty cool.

forresttanaka (dismissed):
Thanks for making those changes. I think this looks very maintainable. And nice changes to the BDD tests.

src/encoded/static/components/search.js
zoldello: facetList is a Set and it does not have join-method. I used [...faceList] to convert faceList to an array then use .join on that. I made it a Set to remove duplicates. Is that acceptable to you?

forresttanaka: Oh, I see. Yeah, that seems good then.

zoldello:
@forresttanaka  :) Back in code review to make collapse optional. It is disabled for summary pages with horizontal facets.

forresttanaka (dismissed):

zoldello:
@forresttanaka  One more quick change to make Assay title opened on load but not Assay Type

forresttanaka (approved):

PR #3390, opened 2020-06-23
ENCD-5358 fix body map thumbnail

emro:
Two fixes:
(1) Fix browser inconsistency (Safari was displaying extra bullets)
(2) Remove searchTerm search bar

### DIFF
--- a/src/encoded/static/components/body_map.js
+++ b/src/encoded/static/components/body_map.js
@@ -620,15 +620,18 @@ class BodyMap extends React.Component {
                     <ul className="body-list-inner">
                         {Object.keys(SystemsList).map(b =>
                             <li key={b}>
-                                <button
-                                    className={`body-list-element ${checkClass(this.state.selectedOrgan, b) ? 'active' : ''}`}
+                                <span
                                     id={b}
+                                    className={`body-list-element ${checkClass(this.state.selectedOrgan, b) ? 'active' : ''}`}
+                                    role="button"
+                                    tabIndex="0"
                                     onClick={e => this.chooseOrgan(e)}
+                                    onKeyPress={e => this.chooseOrgan(e)}
                                     onMouseEnter={e => highlightOrgan(e)}
                                     onMouseLeave={unHighlightOrgan}
                                 >
                                     {b}
-                                </button>
+                                </span>
                             </li>
                         )}
                         <button className="clear-organs" onClick={this.clearOrgans}>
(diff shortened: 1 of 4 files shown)


src/encoded/static/scss/encoded/modules/_matrix.scss
```diff
@@ -271,6 +273,27 @@ $row-data-header-width: 200px;
...
+            font-size: 0.8rem;
+        }
+        .bold-total {
+            font-size: 1.1rem;
+            @media screen and (max-width: $screen-xs-min) {
+                font-size: 0.9rem;
+            }
+        }
+        text-align: left;
+    }
```
forresttanaka, line 296: Shouldn’t cause general problems, but just so you’re aware we’re now mixing desktop first and mobile first styles on encoded. Hopefully it won’t cause too many problems down the line. If we redo all the CSS going forward I would choose desktop first, but mobile first was the style of the day and that’s what bootstrap used which is why they used constants with “min” at the end.

src/encoded/static/components/body_map.js
```diff
@@ -620,15 +620,18 @@ class BodyMap extends React.Component {
...
+                                    role="button"
+                                    tabIndex="0"
                                     onClick={e => this.chooseOrgan(e)}
+                                    onKeyPress={e => this.chooseOrgan(e)}
                                     onMouseEnter={e => highlightOrgan(e)}
                                     onMouseLeave={unHighlightOrgan}
                                 >
                                     {b}
-                                </button>
+                                </span>
```
forresttanaka, line 634: The button was causing the Safari bullet issue?

forresttanaka (approved):
I think the test failure is probably transient. I don’t see that on other branches.

src/encoded/static/components/body_map.js
emro: Yes, ugh. It was some built-in button style that I couldn't figure out how to remove. I tested navigation by tab and I think I managed to get everything working properly without the built-in buttons, though.

src/encoded/static/scss/encoded/modules/_matrix.scss
emro: As I was fixing some minor CSS things this morning I was starting to think that the CSS for the body map is generally pretty messy and should be cleaned up. The trouble is that there is CSS for the body map as displayed on the summary page and different CSS for the body map as a thumbnail, and it should probably all go in a new file (instead of _summary.scss and _matrix.scss) and be clearly identifiable for each implementation. I think if we put the thumbnail on any other page this should be cleaned up for sure. I could also do it as part of this ticket.

forresttanaka: I think the body map is big and isolated enough concept, a new scss file would be a good idea.

PR #3391, opened 2020-06-23
ENCD-5252-add-index-reads-audit

iyoungworth:


src/encoded/tests/test_audit_file.py
```diff
@@ -912,3 +912,123 @@ def test_audit_self_matching_md5sum(testapp, file7):
...
                for error in errors_list)
+
+
+def test_audit_incorrect_index(testapp, fastq_index, single_fastq_indexed, base_experiment):
+    res = testapp.get(fastq_index['@id'] + '@@index-data')
+    errors = res.json['audit']
+    errors_list = []
+    for error_type in errors:
+        errors_list.extend(errors[error_type])
+    assert 'inconsistent index file' not in (error['category']
```
hitz, line 929: this is the test of a correct audit, right?   Should be a comment or something because the test is still named "incorrect_index"

src/encoded/tests/test_audit_file.py
```diff
@@ -912,3 +912,123 @@ def test_audit_self_matching_md5sum(testapp, file7):
...
+        {
+            'dataset': base_experiment['@id']
+        }
+    )
+    res = testapp.get(fastq_index['@id'] + '@@index-data')
+    errors = res.json['audit']
+    errors_list = []
+    for error_type in errors:
+        errors_list.extend(errors[error_type])
+    assert any(error['category'] == 'inconsistent index file'
```
hitz: I feel like you are not testing the detail message to match up the error state to the error message.

hitz (changes requested):
I think you are missing a couple test cases and this code is very complicated; there are 7 error states and 2 "positive" controls (SE case and PE case).   I only count 6 error states and 1 non-error state.

Actually maybe the non-error state is just a pre-error state; it's hard to tell.  
Also - I think you should test the "detail" message for each fail case so that it will be clear which assertion is testing which case; it would be pretty easy to have made a mistake either in the audit code OR the test code and have a false passing test.   Plus, we don't expect many REAL cases so we'll never know...

src/encoded/tests/test_audit_file.py
iyoungworth: I reorganized and wrote more tests to make this clear- test_audit_correct_index gives 3 acceptable states (1 SE fastq, 2 PE fastq paired with each other, 1 PacBio fastq); test_audit_incorrect_index gives all 7 incorrect error states. Some of the error states could have been combined into fewer tests, but the each error is absolutely explicit as written (in the new version)

src/encoded/tests/test_audit_file.py
iyoungworth: Each error state now includes a check for a unique portion of the relevant error detail

src/encoded/tests/test_audit_file.py
iyoungworth: Also it was definitely worth going through more explicitly, the way I was finding 2 PacBio files wasn't quite correct before! But that should be fixed now, it's the only new change to the actual audit code

hitz (dismissed):

hitz (approved):

PR #3397, opened 2020-06-30
ENCD-5373 fix mouse development matrix

emro:
There is new mouse development data and that data has some keys which contain the string "embryo" repeated. In order to determine the correct matrix labels, I parse strings based on "embryo", which does not work as written for this new data. The data will be updated eventually but this is a quick fix in the meantime.

### DIFF
--- a/src/encoded/static/components/matrix_mouse_development.js
+++ b/src/encoded/static/components/matrix_mouse_development.js
@@ -64,6 +64,13 @@ const mouseMatrixColors = {
 };
 
 /**
+ * Parse string (longString) on potentially repeated substring (splitString)
+ */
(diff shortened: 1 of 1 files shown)


forresttanaka (dismissed):
Wow; I never would have expected “embryo embryo.”

emro:
It turned out this needed the string parsing change in a number of other places as well. I think I caught them all. Also there was one small change to use `screen.width` instead of `window.innerWidth` because for some reason `window.innerWidth` wasn't getting the right value and the legend wasn't displaying properly on my phone.

forresttanaka (dismissed):

forresttanaka (approved):

PR #3399, opened 2020-06-30
SNO-158-create-minimally-indexed-deploy

keenangraham:


paul-sud (dismissed):
Looks fine. I do wonder if there is a use case for non-demos with small DBs, in which case the deploy code here would need work, but I don't think it's a big deal.

PR #3401, opened 2020-07-01
ENCD-5359 add hover definitions

emro:
This ticket adds hoverable glossary definitions to the navigation.
A couple of comments / questions about the code:

(1) Is the location for glossary.json good? We could move it elsewhere if it makes more sense somewhere else. (Related note: it seems likely that we will want to generate the glossary page from this glossary.json.)

(2) Is setTimeout working as expected?

(3) I removed onMouseEnter and onMouseLeave from line 88 (https://github.com/ENCODE-DCC/encoded/compare/ENCD-5359-add-hover-definitions?expand=1#diff-c17ab4255c252d6de7bdadac71607f2fL88) because they didn't seem necessary.

(4) I needed a special case for mobile to work the way I wanted it to on the navigation - the wrapper function was preventing the styles from working the way I wanted them to - so I ended up with different implementations for desktop and mobile. Could you think of a simpler way to do this?


src/encoded/static/components/navigation.js
```diff
@@ -36,6 +36,15 @@ HomeBrand.contextTypes = {
...
 
+// Check if device has touch events
+function isMobile() {
+    try {
+        document.createEvent('TouchEvent');
+        return true;
+    } catch (evt) {
+        return false;
+    }
+}
```
forresttanaka: I think this would be a natural extension for browserfeat.js so that the entire code base has access to it. Then the `<html>` tag gets a class on touch-enabled browsers (Modernizr uses the 'touchevents' class: https://modernizr.com/docs and I’ve mostly copied Modernizr class names) so we can use different CSS depending on touchevents. And then from Javascript you can call `BrowserFeat.getBrowserCaps('touchevents'))` to work like this `isMobile`.

src/encoded/static/components/navigation.js
```diff
@@ -217,7 +236,29 @@ const GlobalSections = (props, context) => {
                         if (childAction.id.substring(0, 4) === 'sep-') {
                             return <DropdownMenuSep key={childAction.id} />;
                         }
-
+                        const glossaryMatch = glossary.find(def => def.term === childAction.title);
```
forresttanaka, line 222: I was wondering if the glossary.js entries should include an id, and then match on the ID instead of the title, and add an id to every element that gets a tooltip. If we eventually use this for other things, titles might have conflicts.

src/encoded/static/components/navigation.js
```diff
@@ -47,12 +56,14 @@ export default class Navigation extends React.Component {
             openDropdown: '',
             /** True if test warning banner visible; default depends on domain */
             testWarning: !productionHost[url.parse(context.location_href).hostname],
+            isMobileDevice: false,
```
forresttanaka: I wonder if you could go back to _not_ having this a state, as I think this could cause extra rerenders. I see you use this state to pass to Tooltip, but if you make 'touchevents' a browserfeat property, then Tooltip itself could check whether it gets rendered on a mobile device or not. And Tooltips being visible and needing rerendering _during_ the time you determine whether this runs on a mobile browser or not seems unlikely or impossible. Lately I’ve been trying to reduce the number of states to only the unavoidble ones to reduce extra rerenders.

src/encoded/static/libs/ui/tooltip.js
```diff
@@ -14,107 +14,95 @@ import PropTypes from 'prop-types';
...
-        this.handleTooltipMouseLeave = this.handleTooltipMouseLeave.bind(this);
-        this.handleTriggerFocus = this.handleTriggerFocus.bind(this);
-        this.handleTriggerBlur = this.handleTriggerBlur.bind(this);
-    }
-
-    handleTriggerMouseEnter() {
-        // Mouse entered the tooltip trigger element.
-        this.setState({ tooltipDisplayed: true });
+const Tooltip = (props) => {
+    const [showDefinition, setDefinition] = React.useState(false);
```
forresttanaka, line 17: Yay!

src/encoded/static/libs/ui/tooltip.js
```diff
@@ -14,107 +14,95 @@ import PropTypes from 'prop-types';
...
+            >
+                {trigger}
+            </button>
+            {showDefinition ?
+                <div
+                    className={tooltipCss}
+                    role="tooltip"
+                    id={tooltipId}
+                >
+                    <div className="tooltip-arrow" />
```
forresttanaka, line 90: Does "tooltip-arrow" actually do anything? I’ve felt like it hasn’t in a long time, if ever.

src/encoded/static/libs/ui/tooltip.js
```diff
@@ -14,107 +14,95 @@ import PropTypes from 'prop-types';
...
     position: 'bottom',
     css: '',
+    innerCss: '',
     children: null,
+    timerFlag: true,
+    relativeTooltipFlag: false,
+    isMobileDevice: false,
 };
+
+export default Tooltip;
```
forresttanaka, line 118: This looks really good. Thanks for the long-needed rewrite.

src/encoded/static/components/navigation.js
```diff
@@ -205,9 +224,9 @@ Navigation.childContextTypes = {
     dropdownClick: PropTypes.func, // Called when a dropdown title gets clicked
 };
 
-
 // Main navigation menus
 const GlobalSections = (props, context) => {
+    const glossary = require('./glossary/glossary.json');
```
forresttanaka, line 212: I didn’t know you could do this. This is basically how I imagined it if the JSON was in a .js file, but in a JSON file. This embeds the JSON in the JS as an object, which I think is a good way to do this.

forresttanaka (changes requested):
I requested changes, but if you’d prefer it go in as is I have no issue with that. Just let me know and I’ll approve it as is.

src/encoded/static/components/navigation.js
emro: I added this, this seems like a good solution. I actually thought about doing something like this originally, but then didn't see any other places where isMobile() was implemented. I'm sure it'll be useful in the future though.

src/encoded/static/components/navigation.js
emro: Removed from state! And used React.useEffect() so that was fun.

src/encoded/static/components/navigation.js
emro: I'm not sure what we decided about this, but I can still change it if we want. I don't feel very strongly about it. I sort of think that it's mildly easier to add navigation glossary definitions as is, and think we could always go back and change very easily when we know more about what the glossary and tooltips will look like in the future, but it's also super easy to change and will work well with IDs too.

src/encoded/static/libs/ui/tooltip.js
emro: I was keeping it just because it was part of the original code. I've removed it, after verifying that it has no associated styles. We can always add it back. I assume the original thought was to have little carets above the tooltip pop-ups.

src/encoded/static/components/navigation.js
emro: It's very simple!

src/encoded/static/components/navigation.js
forresttanaka: I agree with that.

forresttanaka (dismissed):
Looks good!

forresttanaka (dismissed):

forresttanaka (approved):

PR #3403, opened 2020-07-02
ENCD-5370-pub-dataset-paged

forresttanaka:
* publication_data is a calculated property in publication objects so it always appears though often with zero elements. I check for this property now, so the Jest test data had to include it to avoid “undefined” crashes during Jest tests.
* I put in a new generic cache in libs so that paging between datasets doesn’t force a request to the server for _every_ page, if that page was already in the cache. I had done caches for similar tables but each had its own cache implementation. They should all be modified to use this generic cache, but there wasn’t enough time this round to test them.
* About the cache, it can cache anything, and has any size you’d like. It doesn’t hash the keys and has no collision detection — a very simple cache implementation I adapted from a blog post I linked to in the comments.
* In `FileTablePaged` I moved the pager from the footer to the header to match the Datasets table on the publication page. And I put the pager in the header because I noticed the pager jumps around when it goes in the footer, because the height of the table can change, page to page.

### DIFF
--- a/src/encoded/static/components/publication.js
+++ b/src/encoded/static/components/publication.js
@@ -1,29 +1,34 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import Cache from '../libs/cache';
+import Pager from '../libs/ui/pager';
 import { Panel, PanelHeading, PanelBody } from '../libs/ui/panel';
+import { CartAddAllElements, CartToggle } from './cart';
 import { auditDecor } from './audit';
 import * as globals from './globals';
 import { Breadcrumbs } from './navigation';
 import { DbxrefList } from './dbxref';
 import { PickerActions, resultItemClass } from './search';
 import Status from './status';
-import { ItemAccessories } from './objectutils';
+import { ItemAccessories, requestObjects } from './objectutils';
 import { SortTablePanel, SortTable } from './sorttable';
-import { FetchedData, Param } from './fetched';
 
 
 const datasetsColumns = {
     accession: {
         display: dataset => <a href={dataset['@id']}>{dataset.accession}</a>,
         title: 'Accession',
+        sorter: false,
     },
     type: {
-        getValue: datasets => (datasets['@type'] || [''])[0], // only the first matters
+        getValue: dataset => (dataset['@type'] || [''])[0], // only the first matters
         title: 'Type',
+        sorter: false,
     },
     target: {
         getValue: dataset => (dataset.target ? dataset.target.label : ''),
         title: 'Target',
+        sorter: false,
     },
     assay_term_name: {
         getValue: (dataset) => {
@@ -34,6 +39,7 @@ const datasetsColumns = {
             return (dataset.assay_term_name || []).join(', ');
         },
         title: 'Assay',
+        sorter: false,
     },
     description: {
         title: 'Description',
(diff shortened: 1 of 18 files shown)


forresttanaka:
Wrong branch.

forresttanaka:
No, I PRed the right branch, but Github gave it the wrong name. Reopening.

forresttanaka:
Active demo example publication: https://encd-5370-pub-dataset-paged-fytanaka.demo.encodedcc.org/publications/11cd568b-e68b-4e21-b261-6fc849899389/

src/encoded/static/components/typeutils.js
```diff
@@ -582,7 +582,7 @@ const derivingCols = {
 };
 
 
-const PAGED_FILE_TABLE_MAX = 50; // Maximnum number of files per page
+const PAGED_FILE_TABLE_MAX = 25; // Maximnum number of files per page
```
emro, line 585: Typo: "maximum"

src/encoded/static/scss/encoded/modules/_tables.scss
```diff
@@ -329,3 +329,21 @@ form.table-filter {
...
+        display: flex;
+        margin-top: -3px;
+        margin-right: -3px;
+        margin-bottom: -3px;
+
+        .pager {
+            margin-left: 5px;
+        }
+    }
+}
```
emro, line 349: Looking at this on mobile, it's good on the Pixel 2 (my phone!) but I might add a `flex-wrap: wrap` on .header-paged-sorrtable for narrower phones. This is pretty not important, though, your call.

Screen Shot 2020-07-02 at 1 30 56 PMScreen Shot 2020-07-02 at 1 31 11 PM

src/encoded/static/libs/ui/pager.js
```diff
@@ -2,47 +2,42 @@ import React from 'react';
...
-    }
-
-    handlePrev() {
-        const { current, updateCurrentPage } = this.props;
+/**
+ * Displays a pager component that lets users navigate a list of items. It displays a left and
+ * right arrow, and the current displayed item number and the total number of items. When the
+ * user clicks a page, it calls the `updateCurrentPage` callback so that the component that uses
+ * <Pager> can react to the click.
+ */
```
emro, line 10: I like how simplified this component is! No constructor(), no render()...

src/encoded/static/components/typeutils.js
```diff
@@ -690,18 +717,17 @@ export const FileTablePaged = ({ fileIds, files, title }) => {
     if (currentPageFiles.length > 0) {
         const headerTitle = typeof title === 'string' ? <h4>{title}</h4> : title;
         const fileCount = fileIds ? fileIds.length : files.length;
-        const fileCountDisplay = <div className="file-table-paged__count">{`${fileCount} file${fileCount === 1 ? '' : 's'}`}</div>;
-
-        // If we have more than one page of files to display, render a pager component in the
-        // footer.
-        const pager = totalPages > 1 ? <Pager total={totalPages} current={currentPageNum} updateCurrentPage={updateCurrentPage} /> : null;
+        const fileCountDisplay = <div className="table-paged__count">{`${fileCount} file${fileCount === 1 ? '' : 's'}`}</div>;
```
emro, line 720: Since this component works for files or datasets, I might have named it something more general, but not worth changing.

src/encoded/static/components/publication.js
```diff
@@ -42,17 +48,179 @@ const datasetsColumns = {
...
+        display: publicationData => publicationData.description || <i>No description</i>,
+        title: 'Description',
+        sorter: false,
+    },
+    status: {
+        display: publicationData => <Status item={publicationData} badgeSize="small" inline />,
+        title: 'Status',
+        sorter: false,
+    },
+};
```
emro, line 91: It would be sort of handy maybe to be able to specify that none of the columns are sortable in cases like these, just to make this less verbose, but I don't think that would be trivial and this works well.

src/encoded/static/components/publication.js
```diff
@@ -42,17 +48,179 @@ const datasetsColumns = {
...
-        display: datasets => <Status item={datasets.status} badgeSize="small" css="result-table__status" />,
+        display: dataset => <Status item={dataset.status} badgeSize="small" inline />,
         title: 'Status',
+        sorter: false,
+    },
+    cart: {
+        title: 'Cart',
+        display: dataset => <CartToggle element={dataset} />,
+        sorter: false,
+    },
```
emro, line 66: Adding the cart here is great.

src/encoded/static/components/publication.js
```diff
@@ -42,17 +48,179 @@ const datasetsColumns = {
...
+        } else {
+            // Page not found in the cache. Request the datasets for the currently displayed page.
+            const pageStartIndex = currentPage * PAGE_DATASET_COUNT;
+            const visibleDatasetIds = datasets.slice(pageStartIndex, pageStartIndex + PAGE_DATASET_COUNT);
+            requestObjects(visibleDatasetIds, '/search/?type=Dataset&limit=all&status!=deleted&status!=revoked&status!=replaced').then((requestedDatasets) => {
+                setVisibleDatasets(requestedDatasets);
+                getPageCache().write(currentPage, requestedDatasets);
+            });
+        }
+    }, [datasets, currentPage]);
```
emro, line 170: The caching code is very cool - I've never coded anything like this before.

emro (dismissed):
Looks great!

src/encoded/static/components/publication.js
forresttanaka: Yeah, could be a new prop on SortTable I suppose.

emro (approved):

PR #3407, opened 2020-07-07
ENCD-5367 clean up body map styles

emro:
Cleaning up the body map code and styles:
- There are two implementations of the body map, the full version and the thumbnail with a pop-up version. All the styles for both of these implementations is now in _body_map_facet.scss instead of in _summary.scss and _matrix.scss. A couple body map related styles are still in the summary / matrix stylesheets but those have to do with those specific pages and where the body map facet is organized on the page. Some of the styles are written to be more generic instead of being page-specific and comments should be clearer.
- I also moved more of the pop-up related JS into body_map.js so that in the future it should be easier to put the thumbnail / pop-up component on a new page. The new <BodyMapThumbnailAndModal> component should be pretty self-contained and easy to include. Plus I re-wrote it to use React Hooks!
- There should be no actual changes to how the body map code works and looks.

### DIFF
--- a/src/encoded/static/components/body_map.js
+++ b/src/encoded/static/components/body_map.js
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import queryString from 'query-string';
 import url from 'url';
+import { svgIcon } from '../libs/svg-icons';
 import QueryString from '../libs/query_string';
 import BodyDiagram from '../img/bodyMap/Deselected_Body';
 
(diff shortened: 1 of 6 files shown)


src/encoded/static/components/body_map.js
```diff
@@ -721,4 +722,109 @@ BodyMap.contextTypes = {
...
+        terms.forEach((term) => {
+            if (CellsList[term] && document.getElementById(term)) {
+                document.getElementById(term).classList.add('active');
+            }
+        });
+    }, [setIsThumbnailExpanded, props.context, props.location]);
+
+    const toggleThumbnail = () => {
+        setIsThumbnailExpanded(!isThumbnailExpanded);
+    };
```
forresttanaka, line 813: You _could_ make an argument for caching `toggleThumbnail` with `React.useCallback()` so that it doesn’t trigger a call to `BodyMapModal` on every render. But I think it makes sense not to cache it in this case.

src/encoded/static/components/matrix_reference_epigenome.js
```diff
@@ -793,12 +708,7 @@ class MatrixPresentation extends React.Component {
                     }
                     <div className="matrix-facet-container">
                         {(this.initialSelectedTab === 'Homo sapiens') ?
-                            <React.Fragment>
-                                <ClickableThumbnail isThumbnailExpanded={this.state.isThumbnailExpanded} toggleThumbnail={this.toggleThumbnail} />
-                                {this.state.isThumbnailExpanded ?
-                                    <BodyMapModal isThumbnailExpanded={this.state.isThumbnailExpanded} toggleThumbnail={this.toggleThumbnail} context={context} />
-                                : null}
-                            </React.Fragment>
+                            <BodyMapThumbnailAndModal context={context} location={this.context.location_href} />
```
forresttanaka, line 711: You could technically have `BodyMapThumbnailAndModal` import context.location_href instead of passing it here. But I wasn’t sure if you wanted the flexibility of having a different source of `location` in some future use.

forresttanaka (approved):
Looks quite good. Nice helpful comments in your scss, and this looks a lot easier to integrate to other pages.

src/encoded/static/components/body_map.js
emro: Ok! I'll keep this in mind in future cases.

src/encoded/static/components/matrix_reference_epigenome.js
emro: Well, I thought about doing this; I originally was trying to use props.context.location_href in the component, but it was always undefined, even when I added location_href to contextTypes everywhere. I think there is a distinction between this.context.location_href and context.location_href which I didn't sort out.

PR #3408, opened 2020-07-07
ENCD-5290-homepage-update

forresttanaka:
In addition to updating the classic image, the primary goal of this ticket involved adding two links to the classic image. I tried to keep the existing code intact as much as possible, but needed to modify the click handler to handle the links, and screen-reader text.


forresttanaka:
Need updates for Design Meeting feedback.

PR #3412, opened 2020-07-09
SNO-159 Allow remote indexing

caseylitton:


ini-templates/base-template.ini
```diff
@@ -15,6 +15,8 @@ postgresql.statement_timeout = 120
 pyramid.default_locale_name = en
 sqlalchemy.url = ${pg_uri}
 retry.attempts = 3
+remote_indexing = ${remote_indexing}
+remote_indexing_threshold = 50
```
hitz: what does this do?

ini-templates/base-template.ini
```diff
@@ -55,7 +57,7 @@ set indexer = true
 set indexer_initial_log = true
 set indexer_initial_log_path = /srv/encoded/initial-indexing-times.txt
 # Used to limit the indexed uuids.  Either leave blank or set to a number like 10 or 1000
-set indexer_short_uuids = 0
+set indexer_short_uuids = 100
```
hitz: what does this do?

ini-templates/base-template.ini
caseylitton: It's used in 
https://github.com/ENCODE-DCC/snovault/blob/SNO-159-allow-remote-indexing/src/snovault/elasticsearch/indexer.py#L157

to determine if indexing should be remote or local based on the number of uuids to index.  There are 8 local indexing processes available, down from 16.

50 was for troubleshooting.  This has been updated to 10000 to start.  We're going to test a full reindexing on 8 procs over the weekend.  This could probably be higher but it really depends on the object type.  10000 files or experiments should probably be offloaded.

ini-templates/base-template.ini
caseylitton: This is a debugging variable that has been around for a bit, Ubuntu 18 update?

It's used here,
https://github.com/ENCODE-DCC/snovault/blob/SNO-159-allow-remote-indexing/src/snovault/elasticsearch/indexer.py#L629

to trim the uuids to get a quicker turn around time when running _indexer_state?reindex=all 

Leaving at 100 was a bug.  0 is the default which like None.  It also just takes the first part of the list which are usually the easiest to index.

hitz (dismissed):

hitz (approved):

PR #3414, opened 2020-07-10
ENCD-5381-fix-publication-object-publication-year

iyoungworth:


src/encoded/types/__init__.py
```diff
@@ -247,7 +247,7 @@ def unique_keys(self, properties):
         "type": "string",
```
hitz: this needs to change to "type": "integer"

src/encoded/types/__init__.py
```diff
@@ -247,7 +247,7 @@ def unique_keys(self, properties):
...
     })
     def publication_year(self, date_published):
-        return date_published.partition(' ')[0]
+        return int(date_published[:4])
```
hitz: I see that you check whether the value can be coerced in the upgrade to a number... but it's not checked here.  You probably need a try/except to prevent this from crashing if somehow date_published doesn't match the regex... On the other hand ... I guess it would return 0 as a valid non-year INT?

hitz: maybe it can just return None if it cannot be cast as int.

hitz (changes requested):
I don't think this is quite right; does it sort the Publication search page publication_year facet by year?

src/encoded/types/__init__.py
iyoungworth: Oops. Fixed this.

src/encoded/types/__init__.py
iyoungworth: I added the same check as in the upgrade and had it return None explicitly if it failed the check

hitz (dismissed):

src/encoded/schemas/publication.json
```diff
@@ -37,7 +38,18 @@
...
+            "description": "The date the publication or communication was published; must be in YYYY-MM-DD format.",
+            "type": "string",
+            "oneOf": [
+                {
+                    "format": "date"
+                },
+                {
+                    "comment": "Historical date formats, not valid for new submissions.",
+                    "permission": "import_items",
+                    "pattern": "^[0-9]{4}[^-]*$"
```
iyoungworth, line 50: The regex previously required a 5th character after the year, updated here to allow just the year in import_items

hitz (approved):

PR #3417, opened 2020-07-13
VALIS-14-show-error-on-invalid-loc

zoldello:


package.json
```diff
@@ -79,7 +79,7 @@
     "dayjs": "^1.8.16",
     "domready": "^1.0.8",
     "form-serialize": "^0.6.0",
-    "genome-visualizer": "git://github.com/ENCODE-DCC/valis-hpgv.git#1.3.0",
+    "genome-visualizer": "git://github.com/ENCODE-DCC/valis-hpgv.git#VALIS-14-show-error-on-invalid-loc",
```
emro, line 82: This is good but it looks like package-lock.json is pointing to commit with SHA:

> "version": "git://github.com/ENCODE-DCC/valis-hpgv.git#b5c9bb34f2747a15a4870b9c6955bfdd08c790a2"

While the SHA for the corresponding VALIS-14 branch is:

> 07b1a6dddc561b66a6b3ae7da0c6f68812a47352

Shouldn't these match?

zoldello: This has been updated.

package-lock.json
```diff
@@ -6720,56 +6622,36 @@
...
-        "gaxios": "^2.0.0",
-        "google-auth-library": "^5.0.0",
-        "pumpify": "^2.0.0",
-        "stream-events": "^1.0.4"
-      }
-    },
     "genome-visualizer": {
-      "version": "git://github.com/ENCODE-DCC/valis-hpgv.git#413ff3f2c16a026a555ea0edf1cb64e3204d044e",
-      "from": "git://github.com/ENCODE-DCC/valis-hpgv.git#1.5.0",
+      "version": "git://github.com/ENCODE-DCC/valis-hpgv.git#84bd9586f98fccef4cc5f3725af62ebd87ee845f",
```
emro: There seem like there are a lot of package-lock changes here. Don't we just want this one package updated? This line looks great though - this is the change that we definitely need.

emro (changes requested):
I'm not sure that we want all of these package-lock updates, but the update to `genome-visualizer` looks good.

emro:
I think that the not-bdd-non-indexing tests might pass if you rebase.

package-lock.json
zoldello: Good point. npm updated these packages itself. Many packages are set up to continuous update minor and patch updates but not major; when you do npm install.

emro (dismissed):
Phil and I discussed the package-lock.json changes and these are required for the new library.

zoldello:
This was merged via ENCD-5596; as it contains all its changes. So there is no need to pull this in.

PR #3418, opened 2020-07-13
ENCD-5390 update fonts sitewide

hitz:
PR for Emma

### DIFF
--- a/src/encoded/static/scss/encoded/_base.scss
+++ b/src/encoded/static/scss/encoded/_base.scss
@@ -1,5 +1,5 @@
 /* Variables */
-$mainThemeColor: 			   #e8ecef; /* i.e. the background color ..for now */
+$mainThemeColor: 			   #fff; /* i.e. the background color ..for now */
 
 $nav-color-bg:                 #0a253d; // Navigation background color
 
@@ -13,14 +13,13 @@ $tableHeadFootBackgroundColor: #f5f5f5;
 
 $std-href-color:               #428bca;
 
-$site-font:                    'Helvetica Neue', sans-serif;
-$brand-font:                   'Helvetica Neue', sans-serif;
+$site-font:                    'Mada', sans-serif;
+$brand-font:                   'Mada', sans-serif;
 
 $Homo-sapiens:                  #63326E;
 $Mus-musculus:                  #3D518C;
 $Caenorhabditis-elegans:        #77878B;
 $Drosophila-melanogaster:       #335C67;
-
 
 // Pass value with CSS units and get just the value back
 @function strip-units($value) {
@@ -110,6 +109,10 @@ a {
     }
 }
 
+b {
+    font-weight: 400 !important;
+}
+
 hr { clear: both; }
 
 figure {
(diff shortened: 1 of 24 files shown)


src/encoded/static/scss/encoded/_base.scss
```diff
@@ -13,14 +13,19 @@ $tableHeadFootBackgroundColor: #f5f5f5;
...
-$brand-font:                   'Helvetica Neue', sans-serif;
+$site-font:                    'Mada', sans-serif;
+$brand-font:                   'Mada', sans-serif;
 
 $Homo-sapiens:                  #63326E;
 $Mus-musculus:                  #3D518C;
 $Caenorhabditis-elegans:        #77878B;
 $Drosophila-melanogaster:       #335C67;
 
+// DELETE ME BEFORE COMMITTING!!!!-----------------------------------------------
```
zoldello: Phil note- Delete this

src/encoded/static/scss/encoded/modules/_facet.scss
```diff
@@ -131,7 +133,12 @@
             flex-direction: row;
             cursor: pointer;
             justify-content: space-between;
-        }  
+            padding: 10px 10px 0;
+            line-height: 1;
+            // h5 {
```
zoldello: Fix this.

zoldello (changes requested):
Minor tweaks

src/encoded/static/scss/encoded/modules/_panels.scss
```diff
@@ -29,6 +29,27 @@ $panel-body-inset:    10px; // Number of pixels padding inside panel body
...
+    .panel {
+        border: 0;
+    }
+}
+
+.Page, .Search, .Matrix, .ChipSeqMatrix, .Summary, .EntexMatrix, .SESCCStemCellMatrix, .ReferenceEpigenomeMatrix, .MouseDevelopmentMatrix {
+    .panel {
+        border: 0;
+    }
+}
```
forresttanaka, line 42: I think these pages should have a new class added to not have a border. This works, but we have to remember to add to this long selector string when we add new matrices.

src/encoded/static/scss/encoded/_base.scss
```diff
@@ -165,6 +192,7 @@ blockquote {
 $disabled-color-factor: 20%;
 
 .btn {
+    font-family: Quicksand;
```
forresttanaka, line 195: Probably should have added a fallback, but should be OK.

forresttanaka (dismissed):

src/encoded/static/scss/encoded/_base.scss
zoldello: I missed this. The fallback I believe defaults to Times New Roman. Is that acceptable?

src/encoded/static/scss/encoded/modules/_panels.scss
zoldello: I do not get what you mean. Is this a must-have?

src/encoded/static/scss/encoded/_base.scss
forresttanaka: I think just “sans-serif” as the default should be fine.

src/encoded/static/scss/encoded/modules/_panels.scss
forresttanaka: Not a must have, but we might want to move to that eventually. Instead of specifying every matrix class in this selector list, and having to maintain it as we get more classes, all matrices and pages should have a class we can target and not have the border.

forresttanaka (dismissed):

forresttanaka (approved):
Hard to tell just from this, but I trust this solves the issues Khine saw.

PR #3419, opened 2020-07-13
ENCD-5393 PublicationData Batch Download

forresttanaka:


forresttanaka:
Up-to-date demo: https://encd-5393-publication-dl-fytanaka.demo.encodedcc.org/

src/encoded/batch_download.py
```diff
@@ -291,6 +328,139 @@ def _get_annotation_metadata(request, search_path, param_list):
...
+    writer.writerow(header)
+
+    # Load the specified PublicationData object and extract its files to build the rows
+    dataset = request.embed(dataset_path, as_user=True)
+    file_ids = dataset.get('files', [])
+    if file_ids:
+        # For each file, load the file object and use its properties to build each row of the table.
+        for file_id in file_ids:
+            # Load the file object and disqualify those we don't handle.
+            file = request.embed(file_id, as_user=True)
```
hitz, line 362: I feel like we might have to batch these in groups of 500 or something. There are publicationDatasets with 10,000+ files.    So like /search/ with @fields=... *&limit=500?  (500 was just an arbitrary guess)

The biggest ones are:

ENCSR226XLF	7360
ENCSR669AMM	3144
ENCSR617ILB	3048
ENCSR456FVU	2670
ENCSR145VDW	2667
ENCSR304ADM	2576
ENCSR369TWP	2128
ENCSR795JHH	2112
ENCSR590NUK	1472
ENCSR038GAB	1389
ENCSR574CRQ	1295
ENCSR870OLK	944

forresttanaka: Does request.embed do a big search? I pass it an individual file `@id` here, so I thought it wasn’t too bad. Granted, the file_ids list could be huge so this request.embed could happen apparently 10,000+ times.

forresttanaka:
Oh, I forgot to mention, the /batch_download/ and /metadata/ queries for PublicationData look like:

```
/batch_download/?type=PublicationData&dataset={publication @id}
```

But I only handle one dataset query here, and I’m guessing I should handle multiple.

src/encoded/batch_download.py
```diff
@@ -140,8 +140,45 @@ def includeme(config):
...
+    ('md5sum', ['files.md5sum']),
+    ('dbxrefs', ['files.dbxrefs']),
+    ('File download URL', ['files.href']),
+    ('Assembly', ['files.assembly']),
+    ('Controlled by', ['files.controlled_by']),
+    ('File status', ['files.status']),
+    ('Derived from', ['files.derived_from']),
+    ('S3 URL', ['files.cloud_metadata.url']),
+    ('Size', ['files.file_size']),
+])
```
hitz, line 167: can we add the following?

    ('Biosample term id', ['biosample_ontology.term_id']),
    ('Biosample term name', ['biosample_ontology.term_name']),
    ('Biosample type', ['biosample_ontology.classification']),
    ('Target', ['target.label']),
    ('No File Available', ['file.no_file_available']),


We really need organism, but I guess it's not in File.

src/encoded/batch_download.py
```diff
@@ -353,8 +523,11 @@ def metadata_tsv(context, request):
     else:
         search_path = '/search/'
     type_param = param_list.get('type', [''])[0]
-    if type_param and type_param.lower() == 'annotation':
-        return _get_annotation_metadata(request, search_path, param_list)
+    if type_param:
```
hitz, line 539: ~~peak_metadata is disabled I think? ~~ never mind fooled by git diff.

src/encoded/static/components/view_controls.js
```diff
@@ -391,20 +391,13 @@ BatchDownloadButton.defaultProps = {
...
-    }
-
-    handleDownloadClick() {
-        const queryString = getQueryFromFilters(this.props.results.filters);
-        this.context.navigate(`/batch_download/?${queryString}`);
-    }
-
-    render() {
-        const { results } = this.props;
+export const BatchDownloadControls = ({ results, queryString }, reactContext) => {
```
hitz, line 394: this is just update of class to ... whatever it's called.

hitz (changes requested):

src/encoded/batch_download.py
forresttanaka: All batch_download versions filter out No File Available=true files, so I left it out.

hitz: puzzling why it's in the TSV specifier then

hitz:
> Oh, I forgot to mention, the /batch_download/ and /metadata/ queries for PublicationData look like:
> 
> ```
> /batch_download/?type=PublicationData&dataset={publication @id}
> ```
> 
> But I only handle one dataset query here, and I’m guessing I should handle multiple.

no no, 1 publicationData at a time is enough; they tend to be very large.

src/encoded/batch_download.py
forresttanaka: I don‘t think any target is available — not in the file nor the PublicationData

hitz: https://www.encodeproject.org/search/?type=File&target.label=*
https://www.encodeproject.org/search/?type=File&biosample_ontology.term_name=*

They are not always there, but they are there if they can be calculated.

hitz: you are correct, publication_data doesn't have anything, which is fine.

forresttanaka:
Closing this draft PR and will update the branch.

PR #3420, opened 2020-07-13
ENCD-5371-biosample-disease-ontology

bonitalam:


src/encoded/audit/biosample.py
```diff
@@ -322,6 +322,30 @@ def audit_biosample_post_differentiation_time(value, system):
...
+        detail = ('Biosample {} specifies a disease_term_id {} '
+            'that is not part of the {} ontology.'.format(
+                audit_link(path_to_text(value['@id']), value['@id']),
+                disease_term_id,
+                disease_term_id.split(':', 1)[0]
+            )
+        )
+        yield AuditFailure('disease_term_id not in ontology', detail,
+                           level='INTERNAL_ACTION')
+        return
```
gabdank: i think you don't need the return statement here

src/encoded/types/biosample.py
```diff
@@ -205,6 +205,20 @@ def age_units(
...
+                            "description": "Ontology term describing the disease affecting the biosample.",
+                            "type": "string",
+                            "notSubmittable": True,
+                        })
+    def disease_term_name(self, request, registry, disease_term_id=None):
+        term_name = None
+        if disease_term_id is not None:
+            if disease_term_id in registry['ontology']:
+                term_name = registry['ontology'][disease_term_id]['name']
+        return term_name
```
gabdank: if you look on biosample_type, you will see that many of the calculated properties assume the term_id exists in the ontology, which ensures that the use of incorrect term_id will result in schema validation error - which is what prevent people from submitting incorrect or non-existing ids. then you don't need the audit checking for a match, please let me know if that makes sense?

gabdank (changes requested):

src/encoded/tests/fixtures/schemas/biosample.py
```diff
@@ -375,4 +375,3 @@ def biosample_23(testapp, lab, award, source, organism, epiblast):
         'source': source['uuid']
     }
     return testapp.post_json('/biosample', item, status=201).json['@graph'][0]
```
gabdank, line 377: please add one new line in the of the file

src/encoded/audit/biosample.py
```diff
@@ -332,7 +332,7 @@ def audit_biosample_post_differentiation_time(value, system):
     'audit_pmi': audit_biosample_pmi,
     'audit_cell_isolation_method': audit_biosample_cell_isolation_method,
     'audit_depleted_in_term_name': audit_biosample_depleted_in_term_name,
-    'audit_post_differentiation_time': audit_biosample_post_differentiation_time,
+    'audit_post_differentiation_time': audit_biosample_post_differentiation_time
```
gabdank: please add back the comma you deleted here

src/encoded/tests/test_audit_biosample.py
```diff
@@ -1,6 +1,5 @@
 import pytest
 
-
 def test_audit_biosample_modifications_whole_organism(
```
gabdank: please - if you haven't modified/touched the file return it to the way it has been before.

gabdank (changes requested):

gabdank (commented):
please also add an insert with biosample using a disease ontology

src/encoded/tests/fixtures/schemas/biosample.py
bonitalam: New line is there

gabdank (approved):

hitz:
This did pass circle-ci, just has a funny report.

PR #3421, opened 2020-07-13
ENCD-5395-check-biosample-characterizations-from-pooled-from-biosample

yunhailuo:


src/encoded/audit/experiment.py
```diff
@@ -1137,9 +1137,9 @@ def check_experiment_chip_seq_standards(
     if 'replication_type' not in experiment or experiment['replication_type'] == 'unreplicated':
         return
 
-    idr_metrics = get_metrics(idr_peaks_files, 'IDRQualityMetric') 
+    idr_metrics = get_metrics(idr_peaks_files, 'IDRQualityMetric')
```
gabdank: I think correcting trailing spaces is great, but please next time do it in a separate ticket.

src/encoded/tests/test_audit_experiment.py
```diff
@@ -2986,14 +2986,72 @@ def test_audit_experiment_tagging_biosample_characterization(
...
+        for error in collect_audit_errors(res, error_types=['ERROR'])
+    )
+    testapp.patch_json(
+        biosample_characterization['@id'],
+        {'characterizes': biosample_1['@id']}
+    )
+    res = testapp.get(base_experiment['@id'] + '@@index-data')
+    assert any(
+        error['category'] == 'missing biosample characterization'
+        for error in collect_audit_errors(res, error_types=['ERROR'])
```
gabdank: I think it would be easier to follow if you make a test fixture of a pooled biosample with characterization or without characterization in parent samples? then the test meaning will be easy to follow.

gabdank (changes requested):

src/encoded/audit/experiment.py
yunhailuo: Let me revert this. I set auto trim in my editor and that's why.

yunhailuo:
Splitted tests into 3 tests with 3 separated fixtures.

gabdank (dismissed):

hitz (approved):
approved after rebase + force push

PR #3426, opened 2020-07-15
ENCD-5400-analysis-update

yunhailuo:


src/encoded/audit/experiment.py
```diff
@@ -4283,6 +4283,36 @@ def check_experiment_atac_encode4_qc_standards(experiment, files_structure):
                     yield AuditFailure('mild to moderate bottlenecking', pbc2_detail, level='WARNING')
 
 
+def audit_analysis_files(value, system, excluded_types):
+    if 'analysis_objects' not in value:
+        return
+    experiment_files = (
+        f['@id']
+        for f in (
```
gabdank: I think that all the files are traversed in "def create_files_mapping(files_list, excluded):"  would not it be more efficient and clear to create another group of files there - and then use this group (experiment_files) here? Another question is about contributing files, I thought Analysis will only include the output files, not controls/references?

src/encoded/audit/experiment.py
```diff
@@ -4283,6 +4283,36 @@ def check_experiment_atac_encode4_qc_standards(experiment, files_structure):
...
+            + value.get('contributing_files', [])
+        )
+    )
+    detail_list = []
+    for analysis in value['analysis_objects']:
+        for f in analysis.get('files', []):
+            if f not in experiment_files:
+                detail_list.append(
+                    'Analysis {} has a file {} which is not belong to this '
+                    'experiment {}.'.format(
```
gabdank: I think it should be "{} which does not belong" instead of "{} which is not belong"

gabdank (changes requested):

buildout.cfg
```diff
@@ -37,7 +37,7 @@ primary_indexer_queue_type = Simple
 primary_indexer_queue_server = true
 primary_indexer_queue_worker = true
 primary_indexer_queue_worker_processes = 16
-primary_indexer_queue_worker_chunk_size = 512
+primary_indexer_queue_worker_chunk_size = 265
```
gabdank: not sure it matters, but should not that be 256

gabdank (dismissed):

buildout.cfg
hitz: it doesn't matter, but we have to revert this before merge.

hitz:
Tests are good, despite github intergration

PR #3427, opened 2020-07-15
ENCD-5288-create-facet-for-cCREs

iyoungworth:


src/encoded/types/dataset.py
```diff
@@ -466,6 +466,48 @@ def files(self, request, original_files, status):
     def superseded_by(self, request, superseded_by):
         return paths_filtered_by_status(request, superseded_by)
 
+    @calculated_property(condition='contributing_files', schema={
+        "title": "Biochemical profile inputs",
+        "description": "The data used to generate a cCRE annotation for a given cell type or biosample.",
+        "type": "string",
+        "notSubmittable": True,
+
```
gabdank: I think this is an extra line you should remove

src/encoded/types/dataset.py
```diff
@@ -466,6 +466,48 @@ def files(self, request, original_files, status):
     def superseded_by(self, request, superseded_by):
         return paths_filtered_by_status(request, superseded_by)
 
+    @calculated_property(condition='contributing_files', schema={
+        "title": "Biochemical profile inputs",
+        "description": "The data used to generate a cCRE annotation for a given cell type or biosample.",
```
gabdank: since cCREs may be cell type agnostic, perhaps you can make less definitive description? Maybe something like: "The input data used to generate the cCRE annotation." ?

src/encoded/types/dataset.py
```diff
@@ -466,6 +466,48 @@ def files(self, request, original_files, status):
...
+        annotation_type,
+        encyclopedia_version=None,
+        contributing_files=None
+    ):
+        # https://encodedcc.atlassian.net/browse/ENCD-5288
+        inputs_list = []
+        if encyclopedia_version is not None and \
+            encyclopedia_version in (
+                'ENCODE v4', 'ENCODE v5', 'ENCODE v6'
+                ):
```
gabdank: I am not sure why you are using set for encyclopedias versions? should not that be a list?

src/encoded/types/dataset.py
```diff
@@ -466,6 +466,48 @@ def files(self, request, original_files, status):
...
+            if annotation_type == 'candidate Cis-Regulatory Elements':
+                if contributing_files is not None:
+                    for input_file in contributing_files:
+                        file = request.embed(input_file, '@@object')
+                        if file['dataset']:
```
gabdank: I think here you can skip calculated properties?
request.embed(input_file, '@@object?skip_calculated=true'

src/encoded/types/dataset.py
```diff
@@ -466,6 +466,48 @@ def files(self, request, original_files, status):
...
+                                    target = request.embed(properties['target'], '@@object')
+                                    inputs_list.append(target['label'])
+                                elif properties['assay_term_name'] == 'DNase-seq':
+                                    inputs_list.append('DNase-seq')
+                            elif file['output_type'] == 'candidate Cis-Regulatory Elements':
+                                if 'derived_from' in file:
+                                    derived_from_file = request.embed(file['derived_from'][0], '@@object')
+                                    if derived_from_file['output_type'] == 'representative DNase hypersensitivity sites (rDHSs)':
+                                        inputs_list.append('rDHS')
+                                    if derived_from_file['output_type'] == 'consensus DNase hypersensitivity sites (cDHSs)':
```
gabdank: if you use request.embed - try to not to do it unless you have to, and if you do then try not to calculate properties unless it is necessary. 

You have to remember that every Experiment will attempt calculating this property - so it is a time consuming step we would like to try nd keep as optimized as possibel

gabdank (changes requested):

src/encoded/types/dataset.py
iyoungworth: Sounds good, I altered it (this is closer to what I have in the changelog also)

src/encoded/types/dataset.py
iyoungworth: Thanks, removed!

src/encoded/types/dataset.py
iyoungworth: Ah agreed, changed it to a list

src/encoded/types/dataset.py
iyoungworth: You are right, I added your suggestion (and it still appears fine on local app)

src/encoded/types/dataset.py
iyoungworth: I swapped the order around a little bit in order to avoid embedding for some files, and added skip_calculated in a couple places to cut down on timing also

gabdank (dismissed):

hitz:
everything passed but not-bdd-not-indexing but I am running locally

hitz:
passed locally merging

PR #3428, opened 2020-07-15
ENCD-5409-lower-indexing-chunksize

hitz:


hitz:
needs to be an option->demo

PR #3431, opened 2020-07-15
ENCD-5413-hide-download-btn-for-large-search

zoldello:


src/encoded/static/components/view_controls.js
```diff
@@ -343,18 +363,21 @@ BatchDownloadModal.propTypes = {
     downloadClickHandler: PropTypes.func.isRequired,
     /** Called when the user does something to close the modal */
     closeModalHandler: PropTypes.func.isRequired,
+    /*** Result total */
```
forresttanaka: Could this be more descriptive? That’s just the name of the property. This also has an ESLint because of the three asterisks.

src/encoded/static/components/view_controls.js
```diff
@@ -309,27 +315,41 @@ const BATCH_DOWNLOAD_PROHIBITED_PATHS = [
...
+                    <p>
+                        The &ldquo;files.txt&rdquo; file can be copied to any server.<br />
+                        The following command using cURL can be used to download all the files in the list:
+                    </p>
+                    <code>xargs -L 1 curl -O -J -L &lt; files.txt</code><br />
+                </>
+            :
+                <>
+                    <p>
+                        This search is too large (&gt;500 datasets) to automatically generate a manifest or metadata file.  We are currently working on methods to download from large searches.
```
forresttanaka: It’d be nice to insert the MAX_DOWNLOADABLE_RESULT constant so that if you change that, this text automatically changes correctly.

src/encoded/static/components/view_controls.js
```diff
@@ -309,27 +315,41 @@ const BATCH_DOWNLOAD_PROHIBITED_PATHS = [
...
-            </p>
-            <p>
-                Further description of the contents of the metadata file are described in the <a href="/help/batch-download/">Batch Download help doc</a>.
-            </p>
-            <p>
-                The &ldquo;files.txt&rdquo; file can be copied to any server.<br />
-                The following command using cURL can be used to download all the files in the list:
-            </p>
-            <code>xargs -L 1 curl -O -J -L &lt; files.txt</code><br />
+            {resultTotal < MAX_DOWNLOADABLE_RESULT ?
```
forresttanaka: I think this should be <= instead of <. With this, 500 datasets would generate the error message, but the error message implies 500 is OK.

forresttanaka (changes requested):
This looks really good; just a couple small changes, and an ESLint error.

forresttanaka (approved):
I’m going to call this approved while not looking _real_ closely, but I think it will serve ENCD-5393 well. I’d like to get this merged so I can pull it into ENCD-5393 before PR if possible. Many thanks to @zoldello for making this be serviceable for that ticket.

PR #3437, opened 2020-07-17
ENCD-5409 lower chunksize option demo

hitz:
I hope this is right... not sure about what flags are used for remote indexer... or if the remote/backup should have same chunk size.  Probably 512 since they don't have ES on them.


keenangraham (approved):

src/encoded/commands/deploy.py
```diff
@@ -1039,6 +1041,12 @@ def hostname(value):
         type=int,
         help="Remote indexing"
     )
+    parser.add_argument(
```
caseylitton, line 1044: This argument will only work for non demos.  If I set it to three for a cluster build then chunk size is three.  But for a demo it will be 256.

Which is fine, it's just a confusing way to make defaults.  And wouldn't be the only variable in here that doesn't work the way someone would expect.  But will be cleared up if we move to using cfg directly per the other comment.

cloud-config/run-scripts/app-encd.sh
```diff
@@ -38,7 +38,7 @@ if [ ! -f "$bin_build_path" ]; then
 fi
 
 # Run bin/buildout
-bin_build_cmd="$ENCD_HOME/bin/buildout -c $ENCD_ROLE.cfg buildout:es-ip=$ENCD_ES_IP buildout:es-port=$ENCD_ES_PORT buildout:pg-uri=$PG_URI buildout:fe-ip=$ENCD_FE_IP buildout:remote_indexing=$ENCD_REMOTE_INDEXING buildout:index_procs=$ENCD_INDEX_PROCS"
+bin_build_cmd="$ENCD_HOME/bin/buildout -c $ENCD_ROLE.cfg buildout:es-ip=$ENCD_ES_IP buildout:es-port=$ENCD_ES_PORT buildout:pg-uri=$PG_URI buildout:fe-ip=$ENCD_FE_IP buildout:remote_indexing=$ENCD_REMOTE_INDEXING buildout:index_procs=$ENCD_INDEX_PROCS buildout:index_chunk_size=$ENCD_INDEX_CHUNK_SIZE"
```
caseylitton, line 41: This pattern is one reason we need to move to demo, test, rc, prod.cfgs.  I just had to tack on two variables for the offloaded indexers.  We should fix this soon since deployments are getting more complex.

caseylitton (approved):
Looks fine.  Demos get 256 and cluster builds get that the arguments says.

cloud-config/run-scripts/app-encd.sh
hitz: totally concur, I was wondering if we maybe even want 2 scripts.  One for deploy demo one for "production" deploy.  Maybe they could import a common module or something.

src/encoded/commands/deploy.py
hitz: yeah that is very sloppy.

hitz:
not-bdd-non-indexing test fails as expected (fix ENCD-5421), merge override.

PR #3440, opened 2020-07-20
ENCD-5420-fix-bottom-shading

zoldello:


forresttanaka (approved):
Cool that works.

hitz:
This failed the test that was fixed in ENCD-5421, merging.

PR #3442, opened 2020-07-20
ENCD-5427 Make title underlines consistent

forresttanaka:
This mostly involved changing `<h2>` tags to `<h1>` as they were the only titles on the page, and should have been `<h1>`.


zoldello (dismissed):
Changes look okay.

zoldello (approved):
Code looks the same as it looked after the previous code review. It still looks fine.

PR #3444, opened 2020-07-21
ENCD-5418 Adjust new fonts

forresttanaka:
I didn’t address the JSON button as it would require some manual offsets. I think the real way to fix this is to use an SVG instead of relying on how font glyphs are designed.

### DIFF
--- a/src/encoded/static/scss/encoded/modules/_modals.scss
+++ b/src/encoded/static/scss/encoded/modules/_modals.scss
@@ -20,7 +20,7 @@
         @at-root #{&}-title {
             flex: 1 1 auto;
 
-            h4 {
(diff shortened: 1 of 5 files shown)


src/encoded/static/scss/encoded/modules/_modals.scss
```diff
@@ -20,7 +20,7 @@
         @at-root #{&}-title {
             flex: 1 1 auto;
 
-            h4 {
+            h2 {
```
zoldello, line 23: Why modify h2 when you can create a new class?

zoldello (approved):
Code looks fine. My comment involves code-setup that existed before this ticket.

PR #3445, opened 2020-07-23
ENCD-5433-refactor-not-bdd-not-indexing-tests

hitz:


src/encoded/tests/features/conftest.py
```diff
@@ -29,7 +29,7 @@ def app(app_settings):
 
 @pytest.mark.fixture_cost(500)
 @pytest.yield_fixture(scope='session')
-def workbook(request, app):
+def index_workbook(request, app):
```
keenangraham, line 32: How does renaming help?

src/encoded/tests/test_search_views.py
```diff
@@ -1,9 +1,9 @@
 import pytest
 
-from encoded.tests.features.conftest import app_settings, app, workbook
+from encoded.tests.features.conftest import app_settings, app, index_workbook
 
 
-def test_search_views_search_view_with_filters(workbook, testapp):
+def test_search_views_search_view_with_filters(index_workbook, testapp):
```
keenangraham, line 7: > -m “indexing” - all tests now pytest.mark.indexing  - includes test_indexing, test_batchdownload, and test_search_views.

Were these supposed to be marked as indexing? Looks like they are still running in `non bdd non indexing tests`.

keenangraham (changes requested):

src/encoded/tests/features/conftest.py
hitz: prevents confusion between the workbook fixture that doesn't index in ES and the one that does

src/encoded/tests/test_search_views.py
hitz: it looks like maybe the search_views tests didn't get pytest.marked as indexing?

hitz: fixed

keenangraham (approved):
Looks like `test_search_views.py` running in `non bdd indexing` now.

PR #3449, opened 2020-07-24
ENCD-5062-add-image-icon-on-report-page

zoldello:


src/encoded/schemas/image.json
```diff
@@ -25,6 +25,22 @@
...
             "type": "string"
+        },
+        "thumbnail": {
+            "title": "Thumbnail",
+            "type": "string"
+        },
+        "downloadUrl": {
+            "title": "Download Url",
+            "type": "string"
+        }
```
forresttanaka, line 28: Are these two properties needed if they’re calculated? Neither the datasets nor the experiment have a “files” property in their schemas, yet they have a calculated `files` property.

src/encoded/types/image.py
```diff
@@ -30,3 +31,19 @@ def unique_keys(self, properties):
...
+    def thumbnail(self, request, attachment):
+        return self.jsonld_id(request) + attachment['href']
+
+    @calculated_property(schema={
+        "title": "Download Url",
+        "description": "Download Url",
+        "type": "string",
+    })
+    def downloadUrl(self, request, attachment):
+        return self.jsonld_id(request) + attachment['href']
```
forresttanaka, line 49: Seems like this should be snake case. You might also want to comment these, as it took me quite a while to figure out why you had two properties with exactly the same value. In the future, I’d like to make the row cells customizable through a registration system, just like facets, and we might not need two properties with the same value at that stage.

src/encoded/static/components/image.js
```diff
@@ -129,6 +129,18 @@ export const Attachment = ({ context, attachment, className, showLink }) => {
...
+        return (
+            <div>
+                <div className="attachment">
+                    <div className="file-thumbnail">
+                        <img src={context.thumbnail} alt="thumbnail" />
+                    </div>
+                </div>
+            </div>
+        );
+    }
```
forresttanaka: I feel like what might be cleaner is if, rather than modifying `Attachment` which expects an attachment, that you modify `Image` to not call `Attachment` anymore, and instead use the code you have here. That way you don’t modify `Attachment` which gets used elsewhere for things with attachments.

Do you need the extra surrounding `<div>`?

src/encoded/static/components/search.js
```diff
@@ -601,11 +601,9 @@ const Image = (props) => {
         <li className={resultItemClass(result)}>
             <div className="result-item">
                 <div className="result-item__data">
-                    <a href={result['@id']} className="result-item__link">{result.caption}</a>
+                    <a href={result['@id']} className="result-item__link">{result['@id']}</a>
                     <Attachment context={result} attachment={result.attachment} />
-                </div>
-                <div className="result-item__meta">
-                    <p className="type meta-title">Image</p>
```
forresttanaka, line 608: I know removing the “Image” type was a request from Meenakshi, but it feels odd that `Image` stands alone as the only search result without a type (I think), and instead I feel we should keep this and add the status. I’ll comment in the ticket about that.

src/encoded/static/components/search.js
```diff
@@ -601,11 +601,9 @@ const Image = (props) => {
         <li className={resultItemClass(result)}>
             <div className="result-item">
                 <div className="result-item__data">
-                    <a href={result['@id']} className="result-item__link">{result.caption}</a>
+                    <a href={result['@id']} className="result-item__link">{result['@id']}</a>
```
forresttanaka, line 603: Meenakshi asked for the UUID here, but I doubt it really matters, and she might have meant @id.

forresttanaka (changes requested):
Nice capability.

src/encoded/static/components/search.js
zoldello: I am showing the @id... right?

forresttanaka: Right. Just saying that’s not what she asked for, but my guess is that it won’t matter to her.

src/encoded/schemas/image.json
hitz: In fact if they are calculated they should NOT be here.  There are a few exceptions for properties that can be submitted or calculated.

src/encoded/types/image.py
hitz: I agree

hitz (changes requested):

forresttanaka (approved):
Though technically, “thumbnail” is one word, not “thumb nail.”

PR #3462, opened 2020-07-31
ENCD-5439-optimize-metadata-endpoint

keenangraham:


src/encoded/reports/metadata.py
```diff
@@ -0,0 +1,365 @@
...
+    qs = QueryString(request)
+    specified_type = qs.get_one_value(
+            params=qs.get_type_filters()
+    )
+    if specified_type == 'Annotation':
+        return _get_annotation_metadata(context, request)
+    elif specified_type == 'PublicationData':
+        return _get_publicationdata_metadata(context, request)
+    else:
+        return _get_metadata(context, request)
```
hitz: Are there tests for these?

src/encoded/tests/test_metadata.py
```diff
@@ -0,0 +1,1034 @@
...
+
+def test_metadata_metadata_report_validate_request(dummy_request):
+    from encoded.reports.metadata import MetadataReport
+    dummy_request.environ['QUERY_STRING'] = (
+        'type=Experiment&files.file_type=bigWig&files.file_type=bam'
+        '&files.replicate.library.size_range=50-100'
+        '&files.status!=archived&files.biological_replicates=2'
+    )
+    mr = MetadataReport(dummy_request)
+    assert mr._validate_request() is None
```
hitz: This is a "passed" validation?  It's not very clear

src/encoded/tests/test_metadata.py
```diff
@@ -0,0 +1,1034 @@
...
+    dummy_request.environ['QUERY_STRING'] = (
+        'type=Experiment&type=Annotation&files.file_type=bigWig&files.file_type=bam'
+        '&files.replicate.library.size_range=50-100'
+        '&files.status!=archived&files.biological_replicates=2'
+    )
+    mr = MetadataReport(dummy_request)
+    with pytest.raises(HTTPBadRequest):
+        mr._validate_request()
```
hitz, line 1066: This is clearly a failed _validate_request.

hitz (dismissed):
PR approved with the understanding that code will be extended to refactor Annotation and PublicationData examples with tests.

src/encoded/reports/metadata.py
keenangraham: Still working on tests. Just PRed to give you a quick look.

src/encoded/tests/test_metadata.py
keenangraham: Updated to return True. This validation actually redundant now with the @allowed_types decorator. But you could add some other validation logic.

src/encoded/reports/metadata.py
```diff
@@ -0,0 +1,387 @@
...
+
+    def writerow(self, row):
+        self.writer.writerow(row)
+        return self.row
+
+    def write(self, row):
+        self.row = row.encode('utf-8')
+
+
+class BatchedSearchGenerator:
```
keenangraham, line 480: This will be for publication_data. Not used yet.

keenangraham:
I cherry-picked from `load-indexing-once` branch to help with indexing tests, but they are still not 100%. Might drop it.

src/encoded/tests/features/conftest.py
```diff
@@ -46,12 +49,14 @@ def index_workbook(request, app):
         'REMOTE_USER': 'TEST',
     }
     testapp = TestApp(app, environ)
-
-    from encoded.loadxl import load_all
-    from pkg_resources import resource_filename
-    inserts = resource_filename('encoded', 'tests/data/inserts/')
-    docsdir = [resource_filename('encoded', 'tests/data/documents/')]
-    load_all(testapp, inserts, docsdir, log_level=log_level)
+    global index_workbook_loaded
```
keenangraham: Cherry-picked from `load-indexing-once` branch. Probably should be merged separately.

src/encoded/tests/test_metadata.py
```diff
@@ -0,0 +1,1565 @@
...
+        ('field', 'files.genome_annotation'),
+        ('field', 'files.platform.title'),
+        ('field', 'files.controlled_by'),
+        ('field', 'files.status'),
+        ('field', 'files.no_file_available'),
+        ('field', 'files.restricted'),
+        ('field', 'files.s3_uri')
+    ]
+    for param in mr._get_field_params():
+        assert param in expected_field_params
```
forresttanaka, line 990: I’m not sure how useful this test is, but it does seem more useful than the one it replaces, which compared two static lists to see we copied and pasted them correctly.

src/encoded/batch_download.py
```diff
@@ -26,122 +26,11 @@
...
-    ('Controlled by', ['files.controlled_by']),
-    ('File Status', ['files.status']),
-    ('Derived from', ['files.derived_from']),
-    ('S3 URL', ['files.cloud_metadata.url']),
-    ('Size', ['files.file_size']),
-    ('No File Available', ['file.no_file_available']),
-    ('Restricted', ['files.restricted'])
-])
-
 _tsv_mapping_publicationdata = OrderedDict([
```
zoldello: I assume you will move this to constants.py in the future..?

src/encoded/batch_download.py
```diff
@@ -546,147 +313,6 @@ def peak_metadata(context, request):
...
-    writer.writerows(rows)
-    return Response(
-        content_type='text/tsv',
-        body=fout.getvalue(),
-        content_disposition='attachment;filename="%s"' % 'metadata.tsv'
-    )
-
-
 @view_config(route_name='batch_download', request_method=('GET', 'POST'))
 def batch_download(context, request):
```
forresttanaka, line 206: batdh_download and metadata_tsv do a lot of similar things, but not the _same_ things. But I had wondered if they could share some code, but it does seem kind of difficult to do, when the shared code might be a line or two here or there.

forresttanaka (dismissed):
I don’t have a lot of feedback to give, but had a couple of thoughts, whatever they’re worth. But I’ll approve and leave the more detailed look to Phil.

src/encoded/reports/constants.py
```diff
@@ -0,0 +1,136 @@
+from collections import OrderedDict
```
zoldello, line 1: Since this is new and speed is the goal, is OrderedDict needed in this file and other files? Beginning in in Python 3.6,  dict has ordering guaranteed.

src/encoded/reports/metadata.py
```diff
@@ -0,0 +1,380 @@
...
+
+
+def make_experiment_cell(paths, experiment):
+    last = []
+    for path in paths:
+        cell_value = []
+        for value in simple_path_ids(experiment, path):
+            if str(value) not in cell_value:
+                cell_value.append(str(value))
+        if last and cell_value:
```
zoldello, line 51: It may be better to write len(last) == 0. While this is correct, it does require a little thinking/pausing especially to developer that work both front and back end. E.g, in Javascript, for last = []. if (last) == true. In Python, its False.

src/encoded/reports/metadata.py
```diff
@@ -0,0 +1,380 @@
...
+                for v in last
+            ]
+        else:
+            last = cell_value
+    return ', '.join(set(last))
+
+
+def make_file_cell(paths, file_):
+    # Quick return if one level deep.
+    if len(paths) == 1 and '.' not in paths[0]:
```
zoldello, line 63: You may want parenthesis after '.' -  if len(paths) == 1 and ('.' not in paths[0])

src/encoded/reports/metadata.py
```diff
@@ -0,0 +1,380 @@
...
+        value = file_.get(paths[0], '')
+        if isinstance(value, list):
```
zoldello, line 65: Maybe:
return value if not isinstance(value, list) else  return ', '.join([str(v) for v in value])

src/encoded/reports/metadata.py
```diff
@@ -0,0 +1,380 @@
...
+                v + ' ' + cell_value[0]
+                for v in last
+            ]
+        else:
+            last = cell_value
+    return ', '.join(sorted(set(last)))
+
+
+def file_matches_file_params(file_, positive_file_param_list):
+    # Expects file_param_list where 'files.' has been
```
zoldello, line 85: Pardon me if I got American and British English confused. 'files.' is the British way. The American way is "file."; with the double quotes since the word has multiple characters.

src/encoded/reports/metadata.py
```diff
@@ -0,0 +1,380 @@
...
+    # stripped off of key (files.file_type -> file_type)
+    # and params with field negation (i.e. file_type!=bigWig)
+    # have been filtered out.
+    for k, v in positive_file_param_list.items():
+        if '.' in k:
+            file_prop_value = list(simple_path_ids(file_, k))
+        else:
+            file_prop_value = file_.get(k)
+        if not file_prop_value:
```
zoldello, line 94: Maybe-

        if not file_prop_value or str(file_prop_value) not in v:
            return False
        if isinstance(file_prop_value, list):
            return any([str(x) in v for x in file_prop_value])

src/encoded/reports/metadata.py
```diff
@@ -0,0 +1,380 @@
...
+            return False
+        if isinstance(file_prop_value, list):
+            return any([str(x) in v for x in file_prop_value])
+        if str(file_prop_value) not in v:
+            return False
+    return True
+
+
+def group_audits_by_files_and_type(audits):
+    grouped_file_audits = defaultdict(lambda: defaultdict(list))
```
zoldello, line 104: Good idea. to use defaultdict.

src/encoded/reports/metadata.py
```diff
@@ -0,0 +1,380 @@
...
+                self.param_list['@id'].extend(
+                    cart.get('elements', [])
+                )
+
+    def _maybe_add_json_elements_to_param_list(self):
+        try:
+            self.param_list['@id'].extend(
+                self.request.json.get('elements', [])
+            )
+        except ValueError:
```
zoldello, line 194: Do you need to comment why ValueError is eaten rather than be allowed to be thrown?

src/encoded/reports/metadata.py
```diff
@@ -0,0 +1,380 @@
...
+        ]
+        return any(conditions)
+
+    def _get_experiment_data(self, experiment):
+        return {
+            column: make_experiment_cell(fields, experiment)
+            for column, fields in self.experiment_column_to_fields_mapping.items()
+        }
+
+    def _get_file_data(self, file_):
```
zoldello, line 269: I assume "file_" as in "you cannot use 'file' so using 'file_'" ..?

zoldello (dismissed):
It looks fine to me. My comments are cosmetic none are required changes.

src/encoded/batch_download.py
keenangraham: I think we need to hoist batch_download endpoints over next. The file filtering stuff that was fixed for metadata will still be broken here for example.

src/encoded/batch_download.py
keenangraham: Yes.

src/encoded/reports/constants.py
keenangraham: That's a good point. I'm going to leave it explicit for now, but should test for correctness and speed after removing these.

src/encoded/reports/metadata.py
keenangraham: I think this is pretty standard.

src/encoded/reports/metadata.py
keenangraham: Think that adds unnecessary noise here.

src/encoded/reports/metadata.py
keenangraham: Not sure two returns on a line is valid.

src/encoded/reports/metadata.py
keenangraham: This is specifically referring to file properties in experiments which are all under `files.`. So like `files.accession` and `files.status`.

src/encoded/reports/metadata.py
keenangraham: This would short circuit the list check.

src/encoded/reports/metadata.py
keenangraham: A lot faster to parse the audits once per experiment rather than crawl through them for every file in every experiment.

src/encoded/reports/metadata.py
keenangraham: This is lifted over from old code. I left it in because I was concerned the ValueError is coming from trying to access `.json` on `request` when it doesn't exist. Could probably clean this up.

src/encoded/reports/metadata.py
keenangraham: Yep.

src/encoded/reports/metadata.py
zoldello: Typo. I meant: 
return value if not isinstance(value, list) else ', '.join([str(v) for v in value])

forresttanaka (dismissed):
I mostly used this as an educational exercise, and it looks _really_ good — way more organized and understandable.

keenangraham:
This should be ready as long as BDD tests pass.

forresttanaka (approved):
This branch, updated to the latest, passed BDD locally for me, so I’m going to pass it here.

PR #3463, opened 2020-08-03
ENCD-4957 Update search buttons

forresttanaka:
* I deleted anything related to the view control registry as it no longer serves any use.
* The mouse development matrix applied query-string parameters from a React state into the query string for the List and Report button links, but the code to process this was in the generic `ViewControls` component. I moved that out of `ViewControls` to the mouse development matrix code, and `ViewControls` and `BatchDownload` now takes an array of filters to add to `results.filters` to reduce logic duplication, and so they don’t both do things specific to one matrix.
* The mouse development matrix had mutated search result props to make BatchDownload process the additional query-string parameters, but React doesn’t allow prop mutation, so `BatchDownload` now takes the additional filters as a property that it applies.
* The view-control buttons now have ids to make WalkMe targeting easier.


forresttanaka:
I updated a demo for this branch https://encd-4957-view-control-fytanaka.demo.encodedcc.org/

A fresh demo will be fully indexed around 12:30 today.

src/encoded/static/components/matrix_mouse_development.js
```diff
@@ -893,10 +893,45 @@ class MatrixPresentation extends React.Component {
             embryoSecondRowWidth = (countEmbryoSecondRow * (embryoDefault + 2)) - 2;
         }
 
+        // Calculate additional filters for the to view control button links
+        const additionalFilters = [];
+        this.state.developmentStageClick.forEach((f) => {
```
zoldello, line 898: The name- this.state.developmentStageClick, implies this is an event. Why is it iterable?

src/encoded/static/components/matrix_mouse_development.js
```diff
@@ -893,10 +893,45 @@ class MatrixPresentation extends React.Component {
...
+            if (['adult', 'postnatal', 'embryo'].includes(f)) {
+                const stageTerm = f === 'embryo' ? 'embryonic' : f;
+                const stageFilterExists = additionalFilters.filter(f2 => f2.term === stageTerm).length > 0;
+                if (!stageFilterExists) {
+                    additionalFilters.push({
+                        term: f === 'embryo' ? 'embryonic' : f,
```
zoldello, line 904: Why not use stageTerm? 'f' and 'f2' is okay for iterators during iterations like for-loop, but can be confusing when hold values for processing.

src/encoded/static/components/view_controls.js
```diff
@@ -1,105 +1,63 @@
...
+ *
+ * @return {string} Type of search page currently in view: list, report, matrix, summary.
  */
-const viewPathMap = {
-    Search: { path: 'search', icon: 'search', label: 'View results as a list' },
-    Report: { path: 'report', icon: 'table', label: 'View tabular report' },
-    Matrix: { path: 'matrix', icon: 'matrix', label: 'View summary matrix' },
-    Summary: { path: 'summary', icon: 'summary', label: 'View summary report' },
+const getSearchPageType = (context) => {
+    const atType = context['@type'][0];
```
zoldello, line 35: A switch may be better

src/encoded/static/components/view_controls.js
```diff
@@ -1,105 +1,63 @@
...
 import { Modal, ModalHeader, ModalBody, ModalFooter } from '../libs/ui/modal';
 import { svgIcon } from '../libs/svg-icons';
 
 
 /**
- * Specifies the possible types of view controls. Extend this object with new view-control types
- * on the rare occasions that a new search view gets implemented. The string value matches the
- * @type of the search-view JSON. This object is visible externally to this module, so enforce its
- * immutability.
+ * Maps the page type to the corresponding view control elements. Extend this array on the rare
```
zoldello, line 9:
```suggestion
 * Maps the page type to the corresponding view control element. Extend this array on the rare
```

src/encoded/static/components/view_controls.js
```diff
@@ -1,105 +1,63 @@
...
  */
-const viewPathMap = {
-    Search: { path: 'search', icon: 'search', label: 'View results as a list' },
-    Report: { path: 'report', icon: 'table', label: 'View tabular report' },
-    Matrix: { path: 'matrix', icon: 'matrix', label: 'View summary matrix' },
-    Summary: { path: 'summary', icon: 'summary', label: 'View summary report' },
+const getSearchPageType = (context) => {
+    const atType = context['@type'][0];
+    if (atType === 'Search') {
+        return 'search';
```
zoldello, line 37: Should this and other result-values be constants?

src/encoded/static/components/view_controls.js
```diff
@@ -1,105 +1,63 @@
...
-    'Dataset',
-    'Donor',
-    'FileSet',
-    'QualityMetric',
-    'Series',
-];
-
 const modalDefaultText = (
     <>
         <p>
```
zoldello, line 63: How come this <p> is not closed?

src/encoded/static/components/view_controls.js
```diff
@@ -201,86 +94,38 @@ const getQueryFromFilters = (filters) => {
...
-                    filters.push({
-                        term: ageTerm,
-                        remove: '',
-                        field: 'replicates.library.biosample.age_display',
-                    });
-                }
-            }
-        });
-    }
+export const ViewControls = ({ results, additionalFilters }) => {
```
zoldello, line 97: Since the parameters can come from outside the file, I do think you need to verify the variables are usable. For example, results can be null and additionalFilter can be a number.

zoldello (approved):
I could not find any significant error. The comments I made are cosmetic.

src/encoded/static/components/matrix_mouse_development.js
forresttanaka: Yeah, I thought that was confusing too, but that’s the existing name.

src/encoded/static/components/matrix_mouse_development.js
forresttanaka: I don’t like single-letter variable names ever, but those are the existing names.

src/encoded/static/components/view_controls.js
forresttanaka: Yeah, that would be a good idea.

src/encoded/static/components/view_controls.js
forresttanaka: What’s not closed?

PR #3468, opened 2020-08-04
ENCD-5458-use-analysis-objects-for-file-graph

yunhailuo:


forresttanaka (dismissed):
My guess is that once the transition to using analysis_objects is complete, we probably won’t need to clone the context object anymore. But for now, that seems like the best way to go.

forresttanaka (approved):
Oh, I see. Didn’t realize assembly and genome_annotation don’t always exist.

PR #3470, opened 2020-08-05
ENCD-5453-fix-missing-read-depth-audit

iyoungworth:
continuing-work

### DIFF
--- a/src/encoded/audit/experiment.py
+++ b/src/encoded/audit/experiment.py
@@ -1148,7 +1148,7 @@ def check_experiment_chip_seq_standards(
         yield from check_file_chip_seq_library_complexity(f)
 
         if target and target['name'] in ['H3K9me3-human', 'H3K9me3-mouse']:
-            break
+            continue
 
         read_depth = get_file_read_depth_from_alignment(f, target, assay_name)
         yield from check_file_chip_seq_read_depth(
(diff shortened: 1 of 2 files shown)


src/encoded/audit/experiment.py
```diff
@@ -1136,9 +1137,32 @@ def check_experiment_chip_seq_standards(
...
+            experiment.get('control_type'),
+            organism_name,
+            target,
+            read_depth,
+            standards_version
+        )
+
+    # Specifically handles read_depth for experiments with H3K9me3 mark
+    for f in unfiltered_alignment_files:
+        if target and target['name'] not in ['H3K9me3-human', 'H3K9me3-mouse']:
```
gabdank: should not the if statement taken out of the for loop?

src/encoded/audit/experiment.py
```diff
@@ -1136,10 +1137,18 @@ def check_experiment_chip_seq_standards(
...
+    # Handles library_complexity for all cases and read_depth for all but H3K9me3
     for f in alignment_files:
-        target = get_target(experiment)
-        if target is False and not experiment.get('control_type'):
-            return
+
+        yield from check_file_chip_seq_library_complexity(f)
+
+        if target and target['name'] in ['H3K9me3-human', 'H3K9me3-mouse']:
+            break
```
gabdank: I think that the original behavior is altered after this code. For instance in H3K9me3 only one "filtered alignment" file will be used to check the library complexity, instead of all of the "filtered alignments"

gabdank (changes requested):

src/encoded/audit/experiment.py
iyoungworth: Agreed, swapped 'continue' in for 'break'

gabdank (approved):

PR #3477, opened 2020-08-11
ENCD-5407-trans-enhancerpage

khzlin:


src/encoded/static/components/experiment.js
```diff
@@ -344,22 +344,26 @@ DatasetConstructionPlatform.propTypes = {
 
 
 /**
- * Renders both Experiment and FunctionalCharacterizationExperiment objects.
+ * Renders both Experiment, FunctionalCharacterizationExperiment, TransgenicEnhancerExperiment objects.
```
forresttanaka: “both”

src/encoded/static/components/experiment.js
```diff
@@ -769,8 +806,19 @@ const ExperimentComponent = ({ context, auditIndicators, auditDetail }, reactCon
...
+            : null}
+
+            {/* Display the biosample table for TransgenicEnhancerExperiment only. */}
+            {isEnhancerExperiment ?
+                <BiosampleTable
+                    title="Biosamples"
+                    items={context.biosamples}
+                    total={context.biosamples.length}
+                />
+            : null}
```
forresttanaka, line 865: `context.biosample_ontology` gets embedded, but context.biosamples.biosample_ontology doesn’t, so you’re missing the data for couple columns in the biosample table. Unless you can embed biosamples.biosample_ontology, you’ll have to do a request to the server to display this table.

src/encoded/static/components/search.js
```diff
@@ -277,39 +277,64 @@ globals.listingViews.register(Biosample, 'Biosample');
...
     let biosamples = [];
     const treatments = [];
 
-    if (result.replicates && result.replicates.length > 0) {
-        biosamples = _.compact(result.replicates.map(replicate => replicate.library && replicate.library.biosample));
-        // flatten treatment array of arrays
-        _.compact(biosamples.map(biosample => biosample.treatments)).forEach(treatment => treatment.forEach(t => treatments.push(t)));
+    if (isEnhancerExperiment) {
+        if (result.biosamples && result.biosamples.length > 0) {
+            biosamples = _.compact(result.biosamples.map(sample => biosamples));
```
forresttanaka: We need to deduplicate with `_.compact` below because `result.replicates.biosample` can really have duplicates. But I don‘t think `result.biosamples` can have duplicates, so you can just assign `result.biosamples` unmodified to `biosamples`.

src/encoded/static/components/search.js
```diff
@@ -277,39 +277,64 @@ globals.listingViews.register(Biosample, 'Biosample');
...
+        }
+    } else {
+        if (result.replicates && result.replicates.length > 0) {
+            biosamples = _.compact(result.replicates.map(replicate => replicate.library && replicate.library.biosample));
+        }
+    // flatten treatment array of arrays
+    _.compact(biosamples.map(biosample => biosample.treatments)).forEach(treatment => treatment.forEach(t => treatments.push(t)));
     }
```
forresttanaka, line 314: Need to indent the last two lines to avoid an ESLint error.

src/encoded/static/components/search.js
```diff
@@ -277,39 +277,64 @@ globals.listingViews.register(Biosample, 'Biosample');
...
-        }));
+    if (isEnhancerExperiment) {
+        if (result.biosamples && result.biosamples.length > 0) {
+            synchronizations = _.uniq(result.biosamples.filter(sample =>
+            sample && sample.synchronization
+            ).map((sample) => {
+                const biosample = sample;
+                return `${biosample.synchronization}${biosample.post_synchronization_time ? ` + ${biosample.age_display}` : ''}`;
+            }));
+        }
```
forresttanaka: In the above section, you can use `biosamples` instead of `result.biosamples`. Especially with `biosamples` being deduplicated (which as I said, you don’t need to do), this would have used the one still containing duplicates if that were possible.

src/encoded/static/components/search.js
```diff
@@ -277,39 +277,64 @@ globals.listingViews.register(Biosample, 'Biosample');
...
-            replicate.library && replicate.library.biosample && replicate.library.biosample.synchronization
-        ).map((replicate) => {
-            const biosample = replicate.library.biosample;
-            return `${biosample.synchronization}${biosample.post_synchronization_time ? ` + ${biosample.age_display}` : ''}`;
-        }));
+    if (isEnhancerExperiment) {
+        if (result.biosamples && result.biosamples.length > 0) {
+            synchronizations = _.uniq(result.biosamples.filter(sample =>
+            sample && sample.synchronization
+            ).map((sample) => {
```
forresttanaka: The last three lines, from the one calling `_.uniq` to `.map((sample)` need to all be on the same line to avoid an ESLint error.

src/encoded/static/components/search.js
```diff
@@ -277,39 +277,64 @@ globals.listingViews.register(Biosample, 'Biosample');
...
+                const biosample = sample;
```
forresttanaka: You can just use `sample` in the return line below to avoid this assignment, or rename the `sample` parameter to `biosample`. We often just use the singular form of the plural array name as the parameter to things like `map`.

src/encoded/static/components/search.js
```diff
@@ -277,39 +277,64 @@ globals.listingViews.register(Biosample, 'Biosample');
...
+        }
+    } else {
+        if (result.replicates && result.replicates.length > 0) {
+            synchronizations = _.uniq(result.replicates.filter(replicate =>
+                replicate.library && replicate.library.biosample && replicate.library.biosample.synchronization
+            ).map((replicate) => {
+                const biosample = replicate.library.biosample;
+                return `${biosample.synchronization}${biosample.post_synchronization_time ? ` + ${biosample.age_display}` : ''}`;
+            }));
+        }
```
forresttanaka, line 325: You probably have to change the `} else { if (...` to `} else if (...) {...}` to avoid an ESLint error.

forresttanaka (changes requested):
It looks quite good. Most of the work going forward will be to deal with the non-embedded properties you’re using, like in characterizations, but also biosample.biosample_ontology.

src/encoded/static/components/experiment.js
khzlin: do you mean like this? 
                `<FetchedItems
                    title="Biosamples"
                    url={`/search/?type=Biosample&DATASET???`}
                    Component={BiosampleTable}
                />`
The only problem is that in the url I don't think biosample has any `dataset` properties. That will allow me to get all the Biosamples used in the TransgenicEnhancerExperiment. Unless I am mistaken.

forresttanaka: Something like that, yes. I think you’ll need to put together a URL that just lists all the biosample URLs in the experiment, like:

http://localhost:6543/search/?type=Biosample&@id=/biosamples/ENCBS356AAA/&@id=/biosamples/ENCBS888RUL/

…and just keep adding on for each biosample in the experiment. There’s a limit to what you can put into the query string though, so you’ll need to do something different if the list of biosamples could get big, like over a 100 or so.

src/encoded/static/components/experiment.js
```diff
@@ -349,20 +353,39 @@ const ExperimentComponent = ({ context, auditIndicators, auditDetail }, reactCon
...
-    if (replicates.length > 0) {
-        biosamples = _.compact(replicates.map((replicate) => {
-            if (replicate.library) {
-                if (replicate.library.documents && replicate.library.documents.length > 0) {
-                    Array.prototype.push.apply(libraryDocs, replicate.library.documents);
+    const biosampleCharacterizations = [];
+    const appliedModifications = [];
+    if (isEnhancerExperiment) {
+        if (context.biosamples) {
+            biosamples = (context.biosamples && context.biosamples.length > 0) ? context.biosamples : [];
```
forresttanaka: Not a real big deal, but since you already tested `context.biosamples` on the line above, you don’t need to test that again here. You can assume it exists and just immediately test the length.

src/encoded/static/components/experiment.js
```diff
@@ -349,20 +353,39 @@ const ExperimentComponent = ({ context, auditIndicators, auditDetail }, reactCon
...
+        }
+        if (biosamples && biosamples.length > 0) {
```
forresttanaka: Similarly, you know `biosamples` exists because you made it yourself above. You just need to test the length.

src/encoded/static/components/experiment.js
```diff
@@ -447,7 +470,11 @@ const ExperimentComponent = ({ context, auditIndicators, auditDetail }, reactCon
     let nameTip = '';
     const names = organismNames.map((organismName, i) => {
         nameTip += (nameTip.length > 0 ? ' + ' : '') + organismName;
-        nameQuery += `${nameQuery.length > 0 ? '&' : ''}replicates.library.biosample.donor.organism.scientific_name=${organismName}`;
+        if (isEnhancerExperiment) {
+            nameQuery += `${nameQuery.length > 0 ? '&' : ''}biosamples.donor.organism.scientific_name=${organismName}`;
+        } else if (!isEnhancerExperiment) {
```
forresttanaka: Just `} else {` works withou the `if` here, since you know `isEnhancerExperiment` would hold false here, guaranteed.

src/encoded/static/components/experiment.js
```diff
@@ -560,6 +587,23 @@ const ExperimentComponent = ({ context, auditIndicators, auditDetail }, reactCon
                                 </div>
                             : null}
 
+                            {isEnhancerExperiment && biosamples && biosamples.length > 0 ?
```
forresttanaka: Don’t need to test for the existence of `biosamples` since you made it. You just need to test its length.

src/encoded/static/components/experiment.js
```diff
@@ -666,6 +715,45 @@ const ExperimentComponent = ({ context, auditIndicators, auditDetail }, reactCon
                                     <dd><a href={context.elements_cloning}>{globals.atIdToAccession(context.elements_cloning)}</a></dd>
                                 </div>
                             : null}
+
+                            {appliedModifications && appliedModifications.length > 0 ?
```
forresttanaka: Same thing; you made `appliedModifications` so you don’t need to see if it exists before checking its length.

forresttanaka (changes requested):
It looks quite good! I did request changes that aren’t super important, but I think we might as well since we’d probably make these changes on the next maintenance in this code.

forresttanaka (approved):
Nice!

forresttanaka:
Sorry! I accidentally deleted this PR. I hope I restored it properly.

PR #3480, opened 2020-08-13
ENCD-5463-make-target-import_items

jenjou:


gabdank (changes requested):
Please add a test that makes sure non admin can not use Target in FCC submission

gabdank (approved):

PR #3482, opened 2020-08-13
ENCD-5466-make-disease-term-id-name-list

bonitalam:


src/encoded/schemas/biosample.json
```diff
@@ -882,10 +882,16 @@
...
         },
         "disease_term_id": {
-            "title": "Disease ontology ID",
-            "description": "An ontology identifier describing the disease affecting the biosample.",
-            "type": "string",
-            "pattern": "^DOID:[0-9]{2,7}$"
+            "title": "Disease ontology IDs",
+            "description": "Ontology identifiers describing the disease(s) affecting the biosample.",
+            "type": "array",
+            "uniqueItems": true,
```
yunhailuo, line 888: Forgot to mention before but it's better to have `"minItems": 1,` to prevent empty array.

src/encoded/types/biosample.py
```diff
@@ -217,13 +217,16 @@ def age_units(
...
+            term_name = list()
+            for term_id in disease_term_id:
+                if term_id in registry['ontology']:
+                    term_name.append(registry['ontology'][term_id]['name'])
+                else:
+                    msg = 'Disease term ID {} is not a valid ID'.format(
+                        term_id
+                    )
+                    raise ValidationFailure('body', ['disease_term_id'], msg)
+            return term_name
```
yunhailuo, line 236: Also need schema change for array right above. Also remember `"minItems": 1,`

yunhailuo (changes requested):

src/encoded/types/biosample.py
bonitalam: Will do

yunhailuo (approved):

PR #3483, opened 2020-08-13
ENCD-5472 Fix AutoSQL file uploading through form

forresttanaka:


src/encoded/static/components/inputs/file.js
```diff
@@ -2,6 +2,39 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 
+/**
+ * Depending on the filename extension, adjust the type and base64-encoded contents from the
+ * defaults that FileReader assigns. Return unchanged results for any file types that don't need
+ * special treatment.
```
zoldello, line 8: @forresttanaka What kind of special treatment?

src/encoded/static/components/inputs/file.js
```diff
@@ -30,10 +63,13 @@ export default class FileInput extends React.Component {
...
         const reader = new FileReader();
         reader.onloadend = () => {
+            // Fill in the attachment values for edited object PUT requests. Some file types need
+            // special adjustments.
+            const { type, href } = adjustFileValues(rFile, reader.result);
             const value = {
                 download: rFile.name,
-                type: rFile.type || undefined,
-                href: reader.result,
+                type: type || undefined,
```
zoldello, line 71: This is philosophical. To me undefined means "value was not set" while null means "value was set to to be empty". Did you mean undefined or null?

src/encoded/static/components/inputs/file.js
```diff
@@ -2,6 +2,39 @@ import React from 'react';
...
+ * @return {object} -
+ *   type - mime type for file
+ *   href - base64-encoded contents of file
+ */
+const adjustFileValues = (rFile, readerResult) => {
+    let type = rFile.type;
+    let href = readerResult;
+
+    // https://stackoverflow.com/questions/190852/how-can-i-get-file-extensions-with-javascript/1203361#answer-12900504
+    const extension = rFile.name.slice((Math.max(0, rFile.name.lastIndexOf('.')) || Infinity) + 1);
```
zoldello, line 21: Or you could use regex. This works though although it took me a while to understand it and I rarely see Infinity.

src/encoded/static/components/inputs/file.js
```diff
@@ -2,6 +2,39 @@ import React from 'react';
...
+ * defaults that FileReader assigns. Return unchanged results for any file types that don't need
+ * special treatment.
+ * @param {object} rFile Object from <input type="file" />
+ * @param {string} readerResult FileReader `result` property.
+ *
+ * @return {object} -
+ *   type - mime type for file
+ *   href - base64-encoded contents of file
+ */
+const adjustFileValues = (rFile, readerResult) => {
```
zoldello, line 16: @forresttanaka  Is this a valid alternative:
const adjustFileValues = (rFile, readerResult) => {
    // https://stackoverflow.com/questions/190852/how-can-i-get-file-extensions-with-javascript/1203361#answer-12900504
 const extension = rFile.name.slice((Math.max(0, rFile.name.lastIndexOf('.')) || Infinity) + 1);
 
const isFileTypeAs =  extension === 'as’;
 const type = isFileTypeAs ?  'text/autosql' :  rFile.type;
 const href = isFileTypeAs ? readerResult.replace(/^(data:).+?(;.*)$/, '$1text/autosql$2') : readerResult

    return {
        type,
        href,
    };
}

zoldello: My spider sense are tingling when I see you only check for .as extension coded but not other types.

zoldello (approved):
@forresttanaka Approved but I do have some questions about some techniques used. They may be nothing or may require change (albeit I bet on the former.)

src/encoded/static/components/inputs/file.js
forresttanaka: The one you see right in this function, and any in the future whose nature can’t be predicted.

src/encoded/static/components/inputs/file.js
forresttanaka: “Value is not set” because it doesn’t exist.

src/encoded/static/components/inputs/file.js
forresttanaka: You could, but the referenced StackOverflow thread’s participants had gone over different methods in detail and came up with a number of solutions. This one I picked not because it was fastest (it was in fact one of the slower ones) but it took care of the most odd cases, such as file names starting with a dot.

src/encoded/static/components/inputs/file.js
forresttanaka: What would you do with the other types? Only .as files need anything special done with them, as documented here and in the ticket.

forresttanaka: As the only function of this line of code is to get a file extension, and the discussion is documented in the link, and for us this doesn’t happen in a tight loop so performance isn’t a concern, I wasn't willing to spend any more time on it.

PR #3489, opened 2020-08-18
ENCD-5476-load-index-fixture-once

keenangraham:


src/encoded/tests/features/test_trackhubs.py
```diff
@@ -1,6 +1,10 @@
 import pytest
 
-pytestmark = [pytest.mark.indexing]
+
+pytestmark = [
+    pytest.mark.bdd,
```
hitz, line 5: so the trackhubs test is now run with BDD?   Does it  use the same fixture?

keenangraham: It's a different case since test_trackhubs is a child of features folder. It means the index_workbook is used as a proper pytest fixutre and not explicitly imported.

hitz (approved):

PR #3492, opened 2020-08-20
ENCD-5437 Highlight file graph arrows

forresttanaka:
* Much of the changes in graph.js and filegallery.js involve adding classes to edges, but also a couple significant existing bug fixes that had no previous effect:
    * Some places tested infoNode against fileNodeID, which wasn't a valid comparison. I’ve fixed these.
    * nodeIdClick in `<Graph>` sometimes called the click handler twice without any noticeable effect. I’ve rearranged the code to prevent this because we now toggle node highlighting when you click the decoration.
* Unexpectedly, npm install --save for dagre-d3 version 1.3.0 removed some parts of the URL in package.json, but it seems to work fine.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4201,17 +4201,18 @@
       }
     },
     "dagre-d3": {
-      "version": "git+https://github.com/ENCODE-DCC/dagre-d3.git#3c99a06af7837611676c98d6cccee3f23127d9d0",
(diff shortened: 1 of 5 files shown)


zoldello (dismissed):
Looks fine.

forresttanaka:
I left out one call when clicking a QC bubble.

zoldello (approved):

PR #3499, opened 2020-08-25
ENCD-5468-IDR-thresholded-peaks-ATAC-audit

iyoungworth:


src/encoded/audit/experiment.py
```diff
@@ -4200,9 +4200,12 @@ def audit_biosample_perturbed_mixed(value, system, excluded_types):
 def check_experiment_atac_encode4_qc_standards(experiment, files_structure):
     # https://encodedcc.atlassian.net/browse/ENCD-5255
     # https://encodedcc.atlassian.net/browse/ENCD-5350
+    # https://encodedcc.atlassian.net/browse/ENCD-5468
     alignment_files = files_structure.get('alignments').values()
     stable_peaks_files = files_structure.get('stable_peaks_files').values()
-    atac_peaks_files = []
+    idr_thresholded_peaks_files = files_structure.get('idr_thresholded_peaks').values()
+    audit_idr_thresholded_peaks_files = []
```
gabdank: I am not sure about the prefix "audit" being used for the variable name. Maybe "idr_thresholded_peak_files_to_audit"? I am thinking of someone reading the code and trying to decipher the meaning of these variable names

src/encoded/audit/experiment.py
```diff
@@ -4424,18 +4439,71 @@ def check_experiment_atac_encode4_qc_standards(experiment, files_structure):
...
-                if rep_peaks < 100000:
-                    yield AuditFailure('insufficient number of reproducible peaks', detail, level='NOT_COMPLIANT')
+    # The better reproducible peaks value in overlap or IDR thresholded peaks is used in audit
+    reproducible_peaks_files = overlap_peaks_files + audit_idr_thresholded_peaks_files
+    peaks_report = {}
+    for f in reproducible_peaks_files:
+        output_type = f['output_type']
+        if 'biological_replicates' in f:
+            reps = f.get('biological_replicates')
+        else: reps = 0
```
gabdank: I think you should take "reps = 0" on the next line

gabdank (changes requested):
in 4405 you loop over peak enrichment metrics and in 4423 you loop over replication metrics, both of which you took form overlap peaks? but I was under impression Frip and self_consistency and rescue ratios are IDR specific?

line 4450 (casting of rep into String) probably could be removed, instead the 4449
else: reps = 0
could become two lines:
else: 
    reps = "0"

I am not sure why you used “.setdefault” if you are not setting any default values?

lines 4486 it is really easy to get confused between “rep” “report” and “peaks_report” maybe “rep” should be “replicate” to reduce confusion?

I was under impression owe usually say “low number of …” and then “insufficient number…” (sometimes we would say extremely low number …). I don’t think we use “moderate number of reproducible….” not because it is incorrect but because it is not consistent with other audits? But maybe I am not up to date with the audit categories?

src/encoded/audit/experiment.py
iyoungworth: Thanks, I renamed this as suggested

src/encoded/audit/experiment.py
iyoungworth: I ended up removing the 'else' statement entirely; I didn't need to use the variable unless the 'if' statement was true

iyoungworth:
> in 4405 you loop over peak enrichment metrics and in 4423 you loop over replication metrics, both of which you took form overlap peaks? but I was under impression Frip and self_consistency and rescue ratios are IDR specific?

self_consistency and rescue_ratio are now checked for in all peaks files, only for replicated experiments (as per discussion this afternoon. FRiP is currently being checked in overlap peaks files, and when in replicated experiments, only for the file derived from multiple replicates

> line 4450 (casting of rep into String) probably could be removed, instead the 4449

This line does actually need to be a string because it starts out as a list (and this reminded me that I needed to sort the list too, to make sure [1,2] and [2,1] are read as the same)

> I am not sure why you used “.setdefault” if you are not setting any default values?

I took this out and made sure to use .get() in cases where the dict key may not have existed

> lines 4486 it is really easy to get confused between “rep” “report” and “peaks_report” maybe “rep” should be “replicate” to reduce confusion?

I corrected this

I left 'moderate' in some audit category names as discussed (we have it in use elsewhere already)

src/encoded/audit/experiment.py
```diff
@@ -4478,57 +4480,98 @@ def check_experiment_atac_encode4_qc_standards(experiment, files_structure):
...
+        detail_items.append(tuple((str(peaks_report['idr_rep_peaks']), peaks_report.get('idr_rep_peaks_file'))))
+    audit_peaks = ' and '.join(m[0] for m in detail_items)
+    file_links = ', '.join(audit_link(path_to_text(m[1]), m[1]) for m in detail_items)
+    detail = (
+        f'According to ENCODE4 standards, ATAC-seq assays processed by the uniform processing '
+        f'pipeline should have either >150k reproducible peaks in an overlap peaks file, or '
+        f'>70k in an IDR thresholded peaks file. 100-150k or 50-70k peaks respectively is '
+        f'acceptable, and <100k or <50k respectively is not compliant. '
+        f'File(s) {file_links} have {audit_peaks} peaks.')
+    if 'pass' in peaks_report.values():
```
gabdank: So in case you have multiple files you will have them listed? will it be clear which is which? I think that it would make more sense to report them separately? (I think you can yield directly form lines 4514-4536) ?

gabdank (changes requested):

src/encoded/audit/experiment.py
iyoungworth: After we talked, the yield is still outside the loop (because all files need to be checked first), but the detail now has an output type descriptor after each ENCFF. This also led me to see that I needed a variable to track files that had been flagged with multiple QC and avoid yielding if any had issues. I also added a couple more comments to make the intent clearer

gabdank (dismissed):

gabdank (approved):

PR #3500, opened 2020-08-25
ENCD-5469-switch-batch-download-mechanism

keenangraham:


src/encoded/tests/data/inserts/expected_publication_data_batch_download.tsv
```diff
@@ -0,0 +1,6 @@
+"http://localhost/metadata/?type=PublicationData&%40id=%2Fpublication-data%2FENCSR727WCB%2F"
+http://localhost/files/ENCFF002ENO/@@download/ENCFF002ENO.tsv
+http://localhost/files/ENCFF002ENS/@@download/ENCFF002ENS.tsv
+http://localhost/files/ENCFF037HIC/@@download/ENCFF037HIC.hic
+http://localhost/files/ENCFF002COS/@@download/ENCFF002COS.bed.gz
+http://localhost/files/ENCFF003COS/@@download/ENCFF003COS.bigBed
```
hitz, line 6: missing newline, not worth fixing by itself

src/encoded/tests/test_reports_metadata.py
```diff
@@ -1155,10 +1259,9 @@ def test_metadata_metadata_report_build_new_request(dummy_request):
         '&field=files.paired_with&field=files.index_of&field=files.derived_from&field=files.file_size'
         '&field=files.lab.title&field=files.md5sum&field=files.dbxrefs&field=files.href'
         '&field=files.genome_annotation&field=files.platform.title&field=files.controlled_by'
-        '&field=files.status&field=files.no_file_available&field=files.restricted'
         '&field=files.s3_uri&%40id=%2Fexperiments%2FENCSR123ABC%2F'
     )
-    new_request.effective_principals == ['system.Everyone']
+    assert new_request.effective_principals == ['system.Everyone']
```
hitz, line 1267: I don't understand how this is a success?  Just a ways to see if the new Request instantiated?

hitz (commented):
Looks good; just curious did you look at the test coverage?  Might be easier to use that to see if anything was missed.   You are going to check that all the column parameters are valid?

src/encoded/tests/test_reports_metadata.py
keenangraham: `effective_principals` is a list that should equal `['system.Everyone']`. Was missing its assert statement.

src/encoded/tests/data/inserts/expected_publication_data_batch_download.tsv
keenangraham: Probably just how I'm writing it from test.

keenangraham:
> Looks good; just curious did you look at the test coverage? Might be easier to use that to see if anything was missed. You are going to check that all the column parameters are valid?

Using `bin/test --cov=src/encoded/reports/ -s -m 'indexing'` pretty close to 100%:

Screen Shot 2020-08-25 at 6 10 32 PM

Actually the couple places it highlights is pretty useful. Can add a test for this.
Screen Shot 2020-08-25 at 6 07 31 PM

This I lifted over from old code, but should be able to test this too.
Screen Shot 2020-08-25 at 6 07 45 PM

keenangraham:
FWIW 100% coverage now

src/encoded/reports/batch_download.py
```diff
@@ -47,7 +47,7 @@ def _get_metadata_link(self):
         )
 
 
-class BatchDownload(MetadataReport, BatchDownloadMixin):
+class BatchDownload(BatchDownloadMixin, MetadataReport):
```
hitz, line 72: is this cosmetic or does it actually matter?  Usually I expect the mixin after.

hitz (approved):

src/encoded/reports/batch_download.py
keenangraham: I think in this case it's important that the method resolution hits the mixin first, and then hits metadata if nothing is found.

PR #3502, opened 2020-08-25
VALIS-7-update-colors-of-plots

zoldello:


src/encoded/static/scss/encoded/modules/_genome_browser.scss
```diff
@@ -67,7 +67,7 @@
 // styles for bigBed files / annotation files
 .hpgv_track-annotation {
     --transcript-arrow: rgba(0, 0, 0, 1.0);
-    --transcript: #BDC5C0;
+    /* --transcript: #BDC5C0; */
```
emro, line 70: I think it's good to leave this as a comment so that there is documentation of how to set the transcript color in CSS though generally I would say we should delete unused CSS.

emro (approved):
Looks great!

emro:
I'm not sure if I should move this ticket to QA ready because of the BDD test failure, but it sounds like everything is failing on that test so I should?

hitz:
> I'm not sure if I should move this ticket to QA ready because of the BDD test failure, but it sounds like everything is failing on that test so I should?

yeah apparently chromedriver update is bugged and circle-ci got it first.  I guess just pass it, although running locally (on chromedriver < 85 locally would be fine as well.

zoldello:
This PR is no longer needed.

PR #3503, opened 2020-08-26
ENCD-5462-qc-grouping

yunhailuo:


src/encoded/static/scss/encoded/modules/_pipeline.scss
```diff
@@ -245,6 +245,10 @@ g.pipeline-node-qc-metric {
         @media screen and (min-width: $screen-md-min) {
             flex: 0 1 33.33%;
         }
+
+        hr {
+            margin-right: 10px;
+        }
```
forresttanaka: Could you move this hr rule into the @media above? Right not there’s too much margin on the right on mobile, and putting it in that @media would disable this rule on mobile.

forresttanaka (changes requested):
Just need to move the hr rule.

src/encoded/static/scss/encoded/modules/_pipeline.scss
yunhailuo: Thanks a lot for checking that. Missed mobile completely.

src/encoded/static/scss/encoded/modules/_pipeline.scss
```diff
@@ -244,7 +244,11 @@ g.pipeline-node-qc-metric {
     @at-root #{&}__data {
         @media screen and (min-width: $screen-md-min) {
             flex: 0 1 33.33%;
+            hr {
+                margin-right: 10px;
```
forresttanaka: Sorry, I guided you wrong on this one. SCSS nesting can get complicated at times, as in this case where the outer nested `.quality-metrics-modal` gets ignored, so now the hr is too narrow on the file page. I tested a change that works: move the hr out of here and into the @media for `.quality-metrics-modal`, just under the `display: flex`.

forresttanaka (changes requested):
Sorry, I led you the wrong way.

src/encoded/static/scss/encoded/modules/_pipeline.scss
yunhailuo: Thank you very much! There is no way I could figure this out myself.

forresttanaka (approved):
Great!

gabdank (approved):

PR #3518, opened 2020-09-09
ENCD-5498-allow-average-fragment-size-remove-small-RNA-title-calc

iyoungworth:


src/encoded/commands/generate_ontology.py
```diff
@@ -213,7 +213,8 @@
     "OBI:0002675": "MPRA",
     "OBI:0002571": "polyA plus RNA-seq",
     "OBI:0002572": "polyA minus RNA-seq",
-    "OBI:0002631": "scRNA-seq"
+    "OBI:0002631": "scRNA-seq",
+    "OBI:0002112": "small RNA-seq"
 }
```
gabdank, line 218: I think that means you need to generate a new ontology to take effect?

src/encoded/audit/experiment.py
```diff
@@ -3609,20 +3618,27 @@ def audit_experiment_spikeins(value, system, excluded_types):
...
+        if 'size_range' in lib:
+            size_range = lib.get('size_range')
+            if size_range != '>200':
+                continue
+        elif 'average_fragment_size' in lib:
+            size_range = lib.get('average_fragment_size')
+            if size_range <= 200:
+                continue
+        else:
             continue
```
gabdank: this is exact repetition of a code above, I think you can extract it into a function.

src/encoded/tests/test_audit_experiment.py
```diff
@@ -3944,3 +3944,14 @@ def test_audit_experiment_analysis_files(
...
+
+def test_audit_experiment_average_fragment_size(testapp, base_experiment, base_replicate, base_library):
+    # average_fragment_size may stand in for size_range, behavior should match
+    testapp.patch_json(base_experiment['@id'], {'assay_term_name': 'RNA-seq'})
+    testapp.patch_json(base_library['@id'], {'average_fragment_size': 220})
+    testapp.patch_json(base_replicate['@id'], {'library': base_library['@id']})
+    res = testapp.get(base_experiment['@id'] + '@@index-data')
+    assert any(error['category'] == 'missing spikeins'
+               for error in collect_audit_errors(res))
+    assert 'missing RNA fragment size' not in collect_audit_errors(res)
```
gabdank: calling twice to collect_audit_errors, perhaps it is more efficient to define a variable and use it twice?

gabdank (changes requested):

src/encoded/commands/generate_ontology.py
iyoungworth: I went through the [directions here ](https://github.com/ENCODE-DCC/encoded/blob/dev/src/encoded/docs/updating_ontologies.md)as per Bonita's direction

src/encoded/audit/experiment.py
iyoungworth: Good idea, I had both sections call a small function instead

src/encoded/tests/test_audit_experiment.py
iyoungworth: I updated this to check a new variable res_errors instead of collecting the errors twice

gabdank (dismissed):

gabdank (dismissed):

gabdank (approved):

PR #3522, opened 2020-09-10
ENCD-5442-improve-factorbook-integration

khzlin:


forresttanaka (changes requested):
For the display part, I think what you have looks complicated, and it took me a while to figure out what each thing was for. You can use the Underscore.js groupBy function to split the dbxrefs array into an object containing three arrays:

```
const groupedDbxrefs = context.dbxrefs && context.dbxrefs.length > 0 ?
    _(context.dbxrefs).groupBy((dbxref) => {
        if (dbxref.startsWith('SCREEN')) {
            return 'SCREEN';
        }
        if (dbxref.startsWith('FactorBook')) {
            return 'FactorBook';
        }
        return 'normal';
    })
: {};
// Result -- if nothing exists in any of these three categories, then
// that category’s key doesn't exist.
{
    normal: [array of normal dbxrefs]
    SCREEN: [array of SCREEN dbxrefs]
    FactorBook: [arrray of FactorBook dbxrefs]
}
```

That replaces the part where you collect the combined, differences, etc. arrays.

Then when you display the normal ones, you can use:

```
{groupedDbxrefs.normal ?
    <div data-test="external-resources">
        <dt>External resources</dt>
        <dd><DbxrefList context={context} dbxrefs={groupedDbxrefs.normal} /></dd>
    </div>
: null}
```

You don’t need to check for `normal`’s length because it must have at least one for the `normal` key to exist at all.

Then to display the new section, you just use:

```
{groupedDbxrefs.SCREEN || groupedDbxrefs.FactorBook ?
    <React.Fragment>
        <div className="panel__split-heading panel__split-heading--experiment">
            <h4>Encyclopedia Integration</h4>
        </div>

        {groupedDbxrefs.SCREEN ?
            <div data-test="external-resources-screen">
                <dt>Registry of cCREs</dt>
                <dd><DbxrefList context={context} dbxrefs={[groupedDbxrefs.SCREEN[0]]} title={'view regulatory elements in this cell type on SCREEN'} /></dd>
            </div>
        : null}

        {groupedDbxrefs.FactorBook ?
            <div data-test="external-resources-factorbook">
                <dt>Factorbook</dt>
                <dd><DbxrefList context={context} dbxrefs={[groupedDbxrefs.FactorBook[0]]} title={'view motifs and integrative analysis'} /></dd>
            </div>
        : null}
    </React.Fragment>
: null}
```

forresttanaka:
Oh also, since you don’t have any variables in the title string, you can just use `title="view regulatory…"` instead of the braces and single quotes.

src/encoded/static/components/dbxref.js
```diff
@@ -369,8 +380,10 @@ DbxrefList.propTypes = {
     dbxrefs: PropTypes.array.isRequired, // Array of dbxref values to display
     context: PropTypes.object.isRequired, // Object containing the dbxref
     addClasses: PropTypes.string, // CSS class to apply to dbxref list
+    title: PropTypes.string,
```
forresttanaka: Could you add a comment here too, and give a reason for its existence, like “Displayed instead of Dbxref string” or something like that. Same for its occurrence above.

forresttanaka (dismissed):

forresttanaka (approved):
experiment.js has a minor ESLint violation, but it’s not worth fixing at this stage. Hopefully someone will fix that the next time they modify experiment.js.

PR #3523, opened 2020-09-11
ENCD-5528-use-analysis-objects-to-draw-file-graph

yunhailuo:


forresttanaka (approved):
I remember that getting changed before, so I think this should do it.

PR #3526, opened 2020-09-11
ENCD-5377-generate-glossary-page-from-glossary-json

emro:
We are adding a page so that encodeproject.org/glossary is generated from data in glossary.json.
This new page will replace encodeproject.org/help/glossary/.

### DIFF
--- a/src/encoded/glossary.py
+++ b/src/encoded/glossary.py
@@ -1,0 +1,23 @@
+from pyramid.view import view_config
+from pyramid.httpexceptions import HTTPBadRequest
+from urllib.parse import (
+    parse_qs,
+    urlencode,
+)
+
+def includeme(config):
+    config.add_route('glossary', '/glossary{slash:/?}')
+    config.scan(__name__)
+
+@view_config(route_name='glossary', request_method='GET', permission='search')
+def glossary(context, request):
+    result = {
+        '@id': '/glossary/',
+        '@type': ['Glossary'],
+        'title': 'Glossary',
+        '@graph': [],
+        'columns': [],
+        'notification': '',
+        'filters': []
+    }
+    return result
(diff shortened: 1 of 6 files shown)


src/encoded/glossary.py
```diff
@@ -0,0 +1,23 @@
+from pyramid.view import view_config
+from pyramid.httpexceptions import HTTPBadRequest
+from urllib.parse import (
+    parse_qs,
+    urlencode,
+)
```
keenangraham: Doesn't look like any of these are used?

src/encoded/glossary.py
```diff
@@ -0,0 +1,23 @@
...
+
```
keenangraham, line 3: Two lines.

src/encoded/glossary.py
```diff
@@ -0,0 +1,23 @@
...
+def includeme(config):
+    config.add_route('glossary', '/glossary{slash:/?}')
+    config.scan(__name__)
+
```
keenangraham, line 7: Two lines.

src/encoded/glossary.py
```diff
@@ -0,0 +1,23 @@
...
+@view_config(route_name='glossary', request_method='GET', permission='search')
+def glossary(context, request):
```
keenangraham, line 10: Looks okay, just fake endpoint for the frontend to hook into?

keenangraham (dismissed):

src/encoded/glossary.py
emro: Yes, exactly

forresttanaka (approved):
Javascript looks good!

keenangraham (approved):

PR #3527, opened 2020-09-11
VALIS-4 rearrange tracks

emro:
There are a number of changes in this ticket:
- Encode points at a new Valis branch which allows you to click the tops and bottoms of tracks to move them up and down in the browser, reordering the browser files
- Sorting buttons are optionally displayed to sort the browser files
- An arbitrary number of sort parameters can be listed in order to sort the files, and they can be ordered in ascending or descending order
- The sort button for "Output type" is secretly also a sort button for "File type", so that all the bigWigs and bigBeds are grouped with each other

The implementation for the experiments has two sort parameters, the defaults, "Replicate" and "Output type", with "Replicate" as the primary sort, in descending order.

The implementation for the carts has three sort parameters, "Assay term name", "Biosample term name", and "Output type", with "Output type" as the primary sort, in descending order. No sort buttons are displayed because the cart browser is paginated so the sort only applies to the files on a page, not all the files in the cart - therefore displaying the sort buttons would be confusing.

### DIFF
--- a/src/encoded/static/components/genome_browser.js
+++ b/src/encoded/static/components/genome_browser.js
@@ -163,21 +163,20 @@ function mapGenome(inputAssembly) {
 // Ordering by replicate is like this: 'Rep 1,2' -> 'Rep 1,3,...' -> 'Rep 2,3,...' -> 'Rep 1' -> 'Rep 2' -> 'Rep N'
 // Multiplication by 1000 orders the replicates with a single replicate at the end
 const sortLookUp = (obj, param) => {
-    if (param === 'Replicates') {
-        if (obj.biological_replicates.length > 1) {
-            return +obj.biological_replicates.join('');
-        }
-        return +obj.biological_replicates * 1000;
-    } else if (param === 'Output type') {
+    switch (param) {
+    case 'Replicates':
+        return obj.biological_replicates.length > 1 ? +obj.biological_replicates.join('') : +obj.biological_replicates * 1000;
+    case 'Output type':
         return obj.output_type.toLowerCase();
-    } else if (param === 'File type') {
+    case 'File type':
         return obj.file_type.toLowerCase();
-    } else if (param === 'Assay term name') {
+    case 'Assay term name':
         return obj.assay_term_name.toLowerCase();
-    } else if (param === 'Biosample term name') {
+    case 'Biosample term name':
         return obj.biosample_ontology.term_name.toLowerCase();
+    default:
+        return null;
     }
-    return null;
 };
 
 /**
@@ -524,7 +523,7 @@ class GenomeBrowser extends React.Component {
             tracks = this.filesToTracks(newFiles, this.props.label, domain);
         }
         this.setState({ trackList: tracks }, () => {
-            if (this.chartdisplay && tracks !== []) {
+            if (this.chartdisplay && tracks.length > 0) {
                 this.drawTracks(this.chartdisplay);
             }
         });
--- a/src/encoded/static/components/genome_browser.js
+++ b/src/encoded/static/components/genome_browser.js
@@ -6,8 +6,6 @@ import { BrowserFeat } from './browserfeat';
 import { filterForVisualizableFiles } from './objectutils';
 import AutocompleteBox from './region_search';
 
-const domainName = 'https://www.encodeproject.org';
-
 // Files to be displayed for local version of browser
 const dummyFiles = [
     {
@@ -28,6 +26,7 @@ const dummyFiles = [
         },
         status: 'released',
         title: 'ENCFF425LKJ',
+        biological_replicates: [1],
     },
     {
         file_format: 'bigWig',
@@ -47,6 +46,7 @@ const dummyFiles = [
         },
         status: 'released',
         title: 'ENCFF638QHN',
+        biological_replicates: [2],
     },
     {
         file_format: 'bigWig',
@@ -66,6 +66,7 @@ const dummyFiles = [
         },
         status: 'released',
         title: 'ENCFF541XFO',
+        biological_replicates: [1],
     },
     {
         file_format: 'bigBed bedRNAElements',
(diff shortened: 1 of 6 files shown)


src/encoded/static/components/genome_browser.js
```diff
@@ -155,6 +159,27 @@ function mapGenome(inputAssembly) {
     return genome;
 }
 
+// Map the name of a sort parameter to a file object property and convert to form that can be compared
+// Ordering by replicate is like this: 'Rep 1,2' -> 'Rep 1,3,...' -> 'Rep 2,3,...' -> 'Rep 1' -> 'Rep 2' -> 'Rep N'
+// Multiplication by 1000 orders the replicates with a single replicate at the end
+const sortLookUp = (obj, param) => {
```
zoldello, line 165: Switch statement may be better as you have more then three conditions and each condition is just one statement

src/encoded/static/components/genome_browser.js
```diff
@@ -155,6 +159,27 @@ function mapGenome(inputAssembly) {
...
+    if (param === 'Replicates') {
+        if (obj.biological_replicates.length > 1) {
```
zoldello, line 167: return obj.biological_replicates.length > 1 ? +obj.biological_replicates.join(''); +obj.biological_replicates * 1000;

src/encoded/static/components/genome_browser.js
```diff
@@ -226,6 +251,8 @@ class GenomeBrowser extends React.Component {
             x1: 59e6,
             pinnedFiles: [],
             disableBrowserForIE: false,
+            sortToggle: this.props.sortParam.map(() => true), // Indicates ascending or descending order for each sort parameter; true is descending and false is ascending
```
zoldello, line 253: What is the purpose of this? Here is a example of what I see:

Code: [3,6,1,6].map(() => true)
Result:  [true, true, true, true]

Are you trying to get all-true?

src/encoded/static/components/genome_browser.js
zoldello, line 165: Switch statement may be better as you have more then three conditions and each condition is just one statement

src/encoded/static/components/genome_browser.js
```diff
@@ -155,6 +159,27 @@ function mapGenome(inputAssembly) {
     return genome;
 }
 
+// Map the name of a sort parameter to a file object property and convert to form that can be compared
+// Ordering by replicate is like this: 'Rep 1,2' -> 'Rep 1,3,...' -> 'Rep 2,3,...' -> 'Rep 1' -> 'Rep 2' -> 'Rep N'
+// Multiplication by 1000 orders the replicates with a single replicate at the end
+const sortLookUp = (obj, param) => {
+    if (param === 'Replicates') {
```
zoldello: return obj.biological_replicates.length > 1 ? +obj.biological_replicates.join(''); +obj.biological_replicates * 1000;

src/encoded/static/components/genome_browser.js
```diff
@@ -456,37 +452,88 @@ class GenomeBrowser extends React.Component {
         }
     }
 
-    compileFiles(domain) {
-        let newFiles = [];
+    // Sort file object by an arbitrary number of sort parameters
```
zoldello, line 454: might be better to use JSDoc as it will can be used in intellisense.

src/encoded/static/components/genome_browser.js
```diff
@@ -456,37 +452,88 @@ class GenomeBrowser extends React.Component {
...
-            if (files.length > 0) {
-                newFiles = [...this.state.pinnedFiles, ...files];
-            }
+            propsFiles = filterForVisualizableFiles(this.props.files);
         }
-        return newFiles;
+        let files = propsFiles;
+
+        // Apply sort parameters
+        orderedSortParam.forEach((param) => {
```
zoldello, line 489: Let's talk about this. forEach is essentially a for-loop. You are over-writing the files-object  in each spin of the loop. Then you call the .value() on the last item saved in files. Either that is very brilliant as you usual do or its overly complicated.

src/encoded/static/components/genome_browser.js
```diff
@@ -456,37 +452,88 @@ class GenomeBrowser extends React.Component {
...
+        let newFiles = [];
+        const domain = `${window.location.protocol}//${window.location.hostname}`;
+        files = this.sortFiles(primarySort, sortDirection, sortIdx, toggleFlag);
+        newFiles = [...this.state.pinnedFiles, ...files];
+        let tracks = [];
+        if (files.length > 0) {
+            tracks = this.filesToTracks(newFiles, this.props.label, domain);
+        }
+        this.setState({ trackList: tracks }, () => {
+            if (this.chartdisplay && tracks !== []) {
```
zoldello: Why are you doing tracks !== []?

zoldello (dismissed):
I have a question about the ENCODE work but its not a show-stopper.

src/encoded/static/components/genome_browser.js
emro: Yes, it seemed like the simplest way to get an array of `true`s that were the length of the input array. It's just for initializing the sorts to be in descending order. Is there a simpler way to achieve this?

src/encoded/static/components/genome_browser.js
emro: We discussed this and agreed that we can leave this code as-is.
(The over-writing of file object is what accomplishes the nested sorting.)

zoldello: I get this now. You are essentially over-riding the files object with a new stored version of itself. So, line 494 is not a "last version" but rather the entire aggregate.

src/encoded/static/components/genome_browser.js
emro: You're right, this should be .length.

zoldello (approved):
Look good.

PR #3530, opened 2020-09-15
ENCD-5530 Badges for all matrix pages

forresttanaka:
* A few badges no longer had corresponding schema values, so I deleted those badges.
* In cases where I had the Adobe Illustrator artwork, I replaced PNG versions of badges with SVG versions. The cases where I got artwork that already existed still use their PNGs.


src/encoded/static/components/matrix_reference_epigenome.js
```diff
@@ -382,13 +382,13 @@ const MatrixHeader = ({ context, showProjects, project }, reactContext) => {
...
             <div className="matrix-header__title">
-                <h1>{context.title}</h1>
-                <div className="matrix-tags">
-                    <MatrixInternalTags context={context} />
-                    {matrixDescription ?
-                        <div className="matrix-description">{matrixDescription}</div>
-                    : null}
+                <div className="matrix-title-badge">
+                    <h1>{context.title}</h1>
+                    {organism ? <MatrixBadges context={context} type={`ReferenceEpigenome-${organism}`} /> : null}
```
zoldello, line 387: Why is the "organism" here but not in the previous version? I thought you were just adding badges (I see  "{`ReferenceEpigenome-${organism}").

src/encoded/static/components/matrix_sescc_stem_cell.js
```diff
@@ -324,15 +324,13 @@ const MatrixHeader = ({ context }) => {
 
     return (
         <div className="matrix-header">
-            <div className="matrix-header__title matrix-header__title-sescc-spacing">
-                <div className="matrix-tag">
+            <div className="matrix-header__title">
+                <div className="matrix-title-badge">
                     <h1>{context.title}</h1>
+                    <MatrixBadges context={context} />
```
zoldello, line 330: Why no type property?

src/encoded/static/components/objectutils.js
```diff
@@ -654,21 +628,108 @@ ImageWithFallback.propTypes = {
 
 
 /**
- * Display internal tag badges from search results.
+ * Maps an internal tag to the corresponding image file name. Exported for Jest testing.
```
zoldello, line 631: Some say you should never modify code to appease a test. If you subscribe to that belief, should the export be removed and internalTagsMap-const replicated in Jest?

src/encoded/static/scss/encoded/modules/_matrix.scss
```diff
@@ -837,6 +807,25 @@ $row-data-header-width: 200px;
 }
 
 
+// Internal tags and type badges for matrix pages.
+.matrix-title-badge {
```
zoldello, line 811: You have used "matrix-title-badge " a number of time. It does not seem to follow the CSS naming standard we use. I forget the name but has a format like <object>_region__subregion.

src/encoded/static/components/matrix_entex.js
```diff
@@ -468,12 +468,12 @@ const MatrixHeader = ({ context }) => {
...
             <div className="matrix-header__title">
-                <h1>{context.title}</h1>
-                <div className="matrix-tags">
-                    <MatrixInternalTags context={context} />
-                    <div className="matrix-description">
-                        ENTEx is a collaboration with the GTEx Consortium to profile approximately 30 overlapping tissues from four donors.
-                    </div>
+                <div className="matrix-title-badge">
+                    <h1>{context.title}</h1>
+                    <MatrixBadges context={context} />
```
zoldello, line 473: type?

zoldello (dismissed):
My comments are observation that need not be addressed. The code looks fine.

src/encoded/static/components/objectutils.js
forresttanaka: That would mean we would need to keep the copy and the main in sync, which seems counter productive.

src/encoded/static/components/matrix_sescc_stem_cell.js
forresttanaka: SESCC displays badges based on internal_tags. We need `type` only for matrices not defined by internal_tags.

src/encoded/static/components/matrix_reference_epigenome.js
forresttanaka: The user could cause an “undefined” if they manually manipulate the query string.

forresttanaka:
Accidentally left the matrix descriptions left aligned on the page. It now goes back to the existing method of right aligned and 40% the width

zoldello (dismissed):
Code looks good. The css naming style was applied.

forresttanaka:
Updated ChIP-seq badge.

zoldello (approved):
Looks good. The change was an svg update

PR #3531, opened 2020-09-15
ENCD-5535 Fix pipeline graph arrows

forresttanaka:
Added a flag, `displayDecoration`, to graph creation to determine whether to add the arrow highlighting decoration or not. Pipeline graphs don’t use this flag while file graphs now do.


emro (approved):
Looks good!

PR #3533, opened 2020-09-15
ENCD-5524 use new life stage age property for mouse dev matrix

emro:
We're now using a new property, the "life_stage_age" property in order to generate the mouse development matrix. This ends up requiring quite a few changes. Because we're no longer using the "biosample_summary" field, we no longer have to worry about "male" and "female" data getting combined, for instance. Also a number of the queries change because we can now use "life_stage_age" instead of "replicates.library.biosample.age_display" and "replicates.library.biosample.age_display".

### DIFF
--- a/src/encoded/static/components/matrix_mouse_development.js
+++ b/src/encoded/static/components/matrix_mouse_development.js
@@ -103,8 +103,8 @@ const analyzeSubCategoryData = (subCategoryData, columnCategory, colMap, colCoun
         rowData[columnCategory].buckets.forEach((value) => {
             if (stageFilter) {
                 stageFilter.forEach((singleFilter) => {
-                    const filterString = singleFilter.replace(/[()]/g, '');
-                    const keyString = rowData.key.replace(/[()]/g, '');
+                    const filterString = singleFilter.replace('embryo', 'embryonic');
+                    const keyString = rowData.key;
                     if (!filterString || keyString.includes(filterString)) {
                         subCategorySums = updateColumnCount(value, colMap, subCategorySums);
                     }
--- a/src/encoded/static/components/matrix_mouse_development.js
+++ b/src/encoded/static/components/matrix_mouse_development.js
@@ -20,7 +20,7 @@ import { SearchControls } from './search';
  * @constant
  */
 const ROW_CATEGORY = 'biosample_ontology.term_name';
-const ROW_SUBCATEGORY = 'biosample_summary';
+const ROW_SUBCATEGORY = 'life_stage_age';
 const COL_CATEGORY = 'assay_title';
 const COL_SUBCATEGORY = 'target.label';
 
@@ -62,13 +62,6 @@ const mouseMatrixColors = {
     thymus: '#881600',
     'urinary bladder': '#F4CF90',
 };
-
-/**
- * Parse string (longString) on potentially repeated substring (splitString)
- */
-function parseString(longString, splitString) {
-    return longString.substring(longString.indexOf(splitString) + splitString.length + 1);
-}
 
 /**
  * Increment subCategorySums value for bucket keys matching a column name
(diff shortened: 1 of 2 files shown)


src/encoded/static/components/matrix_mouse_development.js
```diff
@@ -129,8 +122,10 @@ const analyzeSubCategoryData = (subCategoryData, columnCategory, colMap, colCoun
 // In the future there are likely to be additions to the data which will require updates to this function
 // For instance, ages measured by months will likely be added
 function sortMouseArray(a, b) {
-    const aStage = a.split(' (')[0];
-    const bStage = b.split(' (')[0];
+    const aStage = a.split(/ (.+)/)[0].replace('embryonic', 'embryo');
```
zoldello, line 125: can 'a' or 'b' be null? Do you need to verify the 'a' and 'b' has the desired components prior to parsing?

src/encoded/static/components/matrix_mouse_development.js
```diff
@@ -261,81 +256,24 @@ const convertExperimentToDataTable = (context, getRowCategories, mapRowCategoryQ
...
-                newRowBucketsNames.push(parseString(temp.key, row.key));
-            }
-        });
-        combinedRowCategoryData[rowIdx].biosample_summary.buckets = newRowBuckets;
-    });
-
     // loop through row data to sort buckets by sorted mouse stage/age
-    const newRowCategoryData = JSON.parse(JSON.stringify(combinedRowCategoryData));
-    combinedRowCategoryData.forEach((row, rowIdx) => {
+    const newRowCategoryData = JSON.parse(JSON.stringify(rowCategoryData));
```
zoldello, line 260: I am curious. JSON.stringify converts converts a javascript object to a string. JSON.parse convert a string to a JavaScript object. Why not just use the Javascript object directly?

src/encoded/static/components/matrix_mouse_development.js
```diff
@@ -351,12 +289,9 @@ const convertExperimentToDataTable = (context, getRowCategories, mapRowCategoryQ
     if (stageFilter) {
         stageFilter.forEach((f) => {
             if (['adult', 'postnatal', 'embryo'].includes(f)) {
-                selectedFilters += `replicates.library.biosample.life_stage=${f === 'embroyo' ? 'embryonic' : f}&`;
+                selectedFilters += `replicates.library.biosample.life_stage=${f === 'embryo' ? 'embryonic' : f}&`;
```
zoldello, line 292: I know this is based off old code; string is immutable. Why not store the additions to selectedFilters into an array and then do a join on that when there is no more to add?

zoldello (dismissed):
Look good. None of my comments are significant nor require changes.

src/encoded/static/components/matrix_mouse_development.js
emro: I need to make a copy of the object, or I end up with an infinite loop when I do the .push. This is not the most optimal way to make a copy of an object in general, but the rowCategoryData object has a lot of nesting so it was the only way I could get it to work.

src/encoded/static/components/matrix_mouse_development.js
emro: Well, stageFilter is the array of filters, and this is basically where we need to do the .join(). But it's a little more complicated than a join because in some cases we want it joined by `replicates.library.biosample.life_stage=` and in other cases by `life_stage_age=`, and we have to replace "embryo" by "embryonic".

src/encoded/static/components/matrix_mouse_development.js
emro: I actually think this will work no matter what. It's the sorting function for `mouseAgeFullArray` which is the array of bucket keys (life_stage_age properties). If one of those happened to be the empty string, then a or b would be an empty string and aAge/aStage or bAge/bStage would be empty strings and they'd just get sorted as such.

zoldello (approved):
Syntax looks fine.

PR #3534, opened 2020-09-15
ENCD-5514-fix-miRNA-experiment-audits-on-miRNA-quant-files

iyoungworth:


gabdank (approved):
I think you need to add a test making sure the audit triggers to the microRNA_quantification file

iyoungworth:
> I think you need to add a test making sure the audit triggers to the microRNA_quantification file

In the micro_rna_experiment insert, I patched the file_tsv_1_1 and file_tsv_1_2 inserts to have the 'microRNA quantifications' output_type, so they're checked correctly now in the old tests ([L1282](https://github.com/ENCODE-DCC/encoded/blob/1f87483ac7f068311690d15508ee8c791aeb28a6/src/encoded/tests/test_audit_experiment.py#L1282) and [L1311](https://github.com/ENCODE-DCC/encoded/blob/1f87483ac7f068311690d15508ee8c791aeb28a6/src/encoded/tests/test_audit_experiment.py#L1311))

PR #3539, opened 2020-09-17
ENCD-5538 fix mouse dev queries

emro:
We want to update the queries on the mouse development matrix so that when you click on a mouse stage like "embryo" or "adult", this results in a query for all the mouse ages that correspond to that mouse stage.

I also reorganized some of the code, pulling some of it into functions.

### DIFF
--- a/src/encoded/static/components/matrix_mouse_development.js
+++ b/src/encoded/static/components/matrix_mouse_development.js
@@ -155,6 +155,43 @@ function sortMouseArray(a, b) {
     return aNumerical - bNumerical;
 }
 
+// Generate object listing all ages corresponding to a mouse stage based on rowCategoryData
+const getMouseAgeObject = (rowCategoryData) => {
+    const mouseAgeFullArray = [];
+    rowCategoryData.forEach((datum) => {
+        datum.life_stage_age.buckets.forEach((bucket) => {
+            mouseAgeFullArray.push(bucket.key.replace('embryonic', 'embryo'));
+        });
+    });
+    mouseAgeFullArray.sort(sortMouseArray);
+    const uniqueAgeArray = [...new Set(mouseAgeFullArray)];
+    const mouseAgeObject = {
+        embryo: uniqueAgeArray.filter(age => age.includes('embryo')).map(age => age.replace('embryo', '')),
+        postnatal: uniqueAgeArray.filter(age => age.includes('postnatal')).map(age => age.replace('postnatal', '')),
+        adult: uniqueAgeArray.filter(age => age.includes('adult')).map(age => age.replace('adult', '')),
+    };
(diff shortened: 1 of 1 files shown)


src/encoded/static/components/matrix_mouse_development.js
```diff
@@ -155,6 +155,43 @@ function sortMouseArray(a, b) {
     return aNumerical - bNumerical;
 }
 
+// Generate object listing all ages corresponding to a mouse stage based on rowCategoryData
+const getMouseAgeObject = (rowCategoryData) => {
+    const mouseAgeFullArray = [];
+    rowCategoryData.forEach((datum) => {
```
zoldello, line 161: This is the first time I have ever seen a developer use datum for a single part of data.

zoldello, line 161: Not important, it seems you could of had used a map-function for this.

src/encoded/static/components/matrix_mouse_development.js
```diff
@@ -155,6 +155,43 @@ function sortMouseArray(a, b) {
...
+
+// Convert array of filters into search query
+// For life stage filters, we must append search terms for all corresponding ages
+const generateFilterQuery = (stageFilter, mouseAgeObject) => {
+    let selectedFilters = '';
+    if (stageFilter) {
+        stageFilter.forEach((f) => {
+            if (['adult', 'postnatal', 'embryo'].includes(f)) {
+                mouseAgeObject[f].forEach((mouseAge) => {
+                    selectedFilters += `life_stage_age=${f.replace('embryo', 'embryonic')}${mouseAge}&`;
```
zoldello, line 184: Like a previous review, string are immutable so its best to store values in an array and then do a join. However, its not a big deal for small loop-sizes

src/encoded/static/components/matrix_mouse_development.js
```diff
@@ -155,6 +155,43 @@ function sortMouseArray(a, b) {
     return aNumerical - bNumerical;
 }
 
+// Generate object listing all ages corresponding to a mouse stage based on rowCategoryData
+const getMouseAgeObject = (rowCategoryData) => {
```
zoldello, line 159: "Hungarian notation" notation is frowned upon in the JavaScript community (or at least those that care about irrelevant Computer History so are aware of it)

zoldello (approved):
The code looks good, "replicates.library.biosample" was removed and that is the core of the solution. My comments are not significant.

src/encoded/static/components/matrix_mouse_development.js
emro: Yes, it makes sense not to include "object" in the name because it should be evident that it is an object.

PR #3542, opened 2020-09-21
ENCD-5544-chip-matrix-remove-x-result-and-add-mint

zoldello:


forresttanaka (approved):
The code looks good. I think you might need to do something about the placeholder text though. At some narrow screen widths, it either gets cut off or wraps in an awkward way. This isn’t a new problem — it happens in production too; just one I noticed now. The problem looks like this in Safari 14 which came out last week at a narrowish window width:

Screen Shot 2020-09-21 at 16 20 00

PR #3543, opened 2020-09-22
ENCD-4275 ENCORE matrix

forresttanaka:
Emma reviewed this, and the only thing that has changed since is the link generation in `convertEncoreToDataTable`.

### DIFF
--- a/src/encoded/static/components/matrix_encore.js
+++ b/src/encoded/static/components/matrix_encore.js
@@ -1,0 +1,619 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import _ from 'underscore';
+import url from 'url';
+import QueryString from '../libs/query_string';
+import * as encoding from '../libs/query_encoding';
+import { Panel, PanelBody } from '../libs/ui/panel';
+import { svgIcon } from '../libs/svg-icons';
+import DataTable from './datatable';
+import * as globals from './globals';
+import { MatrixBadges } from './objectutils';
+
+
+/**
+ * Limit the assays to display on the horizontal axis of the main ENCORE matrix to the following:
+ */
+const encoreDisplayedAssays = [
+    'shRNA RNA-seq',
+    'CRISPR RNA-seq',
+    'eCLIP',
+    'TF ChIP-seq',
+    'RNA Bind-n-Seq',
+    'iCLIP',
+];
+
+/**
+ * Limit the assays to display on the horizontal axis of the inset RNA-seq matrix to the following:
+ */
+const rnaSeqDisplayedAssays = [
+    'total RNA-seq',
+    'polyA plus RNA-seq',
+];
+
+/**
+ * Limit the biosample term names to display within each assay to the following:
+ */
+const displayedTermNames = [
+    'K562',
+    'HepG2',
+];
+
+/**
+ * Assays in which we don't display a biosample.
+ */
+const noBiosampleAssays = [
+    'RNA Bind-n-Seq',
+];
+
+
+/**
+ * The ENCORE matrix does front-end filtering, which means we could have columns we can display
+ * according to the `encoreDisplayedAssays` global, but which might only contain data we *don't*
+ * display because it doesn’t have data in terms names we display from the `displayedTermNames`
+ * global. This function checks `colCategoryKey` (assay_title) to see if it has any data for term
+ * names in `displayedTermNames`. If not, we don’t render that column even though it exists in the
+ * matrix JSON.
+ * @param {object} context Matrix data object for the page
+ * @param {string} colCategoryKey Title of assay column we're checking
+ * @param {string} colCategory Column category property e.g. assay_title
+ * @param {string} colSubCategory Column subcategory property e.g. biosample_ontology.term_name
+ * @param {string} rowCategory Row category property e.g. target.label
+ *
+ * @return {bool} True if we should display the given colCategoryKey column
+ */
+const doesRowDataExist = (context, colCategoryKey, colCategory, colSubCategory, rowCategory) => {
+    const isColCategoryDataFound = context.matrix.y[rowCategory].buckets.find((rowCategoryItem) => {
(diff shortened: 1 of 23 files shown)


src/encoded/static/components/matrix_encore.js
```diff
@@ -0,0 +1,619 @@
...
+ * matrix JSON.
+ * @param {object} context Matrix data object for the page
+ * @param {string} colCategoryKey Title of assay column we're checking
+ * @param {string} colCategory Column category property e.g. assay_title
+ * @param {string} colSubCategory Column subcategory property e.g. biosample_ontology.term_name
+ * @param {string} rowCategory Row category property e.g. target.label
+ *
+ * @return {bool} True if we should display the given colCategoryKey column
+ */
+const doesRowDataExist = (context, colCategoryKey, colCategory, colSubCategory, rowCategory) => {
```
zoldello, line 65: May the reduce-function could of had made this clearer..?

src/encoded/static/components/matrix_encore.js
```diff
@@ -0,0 +1,619 @@
...
+const useMatrixScrollHandler = (scrollElement) => {
+    const [scrolledRight, setScrolledRight] = React.useState(true);
+
+    /**
+     * Show a scroll indicator depending on current scrolled position.
+     * @param {object} element DOM element to apply shading to
+     */
+    const handleScrollIndicator = React.useCallback((element) => {
+        // Have to use a "roughly equal to" test because of an MS Edge bug mentioned here:
+        // https://stackoverflow.com/questions/30900154/workaround-for-issue-with-ie-scrollwidth
```
zoldello, line 367: Which of the solutions did you use? It seems like you did parts of the selected solution and the solution by siphon. It is safe to assume developers won't go read it, so maybe explain why you have "scrollDiff <2"..?

src/encoded/static/components/matrix_encore.js
```diff
@@ -0,0 +1,619 @@
...
+            // Right edge of matrix scrolled into view.
+            setScrolledRight(true);
+        } else if (scrollDiff >= 2 && scrolledRight) {
+            // Right edge of matrix scrolled out of view.
+            setScrolledRight(false);
+        }
+    }, [scrolledRight]);
+
+    React.useEffect(() => {
+        if (scrollElement.current) {
```
zoldello, line 379: should this be:
if (scrollElement && scrollElement.current)
?

src/encoded/static/components/matrix_encore.js
```diff
@@ -0,0 +1,619 @@
...
+            const handleScrollEvent = event => handleScrollIndicator(event.target);
+            const handleResizeEvent = () => handleScrollIndicator(scrollElement.current);
+
+            // Cache the reference to the scrollable matrix <div> so that we can remove the "scroll"
+            // event handler on unmount, when scrollElement.current might no longer point at this <div>.
+            const matrixNode = scrollElement.current;
+
+            // Attach the scroll- and resize- event handlers, then force the initial calculation of
+            // `scrolledRight`.
+            scrollElement.current.addEventListener('scroll', handleScrollEvent);
```
zoldello, line 390: In theory, this hook can be called multiple times without the unmounting part called. If that theory is correct. Do you detach the previously binded scroll and resize event somewhere else too? Otherwise, that may be a memory leak.

src/encoded/static/scss/encoded/modules/_matrix.scss
```diff
@@ -1354,3 +1365,285 @@ $divider-color: #000; // Color of dividers between assays with targets
...
+        }
+    }
+}
+
+.encore-column-spacer {
+    border-left: 1px solid $matrix-border-color;
+    width: $encore-spacer-width;
+}
+
+.type-EncoreMatrix {
```
zoldello, line 1644: why isn't this part of line 1627?

src/encoded/tests/data/inserts/experiment.json
```diff
@@ -10,7 +10,7 @@
         "status": "released",
         "date_submitted": "2015-01-20",
         "date_released": "2016-01-01",
-        "internal_tags":["DREAM", "ENCORE"],
+        "internal_tags": ["DREAM"],
```
zoldello, line 13: Why remove "ENCORE"?

zoldello (approved):
Code looks okay. I have some questions but I doubt they are significant. So, I am moving this along.

src/encoded/static/components/matrix_encore.js
forresttanaka: Possibly, but `reduce` has its own clarity costs that I usually reserve for the simplest of cases. This nested case seems to me could seem extremely mind bending with `reduce`.

src/encoded/static/components/matrix_encore.js
forresttanaka: scrollElement gets assigned from useRef on every render, so it has to exist as an object.

src/encoded/static/components/matrix_encore.js
forresttanaka: The unmounting callback gets called before useEffect gets called after the first time. So the order of execution becomes:

addEventListener
... _time passes_
removeEventListener
addEventListener
... _time passes_
removeEventListener
addEventListener
... _and so on_

https://reactjs.org/docs/hooks-reference.html#cleaning-up-an-effect

src/encoded/static/scss/encoded/modules/_matrix.scss
forresttanaka: It should have. I missed it as I desperately tried to make rc1.

src/encoded/tests/data/inserts/experiment.json
forresttanaka: Because I made other test experiments, and this one didn’t really fit ENCORE experiments — someone (me maybe) just added ENCORE to this experiment back in the day.

PR #3546, opened 2020-09-22
ENCD-5374 tooltip improvements

emro:
This ticket accomplishes a number of small updates:
(1) Implements convention of using (?) for tooltips and (i) for pop-ups
(2) Fixes font discrepancy on homepage tooltips and navigation tooltips
(3) Prevents execution of form links when tooltip buttons are clicked on the homepage on mobile
(4) Prevents cropping of tooltips when they land near the edge of a screen
(5) Standardizes delay in hiding tooltips to be 400 ms for all tooltips
(6) Confirms that tooltips are accessible through tab-completion (no changes required)


src/encoded/static/libs/ui/tooltip.js
```diff
@@ -54,6 +74,8 @@ const Tooltip = (props) => {
                     onFocus={e => setDefinitionVisibility(true, e)}
                     onBlur={e => setDefinitionVisibility(false, e)}
                     className={`tooltip-container__trigger ${showDefinition ? 'show' : ''}`}
+                    ref={buttonRef}
+                    type="button"
```
forresttanaka, line 78: Is the `type="button"` required? Shouldn’t the browser know that?

src/encoded/static/scss/encoded/modules/_tooltip.scss
```diff
@@ -45,9 +45,13 @@ $tooltip-color: #000;
     z-index: 1030;
     display: block;
     visibility: hidden;
-    font-size: 12px;
-    line-height: 1.4;
+    font-size: 1rem;
+    line-height: 1.2;
     opacity: 0;
+    @media screen and (max-width: $screen-md-min) {
```
forresttanaka: Should have `min-width` so we don’t mix mobile first and desktop first.

forresttanaka (changes requested):
Looks good. The one change is to reverse the CSS to use min-width instead of max-width. I let that go in the past, but I think if we keep mixing mobile first and desktop first, things will get very confusing.

src/encoded/static/libs/ui/tooltip.js
emro: This surprised me too. Shockingly, it is required. This addresses (3) above. This is where I saw the solution: https://stackoverflow.com/questions/932653/how-to-prevent-buttons-from-submitting-forms

src/encoded/static/scss/encoded/modules/_tooltip.scss
emro: Agreed. I also changed the other instance of "max-width".

forresttanaka (dismissed):
Looks cool. I had to look up “unset” — never saw that before. And thanks for the info on type="button". Didn’t know that one.

emro:
There are two main changes in the latest commit.
(1) More instances of (?) and (i) buttons are changed to ensure consistency - pop-ups are (i)s and black, tooltips are (?) and gray.
(2) I deliberated a bit over how to handle this case: https://www.encodeproject.org/replication-timing-series/ENCSR888VII/ where there are (i) buttons which do not open pop-ups. The issue is that `setInfoNodeId` and `setInfoNodeVisible` are missing on `<FileTable>` on `<DatasetFiles>`. Originally when I tried to solve this, I noticed a property on <DownloadableAccession> which was going unused but was set on many instances of the component, `buttonEnabled`. Unfortunately it seems like this property is not very useful. I think maybe a new ticket should sort out if that property is needed at all and perhaps another new ticket should look into adding the node related functions on the file table for dataset pages. My change should solve the problem for the meantime - at least we won't have buttons that don't execute any click function.

forresttanaka (changes requested):
Generally looks fine as before. But I think the scrolling solution for short tables has issues: if you scroll, the mouse no longer hovers over the icon, and the tooltip quickly disappears before you can read the rest. In my opinion for now we should ignore this problem and simply let the tooltips get cut off for now and not have the scrolling solution. But to solve this longer term:

1. People who write the text shouldn’t make that text so long. If people need a long explanation, they should include a link in the tooltip (like restricted-file tooltips have) to a help-page anchor, and the tooltip shouldn’t disappear as long as people hover over the tooltip itself, again like restricted-file tooltips.

2. We can re-architect tooltips to get inserted at the bottom of the DOM instead of at the tooltip-icon element, so that the tooltip gets its z-index relative to `<body>` and we don’t have to worry about them getting cut off by surrounding elements. The main navigation menus work this way.

And we can do both of these over time.

emro:
I reverted the overflow-y change and made a new ticket to address fixing the short table / tall tooltip problem.

forresttanaka (dismissed):
I’ll write a ticket to re-architect the tooltips.

forresttanaka (dismissed):
Glad to finally be rid of all the different tooltip implementations.

forresttanaka (approved):
OK, the new changes look good.

PR #3547, opened 2020-09-23
ENCD-5542 Display average fragment size

forresttanaka:
Old code needed to generate `<dt>`/`<dl>` pairs in a component without an enclosing div because React.Fragment didn’t exist. That’s why this complicated mechanism exists to display library properties in the experiment summary panel. It should be rewritten to use React.Fragment someday, but I felt like this ticket wasn’t the one to do that, since only one line gets added.

### DIFF
--- a/src/encoded/static/components/experiment.js
+++ b/src/encoded/static/components/experiment.js
@@ -88,6 +88,18 @@ const libraryPropertyExtractors = {
         return { libraryPropertyString, libraryPropertyComponent };
     },
 
+    size_range: (library, replicates) => {
+        // Always display replicate if any other replicate libraries have `average_fragment_size`.
+        const libraryPropertyForceRepDisplay = replicates.some(replicate => replicate.library && replicate.library.average_fragment_size);
+        return { libraryPropertyString: library.size_range, libraryPropertyForceRepDisplay };
+    },
+
+    average_fragment_size: (library, replicates) => {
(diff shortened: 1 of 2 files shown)


emro (dismissed):

forresttanaka:
Now that two different library properties need to adjust their rendering depending on the other property, this change has become more complicated. `libraryPropertyExtractors`, which is used for more complicated library property displays, now optionally returns `libraryPropertyForceRepDisplay` to force that property to display its replicate even if only one exists. Now `size_range` and `average_fragment_size` both check for the existence of the other property and force the replicates to display if it exists. I updated the existing demo (I also deployed a new one):

Single `average_fragment_size`:
https://encd-5542-avg-fragment-fytanaka.demo.encodedcc.org/experiments/ENCSR942YMN/

Multiple `average_fragment_size`:
https://encd-5542-avg-fragment-fytanaka.demo.encodedcc.org/experiments/ENCSR774MGO/

Single `average_fragment_size` _and_ single `size_range`:
https://encd-5542-avg-fragment-fytanaka.demo.encodedcc.org/experiments/ENCSR064VKV/

Multiple `average_fragment_size` _and_ multiple `size_range`:
https://encd-5542-avg-fragment-fytanaka.demo.encodedcc.org/experiments/ENCSR000BXW/

emro (approved):

PR #3550, opened 2020-09-24
ENCD-5347-show-strand-specificity

zoldello:


forresttanaka (dismissed):
Oh, I see. I didn’t realize the definition of strand_specificity had changed in the schema recently.

zoldello:
> Oh, I see. I didn’t realize the definition of strand_specificity had changed in the schema recently.

I did not change the definition of the schema. I just changed how it is displayed on the summary region of the Experiment page. Am I missing something?

forresttanaka:
I don’t know if you’re missing anything. I didn’t say you changed the schema.

src/encoded/static/components/experiment.js
```diff
@@ -137,13 +137,7 @@ const libraryPropertyExtractors = {
...
-        const strandSpecificity = library.strand_specificity;
-        let libraryPropertyString = null;
-
-        // A strand_specificity of "false" is still a displayable value.
-        if (strandSpecificity !== undefined) {
-            libraryPropertyString = strandSpecificity ? 'Strand-specific' : 'Non-strand-specific';
-        }
+        const libraryPropertyString = library.strand_specificity;
         return { libraryPropertyString };
     },
```
forresttanaka, line 142: Seems like you can delete this whole property. All it does now is return what it’s been given.

src/encoded/static/components/experiment.js
```diff
@@ -232,13 +227,26 @@ const LibraryProperties = ({ replicates }) => {
...
+                    if (isStrandSpecificity) {
+                        const strandSpecificValues = libraryPropertyElements.map(element => element.value);
+                        const strandSpecificityFirstValue = strandSpecificValues[0];
+
+                        if (strandSpecificValues.every(value => value === strandSpecificityFirstValue)) {
+                            strandSpecificity = `Strand-specific${strandSpecificityFirstValue === 'strand-specific' ? '' : ` (${strandSpecificityFirstValue})`}`;
+                        } else {
+                            strandSpecificity = 'Strand-specific (mixed)';
+                        }
+                    }
```
forresttanaka, line 241: Just to have this in the Github record, I think a future ticket should move this logic back to `libraryPropertyExtractors` which was designed for this kind of case, to avoid this special case.

forresttanaka (approved):

src/encoded/static/components/experiment.js
zoldello: Good point. Still, I did not add the change to libraryPropertyExtractors because it looks at one strand_specificity at a time. To determine what to display, I needed to be able to examine all strands_specificities.

src/encoded/static/components/experiment.js
zoldello: Correct, bu out of necessity. The old code used to show each strand per line and the old code served that need. The new way has a different format; it is a summary of all the strands. So, I remove the display logic out. The fact that I have to explain it suggest this part of the library code may need some clean up.

src/encoded/static/components/experiment.js
forresttanaka: The `replicates` array gets passed in the second parameter to the library extractors, so you have access to all libraries in the experiment. `size_range` uses this.

src/encoded/static/components/experiment.js
forresttanaka: Right, you moved the display logic out, but kept the property which effectively does nothing now.

PR #3552, opened 2020-09-25
[utils] Command for admin user creation

pedrohr:
Useful for bootstrapping an encoded application. Must be run as a pyramid command


keenangraham (approved):

keenangraham:
Not sure the use case but code looks okay. Could probably generalize a groups arg to create any kind of user.

vsmalladi:
So does this run when we launched?

hitz:
> So does this run when we launched?

no you have to hack it and run it manually (it uses Cherry lab etc.)... in should be in bin/create-admin-user after buildout

pedrohr:
It was merged on Snovault: https://github.com/ENCODE-DCC/snovault/pull/293

PR #3555, opened 2020-09-28
ENCD-5537 save browser position on sort

emro:
Currently when you click on the sort buttons on the Valis browser, the position is reset to the default position. If the user has scrolled the browser to a new location, we want the browser to preserve this new location when the sort buttons are clicked.

### DIFF
--- a/src/encoded/static/components/genome_browser.js
+++ b/src/encoded/static/components/genome_browser.js
@@ -522,7 +522,22 @@ class GenomeBrowser extends React.Component {
         if (files.length > 0) {
             tracks = this.filesToTracks(newFiles, this.props.label, domain);
         }
-        this.setState({ trackList: tracks }, () => {
+        let contig = this.state.contig;
+        let x0 = this.state.x0;
+        let x1 = this.state.x1;
+        if (this.chartdisplay) {
+            const scrollLocation = this.chartdisplay.getElementsByClassName('hpgv_panel-header')[0].getElementsByTagName('span')[0].innerText;
+            const splitScrollLocation = scrollLocation.split(' ');
+            contig = splitScrollLocation[0];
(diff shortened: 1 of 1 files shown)


src/encoded/static/components/genome_browser.js
```diff
@@ -522,7 +522,22 @@ class GenomeBrowser extends React.Component {
         if (files.length > 0) {
             tracks = this.filesToTracks(newFiles, this.props.label, domain);
         }
-        this.setState({ trackList: tracks }, () => {
+        let contig = this.state.contig;
+        let x0 = this.state.x0;
+        let x1 = this.state.x1;
+        if (this.chartdisplay) {
+            const scrollLocation = this.chartdisplay.getElementsByClassName('hpgv_panel-header')[0].getElementsByTagName('span')[0].innerText;
```
zoldello, line 529: The older means of referencing DOM like using "getElementsByClassName" is done when you need I speed. The speed difference is negligible unless you are using it in a loop with tens of thousands of iterations. Is there a special reason you are using it?

zoldello (approved):
This make sense. During sort, you save relevant parameter in the state. My comment is in the PR is not significant.

src/encoded/static/components/genome_browser.js
```diff
@@ -522,7 +522,22 @@ class GenomeBrowser extends React.Component {
         if (files.length > 0) {
             tracks = this.filesToTracks(newFiles, this.props.label, domain);
```
zoldello, line 523: I take it that this PR is part of an addition to a different ticket...?

zoldello (approved):

src/encoded/static/components/genome_browser.js
emro: No, this is the only change.

src/encoded/static/components/genome_browser.js
emro: Do you mean that it would be better to use a ref? I think that could be tricky since I'm referring to an element generated in a component in a library. I thought about making a method to return the browser position but this was the simplest solution since it didn't require any changes to Valis and I think it should work consistently.

PR #3556, opened 2020-09-28
ENCD-5526 Display motified_site_by_gene_id on GM pages

forresttanaka:
I moved the `renderers` static object out of `ModificationSiteItems` to the module name space where I should have put that object to begin with and removed the redundant `modificationSitePropNames` variable.

### DIFF
--- a/src/encoded/static/components/genetic_modification.js
+++ b/src/encoded/static/components/genetic_modification.js
@@ -54,9 +54,55 @@ IntroducedTags.propTypes = {
 };
 
 
-// All possible genetic modification object modification site property names. This might need to
-// change if the genetic_modification.json schema changes.
-const modificationSitePropNames = ['modified_site_by_target_id', 'modified_site_by_coordinates', 'modified_site_by_sequence', 'modified_site_nonspecific'];
+const modifiedSiteRenderers = {
+    modified_site_by_target_id: (gm) => {
+        let targetName;
+        let targetLink;
+
+        if (typeof gm.modified_site_by_target_id === 'string') {
+            // Non-embedded target; get its name from its @id.
+            targetName = globals.atIdToAccession(gm.modified_site_by_target_id);
+            targetLink = gm.modified_site_by_target_id;
+        } else {
+            // Embedded target; get its name from the embedded object.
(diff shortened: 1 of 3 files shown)


src/encoded/static/components/genetic_modification.js
```diff
@@ -54,73 +54,74 @@ IntroducedTags.propTypes = {
...
+            <dt>Gene</dt>
+            <dd className="sequence"><a href={gm.modified_site_by_gene_id['@id']}>{gm.modified_site_by_gene_id.symbol}</a></dd>
+        </div>
+    ),
+    modified_site_by_coordinates: (gm) => {
+        const { assembly, chromosome, start, end } = gm.modified_site_by_coordinates;
+        return (
+            <div data-test="mscoords">
+                <dt>Coordinates</dt>
+                <dd>{`${assembly} chr${chromosome}:${start}-${end}`}</dd>
```
zoldello, line 89: Do you need space between start and end?

src/encoded/static/components/genetic_modification.js
```diff
@@ -54,73 +54,74 @@ IntroducedTags.propTypes = {
 };
 
 
-// All possible genetic modification object modification site property names. This might need to
-// change if the genetic_modification.json schema changes.
-const modificationSitePropNames = ['modified_site_by_target_id', 'modified_site_by_coordinates', 'modified_site_by_sequence', 'modified_site_nonspecific'];
+const modifiedSiteRenderers = {
```
zoldello, line 57: I have not see someone use an object like this in a long time. It was common pattern used before ES6 was widely accepted in forms of Patterns like revealing module pattern (not exactly but the core idea). Nowadays, people would just create a new class. Your keys would be the methods in the class. The value would be the body.

zoldello (approved):
Looks good. My comment is not significant.

src/encoded/static/components/genetic_modification.js
forresttanaka: Not sure how a class would work for this case. Normally you instantiate a class to make an object containing methods that operate on that object. In this case, this object’s “methods” get used on different things that aren’t necessarily objects, and certainly not connected to methods themselves.

PR #3560, opened 2020-09-29
ENCD-5529-add-SOP-for-adding-new-assay

iyoungworth:
minor-changes

added-2-missing-links

link-fixes-and-new-code-block

stylistic-changes-and-added-more-links

### DIFF
--- a/src/encoded/docs/adding_new_assays.md
+++ b/src/encoded/docs/adding_new_assays.md
@@ -16,9 +16,48 @@ Guide to where to edit Source Code
 Steps to add a new assay
 ---------------- 
 
-1. Schemas and changelogs:
+1. Ontology:
 
-    Assays are identified by precise ontological terms that we list in a property *assay_term_name*. This is used in many different objects on the portal, and in most cases a mixin property ```assay``` provides this enum list. You will need to add your new assay type to the enum in mixins.json and update the changelogs of the objects that import this property.
+    Assays are identified by precise ontological terms that we list in a property *assay_term_name*.
+    First, check to see if the new assay type has an entry in [OBI].
+    * If there is an entry, you will use the OBI identifier for your term (e.g., OBI:0002039).
+    * If there is no entry, write a New Term Request ticket on the [NTR board], and provide a publication describing the assay type. The ticket number will be used as your assay term identifier (e.g., NTR-564 becomes NTR:0000564).
+
+    Add the assay term name and identifier to [types/assay_data.py].
+
+    Your term identifier will be associated with three properties:
+    * ```name```: the name for your assay type in its non-abbreviated form (e.g., *single-nucleus ATAC-seq*)
+    * ```preferred_name```: your term, the shortened form of the assay name (e.g., *snATAC-seq*)
+    * ```assay```: the parent term for that your assay term will slim to
+
+    If you are using an OBI term, you will need to check if ```preferred_name``` and ```assay``` are already specified. Check for these properties in the OBI entry for your term in the ENCODE ```ontology.json``` (see [Updating ontologies] for details). If they're absent, you will need to manually provide these terms in these files:
+    * Associate the OBI term with the desired ```preferred_name``` in [src/encoded/commands/generate_ontology.py]
+    * Assign an ```assay``` slim in [src/encoded/commands/manual_slims.py]
+
+    If instead you are using an NTR identifier, add an entry to [src/encoded/commands/ntr_terms.py] providing the three properties (```name```, ```preferred_name```, and ```assay```).
+
+    For **CUT&Tag**, the entry would appear as follows:
+    ```
+     "NTR:0000564": {
+        "assay": ['DNA binding'],
+        "category": [],
+        "developmental": [],
+        "name": "Cleavage Under Targets and Tagmentation",
+        "objectives": [],
+        "organs": [],
+        "preferred_name": "CUT&Tag",
+        "slims": [],
+        "synonyms": [],
+        "systems": [],
+        "types": []
+    },
+    ```
+    Follow the instructions in [Updating ontologies] to include your new assay in the encoded application ```ontology.json```.
+
+
+2. Schemas and changelogs:
+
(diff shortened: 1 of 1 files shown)


src/encoded/docs/adding_new_assays.md
```diff
@@ -0,0 +1,283 @@
...
+    * **types** -  business logic for dispatching URLs and producing the correct JSON
+    * **docs** -  SOPs and additional information on schemas, audits, ontologies, and annotations
+
+Steps to add a new assay
+---------------- 
+
+1. Schemas and changelogs:
+
+    Assays are identified by precise ontological terms that we list in a property *assay_term_name*. This is used in many different objects on the portal, and in most cases a mixin property ```assay``` provides this enum list. You will need to add your new assay type to the enum in mixins.json and update the changelogs of the objects that import this property.
+    The objects **Award**, **Pipeline**, and **Software** list ```assay_term_name``` in unique objects without importing the ```assay``` property, so both the schema profile and changelog will need to be updated in these cases.
```
bonitalam: I think you mean "in unique properties" for those objects?

src/encoded/docs/adding_new_assays.md
```diff
@@ -0,0 +1,283 @@
...
+
+        ```
+        testapp.patch_json(base_experiment['@id'], {'assay_term_name': 'CUT&Tag'})
+            res = testapp.get(base_experiment['@id'] + '@@index-data')
+            assert any(error['category'] == 'missing target'
+                for error in collect_audit_errors(res))
+        ```
+        * If you prefer, you can also write a new experiment fixture which you would add to [tests/fixtures/schemas/experiment.py]. For more detailed instructions, please refer to [making_audits.md].
+
+4. Ontology:
```
bonitalam: I think you might have to also add a section that sometimes official terms will require slimming as well, if what you want isn't reflected in the official ontology term yet (e.g. the assay does slim to "DNA binding" but it should). You can see examples here: https://github.com/ENCODE-DCC/encoded/blob/dev/src/encoded/commands/generate_ontology.py#L177-L249 when we slimmed official assays.

src/encoded/docs/adding_new_assays.md
```diff
@@ -0,0 +1,283 @@
...
+
+[samtools_flagstats_quality_metric.md]
+
+[samtools_stats_quality_metric.md]
+
+[star_quality_metric.md]
+
+[transgenic_enhancer_experiment.md]
+
+[trimming_quality_metric.md]
```
bonitalam: Yeah, I'm in favor of removing this.

bonitalam (changes requested):
Just a few comments but overall looks great.

src/encoded/docs/adding_new_assays.md
iyoungworth: Thank you, yes! I corrected it to 'properties'

src/encoded/docs/adding_new_assays.md
iyoungworth: Good point, I added links for where to add a manual slim and/or preferred_name.

I wasn't absolutely explicit about how to search our ontology.json for the OBI term of interest, and just added another pointer to updating_ontologies.md. I can add more detail if it feels lacking

bonitalam: I don't think more detail is necessary.

src/encoded/docs/adding_new_assays.md
```diff
@@ -0,0 +1,156 @@
...
+
+    Check to see if the new assay type has an entry in [OBI].
+    * If there is an entry, you will use the OBI identifier (e.g., OBI:0002039).
+    * If there is no entry, write a New Term Request ticket on the [NTR board], and provide a publication describing the assay type. The ticket number will be used as your assay term identifier (e.g., NTR-564 becomes NTR:0000564).
+
+    Add the assay term name and identifier to [types/assay_data.py].
+
+    If you using an NTR identifier, add an entry to [src/encoded/commands/ntr_terms.py]. You will need to provide:
+    * The name for your assay type in the ```name``` property
+    * A shortened form of the name in ```preferred_name```
```
bonitalam: I think clarification here would be good. You could say something about how the name is usually the non-abbreviated name of the assay. What you have as preferred_name should match what you would like added as the assay_term_name in all of the schemas.

src/encoded/docs/adding_new_assays.md
```diff
@@ -0,0 +1,156 @@
...
+
+        ```
+        testapp.patch_json(base_experiment['@id'], {'assay_term_name': 'CUT&Tag'})
+            res = testapp.get(base_experiment['@id'] + '@@index-data')
+            assert any(error['category'] == 'missing target'
+                for error in collect_audit_errors(res))
+        ```
+        * If you prefer, you can also write a new experiment fixture which you would add to [tests/fixtures/schemas/experiment.py]. For more detailed instructions, please refer to [making_audits.md].
+
+4. Ontology:
```
bonitalam: I actually think probably this section should be added to the beginning, because you really need to make sure you have an ontological term (OBI or NTR) before adding it to the schemas. You also mention in the current first section "Assays are identified by precise ontological terms that we list in a property *assay_term_name*." It's probably important to address how to get the ontological term first.

bonitalam (changes requested):
Sorry had a change of heart about the ordering of the SOP. I think it would make more sense to have ontology section first. I also had one other comment to add more clarification about name and preferred_name for assays.

src/encoded/docs/adding_new_assays.md
iyoungworth: I shifted it to be the first section, clarified name/preferred_name as mentioned in the other comment, and rewrote the section a little to make more sense with the reordering.

iyoungworth: I reordered the sections to place this first, clarified name/preferred_name as requested in another comment, and did some rewording of the ontology section in particular to make the swap make more sense

bonitalam (approved):

PR #3562, opened 2020-09-30
ENCD-5561 Clean up app config per deployment

caseylitton:
I'm still testing this but all the ini files look the same before and after.


README.md
```diff
@@ -28,7 +28,7 @@ For issues see [Snovault OSX App Installation][] first.
 
 1. Build Application
     ```
-    $ make clean && buildout bootstrap && bin/buildout
+    $ make clean && buildout bootstrap && bin/buildout -c development.cfg
```
hitz: what are the other options? candidate, rc, demo, production?  development is for local install I guess.  Can this just be the default?

caseylitton: The base.ini becomes static with this update.   And is not usable for buildout.  So buildout must be overridden with development, demo, rc, test, or production(previously candidate).   Which sound bad, but is really a feature.  The base ini is a template then user need to configure to use.  

If development was the base(default) then we'd need to update the base.ini when updating development config.  Having then all separate allows individual configuration for QA demos, local demos, and the various clusters.

paul-sud (commented):
Changes look good to me.

src/encoded/commands/deploy.py
```diff
@@ -715,6 +718,21 @@ def main():
         )
         for key, val in tags_dict.items():
             print(f"{key:28}:'{val}'")
+        PG_URI='postgresql:///encoded'
+        if run_vars['ENCD_BUILD_TYPE'] == 'app' or run_vars['ENCD_BUILD_TYPE'] == 'app-es':
+            PG_URI="postgresql://run_vars['ENCD_PG_IP']/encoded"
+        bin_build_cmd= "bin/buildout -c {ENCD_ROLE}.cfg buildout:es-ip={ENCD_ES_IP} buildout:es-port={ENCD_ES_PORT} buildout:pg-uri={PG_URI} buildout:fe-ip={ENCD_FE_IP} buildout:remote_indexing={ENCD_REMOTE_INDEXING} buildout:index_procs={ENCD_INDEX_PROCS} buildout:index_chunk_size={ENCD_INDEX_CHUNK_SIZE}".format(
```
paul-sud: Does this obviate the need to put these variables in `env.yml`? https://github.com/ENCODE-DCC/encoded/blob/dev/cloud-config/template-parts/env.yml

Since they are templated into the ini file here it looks like the buildout run on the AMIs could be simplified: https://github.com/ENCODE-DCC/encoded/blob/dev/cloud-config/run-scripts/app-encd.sh#L41

Finally, beyond the scope of the ticket but I feel like some of these variables aren't that dynamic and should probably live statically in the config files.

src/encoded/commands/deploy.py
```diff
@@ -696,10 +696,13 @@ def main():
...
         print(f'run_args dict keys: {run_args.keys()}')
         print(f'\nRun Variables.  In /etc/environment on instance')
+        run_vars = {}
         for line in run_args['user_data'].split('\n'):
             line = line.strip()
             if line[:5] == 'ENCD_':
-                print(line)
+                key_val = line.split('=')
+                run_vars[key_val[0]] = key_val[1]
+                print(f"{key_val[0]}={run_vars[key_val[0]]}")
```
paul-sud: This would be more legible with different unpacking:
```suggestion
                key, value, *_ = line.split('=')
                run_vars[key] = value
                print(f"{key}={value}")
```

demo.cfg
```diff
@@ -1,2 +1,30 @@
 [buildout]
 extends = buildout.cfg
+# Uncomment to just test ini build
+extends = ini-templates/ini_buildout.cfg
+ini_template = ini-templates/production-template.ini
+ini_name = demo.ini
+# app
+accession_factory = encoded.server_defaults.test_accession
```
paul-sud: If variables like these are going to just be static, then wouldn't it be better to put them in the ini-template? It will be easier to move to a different templating engine later if they stay in `ini` files, rather `cfg` which is more buildout specific. Same goes for the other cfg/ini files.

paul-sud (commented):
I know this is still in progress, but I've been working with buildout a little for ENCD-5593 so it's fresh in my mind, just wanted to share my thoughts.

src/encoded/commands/deploy.py
caseylitton: hmm.  I think they buildout vars should end up in the env.  Then we could do something like

[buildout:os.environ.get("SOME_VARIABLE") == "123"]

which I've never tested.  Or just access the env directly from the application.

caseylitton: and yes some of the vars like ports have never changed so I'll make those static.

src/encoded/commands/deploy.py
caseylitton: That's neat unpacking.  If this ends up in final refactor I'll use it.

demo.cfg
caseylitton: I think was resolved in current update.

PR #3563, opened 2020-09-30
ENCD-5515-ucsc-visualization-fcc-files

bonitalam:


src/encoded/vis_defines.py
```diff
@@ -145,8 +145,8 @@
 HIC_FILE_TYPES = ['hic']
 BIGINTERACT_FILE_TYPES = ['bigInteract']
 VISIBLE_FILE_FORMATS = BIGBED_FILE_TYPES + BIGWIG_FILE_TYPES + HIC_FILE_TYPES + BIGINTERACT_FILE_TYPES
-VISIBLE_DATASET_TYPES = ["Experiment", "Annotation"]
-VISIBLE_DATASET_TYPES_LC = ["experiment", "annotation"]
+VISIBLE_DATASET_TYPES = ["Experiment", "Annotation", "FunctionalCharacterizationExperiment"]
+VISIBLE_DATASET_TYPES_LC = ["experiment", "annotation", "functionalcharacterizationexperiment"]
```
yunhailuo: The type here needs to match `item_type` defined in `FunctionalCharacterizationExperiment` class: https://github.com/ENCODE-DCC/encoded/blob/dev/src/encoded/types/functional_characterization_experiment.py#L47

src/encoded/static/vis_defs/STARR_vis_def.json
```diff
@@ -0,0 +1,78 @@
...
+                    "autoScale": "off",
+                    "maxHeightPixels": "64:18:8",
+                    "windowingFunction": "mean+whiskers",
+                    "output_type": [ "control normalized signal" ]
+                }
+            }
+        },
+        "other_groups":  {
+            "dimensions": { "Biosample": "dimY", "Replicates": "dimA" },
+            "dimensionAchecked": "first",
```
yunhailuo, line 54: Just to confirm, we decided to show only the first rep by default? Also which output types are `full`, `dense` and `hide` is decided?

src/encoded/static/vis_defs/PCS_vis_def.json
```diff
@@ -0,0 +1,51 @@
...
+        "Views":  {
+            "tag": "view",
+            "group_order": [ "Signal of unique reads" ],
+            "groups": {
+                "Signal of unique reads": {
+                    "tag": "aSIG",
+                    "visibility": "full",
+                    "type": "bigWig",
+                    "viewLimits": "0:10",
+                    "autoScale": "off",
```
yunhailuo: Not sure how important this is. I did noticed we typically turn off autoScale and set a limit of 0:10. But I also noticed on the demo the signal could be off chart if we do that. Again, more like a confirmation: we indeed want to do that?

yunhailuo (changes requested):

src/encoded/static/vis_defs/PCS_vis_def.json
```diff
@@ -0,0 +1,51 @@
...
+                        "replicate": {
+                            "title_mask": "Replicate_{replicate_number}",
+                            "combined_title": "Pooled"
+                        }
+                    }
+                },
+                "Biosample": {
+                    "tag": "BS",
+                    "groups": { "one": { "title_mask": "{biosample_term_name}"} }
+                }
```
yunhailuo: Isn't that pooled clone sequencing typically don't have real biosample?

bonitalam: You are right they do not have a real biosample in the conventional sense. They aren't cell free though. Are you suggesting we remove this group? (also biosample info in the long and short labels?)

src/encoded/static/vis_defs/PCS_vis_def.json
bonitalam: Good catch. I could just set autoScale to be on and take away the viewLimits to make sure the signal is viewed ok. Probably the signal of unique reads in the STARR file should be changed as well. I guess we usually don't turn on autoScale to avoid the track being distorted by outliers? But it seems like for the FCC signal of unique read/bigWig files this isn't the issue.

src/encoded/vis_defines.py
bonitalam: Thanks, will change that!

src/encoded/static/vis_defs/STARR_vis_def.json
bonitalam: I think most experiments showed only the first rep by default, so I went with that convention. We can show all if that's better? I don't really have a strong preference either way. 

For the bigBed (peaks) file I figured that it would be good to have it viewed as dense - track as a single line. The other bigWig (signal) files - the decision to use full or hide was just determined in that I thought probably most users would be most interested in the control normalized signal which would have full visibility vs. the other signal files (hide).

src/encoded/vis_defines.py
```diff
@@ -146,8 +146,8 @@
 HIC_FILE_TYPES = ['hic']
 BIGINTERACT_FILE_TYPES = ['bigInteract']
 VISIBLE_FILE_FORMATS = BIGBED_FILE_TYPES + BIGWIG_FILE_TYPES + HIC_FILE_TYPES + BIGINTERACT_FILE_TYPES
-VISIBLE_DATASET_TYPES = ["Experiment", "Annotation"]
-VISIBLE_DATASET_TYPES_LC = ["experiment", "annotation"]
+VISIBLE_DATASET_TYPES = ["Experiment", "Annotation", "Functional_Characterization_Experiment"]
```
yunhailuo: I think this should match the class name which doesn't have `_`, though it's hidden now and won't trigger any errors.

yunhailuo (changes requested):

yunhailuo (approved):

PR #3567, opened 2020-10-02
ENCD-5545 adding Gene Silencing Series

gabdank:


src/encoded/tests/fixtures/schemas/gene_silencing_series.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+@pytest.fixture
+def base_gene_silencing_series(testapp, lab, base_experiment_submitted, award):
+    item = {
+        'award': award['uuid'],
+        'lab': lab['uuid'],
+        'related_datasets': [base_experiment_submitted['@id']]
+    }
+    return testapp.post_json('/single_cell_rna_series', item, status=201).json['@graph'][0]
```
yunhailuo: Am I reading this wrong? `single_cell_rna_series`?

yunhailuo (changes requested):

src/encoded/tests/fixtures/schemas/gene_silencing_series.py
gabdank: @yunhailuo thanks for the catch, fixing

yunhailuo (approved):

PR #3568, opened 2020-10-02
ENCD-5513-gm-doc-in-exp-page

khzlin:


src/encoded/static/components/experiment.js
```diff
@@ -917,6 +927,19 @@ const ExperimentComponent = ({ context, auditIndicators, auditDetail }, reactCon
                 title={`Functional characterization experiments with ${context.accession} as an elements cloning`}
             />
 
+            {isFunctionalExperiment && plasmidMapDocs.length > 0 ?
+                <div>
```
emro: In general it's better to avoid empty wrapper `<div>`s. Normally you can use `<React.Fragment>` or in this case I'm not sure the `<div>`s are even necessary. However, it looks like this duplicates the pattern for biosample characterizations code (line 900) so probably this should be left as-is.

src/encoded/static/components/experiment.js
```diff
@@ -428,13 +428,23 @@ const ExperimentComponent = ({ context, auditIndicators, auditDetail }, reactCon
         });
     }
 
-    // Collect biosample docs.
+    // Collect biosample docs
+    // Collect plasmid maps from applied_modifictions documents.
```
emro: Typo: "applied_modifictions" should be "applied_modifications"
Also, I generally try to be consistent on comments about things like whether there are periods at the ends of lines or not, and put periods at the ends of every line/sentence or none of them

src/encoded/static/components/experiment.js
```diff
@@ -428,13 +428,23 @@ const ExperimentComponent = ({ context, auditIndicators, auditDetail }, reactCon
...
         biosampleDocs = biosampleDocs.concat(CollectBiosampleDocs(biosample));
         if (biosample.part_of) {
             biosampleDocs = biosampleDocs.concat(CollectBiosampleDocs(biosample.part_of));
         }
+        if (biosample.applied_modifications && biosample.applied_modifications.length > 0) {
+            biosample.applied_modifications.forEach((am) => {
+                const plasmidMapDoc = (am.documents && am.documents.length > 0) ? am.documents.filter(doc => doc.document_type === 'plasmid map') : [];
+                plasmidMapDocs = plasmidMapDocs.concat(plasmidMapDoc);
+            }
+            );
```
emro: The `}` and `);` should be on the same line. Did you install a linting plug-in? Linting should help you catch these things.

emro (changes requested):
Looks good! I think the two minor changes are probably worth making, the third is more of a note not a requested change.

src/encoded/static/components/experiment.js
khzlin: yes, I have the listing installed and activated. It didn't catch this.

emro (dismissed):
Great!

src/encoded/static/components/experiment.js
```diff
@@ -927,16 +927,7 @@ const ExperimentComponent = ({ context, auditIndicators, auditDetail }, reactCon
...
-                    <Panel>
-                        <PanelHeading>
-                            <h4>Plasmid maps</h4>
-                        </PanelHeading>
-                        <PanelBody addClasses="panel-body-doc-interior">
-                            <DocumentsSubpanels documentSpec={{ documents: plasmidMapDocs }} />
-                        </PanelBody>
-                    </Panel>
-                </div>
+                <DocumentsPanelReq documents={plasmidMapDocs} title="Plasmid Maps" />
```
emro: Is there a reason you changed the title from "Plasmid maps" to "Plasmid Maps"? I don't know if we have a definite convention about this but I think in general we have the first word capitalized and second words uncapitalized. Like we have "Isogenic replicates" instead of "Isogenic Replicates".

src/encoded/static/components/doc.js
```diff
@@ -86,17 +86,19 @@ DocumentsPanel.defaultProps = {
 const DocumentsPanelRenderer = (props) => {
     const documents = props.documentSearch['@graph'];
     if (documents && documents.length > 0) {
-        return <DocumentsPanel documentSpecs={[{ documents }]} />;
+        return <DocumentsPanel documentSpecs={[{ documents }]} title={props.title} />;
```
emro, line 89: This is a very good change

src/encoded/static/components/doc.js
```diff
@@ -86,17 +86,19 @@ DocumentsPanel.defaultProps = {
...
     }
     return null;
 };
 
 DocumentsPanelRenderer.propTypes = {
     documentSearch: PropTypes.object, // Search result object; this uses its @graph to get the documents,
+    title: PropTypes.string, // Title of document panel, optional
```
emro: If "title" weren't optional, this line would have a `.isRequired` after it

src/encoded/static/components/experiment.js
emro: Using this component instead of `<DocumentsSubpanels>` is much better.

emro (commented):
I think the title should probably be changed.

zoldello (approved):
Code looks good. The changes were very minor.

PR #3570, opened 2020-10-05
SNO-124-make-elasticsearch-shards-configurable

Bek:


src/encoded/commands/deploy.py
```diff
@@ -1009,6 +1013,8 @@ def hostname(value):
     parser.add_argument('--cluster-size', type=int, default=5, help="Elasticsearch cluster size")
     parser.add_argument('--es-ip', default='localhost', help="ES Master ip address")
     parser.add_argument('--es-port', default='9201', help="ES Master ip port")
+    parser.add_argument('--primary-shards', default=3, type=int, help="ES number of primary shards")
+    parser.add_argument('--replicate-shards', default=1, type=int, help="ES number of replicate shards")
```
keenangraham, line 1020: Changing defaults from 5/2?

src/encoded/commands/deploy.py
```diff
@@ -1012,6 +1016,8 @@ def hostname(value):
     parser.add_argument('--cluster-size', type=int, default=5, help="Elasticsearch cluster size")
     parser.add_argument('--es-ip', default='localhost', help="ES Master ip address")
     parser.add_argument('--es-port', default='9201', help="ES Master ip port")
+    parser.add_argument('--primary-shards', default=3, type=int, help="ES number of primary shards")
```
keenangraham, line 1019: I guess I'm still concerned that we're not going to have any replicates in production anymore?

keenangraham (changes requested):

src/encoded/commands/deploy.py
Bek: yes, 5/2 was pushing us beyond the 1000 shard (primary and replicate) limit per cluster.

src/encoded/commands/deploy.py
Bek: We have one replicate for each primary shard. Primary and replicate will be hosted in different nodes, and does provide some redundancy in case one goes down.

PR #3574, opened 2020-10-06
ENCD-5534 Allow annotations and FCCs in carts

forresttanaka:
I moved `rawdataFiles` and `selectedVisualizableFiles` out of state and now calculate them each time something they depend on changes, to remove potential extra re-renderings from state changes. Those variables’ contents directly rely on the contents of other states.

I used to keep a small subset of properties of the file objects involved with the datasets, and then before having Valis render a page of tracks, I requested real file objects for that page of files. I no longer do this, and instead keep a larger subset of properties of the file objects — enough to give to `<GenomeBrowser>` and Valis directly. As a side effect, we now have a mechanism to sort the files in Valis in the cart by presorting them before giving them to `<GenomeBrowser>`. But that also means cart page needs to own the sorting controls, or the sorting controls have to call the parent when sorting options change.

### DIFF
--- a/src/encoded/static/components/cart/add_multiple.js
+++ b/src/encoded/static/components/cart/add_multiple.js
@@ -63,8 +63,8 @@ class CartAddAllSearchComponent extends React.Component {
                         // Not logged in, so test whether the merged new and existing carts would have
                         // more elements than allowed in a logged-out cart. Display an error modal if
                         // that happens.
-                        const margedCarts = mergeCarts(this.props.elements, elementsForCart);
-                        if (margedCarts.length > CART_MAXIMUM_ELEMENTS_LOGGEDOUT) {
+                        const mergedCarts = mergeCarts(this.props.elements, elementsForCart);
+                        if (mergedCarts.length > CART_MAXIMUM_ELEMENTS_LOGGEDOUT) {
                             this.setState({ overMaximumError: true });
                             return;
                         }
--- a/src/encoded/static/components/cart/add_multiple.js
+++ b/src/encoded/static/components/cart/add_multiple.js
@@ -5,7 +5,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
-import * as encoding from '../../libs/query_encoding';
+import url from 'url';
+import QueryString from '../../libs/query_string';
 import { addMultipleToCartAndSave, cartOperationInProgress } from './actions';
 import { atIdToType } from '../globals';
 import { requestSearch } from '../objectutils';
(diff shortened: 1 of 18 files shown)


src/encoded/static/libs/ui/button.js
```diff
@@ -287,7 +291,7 @@ export const Selected = ({ labels, execute, id, triggerVoice, css, inline, child
     return (
         <div className={`dropdown-button${css ? ` ${css}` : ''}`} style={inline ? { display: 'inline-flex' } : null}>
             <div className="dropdown-button__composite" onMouseEnter={actions.handleMouseEnter} onMouseLeave={actions.houseMouseLeave}>
-                <button className="dropdown-button__composite-execute" onClick={handleExecute} onKeyUp={actions.handleKey}>
+                <button className="dropdown-button__composite-execute" onClick={handleExecute} onKeyUp={actions.handleKey} disabled={disabled}>
```
emro, line 294: Can't you just have `disabled` here?

src/encoded/static/components/cart/add_multiple.js
```diff
@@ -36,56 +37,40 @@ class CartAddAllSearchComponent extends React.Component {
...
+                if (elementsForCart.length > 0) {
+                    // We should always have elements qualified to exist in a cart because we
+                    // wouldn't have shown the "Add all items to cart" button if we didn't know we
+                    // had qualfied elements in the search results, but just in case.
+                    if (!loggedIn) {
+                        // Not logged in, so test whether the merged new and existing carts would have
+                        // more elements than allowed in a logged-out cart. Display an error modal if
+                        // that happens.
+                        const margedCarts = mergeCarts(this.props.elements, elementsForCart);
+                        if (margedCarts.length > CART_MAXIMUM_ELEMENTS_LOGGEDOUT) {
```
emro: I feel like you meant to have this called `mergedCarts`.

src/encoded/static/components/genome_browser.js
```diff
@@ -549,14 +560,15 @@ class GenomeBrowser extends React.Component {
             let labelLength = 0;
             const defaultHeight = 34;
             const extraLineHeight = 12;
-            const maxCharPerLine = 30;
```
emro, line 552: There are still some cases with overflow which should be fixed:

Screen Shot 2020-10-06 at 1 37 53 PM

src/encoded/static/scss/encoded/modules/_cart.scss
```diff
@@ -1156,3 +1189,67 @@ $file-type-colors: (
...
+
+        button {
+            width: 100%;
+        }
+    }
+}
+
+.cart-dataset-selector {
+    width: 100%;
+    margin-bottom: 5px;
```
emro, line 1220: I would add a little padding here on the right and left.

src/encoded/static/components/cart/cart.js
```diff
@@ -82,9 +97,22 @@ const requestedFacetFields = displayedFacetFields.concat([
...
 
 
+/**
+ * Maps long annotation_type values to shorter versions for Valis track labels. Any not included
+ * remain unchanged.
+ */
+const annotationTypeMap = {
+    'candidate Cis-Regulatory Elements': 'cCRE',
+    'representative DNase hypersensitivity sites (rDHSs)': 'rDHS',
+};
```
emro: Maybe we will want to change the labels on the annotation pages to match those in the cart and then this could go in globals.js or objectutils.js.

src/encoded/static/components/cart/cart.js
```diff
@@ -156,27 +184,27 @@ CartSearchResults.defaultProps = {
...
-    }, [files, assembly, pageNumber]);
+    });
+
     const sortParam = ['Assay term name', 'Biosample term name', 'Output type'];
     return <GenomeBrowser files={pageFiles} label={'cart'} assembly={assembly} expanded sortParam={sortParam} />;
 };
 
 CartBrowser.propTypes = {
-    /** File @ids of all visualizable tracks, not just on the displayed page */
+    /** Files of all visualizable tracks, not just on the displayed page */
```
emro, line 197: Did we want to add the sort buttons now, since we're collecting all the visualizable tracks?
Edit: I see you addressed this above. This is probably a new ticket / worth a discussion.

emro (commented):
Most of these are suggestions, not necessary changes. I think maybe the overflow is the only one that is relatively important. It looks great overall!

src/encoded/static/libs/ui/button.js
forresttanaka: That would leave the button permanently disabled, or am I misunderstanding?

src/encoded/static/components/genome_browser.js
forresttanaka: How should this be fixed? Make `maxCharPerLIne` even smaller?

forresttanaka: How did you get the track labels to not show the expander icons?

forresttanaka: OK, I see now where that depends on a lot of things. I updated the height.

src/encoded/static/scss/encoded/modules/_cart.scss
forresttanaka: Added 5px left and right.

src/encoded/static/components/cart/cart.js
forresttanaka: I moved this object to genome_browser.js, since it seems unlikely this module would import from another module that imports from it (not that this would cause problems, as we found several months ago; that still makes me uncomfortable).

src/encoded/static/components/cart/add_multiple.js
forresttanaka: Changed.

emro (dismissed):
Looks good!

src/encoded/static/libs/ui/button.js
emro: Oh yes, my mistake.

src/encoded/static/components/genome_browser.js
```diff
@@ -497,11 +507,13 @@ class GenomeBrowser extends React.Component {
...
-                .sortBy(obj => sortLookUp(obj, param));
-        });
-        files = files.value();
+        if (this.props.displaySort) {
+            orderedSortParam.forEach((param) => {
+                files = _.chain(files)
+                    .sortBy(obj => sortLookUp(obj, param));
+            });
+            files = files.value();
+        }
```
emro, line 516: I think this is ok. I added in displaySort so that we could turn off the sort buttons and leave the cart sorting. Currently the carts do sort results, just on a page-by-page basis and we just don't bring the users' attention to it. This all needs to be revisited when sorting buttons are added to the cart so it seems ok to get rid of page-by-page sorting for now. The comment explanation of displaySort is now misleading, though, because it's not just turning off the sort buttons, it's now turning off the sort entirely.

Screen Shot 2020-10-08 at 8 58 10 AM

emro (commented):
If we are ok with turning off cart page-by-page sorting (and I think that's fine because cart sort buttons are a near-future plan), this is all good.

src/encoded/static/components/genome_browser.js
forresttanaka: OK, I see.

emro (approved):

PR #3577, opened 2020-10-07
ENCD-5562 add screen links

emro:
Adding links to SCREEN and Motifs to the experiment pages.
Local data was updated for debugging purposes.

### DIFF
--- a/src/encoded/static/components/search.js
+++ b/src/encoded/static/components/search.js
@@ -349,8 +349,8 @@ const ExperimentComponent = (props, reactContext) => {
     }
 
     // Get SCREEN link and FactorBook link if they exist
-    const screenLink = (result.dbxrefs && result.dbxrefs.filter(dbxref => (dbxref.indexOf('SCREEN') > -1)).length > 0) ? result.dbxrefs.filter(dbxref => (dbxref.indexOf('SCREEN') > -1))[0] : null;
-    const motifsLink = (result.dbxrefs && result.dbxrefs.filter(dbxref => (dbxref.indexOf('FactorBook') > -1)).length > 0) ? result.dbxrefs.filter(dbxref => (dbxref.indexOf('FactorBook') > -1))[0] : null;
+    const screenLink = (result.dbxrefs && result.dbxrefs.some(dbxref => (dbxref.indexOf('SCREEN') > -1))) ? result.dbxrefs.filter(dbxref => (dbxref.indexOf('SCREEN') > -1))[0] : null;
+    const motifsLink = (result.dbxrefs && result.dbxrefs.some(dbxref => (dbxref.indexOf('FactorBook') > -1))) ? result.dbxrefs.filter(dbxref => (dbxref.indexOf('FactorBook') > -1))[0] : null;
     let screenSearch = '';
     let screenAssembly = '';
     if (screenLink) {
--- a/src/encoded/static/components/search.js
+++ b/src/encoded/static/components/search.js
@@ -235,16 +235,16 @@ class BiosampleComponent extends React.Component {
                             {`${separator}${lifeStage}${ageDisplay})`}
                         </a>
                         <div className="result-item__data-row">
-                            <div><strong>Type: </strong>{result.biosample_ontology.classification}</div>
-                            {result.summary ? <div><strong>Summary: </strong>{BiosampleSummaryString(result)}</div> : null}
-                            {rnais.length > 0 ? <div><strong>RNAi targets: </strong>{rnais.join(', ')}</div> : null}
-                            {constructs.length > 0 ? <div><strong>Constructs: </strong>{constructs.join(', ')}</div> : null}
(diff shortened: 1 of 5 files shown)


src/encoded/static/components/search.js
```diff
@@ -345,6 +345,20 @@ const ExperimentComponent = (props, reactContext) => {
         );
     }
 
+    // Get SCREEN link and FactorBook link if they exist
+    const screenLink = result.dbxrefs.filter(dbxref => (dbxref.indexOf('SCREEN') > -1))[0];
```
zoldello: can dbxrefs ever be falsey? If yes, this will throw an exception.

src/encoded/static/components/search.js
```diff
@@ -407,6 +421,12 @@ const ExperimentComponent = (props, reactContext) => {
                                         )}
                                     </div>
                                 )}
+                                {screenLink ?
+                                    <div><strong>candidate Cis-Regulatory Elements (cCREs): </strong><a href={`https://screen.encodeproject.org/search?q=${screenSearch}&assembly=${screenAssembly}`}>SCREEN</a></div>
```
zoldello: <strong> or have a css class to bolden the text?

src/encoded/tests/data/inserts/experiment.json
```diff
@@ -52,7 +52,8 @@
         "dbxrefs": [
             "UCSC-ENCODE-mm9:wgEncodeEM002001",
             "GEO:GSM798326",
-            "GEO:GSE32218"
+            "GEO:GSE32218",
```
zoldello, line 55: Do you to also update the unit and BDD experiment search pages?

zoldello (dismissed):
Code looks good. My comments may not necessarily require changes.

src/encoded/static/components/search.js
emro: Yes, this would be a lot better. I'll change that on all of these.

src/encoded/tests/data/inserts/experiment.json
emro: I updated the batch download test. All the other tests passed. I think it's ok, then?

src/encoded/static/components/search.js
emro: I don't think so, but I'll account for that just in case.

src/encoded/static/components/search.js
```diff
@@ -235,16 +235,16 @@ class BiosampleComponent extends React.Component {
                             {`${separator}${lifeStage}${ageDisplay})`}
                         </a>
                         <div className="result-item__data-row">
```
zoldello, line 237: Why are you making these bold? I did not see the request in the ticket.

src/encoded/static/scss/encoded/modules/_search.scss
```diff
@@ -62,6 +62,10 @@ $form-bg: #f3f3f3;
         position: relative;
     }
 
+    @at-root #{&}__property-title {
+        font-weight: 700;
```
zoldello, line 66: Not for now, but in the future, you can use "bold" as bold also has a a value of 700 (https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight)

src/encoded/static/components/search.js
```diff
@@ -346,8 +346,8 @@ const ExperimentComponent = (props, reactContext) => {
     }
 
     // Get SCREEN link and FactorBook link if they exist
```
zoldello, line 351: Are you trying to do this:
const screenLink  = result.dbxrefs  && result.dbxrefs.find( dbxref => (dbxref.indexOf('SCREEN') > -1)  )

If not, you can use "some" to check for any dbxref and find to get the first instance, like:
const screenLink = (result.dbxrefs && result.dbxrefs.**some**(dbxref => (dbxref.indexOf('SCREEN') > -1))) ? result.dbxrefs.**find**(dbxref => (dbxref.indexOf('SCREEN')))[0] : null;

I did not test it out so it may be off by a parenthesis or two.

zoldello (dismissed):
My comments are suggestions. But the code is fine if they are deemed not worth consideration.

src/encoded/static/components/search.js
emro: They were already bold, I'm just making it consistent that we're using class names to make them bold instead of `<strong>`. You suggested doing that in one place in your first code review and I wanted to do it everywhere if we were doing it one place.

zoldello: Thanks for catching me on that :)

src/encoded/static/components/search.js
emro: I changed this. I always forget about `.some`, somehow.

zoldello (dismissed):
Look good. Did you do an interactive rebase to change the commit history?

zoldello (approved):
Code looks okay.

PR #3583, opened 2020-10-12
VALIS-26-enable-plotting-of-ENCD-bigbed-types

zoldello:


src/encoded/static/components/objectutils.js
```diff
@@ -884,10 +884,7 @@ export function computeAssemblyAnnotationValue(assembly, annotation) {
  */
 export const isFileVisualizable = file => (
     (file.file_format === 'bigWig' || file.file_format === 'bigBed')
-        && (file.file_format_type !== 'bedMethyl')
         && (file.file_format_type !== 'bedLogR')
```
zoldello, line 835: @emro file types like bedLogR were not enabled as they were deemed either obsolete or not use enough to be worth enabling. But in case any one of them need to be enabled, you can enable the genomic-reader (npm package that reads the bigbed file) by following this instruction I wrote- https://github.com/ENCODE-DCC/bigwig-reader#creating-custom-bigbed-formats

src/encoded/static/components/genome_browser.js
```diff
@@ -627,6 +627,7 @@ class GenomeBrowser extends React.Component {
             trackObj.displayLabels = false;
             trackObj.heightPx = labelLength > 0 ? (defaultHeight + (extraLineHeight * labelLength)) : defaultHeight;
             trackObj.expandedHeightPx = 140;
+            trackObj.fileFormatType = file.file_format_type;
```
zoldello, line 681: @emro Before, bigbed files was all understood as following the column format of UCSC. With this update, there are 6 recognizable Bigbed files-    BigBed (as defined by UCSC) , Broad Peak, Narrow Peak, Methyl, Tss Peak, Idr Peak. If all bigbed variants not part of this list default to bigbed as defined by UCSC

src/encoded/static/components/objectutils.js
emro: This is terrific! The documentation will be very helpful.

emro (changes requested):
The Valis change looks great! I'm not sure we want to update all the packages as part of this ticket, though.

emro (dismissed):
Package-lock.json changes seem to all be required by library update. Looks good!

PR #3584, opened 2020-10-13
ENCD-5559 FCC processed data facet

forresttanaka:
The facet mechanism now has _two_ registries: the existing `FacetRegistry` and a new `SpecialFacetRegistry` that you use for facets not based on the contents of the `facet` property of search results, at least not directly. Special facets always appear above the normal facets, though I could see an option to put certain special facets at the end in the future.

### DIFF
--- a/src/encoded/static/components/facets/audit_processed_data.js
+++ b/src/encoded/static/components/facets/audit_processed_data.js
@@ -1,0 +1,106 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import url from 'url';
+import QueryString from '../../libs/query_string';
+import BooleanSwitch from '../../libs/ui/boolean-switch';
+import { SpecialFacetRegistry } from './registry';
+
+
+/**
+ * Special facet to display a switch to select between having the search result query string
+ * contain: "audit.WARNING.category!=lacking processed data" and without that term at all, with the
+ * boolean switch component showing the "on" state with the former and the "off" state with the
+ * latter.
+ */
+const AuditProcessedData = ({ facet, results, mode, pathname, queryString, isExpanded, handleExpanderClick, handleKeyDown, isExpandable }, reactContext) => {
+    /** Keeps track of the switch state for `handleSwitch` */
+    const switchState = React.useRef(false);
+
+    /**
+     * Called when the user clicks the term to flip the switch and modify the query string
+     * accordingly.
+     */
+    const handleSwitch = () => {
+        const parsedUrl = url.parse(results['@id']);
+        const query = new QueryString(parsedUrl.query);
+        if (switchState.current) {
+            query.deleteKeyValue('audit.WARNING.category');
+        } else {
+            query.replaceKeyValue('audit.WARNING.category', 'lacking processed data', true);
+        }
+        const href = `?${query.format()}`;
+        reactContext.navigate(href, { noscroll: true });
+    };
+
+    // Make sure a single type=FunctionalCharacterization experiment exists, or display no facet.
+    const fccTypeFilters = results.filters.filter(filter => filter.field === 'type' && filter.term === 'FunctionalCharacterizationExperiment');
+    if (fccTypeFilters.length === 1) {
+        // Determine the state of the switch from the query string, and save it so that
+        // `handleSwitch` doesn't have to do this same calculation later.
+        const parsedUrl = url.parse(results['@id']);
+        const query = new QueryString(parsedUrl.query);
+        const auditWarningValues = query.getKeyValues('audit.WARNING.category', true);
+        switchState.current = auditWarningValues.length === 1 && auditWarningValues[0] === 'lacking processed data';
+
+        // Retrieve reference to the registered facet title component for this facet.
+        const TitleComponent = SpecialFacetRegistry.Title.lookup();
+
+        return (
+            <div className="facet facet--audit-warning">
+                <div
+                    className="facet__expander--header"
+                    tabIndex="0"
+                    role="button"
+                    aria-label="audit.WARNING.category"
+                    aria-pressed={isExpanded}
+                    onClick={e => handleExpanderClick(e, isExpanded, facet.field)}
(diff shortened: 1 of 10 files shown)


src/encoded/static/libs/ui/boolean-switch.js
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * Renders a boolean switch in the style of iOS. This keeps no internal state -- the parent
+ * component keeps this state and passes it down. The "frame" of the switch represents the space in
+ * which the actuator slides. The actuator represents the "physical" switch you would slide in real
+ * life.
+ */
+const DEFAULT_SWITCH_HEIGHT = 22;
+const DEFAULT_SWITCH_WIDTH = DEFAULT_SWITCH_HEIGHT * 1.6;
+const BooleanSwitch = ({ id, state, title, triggerHandler, options: { width, height, cssTitle, cssFrame, cssActuator } }) => {
```
emro, line 13: Unless I'm missing something, this component is not accessible. I tried adding "tabindex" properties and that wasn't it so it may be a little more complicated than that.

src/encoded/static/libs/ui/boolean-switch.js
```diff
@@ -0,0 +1,88 @@
...
+
+    // Calculate the inline styles for the switch and actuator.
+    const switchWidth = width || DEFAULT_SWITCH_WIDTH;
+    const switchHeight = height || DEFAULT_SWITCH_HEIGHT;
+    const triggerSize = switchHeight - 4;
+    const switchStyles = {
+        width: switchWidth,
+        height: switchHeight,
+        borderRadius: switchHeight / 2,
+        backgroundColor: state ? '#65afeb' : '#e9e9eb',
```
emro: We already have so many blues floating about - the darker gray-blue for buttons, the link blue, etc. What do you think about using the blue that is the background for the selected facet terms?

Screen Shot 2020-10-14 at 1 44 31 PM

src/encoded/static/components/facets/audit_processed_data.js
```diff
@@ -0,0 +1,106 @@
...
+import BooleanSwitch from '../../libs/ui/boolean-switch';
+import { SpecialFacetRegistry } from './registry';
+
+
+/**
+ * Special facet to display a switch to select between having the search result query string
+ * contain: "audit.WARNING.category!=lacking processed data" and without that term at all, with the
+ * boolean switch component showing the "on" state with the former and the "off" state with the
+ * latter.
+ */
```
emro, line 10: I gather that it was discussed how there's no control for getting to `&audit.WARNING.category=lacking+processed+data` and that's ok. Probably that would only be useful to wranglers anyway? It looks like `=lacking+processed+data` has the same audit facet display as no audit warning term and that seems fine since you can't get there without url modification and maybe we don't want to think very hard about this, but maybe it could be colored red or something if that is the setting?

src/encoded/static/components/facets/audit_processed_data.js
```diff
@@ -0,0 +1,106 @@
...
+        } else {
+            query.replaceKeyValue('audit.WARNING.category', 'lacking processed data', true);
+        }
+        const href = `?${query.format()}`;
+        reactContext.navigate(href, { noscroll: true });
+    };
+
+    // Make sure a single type=FunctionalCharacterization experiment exists, or display no facet.
+    const fccTypeFilters = results.filters.filter(filter => filter.field === 'type' && filter.term === 'FunctionalCharacterizationExperiment');
+    if (fccTypeFilters.length === 1) {
```
emro, line 31: I feel like this could be >0 instead of ===1 but I don't think it really matters since there should only be one.

emro (commented):
The code looks great, just have a few comments. The most important question is the one about accessibility.

src/encoded/static/libs/ui/boolean-switch.js
forresttanaka: What do you see when you try to access it with tab? I go from focusing on the “Processed data” collapsing title to focusing on the entire line for “with processed data,” and VoiceOver says, “with processed data checked/unchecked checkbox” (depending on its current state). Hitting space bar toggles the switch which navigates to the new search page. Do you see something different? I see this in Safari, Chrome, Edge, and Firefox.

src/encoded/static/components/facets/audit_processed_data.js
forresttanaka: @hitz Do you have an opinion on this? This seemed OK to me since you have the full control at the bottom of the facets, and I’m not sure if a different color would mean anything to the user. So right now if you have no “audit.WARNING.category” or “audit.WARNING.category=lacking+processed+data” the switch appears off. If you have “ audit.WARNING.category!=lacking+processed+data” the switch appears on.

src/encoded/static/components/facets/audit_processed_data.js
forresttanaka: You can have more than one if you enter multiple type= in the query string, but as multiple “type=FunctionalCharacterizationExperiment” possibly combined with the != version means the user did something weird, I didn’t want the facet to appear unless the query string gets entered correctly, or comes from the menu item.

src/encoded/static/libs/ui/boolean-switch.js
emro: Oh, I see! I was trying to toggle it using the return key. It does work! Thank you!

emro (dismissed):
I think it's good. If you make any changes for the color or the controls, I can re-review.

forresttanaka:
Patched demo at https://encd-5559-fcc-facet-fytanaka.demo.encodedcc.org/ with the updated color.

emro (dismissed):

forresttanaka:
Changed the boolean switch to a dropdown. I decided to keep the boolean_switch.js file since it has already seen code review _and_ testing, and we’ll probably need it someday.

emro (dismissed):
Looks great! I think the new UI is clearer and the code is simpler too.

hitz (approved):
rebase approved

PR #3585, opened 2020-10-13
ENCD-5576 Matrix disclosure arrow alignment

forresttanaka:
A button font-size change left these arrows unaligned. This change prevents this element from getting affected by font-size changes in buttons.

### DIFF
--- a/src/encoded/static/scss/encoded/modules/_matrix.scss
+++ b/src/encoded/static/scss/encoded/modules/_matrix.scss
@@ -621,6 +621,7 @@ $matrix-border-color: #f0f0f0;
             padding: 0;
             flex: 0 0 auto;
             height: 10px;
+            font-size: 0;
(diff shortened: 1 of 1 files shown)


src/encoded/static/scss/encoded/modules/_matrix.scss
```diff
@@ -621,6 +621,7 @@ $matrix-border-color: #f0f0f0;
             padding: 0;
             flex: 0 0 auto;
             height: 10px;
+            font-size: 0;
```
zoldello, line 624: This works. But it may be confusing as you are making  font-size = 0 and that moves the arrow. I would recommending modifying:

`matrix__category-expander svg`

and using something like 

`display:flex`

zoldello (approved):
The comment I made is optional.

PR #3586, opened 2020-10-13
ENCD-4700-hide-checkbx-in-proceeded-data

zoldello:


forresttanaka (dismissed):
A Javascript warning can appear when selecting the Juicebox browser, though I think real data makes this situation very unlikely or impossible, and this bug only tangentially gets triggered by this ticket. If you look at http://localhost:6543/experiments/ENCSR000AJK/ and choose Ensembl and then choose Juicebox, the Javascript warning appears. Unless you want to fix this (change `onClick` to `onChange`), I don’t think this bug has much importance.

src/encoded/static/components/filegallery.js
```diff
@@ -331,12 +330,14 @@ FileTable.contextTypes = {
...
         title: 'Visualize',
+        hide: (list, columns, meta) => meta && meta.browserOptions && (meta.browserOptions.browserFileSelectHandler === null || meta.browserOptions.currentBrowser !== 'hic'),
         display: (item, meta) => {
             const selectedFile = meta.browserOptions.selectedBrowserFiles.find(file => file['@id'] === item['@id']);
             const selectable = visFileSelectable(item, meta.browserOptions.selectedBrowserFiles, meta.browserOptions.currentBrowser);
             return (
                 <div className="file-table-visualizer">
-                    <input name={item['@id']} type="checkbox" checked={!!selectedFile} disabled={!selectable && !selectedFile} onClick={meta.browserOptions.browserFileSelectHandler} />
+                    // https://stackoverflow.com/a/39020641/178550
+                    <input name={item['@id']} type="checkbox" checked={!!selectedFile} disabled={!selectable && !selectedFile} onClick={meta.browserOptions.browserFileSelectHandler} onChange={() => {}} />
```
forresttanaka: Should we instead change the existing onClick to an onChange? That should work I would think, and then we wouldn’t need the empty function.

zoldello: @forresttanaka Updated. You are right; my way could be confusing in the future and should be addressed even if unrelated to the ticket. The change looked fine to me on the UI.

forresttanaka (approved):

PR #3588, opened 2020-10-13
ENCD-5566 fix cropped tooltips

emro:
To prevent tooltips from being cropped when their containers do not display overflow, we are now attaching tooltips to the document body.

### DIFF
--- a/src/encoded/static/components/objectutils.js
+++ b/src/encoded/static/components/objectutils.js
@@ -6,6 +6,7 @@ import * as encoding from '../libs/query_encoding';
 import { CartToggle, cartGetAllowedTypes } from './cart';
 import * as globals from './globals';
 import { BrowserFeat } from './browserfeat';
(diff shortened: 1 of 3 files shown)


forresttanaka (dismissed):
That looks really good. Glad this didn’t involve _too_ much rewriting.

forresttanaka (approved):
This looks pretty good. I know of one future enhancement to tooltips for when a timer keeps one tooltip open and then you enter the trigger for another tooltip. That should immediately cancel the first tooltip. I have a [personal repo](https://github.com/forresttanaka/ctxmod) we can include that could make this easier, by having tooltips populate a global context so that different tooltips can know about and cancel each other. That repo doesn’t have a direct connection with tooltips — it lets us add multiple React context providers without having to nest each one. I’ve been hesistant to add context providers because of the nesting issue, but that repo lets us use one provider for lots of independent uses.

PR #3593, opened 2020-10-20
ENCD-5345-pipeline-prop-on-reference-obj

jnau:
I didn't add inserts due to how the inserts are loaded in loadxl.py. Since reference objects are loaded before pipeline objects, this causes an error when specifying a pipeline in the related pipelines property. I can't move reference to be after pipeline in loadxl.py because the reverse link is already in place from ENCD-5298, which will invalidate the pipeline object that has anything in the reference_fileset property.

### DIFF
--- a/src/encoded/loadxl.py
+++ b/src/encoded/loadxl.py
@@ -641,7 +641,7 @@ PHASE1_PIPELINES = {
         remove_keys('related_files'),
     ],
     'reference': [
-        remove_keys('related_files', 'software_used'),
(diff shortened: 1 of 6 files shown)


gabdank (changes requested):
please add related_pipelines to one of the reference datasets in the inserts. You can modify the loadxl.py in a way that will allow you to load references, despite of the pipeline not being yet specified. Will be happy to explain how to do it.

gabdank (approved):

PR #3594, opened 2020-10-20
ENCD-5585-fix-views-with-no-at-type

forresttanaka:
The branch ENCD-5585-view-type-error copies this branch but reintroduces the bug in the front end to demonstrate that the new BDD test catches this bug.


src/encoded/tests/test_no_attype.py
```diff
@@ -0,0 +1,18 @@
...
+]
+
+
+'''
+    Used to test views with no @type property.
+'''
+def test_no_attype(testapp):
+    res = testapp.get('/searchv2_raw/?type=Experiment', status=200)
+    assert res.content_type == 'application/json'
+    assert res.text
```
keenangraham: I wonder if this is testing anything more than e.g. https://github.com/ENCODE-DCC/encoded/blob/1a093913abc99ea634657a005f4c5172a833bc9a/src/encoded/tests/test_search_views.py#L173 which implicitly assumes 200 response. Does `res.text` fail without JS fix?

keenangraham (dismissed):
Looks fine, just a couple questions.

src/encoded/tests/test_no_attype.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+pytestmark = [
+    pytest.mark.indexing,
+    pytest.mark.usefixtures('index_workbook'),
+]
+
+
+'''
+    Used to test views with no @type property.
```
keenangraham: Should probably be comment instead of docstring, or put docstring into function. https://www.python.org/dev/peps/pep-0257/

src/encoded/tests/test_no_attype.py
forresttanaka: Oh, I had meant to delete this file. I looked at this test first before realizing we basically already did the same thing above, so I went to a purely BDD test,

keenangraham (approved):

PR #3596, opened 2020-10-20
ENCD-5571-display-location-Gene

khzlin:


src/encoded/static/components/gene.js
```diff
@@ -119,6 +119,19 @@ class Gene extends React.Component {
                           </div>
                         : null}
 
+                        {context.locations && context.locations.length > 0 ?
+                            <div data-test="locations">
+                                <dt>Gene locations</dt>
+                                <dd>
+                                    <ul>
+                                        {context.locations.map((location, l) => (
```
zoldello, line 127: I tend not to use 'l' as an index because it can be misinterpreted as '1'. But the code is fine as-is.

src/encoded/static/components/gene.js
```diff
@@ -119,6 +119,19 @@ class Gene extends React.Component {
...
+                                            <li key={l} className="multi-comma">{`${location.assembly} ${location.chromosome}:${location.start}-${location.end}`}</li>
```
zoldello, line 128: Is the lack of space between ':' and '-' intended?

zoldello (approved):
Code looks good. My comments are observatory and option.

src/encoded/static/components/gene.js
khzlin: yeah, I think so. There's a similar one in GeneticModifications for modified_site_by_coordinates so I kept the same format. Here's an example: https://www.encodeproject.org/genetic-modifications/ENCGM592MOY/

PR #3597, opened 2020-10-21
ENCD-5579 Update Dataset Search button types

forresttanaka:
I moved the collection of the dataset types actually included in the cart to the parent component, because now two child components need that information.

### DIFF
--- a/src/encoded/static/components/cart/cart.js
+++ b/src/encoded/static/components/cart/cart.js
@@ -849,19 +849,21 @@ FileFacets.defaultProps = {
  * FCCs for now.
  */
 const MAX_SEARCHABLE_DATASETS = 125; // Maximum number of datasets for search-results links.
-const CartDatasetSearch = ({ elements }, reactContext) => {
+const CartDatasetSearch = ({ elements, usedDatasetTypes }, reactContext) => {
     const [isWarningVisible, setIsWarningVisible] = React.useState(false);
 
-    // * For now only include experiments. Might need to expand this in the future.
-    const experimentElements = elements.filter(element => element.includes('/experiments/'));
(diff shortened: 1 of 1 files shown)


src/encoded/static/components/cart/cart.js
```diff
@@ -943,9 +941,16 @@ const CartTools = ({ elements, selectedTerms, selectedDatasetType, typeChangeHan
     // datasets."
     const disabledMessage = selectedDatasetType === DEFAULT_DATASET_TYPE ? 'Select dataset type to download' : '';
     const selectedType = datasetTypes[selectedDatasetType].type;
+
+    // Make a list of all the dataset types currently in the cart.
+    const usedDatasetTypes = [Object.keys(defaultDatasetType)[0]].concat(elements.reduce((types, elementAtId) => {
+        const type = atIdToType(elementAtId);
+        return types.includes(type) ? types : types.concat(type);
+    }, []));
```
emro, line 949: I don't quite follow the distinction between "usedDatasetTypes" and "actualDatasetTypes" - why can't we compute "actualDatasetTypes" here and use it for both Cart components? CartTypeSelector needs the "all" type?

emro (approved):
The code all looks good! I would move to QA Ready but you made a note on the ticket about trying to get the Indexer test to pass. Should I hold off then?

forresttanaka:
Seems like we’re allowing the indexer test to fail for now.

src/encoded/static/components/cart/cart.js
forresttanaka: Yeah, the only difference is that `usedDatasetTypes` includes "all" while `actualDatasetTypes` does not.

PR #3598, opened 2020-10-21
SNO-175-fix-parametrize-unhashable-list-errors

caseylitton:
https://github.com/ENCODE-DCC/snovault/pull/301

failing indexer tests are the typcial region/vis we're having trouble with.  I restarted the bdd.


keenangraham (dismissed):

caseylitton:
could become an issue when we update pytest

PR #3599, opened 2020-10-21
SNO-174-add-local-storage-redis

caseylitton:


ini-templates/base-template.ini
```diff
@@ -74,7 +74,7 @@ set remote_indexing_threshold = ${remote_indexing_threshold}
 use = egg:encoded#indexer
 app = app
 path = /index_vis
-timeout = 60
+set timeout = 60
```
caseylitton, line 79: these can't change.  
- 'set timeout = #' allows us to read the timeout in the config in the app
- 'timeout = #' is what the es_indexer_listener has access too.

pedrohr (dismissed):
💯

pedrohr (dismissed):
💯 💯

buildout.cfg
```diff
@@ -55,14 +57,15 @@ subprocess-middleware-node = git https://github.com/lrowe/subprocess-middleware-
 pyramid = git https://github.com/Pylons/pyramid.git
 pyramid_multiauth = git https://github.com/mozilla-services/pyramid_multiauth.git
 pytest = hg https://bitbucket.org/hpk42/pytest
+pytest_exact_fixtures = git https://github.com/ENCODE-DCC/pytest_exact_fixtures.git
```
keenangraham: We are still adding this?

caseylitton: no, that's removed now. thx

PR #3601, opened 2020-10-23
ENCD-5591-Skip-vis-region-tests

caseylitton:


hitz (approved):
sad to leave this untested but it's clearly a pain point.

caseylitton:
not needed found other fix

PR #3604, opened 2020-10-23
ENCD-5457-lower-audit-severity-RNA-frag-size-for-Bru

iyoungworth:


src/encoded/audit/experiment.py
```diff
@@ -3973,7 +3973,10 @@ def audit_library_RNA_size_range(value, system, excluded_types):
                     audit_link(path_to_text(rep['library']['@id']), rep['library']['@id'])
                 )
             )
-            yield AuditFailure('missing RNA fragment size', detail, level='NOT_COMPLIANT')
+            if value.get('assay_term_name') in ['BruChase-seq', 'Bru-seq', 'BruUV-seq']:
```
gabdank: I think it is more appropriate to define the level depending on assay_term, and then yield AuditFailure once (instead of  duplicating this code)

gabdank (changes requested):

src/encoded/audit/experiment.py
iyoungworth: That's a good point, I missed the opportunity to supply 'level' with a variable. Change made.

gabdank (approved):

PR #3607, opened 2020-10-26
SNO-176-invalidate-object-when-unlinked

keenangraham:


hitz:
I cannot merge the encoded branch because it will conflict with the snovault test loggin disablement (SNO-172).  Otherwise the code looks fine.

PR #3610, opened 2020-11-02
ENCD-5583-reorder-pipeline-on-experiment

zoldello:


src/encoded/static/components/filegallery.js
```diff
@@ -2192,7 +2192,10 @@ const AnalysesSelector = ({ analyses, selectedAnalysesIndex, handleAnalysesSelec
             <div className="analyses-selector analyses-selector--file-gallery-facets">
                 <h4>Choose analysis</h4>
                 <select className="analyses-selector" value={selectedAnalysesIndex} onChange={handleSelection}>
-                    {analyses.map((analysis, index) => (
+                    // In ENCODE releases, release year is not in the data. To get around this, if you strip out the text and
```
forresttanaka: Good comment, but it doesn‘t show up as a comment because it’s embedded in JSX instead of Javascript. You either need to convert it to {/* comment */} or move it inside the analyses.sort loop. It also has a couple trailing spaces at the ends of the lines.

src/encoded/static/components/filegallery.js
```diff
@@ -2192,7 +2192,10 @@ const AnalysesSelector = ({ analyses, selectedAnalysesIndex, handleAnalysesSelec
             <div className="analyses-selector analyses-selector--file-gallery-facets">
                 <h4>Choose analysis</h4>
                 <select className="analyses-selector" value={selectedAnalysesIndex} onChange={handleSelection}>
-                    {analyses.map((analysis, index) => (
+                    // In ENCODE releases, release year is not in the data. To get around this, if you strip out the text and 
+                    // leave only numbers, you will consistently have a means to identify ealier releases from latter ones. This 
+                    // approached was used to sort.
+                    {analyses.sort((a,b) => parseInt(b.title.replace(/\D/g,'')) - parseInt(a.title.replace(/\D/g,''))).map((analysis, index) => (
```
forresttanaka: This has a couple ESLint warnings; one for no space after the comma in the sort parameters, and one for not including a radix in parseInt. parseInt no longer requires it, but ESLint does. You could also use Math.trunc() if Babel has polyfilled it for IE11, which I’m not sure if they have.

forresttanaka (changes requested):

src/encoded/static/components/filegallery.js
zoldello: I did it the way I did because comments add download time to HTML pages. But if comments are needed in downloaded HTML file, that is doable.

forresttanaka (dismissed):
I did a clean build of your branch and looked at the final generated bundle.xxxxx.js and it looks like those JSX-embedded comments get stripped our during the uglification process like all other comments.

src/encoded/static/components/filegallery.js
```diff
@@ -2163,7 +2163,11 @@ export const compileAnalyses = (experiment, files) => {
             });
         }
     }
-    return _(compiledAnalyses).sortBy(compiledAnalysis => -compiledAnalysis.assemblyAnnotationValue);
+
+    // In ENCODE releases, release year is not in the data. To get around this, if you strip out the text and
+    // leave only numbers, you will consistently have a means to identify ealier releases from latter ones. This
+    // approached was used to sort. "Lab custom" appear at the bottom as intented
+    return _(compiledAnalyses).sortBy(a => parseInt(a.title.replace(/\D/g, ''), 10)).reverse();
```
forresttanaka: I feel like it’s sort of a shame that we do the sorting in `computeAssemblyAnnotationValue` and then ignore it with this sort. Could you instead do a two-key sort, with `computeAssemblyAnnotationValue` doing the first stage, and then sort those by project? What do you think?

zoldello: @forresttanaka 
Do you mean:

```
 return _(compiledAnalyses)
        .sortBy(compiledAnalysis => -compiledAnalysis.assemblyAnnotationValue)
        .sortBy(a => parseInt(a.title.replace(/\D/g, ''), 10)).reverse();

```
That is doable but I believe the second sort will override the first

forresttanaka: Yeah, this still overrides the first sort. If you could just sort on the project/rfa without affecting the order within each project/rfa that compiledAnalysis did, I think that would be great. I think `computeAssemblyAnnotationValue` could result in something like:

ENCODE3 GRCh38
ENCODE4 GRCh38
ENCODE3 hg19
ENCODE4 hg19

And then if you could keep that existing sorting while sorting on the project/rfa:

ENCODE4 GRCh38
ENCODE4 hg19
ENCODE3 GRCh38
ENCODE3 hg19

Right now your `replace` redoes the sorting that `computeAssemblyAnnotationValue`, making calling `computeAssemblyAnnotationValue` pointless. I used `computeAssemblyAnnotationValue` because the assembly/annotation facet in the file gallery uses this same order. So if that function needs modification for some future assembly/annotation, then this menu follows that order within a project/rfa.

zoldello: Is this along the lines you were thinking?

zoldello:
Closing it is no longer needed.

PR #3611, opened 2020-11-03
SNO-179-allow-devserver-set-es-config-dir

caseylitton:


caseylitton:
not needed atm since not merging es7

PR #3612, opened 2020-11-03
VALIS-28-update-fonts-to-match-ENCODE

zoldello:


emro (changes requested):
Shouldn't there be an update in package-lock.json also?

zoldello:
@emro package-lock was addressed.

zoldello:
This was merged via ENCD-5596; as it contains all its changes. So there is no need to pull this in.

PR #3613, opened 2020-11-06
ENCD-4660-hide-leave-modal-on-download

zoldello:


forresttanaka (approved):
Could also use a ref to keep track of onbeforeunload since you don’t use that state in the JSX, but the state seems to work fine too.

PR #3616, opened 2020-11-06
ENCD-5444 Rewrite the report page

forresttanaka:
* Nearly a complete rewrite of the /report/ page with these goals:
    * Document
    * Introduce a registry so alternate cell renderers can override default rendering without needing special-case code inline. Examples include the @id renderer that converts the @id of any object into a link to that object regardless of the type, and the image thumbnail renderer that only overrides the default rendering if the current report’s type is Image.
    * Include a new pagination mechanism so that users don’t need to scroll to the bottom to click the “Load More” button, and so the user can select the maximum number of items to view per page.
    * Eliminate the page blanking when making facet or sorting selections.
    * Make the data-extraction method — the process of converting property names from the schema columns to values in a displayable form — easier to maintain.
* The existing pager that only shows the current page between paging arrows still exists where it has already existed, but now both that and the new multi-page pager reside in pager.js, and you can select which pager to use within the Pager component.
* The data-extraction method now has two stages. One uses the dotted notation to extract embedded properties and flattens any array properties within. The second stage takes any complex values from the first stage and, depending on what the value looks like, displays a comma-separated list, an @id, or lists of stringified objects.

### DIFF
--- a/src/encoded/static/components/globals.js
+++ b/src/encoded/static/components/globals.js
@@ -107,7 +107,7 @@ export function atIdToAccession(atId) {
     if (matched && matched.length === 2) {
         return matched[1];
     }
-    return '';
+    return atId;
 }
 
 
@@ -359,4 +359,51 @@ export const viewToSvg = {
     table: 'table',
     summary: 'summary',
     th: 'matrix',
+};
+
+
+// Media query breakpoints to match those in style.scss.
+const SCREEN_XS = 480;
+const SCREEN_SM = 768;
+const SCREEN_MD = 960;
+const SCREEN_LG = 1160;
+const SCREEN_XL = 1716;
+
+
+/**
+ * Determine whether the current browser-window width activates the given Bootstrap-based media
+ * query code. The component this function gets called from must have mounted.
+ * @param min {string} Bootstrap-based code for each breakpoint
+ *
+ * @return {bool} True if specified breakpoint is active
+ */
+export const isMediaQueryBreakpointActive = (min) => {
+    if (window.matchMedia) {
+        let breakpoint;
+        switch (min) {
+        case 'XS':
+            breakpoint = SCREEN_XS;
+            break;
+        case 'SM':
+            breakpoint = SCREEN_SM;
+            break;
+        case 'MD':
+            breakpoint = SCREEN_MD;
+            break;
+        case 'LG':
+            breakpoint = SCREEN_LG;
+            break;
+        case 'XL':
+            breakpoint = SCREEN_XL;
+            break;
+        default:
+            // Undefined code; default to mobile.
+            breakpoint = 0;
+        }
+        const mql = window.matchMedia(`(min-width: ${breakpoint}px)`);
+        return mql.matches;
+    }
+
+    // Browser doesn't support matchMedia, so default to mobile.
+    return false;
 };
--- a/src/encoded/static/components/globals.js
+++ b/src/encoded/static/components/globals.js
@@ -101,7 +101,7 @@ export const sanitizeId = id => (id ? `${id.replace(/\s/g, '_')}` : '-');
 
 
 // Take an @id and return the corresponding accession. If no accession could be found in the @id,
-// the empty string is returned.
+// the unchanged @id is returned.
 export function atIdToAccession(atId) {
     const matched = atId.match(/^\/.+\/(.+)\/$/);
     if (matched && matched.length === 2) {
--- a/src/encoded/static/components/objectutils.js
+++ b/src/encoded/static/components/objectutils.js
@@ -493,13 +493,11 @@ export const DownloadableAccession = (props) => {
 
 DownloadableAccession.propTypes = {
     file: PropTypes.object.isRequired, // File whose accession to render
-    buttonEnabled: PropTypes.bool, // Check if button is enabled
     clickHandler: PropTypes.func, // Function to call when button is clicked
     loggedIn: PropTypes.bool, // True if current user is logged in
 };
 
 DownloadableAccession.defaultProps = {
-    buttonEnabled: true,
     clickHandler: null,
     loggedIn: false,
 };
@@ -899,3 +897,15 @@ export const isFileVisualizable = file => (
 export function filterForVisualizableFiles(fileList) {
     return fileList.filter(file => isFileVisualizable(file));
 }
+
+
+/**
+ * Displays an item count intended for the tops of table, normally reflecting a search result count.
+ */
+export const TableItemCount = ({ count }) => (
+    <div className="table-item-count">{count}</div>
+);
+
+TableItemCount.propTypes = {
+    count: PropTypes.string.isRequired,
+};
--- a/src/encoded/static/components/report.js
+++ b/src/encoded/static/components/report.js
@@ -174,11 +174,16 @@ ReportDataRegistry.register({ field: 'download_url', type: 'Image' }, ImageDownl
  * `files` renderer for type=Experiment. Displays a comma separated list of accessions that link to
  * the corresponding file page.
  */
-const ExperimentFilesRenderer = ({ value }) => {
-    if (value && typeof value === 'object' && Array.isArray(value) && value.length > 0) {
-        return value.map(file => <a key={file} href={file}>{globals.atIdToAccession(file)}</a>).reduce((prev, curr) => [prev, ', ', curr]);
+const ExperimentFilesRenderer = ({ item }) => {
+    if (item.files && item.files.length > 0) {
+        return item.files.map(file => <a key={file['@id']} href={file['@id']}>{globals.atIdToAccession(file['@id'])}</a>).reduce((prev, curr) => [prev, ', ', curr]);
     }
     return null;
+};
+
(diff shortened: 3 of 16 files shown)


forresttanaka:
Patched demo: https://encd-5444-report-rewrite-fytanaka.demo.encodedcc.org/

src/encoded/static/components/globals.js
```diff
@@ -107,7 +107,7 @@ export function atIdToAccession(atId) {
     if (matched && matched.length === 2) {
         return matched[1];
     }
-    return '';
+    return atId;
```
emro, line 110: The explanation for the function isn't accurate any more: "if no accession could be found, the empty string is returned".

src/encoded/static/components/objectutils.js
```diff
@@ -899,3 +897,15 @@ export const isFileVisualizable = file => (
 export function filterForVisualizableFiles(fileList) {
     return fileList.filter(file => isFileVisualizable(file));
 }
+
+
+/**
+ * Displays an item count intended for the tops of table, normally reflecting a search result count.
+ */
+export const TableItemCount = ({ count }) => (
+    <div className="table-item-count">{count}</div>
```
emro, line 906: There are probably other places we can use this too, right?

src/encoded/static/components/objectutils.js
```diff
@@ -493,13 +493,11 @@ export const DownloadableAccession = (props) => {
 
 DownloadableAccession.propTypes = {
     file: PropTypes.object.isRequired, // File whose accession to render
-    buttonEnabled: PropTypes.bool, // Check if button is enabled
     clickHandler: PropTypes.func, // Function to call when button is clicked
     loggedIn: PropTypes.bool, // True if current user is logged in
 };
 
 DownloadableAccession.defaultProps = {
-    buttonEnabled: true,
```
emro, line 502: This seems like a good thing to get rid of. I kept the `buttonEnabled` on `RestrictedDownloadButton` which I think we don't actually use, despite setting it in various places. In any case, I think we won't have any merge conflicts here with ENCD-5566.

src/encoded/static/components/globals.js
```diff
@@ -360,3 +360,50 @@ export const viewToSvg = {
...
+            // Undefined code; default to mobile.
+            breakpoint = 0;
+        }
+        const mql = window.matchMedia(`(min-width: ${breakpoint}px)`);
+        return mql.matches;
+    }
+
+    // Browser doesn't support matchMedia, so default to mobile.
+    return false;
+};
```
emro, line 409: This looks very useful. Potentially I could be using this in `tooltip.js` where I check to see if the screen width is less than 960px.

src/encoded/static/components/report.js
```diff
@@ -3,445 +3,586 @@ import PropTypes from 'prop-types';
...
+    let result = rawValue;
+    if (typeof rawValue === 'object') {
+        if (Array.isArray(rawValue)) {
+            // Flatten and dedupe arrays.
+            const flattenedValue = _.chain(rawValue)
+                .flatten()
+                .uniq()
+                .value();
+
+            // Take care of arrays by concatenting their elements into one string.
```
emro: "concatenating" not "concatenting"

src/encoded/static/components/report.js
```diff
@@ -3,445 +3,586 @@ import PropTypes from 'prop-types';
...
+
+/**
+ * Display all the data rows of the report table.
+ */
+const ReportData = ({ context, visibleFields, type }) => (
+    <tbody>
+        {context['@graph'].map(item => (
+            <tr key={item['@id']}>
+                {visibleFields.map((field) => {
+                    // Get the value of the property in `item` and see if the currenet field and
```
emro: "current" not "currenet"

src/encoded/static/libs/ui/pager.js
```diff
@@ -40,4 +42,138 @@ Pager.propTypes = {
...
+ *
+ * For cases with more than nine pages, Pager attempts to keep a cluster of visible page numbers
+ * surrounding the current page so that the user can see and select the two preceding and two
+ * succeeding page numbers, except when the current page gets to the extreme ends of the page
+ * range and fewer than two page numbers exist on one side of the cluster. When the current page
+ * approaches the ends of the page range, more than two visible pages appear in the cluster on the
+ * side facing the near end of the page range. This allowance keeps the width of the entire Pager
+ * component consistent regardless of the current page number, so that the previous/next buttons
+ * don't shift around horizontally.
+ */
```
emro, line 72: I think this works really well. When I looked at the demo prior to reading the code, I expected the ellipses to be clickable, but when they weren't, the way that they actually work is pretty intuitive. It does take two clicks to get to some pages, but if you had to click on an ellipsis that would be two clicks too and I'm not sure why you'd need to get to page 11 specifically anyway. I like how the previous/next buttons don't move horizontally.

src/encoded/static/libs/ui/pager.js
```diff
@@ -40,4 +42,138 @@ Pager.propTypes = {
...
+    // React keys. No point memoizing `pageNumbers` as it needs recalculating when any prop changes.
+    let pageNumbers;
+    if (total <= 9) {
+        // A total page count of nine or fewer has no ellipses -- just a straight array of
+        // sequential numbers.
+        pageNumbers = _.range(1, total + 1); // _.range ends at max - 1
+    } else {
+        // With more than nine pages, build a cluster of pages around the current page, first by
+        // determining the minimum and maximum page numbers for the cluster. Allow for filling in
+        // extra numbers in the cluster for cases where the current page number approches the ends
```
emro: "approaches"

src/encoded/static/libs/ui/pager.js
```diff
@@ -40,4 +42,138 @@ Pager.propTypes = {
...
+                })}
+                <li className={`pager-multi__page pager-multi__page--arrow${currentPage === total ? ' pager-multi__page--arrow-disabled' : ''}`}>
+                    <button type="button" onClick={nextClick} aria-label="Next page" aria-disabled={currentPage === total}>
+                        {svgIcon('chevronRight')}
+                    </button>
+                </li>
+            </ul>
+        </nav>
+    );
+};
```
emro, line 170: I think this works pretty well on mobile, but I did have a thought which is more general than this ticket, which is that it is frustrating how clicking on a page number scrolls you back to the top and then you have to scroll back through all of the many facets. I feel like it could be better to have the facets hidden and expandable on mobile.

src/encoded/static/scss/encoded/modules/_pager.scss
```diff
@@ -57,3 +57,99 @@ $pager-dimension: 24px;
...
+                    height: 100%;
+                    padding-left: 5px;
+                    padding-right: 5px;
+                    line-height: 1rem;
+                    font-size: 1rem;
+                    font-weight: bold;
+                    border: none;
+                    background-color: transparent;
+                    border-radius: 0;
+                    cursor: pointer;
```
emro, line 101: I like these button styles. I think they go well with our existing button styles.

emro (commented):
It all looks great. I had a couple of big-picture comments, some not directly related to this ticket. Also I found a few typos which are not particularly important. This seems like a big improvement and I think the pagination could be very useful in other places too.

forresttanaka:
Thanks! I’ll fix the typos and push again.

src/encoded/static/libs/ui/pager.js
forresttanaka: Yeah, and maybe expandable per facet might be good. That will take some thought.

src/encoded/static/libs/ui/pager.js
forresttanaka: I saw a variation where clicking an ellipses brings up a tooltip with all the missing page numbers as selectable. In some cases though, we would have a couple thousand pages behind an ellipsis, so I didn’t see that as practical.

src/encoded/static/components/globals.js
forresttanaka: Yeah, I wanted to see if this worked well as a generic function. I’m surprised I’ve never needed this before.

src/encoded/static/components/objectutils.js
forresttanaka: Yes. I copied this item-count header style from the file table in FileGallery and figured we could use this for tables that needs counts. I feel like using this in the head of list and matrix views could make placing the “Showing x of y results” a non-issue. Placing that text in a logical place in matrix headers always seems difficult.

emro (dismissed):
Looks good! Some of the new functions will be useful across the site, hopefully we can propagate those as we continue to update other pages.

forresttanaka:
The data-extraction method I made doesn’t work for cases Meenakshi had, and I think fixing it to make it work will make it even more complicated than what we had, so I restored David Glick’s method, `lookupColumn`. I removed the reportCell lookup it had though; no one used it as it wasn’t really flexible enough. Hopefully people will use the new method for future exceptions.

forresttanaka:
Need to add some changes from Ben and check performance again before PR.

emro (approved):
Everything looks good, no changes needed code-wise.

PR #3617, opened 2020-11-09
ENCD-5599-adjust-madqc-auditing-for-technical-replicates

iyoungworth:


src/encoded/audit/experiment.py
```diff
@@ -1373,11 +1395,20 @@ def check_experiment_cage_rampage_standards(experiment,
...
+            mad_metrics, 'RAMPAGE (paired-end, stranded)', 0.9)
+
+        return
+
+    else:
+        yield from check_spearman(
+            mad_metrics, experiment['replication_type'],
+            0.9, 0.8, 'RAMPAGE (paired-end, stranded)')
+
+        return
```
gabdank: I think the technical MAD should be limited to bulk RNA only. 
we are not planning to run and we don't have a new RAMPAGE pipeline.

src/encoded/audit/experiment.py
```diff
@@ -1315,11 +1327,21 @@ def check_experiment_small_rna_standards(experiment,
...
+            0.9)
+
+        return
+
+    else:
+        yield from check_spearman(
+            mad_metrics, experiment['replication_type'],
+            0.9, 0.8, 'Small RNA-seq single-end pipeline')
+
+        return
```
gabdank: I think the technical MAD should be limited to bulk RNA only.
we are not planning to run small RNA-seq single-end pipeline (it is the old RNA pipeline, I don't think it has to handle the tech rep MAD case)

src/encoded/audit/experiment.py
```diff
@@ -1252,18 +1252,25 @@ def check_experiment_long_rna_standards(experiment,
...
+        mad_metrics = get_metrics(gene_quantifications,
+                                  'MadQualityMetric',
+                                  desired_assembly,
+                                  desired_annotation)
+        replicates = experiment.get('replicates')
+
+        if experiment['replication_type'] == 'unreplicated' and len(replicates) > 1:
+            yield from check_spearman_technical_replicates(
+                mad_metrics, pipeline_title, 0.9)
+
```
gabdank: not a big deal, but I think you can drop the empty lines

gabdank (changes requested):

src/encoded/audit/experiment.py
iyoungworth: Got it, I've removed the empty lines in this section

gabdank (approved):

PR #3618, opened 2020-11-09
ENCD-5616-update-ckeditor

paul-sud:


forresttanaka (dismissed):
Looks good to me. Nice cleanup of the old mechanism too.

PR #3627, opened 2020-11-12
ENCD-5629-add-file-output-type-entries-to-glossary

iyoungworth:
minor-corrections


gabdank (dismissed):

gabdank (changes requested):
I think the updated description of alignments, should be applied also for unfiltered and transcriptome alignments.

iyoungworth:
Thank you for noticing that, I updated the unfiltered and transcriptome alignments definitions and rebased

src/encoded/static/components/glossary/glossary.json
```diff
@@ -1,10 +1,512 @@
 [
     {
-		"term": "Functional Genomics data",
+		"term": "functional genomics data",
 		"definition": "Data generated by assays investigating processes such as transcription, translation and epigenetic regulation on a genome-wide scale. Examples of assays generating functional genomics data: RNA-seq, ChIP-seq, DNase-seq, ATAC-seq, WGBS, HiC, and ChIA-PET."
 	},
 	{
-		"term": "Functional Characterization data",
+		"term": "functional characterization data",
 		"definition": "Data generated by assays (i.e STARR-seq, MPRA, and CRISPR screen) investigating the relationship between DNA sequences and their regulatory activities."
```
gabdank: @iyoungworth can  you please align these two old entries with the rest of the entries in the file? It looks like the indentation of these entries is different.

gabdank (changes requested):
please correct the indentation

src/encoded/static/components/glossary/glossary.json
iyoungworth: Sure thing- it looks like my text editor switched between tabs and spaces. They should all be consistent now.

gabdank (approved):

PR #3628, opened 2020-11-14
ENCD-4529-change-timing-of-unloggedin-cart-count-max-error

zoldello:


forresttanaka (commented):
It seems to work OK, both logged in and not logged in. I’m getting 503 errors with cart requests on your demo while logged in, like if I try to add thousands of experiments to my logged-in cart.

PR #3631, opened 2020-11-17
ENCD-5631 Add audit for treatment series for mixed units

jenjou:


src/encoded/audit/series.py
```diff
@@ -0,0 +1,78 @@
...
+                    if rep['status'] not in ['deleted'] and \
+                       'library' in rep and 'biosample' in rep['library']:
+                        biosample_object = rep['library']['biosample']
+                        if 'treatments' in biosample_object:
+                            if len(biosample_object['treatments']) != 0:
+                                for t in biosample_object['treatments']:
+                                    treatment_duration_units.add(t['duration_units'])
+    if len(treatment_duration_units) > 1:
+        detail = ('Treatments associated with series {} '
+            ' use inconsistent duration units.'.format(
```
gabdank: I think there is an extra space in the detail string?

gabdank: and I think we agreed to use f'strings?

src/encoded/audit/series.py
```diff
@@ -0,0 +1,78 @@
...
+                       'library' in rep and 'biosample' in rep['library']:
+                        biosample_object = rep['library']['biosample']
+                        if 'treatments' in biosample_object:
+                            if len(biosample_object['treatments']) != 0:
+                                for t in biosample_object['treatments']:
+                                    treatment_amount_units.add(t['amount_units'])
+    if len(treatment_amount_units) > 1:
+        detail = ('Treatments associated with series {} '
+            ' use inconsistent amount units.'.format(
+                audit_link(path_to_text(value['@id']), value['@id']),
```
gabdank: extra space in the detail string, and perhaps it should be in f'string?

src/encoded/audit/series.py
```diff
@@ -0,0 +1,78 @@
...
+            )
+        )
+        yield AuditFailure('mixed amount units in treatment concentration series',
+                           detail, level='WARNING')
```
gabdank: we may consider having a shorter audit category. instead of "mixed amount units in treatment concentration series" you can do something like: "inconsistent treatment units" - that will fit better in facet, and will cover both for time and concentration

src/encoded/audit/series.py
```diff
@@ -0,0 +1,78 @@
...
+                                    treatment_duration_units.add(t['duration_units'])
+    if len(treatment_duration_units) > 1:
+        detail = ('Treatments associated with series {} '
+            ' use inconsistent duration units.'.format(
+                audit_link(path_to_text(value['@id']), value['@id'])
+            )
+        )
+        yield AuditFailure('mixed duration units in treatment time series',
+                           detail, level='WARNING')
+        return
```
gabdank: we may consider having a shorter audit category. instead of "mixed duration units in treatment time series" you can do something like: "inconsistent treatment units" - that will fit better in facet, and will cover both for time and concentration

gabdank (changes requested):

gabdank (dismissed):

gabdank (approved):

PR #3632, opened 2020-11-17
ENCD-4525 Cart improvements

forresttanaka:
* _A lot_ of files have changed, but the vast majority have a small change involving a new `<TopAccessories>` component that contains breadcrumbs instead of calling `<BreadCrumbs>` directly. The released status now gets handled inside `<TopAccessories>`. In a couple cases, I removed breadcrumbs where they served no purpose.
* I removed breadcrumbs from genetic_modification.js because the properties it used no longer exist.
* I converted `NameInput` to functional component with hooks so that we focus on the field with `useEffect` _after_ the modal gets painted instead of before.
* I added a `rowKeys` property to `<SortTable>` so that the table doesn’t _need_ to rely on a loop index as the React key, though you still can use the loop index if you need to.
* The modal code now handles the return key for submissions in addition to the ESC key for closing.

### DIFF
--- a/src/encoded/static/components/cart/add_multiple.js
+++ b/src/encoded/static/components/cart/add_multiple.js
@@ -8,7 +8,7 @@ import { connect } from 'react-redux';
 import url from 'url';
 import QueryString from '../../libs/query_string';
 import { addMultipleToCartAndSave, cartOperationInProgress } from './actions';
-import { atIdToType } from '../globals';
+import { atIdToType, truncateString } from '../globals';
 import { requestSearch } from '../objectutils';
 import {
     MaximumElementsLoggedoutModal,
@@ -86,7 +86,8 @@ class CartAddAllSearchComponent extends React.Component {
         const { savedCartObj, inProgress } = this.props;
         const cartName = (savedCartObj && Object.keys(savedCartObj).length > 0 ? savedCartObj.name : '');
         return (
-            <React.Fragment>
+            <div className="cart-add-all">
+                {savedCartObj && savedCartObj.name ? <div className="cart-toggle__name">{truncateString(savedCartObj.name, 22)}</div> : null}
                 <button
                     disabled={inProgress || savedCartObj.locked}
                     className="btn btn-info btn-sm"
@@ -98,7 +99,7 @@ class CartAddAllSearchComponent extends React.Component {
                 {this.state.overMaximumError ?
                     <MaximumElementsLoggedoutModal closeClickHandler={this.handleErrorModalClose} />
                 : null}
-            </React.Fragment>
+            </div>
         );
     }
 }
@@ -177,7 +178,8 @@ const CartAddAllElementsComponent = ({ savedCartObj, elements, inProgress, addAl
 
     const cartName = (savedCartObj && Object.keys(savedCartObj).length > 0 ? savedCartObj.name : '');
     return (
-        <div className="cart__add-all-element-control">
+        <div className="cart-add-all">
+            {savedCartObj && savedCartObj.name ? <div className="cart-toggle__name">{truncateString(savedCartObj.name, 22)}</div> : null}
             <button
                 disabled={inProgress || savedCartObj.locked}
                 className="btn btn-info btn-sm"
--- a/src/encoded/static/components/cart/manager.js
+++ b/src/encoded/static/components/cart/manager.js
@@ -49,6 +49,7 @@ class CurrentCartButtonComponent extends React.Component {
                 value={cart['@id']}
                 onChange={this.handleRadioButtonChange}
                 checked={selected}
+                aria-label="Current cart"
                 className="cart-manager-table__current-button"
                 disabled={cart.status === 'deleted' || cart.status === 'disabled' || inProgress}
             />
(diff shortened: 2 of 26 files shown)


forresttanaka:
Patched demo at:
https://encd-4525-cart-improvements-fytanaka.demo.encodedcc.org/

src/encoded/static/components/cart/manager.js
```diff
@@ -706,31 +715,50 @@ const cartTableColumns = {
...
-    const cartCount = cartManager['@graph'].reduce((sum, cart) => (cart.status !== 'deleted' ? sum + 1 : sum), 0);
-    return (
-        <div className="cart-counts">
-            {cartCount} cart{cartCount !== 1 ? 's' : ''} ({cartManager.cart_user_max} maximum)
-        </div>
-    );
-};
+const CartCounts = ({ cartManager, cartCount }) => (
+    <div className="cart-counts">
+        {cartCount} cart{cartCount !== 1 ? 's' : ''} ({cartManager.cart_user_max} maximum)
```
zoldello, line 720: We may need to create a utility function for this the code-  !== 1 ? 's' : ''

src/encoded/static/libs/ui/modal.js
```diff
@@ -387,28 +387,35 @@ export class Modal extends React.Component {
...
+        if (!this.props.actuator || this.state.modalOpen) {
+            if (e.keyCode === 27) {
+                // User typed ESC.
+                if (this.props.closeModal) {
+                    this.props.closeModal();
+                } else {
+                    this.closeModal();
+                }
+            } else if (e.keyCode === 13) {
+                // User typed RETURN.
```
zoldello, line 415: It may be a good idea to also reference the "Enter" key as some developers may use Windows or Linux.

src/encoded/static/components/objectutils.js
```diff
@@ -753,6 +743,31 @@ ItemAccessories.contextTypes = {
 };
 
 
+/**
+ * Display the top section containing the breadcrumb links on summary pages.
+ */
+export const TopAccessories = ({ context, crumbs }) => {
```
zoldello, line 749: Can this be named "TopBreadCrumbSection"? "TopAccessories" seems very generic

src/encoded/static/components/cart/add_multiple.js
```diff
@@ -86,7 +86,8 @@ class CartAddAllSearchComponent extends React.Component {
         const { savedCartObj, inProgress } = this.props;
         const cartName = (savedCartObj && Object.keys(savedCartObj).length > 0 ? savedCartObj.name : '');
         return (
-            <React.Fragment>
+            <div className="cart-add-all">
```
zoldello, line 89: Does "cart-add-all" follow the naming convention?

src/encoded/static/components/cart/add_multiple.js
```diff
@@ -86,7 +86,8 @@ class CartAddAllSearchComponent extends React.Component {
...
+                {savedCartObj && savedCartObj.name ? <div className="cart-toggle__name">{truncateString(savedCartObj.name, 22)}</div> : null}
```
zoldello, line 90: The Jira ticket tells me why 22 is used? But a future code-reader may not know why. Maybe a const with an appropriate name would make it clearer.

src/encoded/static/components/cart/manager.js
```diff
@@ -293,17 +288,24 @@ class NameCartButtonComponent extends React.Component {
...
      */
     handleSubmit() {
-        const { create, onCreate, onRename, updateCartManager } = this.props;
+        const { create, onCreate, onRename, onSwitchCart, updateCartManager, user } = this.props;
         const { newName, newIdentifier } = this.state;
         const submitFunc = create ? onCreate : onRename;
-        submitFunc(newName.trim(), newIdentifier).then(() => {
-            // Successfully changed the cart name/identifier. Indicate the cart manager needs
-            // reloading so it displays the new cart information.
+        submitFunc(newName.trim(), newIdentifier).then((cartAtId) => {
```
zoldello, line 294: You are modifying a parameter. That is an anti-pattern.

src/encoded/static/components/cart/manager.js
```diff
@@ -293,17 +288,24 @@ class NameCartButtonComponent extends React.Component {
...
             this.handleClose();
+
+            // Successfully changed the cart name/identifier. Indicate the cart manager needs
+            // reloading so it displays the new cart information, and set the new cart as the
+            // current cart. Null user only on initial load, but make sure to handle a null user
```
zoldello, line 299: "Nullify user"?

src/encoded/static/components/cart/manager.js
```diff
@@ -541,7 +550,7 @@ DeleteCartButtonComponent.propTypes = {
     setInProgress: PropTypes.func.isRequired,
     /** Function to call to update the cart manager page */
     updateCartManager: PropTypes.func.isRequired,
-    /** System function for xhr requests */
+    /** System function for XHR requests */
```
zoldello, line 553: XHR refers to XMLHttpRequest. Fetch API is a different architecture.

src/encoded/static/components/cart/manager.js
```diff
@@ -706,31 +715,50 @@ const cartTableColumns = {
...
+const CartCounts = ({ cartManager, cartCount }) => (
+    <div className="cart-counts">
+        {cartCount} cart{cartCount !== 1 ? 's' : ''} ({cartManager.cart_user_max} maximum)
+    </div>
+);
 
 CartCounts.propTypes = {
     /** cart-manager object from database */
     cartManager: PropTypes.object.isRequired,
+    /** Number of carts not deleted and not auto save */
```
zoldello, line 727: "saved"?

src/encoded/static/components/cart/manager.js
```diff
@@ -706,31 +715,50 @@ const cartTableColumns = {
...
-    <div className="cart-manager-table__legend">
-        <div className="cart-manager-table__legend-item"><div className="cart-manager-table__chip--current" />Current</div>
-        <div className="cart-manager-table__legend-item"><div className="cart-manager-table__chip--autosave" />Auto Save</div>
+const CartManagerFooter = ({ adminUser, isDeletedVisible, deletedVisibleChangeHandler }) => (
+    <div className="cart-manager-table__footer">
+        <div className="cart-manager-table__footer-item">
+            {adminUser ?
+                <React.Fragment>
+                    <input id="check-deleted-visible" type="checkbox" checked={isDeletedVisible} onChange={deletedVisibleChangeHandler} />
+                    <label htmlFor="check-deleted-visible">Show deleted carts</label>
```
zoldello, line 741: Good idea.

zoldello (approved):
Code looks good. My comments cover nothing that must be changed.

src/encoded/static/components/cart/manager.js
forresttanaka: Yeah, that’s probably a good idea.

src/encoded/static/components/objectutils.js
forresttanaka: Intentionally generic. Conceivably, you could have a cart button without breadcrumbs.

src/encoded/static/components/cart/add_multiple.js
forresttanaka: Yes. “cart-add-all” represents a Block in BEM.

src/encoded/static/components/cart/manager.js
forresttanaka: You mean trim()? According to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim#Description), trim() makes a new string.

src/encoded/static/components/cart/manager.js
forresttanaka: Yeah, that’s true.

src/encoded/static/components/cart/manager.js
forresttanaka: I meant the proper noun “Auto Save cart,” though I should have capitalized.

PR #3638, opened 2020-11-19
ENCD-5607-warning-about-cart-title-private-info

zoldello:


forresttanaka:
We’re not going to do anything about the default last name being their email address? You can still see it with a User search. I don’t know what the last decision about this was.

src/encoded/static/components/cart/manager.js
```diff
@@ -135,12 +136,15 @@ NameInput.propTypes = {
     nonValidMessage: PropTypes.string,
     /** True if label and input on the same line */
     inline: PropTypes.bool,
+    /** HTML placeholder content */
+    placeholder: PropTypes,
```
forresttanaka: `PropTypes.string` I think.

forresttanaka (changes requested):
One small change. I’m surprised no ESLint error on that.

forresttanaka (approved):

PR #3642, opened 2020-11-25
ENCD-5583-reorder-pipeline-on-experiment

zoldello:


src/encoded/static/components/filegallery.js
```diff
@@ -2180,6 +2181,36 @@ const AnalysesSelector = ({ analyses, selectedAnalysesIndex, handleAnalysesSelec
         }
     });
 
+    /**
+     * Selects certain awards as the default if listed.
+    */
+    useMount(() => {
+        const awardSuffix = ['ENCODE4', 'ENCODE3', 'ENCODE2', 'Lab custom']; // ordered in decreasing precidence
+        const analysesAwarded = analyses.filter(analysis => awardSuffix.some(award => analysis.title.includes(award)));
```
forresttanaka, line 2189: Would it help if you modified `compileAnalyses` so that it kept track of `pipeline_award_rfas` for non-mixed cases, and passed that through to this `analyses` parameter? That way you don’t have to extract the number from the title, and instead compare it against the contents of `awardSuffix`?

zoldello: The issue is that title is the only sure way to determine the ordering. For example, let say you have to entries: 
```
[
	{
		title: ‘ENCODE3 Uniform’,
		pipeline_award_rfas: ‘ENCODE3’,
	},
	{
		title: ‘Mixed GRch38’,
		pipeline_award_rfas: ‘ENCODE3’,
	}
]

```
Both of these have the same pipeline. However, the first entry supposed to have higher precedence over the second.

forresttanaka: Doesn’t the “Mixed” case not get sorted anyway? You would only sort on things the RFA matches. I’m not sure the last decision on how this was supposed to get sorted.

zoldello: I missed your last message. The goal is no longer to sort. The goal is to identify ansuitable entry to be auto-selected. This variable:

`const awardSuffix = ['ENCODE4', 'ENCODE3', 'ENCODE2', 'Lab custom', 'Mixed'];`

is the descending order of precedence.

forresttanaka: OK, I see that what you do achieves in one step what I would have done in two, and I can see the argument for that. I still would have done it in two, but again, I can see the argument for your way. I’ll continue the code review.

forresttanaka (approved):
That has become a very complex menu. Thanks for the really good code comments too.

PR #3646, opened 2020-12-01
ENCD-4551-disable-visit-sharable-cart-during-cart-ops

zoldello:


forresttanaka (changes requested):
The Cart Manager page also has a share button for each cart in the list, and it opens the same modal, but it still shows the problem described in the ticket because the Cart Manager doesn’t pass the `disableFooterSubmitBtn` boolean.

You can cover both cases by having `CartShare` get its own copy of `inProgress` by grabbing `state.inProgress` in its `mapStateToProps` function, the same way `CartStatus` does. Then you won’t need everyone to pass it the `disableFooterSubmitBtn` property.

src/encoded/static/components/cart/status.js
```diff
@@ -137,7 +137,7 @@ class CartStatusComponent extends React.Component {
                     <DropdownMenu label="cart-control">
                         {menuItems}
                     </DropdownMenu>
-                    {this.state.shareModalOpen ? <CartShare userCart={savedCartObj} closeShareCart={this.closeShareCart} /> : null}
+                    {this.state.shareModalOpen ? <CartShare disableFooterSubmitBtn={inProgress} userCart={savedCartObj} closeShareCart={this.closeShareCart} /> : null}
```
forresttanaka: `<CartShare>` doesn’t use `disableFooterSubmitBtn`, and it doesn’t need to because you’re already retrieving the store `inProgress` state.

forresttanaka (changes requested):
All looks good except for the unneeded `disableFooterSubmitBtn` property.

forresttanaka (approved):

PR #3647, opened 2020-12-02
ENCD-5543-add-series-page

emro:
This update primarily adds a search page for series, but it also includes style changes on every page with tabs.
- New icons designed by Khine are used for organisms and series
- Tabs are left aligned
- Tabs that are also headlines use the Oswald font
- Tabs don't have the lime green top border
- Active tabs have a white background and non-active tabs have a gray background
- All tab panels have left/right/bottom borders
- Tabs turn into buttons on mobile / vertical tablets
- Colors on ChIP-seq tab with low contrast are eliminated
- Tab code is consolidated


src/encoded/static/components/search.js
```diff
@@ -1213,6 +1215,7 @@ export class ResultTable extends React.Component {
                     filters={filters}
                     searchBase={searchBase ? `${searchBase}&` : `${searchBase}?`}
                     onFilter={this.onFilter}
+                    seriesFlag={seriesFlag}
```
forresttanaka: I’d prefer to call this something like of `hideDocType` in case we need this for things not having anything to do with Series, similar to `hideBrowserSelector` above.

src/encoded/glossary.py
```diff
@@ -18,3 +19,17 @@ def glossary(context, request):
         'filters': []
     }
     return result
+
+
+@view_config(route_name='series_search', request_method='GET', permission='search')
+def glossary(context, request):
```
forresttanaka: I’m not sure of the connection /series-search/ has with glossary. Is there one?

src/encoded/static/components/series_search.js
```diff
@@ -0,0 +1,165 @@
...
+    /* eslint-enable jsx-a11y/click-events-have-key-events */
+};
+
+SeriesSearch.contextTypes = {
+    location_href: PropTypes.string,
+    navigate: PropTypes.func,
+    fetch: PropTypes.func,
+};
+
+globals.contentViews.register(SeriesSearch, 'SeriesSearch');
```
forresttanaka, line 153: We would really need Keenan to confirm, but I feel like you could simplify this by making /series-search/ a real search view, adding to search_view.py. The advantage is that the returned JSON would know to go to /search-view/ instead of /search/ and you won’t have to have that `<div>` wrapper and click handler to modify the URL. So I’m thinking `/series-search/?type=OrganismDevelopmentSeries`. Each facet and tab click would just use links instead of GET requests. I could easily be missing something obvious that makes this impractical, but I feel like that would be simpler if it works.

src/encoded/static/components/search.js
emro: I thought maybe we'd want more updates to the <ResultTable> also but ended up just doing the one change so this is better.

src/encoded/static/components/series_search.js
emro: Updated.

src/encoded/glossary.py
emro: Updated.

emro:
Recent updates include:
- linking API and filters/buttons
- using <TabPanel> for summary page and series-search page
- add back `disabled` styles
- clear filters on series-search tab clicks
- default to organism development series if no series is specified in url
- fix tests

src/encoded/static/components/search.js
```diff
@@ -1213,6 +1217,7 @@ export class ResultTable extends React.Component {
...
```
forresttanaka: Looks like `<FacetList>` takes `hideDocType` but you pass `seriesFlag` to it.

src/encoded/static/components/summary.js
```diff
@@ -178,7 +178,7 @@ SummaryStatusChart.contextTypes = {
 
 // Render the horizontal facets.
 // Note: these facets are not necessarily horizontal, it depends on the screen width
-const SummaryHorizontalFacets = ({ context, facetList }, reactContext) => {
+const SummaryHorizontalFacets = ({ context, facetList }) => {
```
forresttanaka, line 181: You can also delete the `contextTypes` object below since we don’t use it here anymore.

src/encoded/static/components/summary.js
```diff
@@ -330,78 +330,110 @@ class SummaryBody extends React.Component {
...
-                        <div className="results-count">There {this.props.context.total > 1 ? 'are' : 'is'} <b className="bold-total">{this.props.context.total}</b> result{this.props.context.total > 1 ? 's' : ''}.</div>
-                        <div className="results-table-control results-table-control--centered">
-                            <div className="results-table-control__main">
-                                <ViewControls results={this.props.context} />
-                            </div>
+                                <div>
+                                    <div className={`results-controls ${this.state.selectedOrganism.length > 0 ? `${this.state.selectedOrganism.replace(' ', '-')}` : ''}`}>
+                                        <div className="results-count">There {this.props.context.total > 1 ? 'are' : 'is'} <b className="bold-total">{this.props.context.total}</b> result{this.props.context.total > 1 ? 's' : ''}.</div>
+                                        <ClearFilters searchUri={this.props.context.clear_filters} enableDisplay={!!clearButton} />
```
forresttanaka: Technically `clearButton` is already boolean, so the `!!` does nothing.

src/encoded/static/components/series_search.js
```diff
@@ -0,0 +1,161 @@
...
+        if (!(query.getKeyValues('type')[0])) {
+            query.addKeyValue('type', selectedSeries);
+            const href = `?${query.format()}`;
+            context.navigate(href);
+        }
+    }, [context, context.fetch, query, selectedSeries]);
+
+    React.useEffect(() => {
+        setSeriesTabs(buildTabs());
+    }, [buildTabs]);
```
forresttanaka: Doesn’t look to me like `buildTabs` needs to wait for component mount to run, or does it? Could its code just run inline on every render, with `seriesTabs` just a regular variable instead of a state? That would also avoid modifying the `seriesTabs` state variable directly, which isn’t allowed under React rules. The newest ESLint does _very_ strict tests for that.

forresttanaka (changes requested):
The new endpoint search makes that looks a lot easier.

src/encoded/search_views.py
```diff
@@ -103,6 +104,37 @@ def search(context, request):
     )
     return fr.render()
 
+@view_config(route_name='series_search', request_method='GET', permission='search')
```
keenangraham, line 108: Looks good but usually put two lines between functions.

forresttanaka (dismissed):

keenangraham (dismissed):
By the way you might want to add a `series_search` test following this pattern just to make sure the backend is returning what you expect: https://github.com/ENCODE-DCC/encoded/blob/d12489031f98c7f3ab82ab68a0f867a12cbea0f5/src/encoded/tests/test_search_views.py#L12-L32

forresttanaka (approved):

PR #3649, opened 2020-12-02
Update to 110.0

hitz:


CHANGES.rst
```diff
@@ -1,5 +1,30 @@
 Changes
 =======
+109.0 (released) snovault 1.0.51
```
paul-sud: This should be 110 right?
```suggestion
110.0 (released) snovault 1.0.51
```

paul-sud (changes requested):
Think there is a small change needed to version in changelog.

CHANGES.rst
hitz: ya

paul-sud (approved):

PR #3653, opened 2020-12-07
ENCD-5655-page-editor-fix

paul-sud:


src/encoded/static/components/blocks/richtext.js
```diff
@@ -20,7 +20,7 @@ export default class RichTextBlockView extends React.Component {
     componentDidMount() {
         const $script = require('scriptjs');
         if (this.context.editable) {
-            $script('./node_modules/ckeditor4/ckeditor.js', this.setupEditor);
+            $script("ckeditor/ckeditor.js", this.setupEditor);
```
forresttanaka: Need single quotes to avoid ESLint errors.

src/encoded/static/components/blocks/richtext.js
```diff
@@ -46,9 +46,9 @@ export default class RichTextBlockView extends React.Component {
     }
 
     setupEditor() {
-        const ck = require('ckeditor4');
-        ck.disableAutoInline = true;
-        this.editor = ck.inline(this.domNode, {
+        require('ckeditor4');
+        CKEDITOR.disableAutoInline = true;
```
forresttanaka, line 51: To avoid the ESLint error for CKEDITOR being undefined, you have to disable the `no-undef` rule in this area. We’ve usually done it by putting `/* eslint-disable no-undef */` before the function/method, and `/* eslint-enable no-undef */` after. Sounds like ckeditor5 makes integrating with webpack better, but not sure if it changed its API too much to use here.

webpack.config.js
```diff
@@ -13,13 +14,27 @@ const PATHS = {
...
 const plugins = [];
 
 // To get auth0 v11 to build correctly:
 // https://github.com/felixge/node-formidable/issues/337#issuecomment-183388869
 plugins.push(new webpack.DefinePlugin({ 'global.GENTLY': false }));
+plugins.push(
+    new CopyPlugin({
+        patterns: [
+            {
+                // Adapted from https://github.com/ckeditor/ckeditor4-webpack-template/blob/master/webpack.config.js
```
forresttanaka, line 30: Thanks for the link; that helps a lot. I’m not very good at configuring webpack.

forresttanaka (changes requested):
Just a couple small changes to get ESLint to quiet down.

webpack.config.js
paul-sud: Me neither apparently, took a while mucking about with webpack to get it to work.

src/encoded/static/components/blocks/richtext.js
paul-sud: Added, thanks.

src/encoded/static/components/blocks/richtext.js
paul-sud: Ah yes. Had ESLint installed but it wasn't working for some reason. Fixed now.

forresttanaka (approved):

paul-sud:
Branch name is wrong, closing this PR.

PR #3655, opened 2020-12-08
ENCD-5647-add-capped-mRNA-to-nucleic-acid-term-name

iyoungworth:
added-dependency

minor-correction

### DIFF
--- a/src/encoded/upgrade/library.py
+++ b/src/encoded/upgrade/library.py
@@ -280,4 +280,4 @@ def library_14_15(value, system):
         if 'polyadenylated mRNA' in value['depleted_in_term_name']:
             if value['nucleic_acid_term_name'] == 'polyadenylated mRNA':
                 value['nucleic_acid_term_name'] = 'RNA'
-                value['notes'] = (notes + ' The nucleic_acid_term_name of this library was automatically upgraded by ENCD-5647.').strip()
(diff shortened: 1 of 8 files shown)


gabdank (changes requested):
the changes look good. please add a test for the introduced dependency.

src/encoded/upgrade/library.py
```diff
@@ -270,3 +270,14 @@ def library_13_14(value, system):
...
+
+@upgrade_step('library', '14', '15')
+def library_14_15(value, system):
+    # https://encodedcc.atlassian.net/browse/ENCD-5647
+    notes = value.get('notes', '')
+    if 'depleted_in_term_name' in value:
+        if 'polyadenylated mRNA' in value['depleted_in_term_name']:
+            if value['nucleic_acid_term_name'] == 'polyadenylated mRNA':
+                value['nucleic_acid_term_name'] = 'RNA'
+                value['notes'] = (notes + ' The nucleic_acid_term_name of this library was automatically upgraded by ENCD-5647.').strip()
```
gabdank: I think it would be better to give human-readable reason than refer to a system (Jira) that one day is not going to be there anymore.

gabdank (changes requested):

src/encoded/upgrade/library.py
iyoungworth: Sounds good, I adjusted to something more meaningful

gabdank (approved):

PR #3657, opened 2020-12-10
ENCD-5593-move-install-to-setup

paul-sud:


caseylitton (dismissed):

keenangraham:
I think this is going to require updating some stuff in devops. For sure the cluster deploy stuff e.g.: https://github.com/ENCODE-DCC/devops/blob/46f1923a7185d3fda7c9eae329693f93d436dc94/deployment/docs/cluster_deploy.sh#L35

keenangraham:
And deployment docs that still use `bin/batchupgrade` for example.

paul-sud:
@keenangraham see https://github.com/ENCODE-DCC/devops/pull/80 for devops PR.

README.md
```diff
@@ -19,28 +19,34 @@ For issues see [Snovault OSX App Installation][] first.
...
 
-1. Build Application
+3. Build Application
     ```
-    $ make clean && buildout bootstrap && bin/buildout
+    $ make clean && pip install -e '.[dev]' && buildout
     ```
 
-1. Run Application
+If you need to develop `snovault` side by side you can use the following commands instead, assuming `encoded` and `snovault` are present at the same level in your filesystem and virtual environment is activated.
```
ottojolanki, line 33: Does "at the same level" mean that I have  repos located like `/home/user/repos/encoded/` and `/home/user/repos/snovault/`?

README.md
```diff
@@ -50,6 +56,18 @@ For issues see [Snovault OSX App Installation][] first.
...
     ```
 
+You can also invoke `pytest` directly if you need more granular control over which Python tests to run.
+
+```bash
+# Make sure you are in your venv
+# Run a specific test in a specific file
+$ pytest TEST_FILE_PATH::TEST_NAME
+# Run tests with a name matching the expression
+$ pytest -k $EXPRESSION
```
ottojolanki: What does expression mean here? So if I would have `test_foo_1` and `test_foo_2` then `pytest -k test_foo_*` would run both of those?

cloud-config/template-parts/packages_all.yml
```diff
@@ -16,6 +16,7 @@
 - libssl-dev
 - libsqlite3-dev
 - python3-dev
+- python3-pip
```
ottojolanki, line 19: I'll add this to the AMIs.

ottojolanki (dismissed):
Couple of clarifications if you like. Looks good to me.

README.md
paul-sud: yes

README.md
paul-sud: It is based on substring matching + python operators. To run both you would use `-k test_foo`. However you can have expressions like `not test_bar`, `foo and bar`, etc. I agree it's not clear what expression is supposed to be, let me add a link to the pytest docs.

cloud-config/template-parts/packages_all.yml
paul-sud: Great, thanks.

ottojolanki (approved):
LGTM

PR #3659, opened 2020-12-11
SNO-185-log-connection-error

caseylitton:


gabdank (commented):
@caseylitton I am not sure how the changes in region_indexer.py and vis_indexer.py are related to the exception handling you added in snovault for SNO-185 ?

keenangraham (approved):
I think would be good to rebase.

PR #3664, opened 2021-01-06
ENCD-5652-AMI-based-deployment

ottojolanki:
Deploy from AMI fetched from S3. AMI building and sharing implemented in DEV-18.


caseylitton (dismissed):
This all makes sense, just removing what was placed in packer repo.  So long as no packages were missed.  Really just need to build and test the app and services.  There were some packages missing when updating to ubuntu18 but we didn't notice for six months.

src/encoded/commands/deploy.py
```diff
@@ -21,6 +21,19 @@
...
+    """Fetch the current ami-id from S3.
+    Args:
+        bucket: name of S3 bucket where ami-id is stored
+        key: location of the ami-id file within the bucket
+    Returns:
+        ami_id: string containing the contents current ami id S3 object.
+    """
+    s3 = boto3.resource("s3")
+    ami_id = s3.ObjectSummary(bucket_name=bucket, key=key).get()["Body"].read().decode().strip()
+    print(f"Fetched ami: {ami_id}")
```
caseylitton, line 34: Might wanna add the image id to the instance tags in the 'def _tag_ec2_instance' function

src/encoded/commands/deploy.py
```diff
@@ -789,54 +802,25 @@ def main():
...
         print("ssh ubuntu@{}{}".format(instance_info['public_dns'], tail_cmd))
         print('')
         helper_vars.append("datam='{}'".format(instance_info['instance_id']))
         for index in range(main_args.cluster_size):
-            str_index = str(index)
-            key_name = 'cluster_node_' + str_index
+            key_name = 'cluster_node_{}'.format(index)
             node_info = instances_info[key_name]
             helper_vars.append("data{}='{}'  # {}".format(index, node_info['instance_id'], key_name))
             if index == 0:
```
paul-sud, line 826: Found a stale reference to `main_args.build_ami` below this, probably should remove also?

https://github.com/ENCODE-DCC/encoded/blob/61f82f1daa4a86c25b7deecb99605a0291229706/src/encoded/commands/deploy.py#L827

paul-sud (changes requested):
I think there is a little more to clean up in the deploy.py, otherwise looks good to me.

src/encoded/commands/deploy.py
ottojolanki: `main_args.build_ami` means that after making the ES machine, it will be saved as an AMI. I don't think this is made any more stale by this PR. I agree that it is most likely not useful, but I would prefer to leave cleaning this up to the "official clean up ticket" that will follow after we confirm the that this way of deploying works.

paul-sud: Ok just seemed a little inconsistent with removing similar conditionals from the single node deployments, I'm fine with leaving it for a subsequent PR.

paul-sud (approved):

PR #3666, opened 2021-01-06
ENCD-4179-batchdownload-link-on-dataset-pages

zoldello:


src/encoded/static/components/app.js
```diff
@@ -946,7 +946,7 @@ class App extends React.Component {
     }
 
     handleBeforeUnload() {
-        if (this.state.unsavedChanges.length > 0 || cartIsUnsaved()) {
+        if (this.state.unsavedChanges.length > 0 || (cartIsUnsaved && cartIsUnsaved())) {
```
forresttanaka: You can just delete all references to `cartIsUnsaved` — it doesn’t exist anymore.

src/encoded/static/components/filegallery.js
```diff
@@ -1134,33 +1134,83 @@ VisualizationControls.defaultProps = {
...
         this.props.handleAssemblyAnnotationChange(e.target.value);
     }
 
+    setDownloadModalVisibility(isOpen) {
+        this.setState({ isBatchDownloadOpen: isOpen });
+    }
+
+    handleDownloadClick() {
+        const { context, filters } = this.props;
+        const accession = context.accession;
```
forresttanaka: This will need to use destructuring to work with ENCD-5661.

src/encoded/static/components/filegallery.js
```diff
@@ -1134,33 +1134,83 @@ VisualizationControls.defaultProps = {
...
+        const assemblies = filters.assembly && filters.assembly.length > 0 && filters.assembly[0] !== 'All assemblies' ?
+            `&files.assembly=${filters.assembly[0].replace(/\s+/g, '')}` :
```
forresttanaka: I think it’s better to just always use `encodedURIComponent` for the right-hand-side of all query-string elements, which takes care of all these issues in a standard way. This applies to `output_type`, `file_type`, etc.

src/encoded/static/components/filegallery.js
```diff
@@ -3006,6 +3065,7 @@ class FileGalleryRendererComponent extends React.Component {
                                     handleInclusionChange={this.handleInclusionChange}
                                     browserChangeHandler={this.handleBrowserChange}
                                     visualizeHandler={this.handleVisualize}
+                                    context={this.props.context}
```
forresttanaka: Not a real big deal, but `context` has already been defined through destructuring at the top of `render()`, so you can use that here instead of `this.props.context`. The same applies below.

src/encoded/static/scss/encoded/modules/_panels.scss
```diff
@@ -496,6 +496,31 @@ $file-gallery-controls-height: 34px; // Heit of controls inside file-gallery-con
             height: $file-gallery-controls-height;
             vertical-align: top;
         }
+
+        @media screen and (max-width: $screen-sm) {
```
forresttanaka: Need to reverse the logic and use `min-width` to align with all the other uses of mobile-first that we inherited from Bootstrap. Could use the more common `$screen-sm-min` as a reminder that these should always be `min-width`. Mixing `min-width` and `max-width` will make maintenance a nightmare.

src/encoded/static/components/filegallery.js
```diff
@@ -1187,6 +1237,14 @@ FilterControls.defaultProps = {
     currentBrowser: '',
 };
 
+FilterControls.propTypes = {
+    /** Object within which the attachment is to be displayed */
+    context: PropTypes.object.isRequired,
+}
```
forresttanaka: You should just add `context` to the existing `FilterControls.propTypes` object instead of overwriting the existing one with this one.

forresttanaka (changes requested):

src/encoded/static/components/filegallery.js
zoldello: Done

src/encoded/static/components/app.js
zoldello: Done

src/encoded/static/components/filegallery.js
zoldello: I used "encodedURIComponent" to make the links more secure. But I left the regex. I do need to remove white space; encodedURIComponent would just URI encode it which was not needed.

src/encoded/static/scss/encoded/modules/_panels.scss
zoldello: Thanks for this feedback. It was pretty helpful.

src/encoded/static/components/filegallery.js
zoldello: Done.

src/encoded/static/components/app.js
```diff
@@ -945,13 +943,6 @@ class App extends React.Component {
         return true;
     }
```
forresttanaka: We still need the `handleBeforeUnload` method here. The part we didn’t need was the call to `cartIsUnsaved`. Without `handleBeforeUnload`, if you reload the page after changing an object with the form, we don't get an alert letting the user back out of reloading.

src/encoded/static/components/filegallery.js
```diff
@@ -1134,37 +1135,89 @@ VisualizationControls.defaultProps = {
...
     handleAssemblyAnnotationChange(e) {
         this.props.handleAssemblyAnnotationChange(e.target.value);
     }
 
+    handleDownloadClick() {
+        const { context, filters } = this.props;
+        const { accession } = context;
+        const assemblies = filters.assembly && filters.assembly.length > 0 && filters.assembly[0] !== 'All assemblies' ?
+            `&files.assembly=${encodedURIComponent(filters.assembly[0].replace(/\s+/g, ''))}` :
+            '';
```
forresttanaka: This doesn’t work though. The `filters.assembly` combines `assembly` and `genome_annotation` which doesn’t match anything. By deleting the space with `filters.assembly[0]` === "GRCh38 V19" you get "GRCh38V19" which matches no assembly. You have to pass `files.assembly=GRCh38&files.genome_annotation=V19`. What would be great is if the individual selected assembly and genome annotation was available still split out so you can pass those two as separate query-string parameters, but if it isn't it'd be great if you could add those in somehow, in parallel with `filters`.

forresttanaka (changes requested):

src/encoded/static/components/filegallery.js
forresttanaka: Looks like `props.filterOptions` might give you the selected split assembly and annotation, as long as `filterOptions` keeps the same order as `filters`, which at a glance it looks like it does. The problem is knowing which `filterOptions` is currently selected. `props.selectedFilterValue` seems promising, but when I selected the second assembly/annotation combination, `selectedFilterValue` was "0" which doesn't help. Maybe that just needs fixing, or maybe it isn't the value I think it is.

zoldello: selectFilterValue applies to the filter by the buttons. I decided to use a different option.

forresttanaka (dismissed):

zoldello:
@forresttanaka This did not have v111 released tag. So I rebased. But there was a problems with it so I had to fix my new commit with the original commit you saw.

src/encoded/static/components/filegallery.js
```diff
@@ -1132,37 +1135,99 @@ VisualizationControls.defaultProps = {
...
+        if (filters.assembly && filters.assembly.length > 0 && filters.assembly[0] !== 'All assemblies') {
+            const analysis = filters.assembly[0].split(' ');
+            const assembly = analysis[0] ? `&files.assembly=${encodedURIComponent(analysis[0])}` : '';
+            const genomeAnnotation = analysis[1] ? `&files.genome_annotation=${encodedURIComponent(analysis[1])}` : '';
+            assemblies = `${[assembly, genomeAnnotation].filter(a => a !== '').join('')}`;
+        }
+        const fileTypes = filters.file_type && filters.file_type.length > 0 ?
+            filters.file_type.map(fileType => `&files.file_type=${encodedURIComponent(fileType)}`).join('') :
+            '';
+        const biologicalReplicates = filters.biological_replicates && filters.biological_replicates.length > 0 ?
```
forresttanaka, line 1178: Just FYI because I checked this just a few days ago, AirBnB [seems to recommend](https://github.com/airbnb/javascript#comparison--nested-ternaries) the question mark and colon in ternary expressions at the beginning of the line. ESLint doesn’t check this though.

forresttanaka (dismissed):

zoldello:
@forresttanaka ~~If reviewing this, please hold on. I have a new requirement.~~. Added the new request

src/encoded/static/components/filegallery.js
```diff
@@ -1195,7 +1195,9 @@ class FilterControls extends React.Component {
 
     render() {
         const { filterOptions, selectedFilterValue, browsers, currentBrowser, browserChangeHandler, visualizeHandler, context } = this.props;
-        const contextFiles = context.files || [];
+        const filterIncArchive = document.querySelector('[name="filterIncArchive"]');
+        const filterIncArchiveStatus = filterIncArchive ? filterIncArchive.checked : false;
```
forresttanaka: I feel like instead of querying the DOM, you should pass the FileGalleryRendererComponent `inclusionOn` state as a prop to `FilterControls`. Seems more “Reactic,” and also when `inclusionOn` changes state, `FilterControls` gets notified of this. Same thing for `handleDownloadClick`.

forresttanaka (changes requested):

src/encoded/static/components/filegallery.js
zoldello: Done. Good idea.

forresttanaka (dismissed):

zoldello:
@forresttanaka I apologize for this. There is a fix for the download file I added.

forresttanaka (approved):
Looks like it solves the problem with deprecated files.

PR #3668, opened 2021-01-07
ENCD-5623 add section breaks to glossary

emro:
Updates:
- Glossary entries can now include lists of file formats and additional information (links) in addition to their definitions
- Definitions are grouped based on a "section" property (if there is no "section" property they are grouped under "General terms") and are alphabetized within those groups
- Section links appear as headers and are displayed at the top so you can scroll to any particular section quickly
- Navigation terms are capitalized in the navigation but not in the glossary so there was a small fix for that

### DIFF
--- a/src/encoded/static/components/glossary.js
+++ b/src/encoded/static/components/glossary.js
@@ -1,24 +1,77 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import _ from 'underscore';
 import * as globals from './globals';
+
+const GlossaryTerm = (props) => {
+    const term = props.glossaryEntry.term;
+    const definition = props.glossaryEntry.definition;
+    return (
+        <div className="glossary-element">
+            <h4 className="glossary-term">{term}</h4>
+            <p className="glossary-definition">{definition}</p>
+            {props.glossaryEntry.file_formats ?
+                <p className="glossary-extra-info">
+                    <span className="glossary-label">File formats: </span>
+                    <span className="glossary-field">{props.glossaryEntry.file_formats}</span>
+                </p>
+            : null}
+            {props.glossaryEntry.additional_information ?
+                <p className="glossary-extra-info">
+                    <span className="glossary-label">Additional information: </span>
+                    {props.glossaryEntry.additional_information.map(href =>
+                        <span className="glossary-field" key={href.url}>
+                            <a href={href.url}>{href.title}</a>
+                        </span>
+                    )}
+                </p>
+            : null}
+        </div>
+    );
+};
+
+GlossaryTerm.propTypes = {
+    glossaryEntry: PropTypes.object.isRequired,
+};
 
 const Glossary = (props) => {
     const { context } = props;
     const glossary = require('./glossary/glossary.json');
+    const glossaryBySection = _(glossary).groupBy(entry => entry.section || 'General terms');
 
     return (
         <div className="layout">
             <div className="layout__block layout__block--100">
-                <div className="ricktextblock block" data-pos="0,0,0">
+                <div className="richtextblock block" data-pos="0,0,0">
                     <center>
                         <h1>{context.title}</h1>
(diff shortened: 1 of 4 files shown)


src/encoded/static/components/glossary.js
```diff
@@ -1,24 +1,77 @@
...
 const Glossary = (props) => {
     const { context } = props;
     const glossary = require('./glossary/glossary.json');
+    const glossaryBySection = _(glossary).groupBy(entry => entry.section || 'General terms');
 
     return (
         <div className="layout">
             <div className="layout__block layout__block--100">
-                <div className="ricktextblock block" data-pos="0,0,0">
+                <div className="richtextblock block" data-pos="0,0,0">
```
zoldello, line 45: Is there a particular reason why "richtextblock" has no dashes to separate word in the name? CSS community typically use dash to separate words in CSS style names like "file-gallery-tab-bar"

src/encoded/static/components/glossary.js
```diff
@@ -1,24 +1,77 @@
...
+                                {(sectionIdx < Object.keys(glossaryBySection).length - 1) ?
+                                    <span> | </span>
+                                : null}
+                            </React.Fragment>
+                        )}
+                    </h4>
+                    {Object.keys(glossaryBySection).map(section =>
+                        <div className="glossary-block" key={section} id={section}>
+                            <h2 id={`${section.toLowerCase().split(' ').join('-')}`}>{section}</h2>
+                            {(section === undefined) ?
```
zoldello, line 62: In JavaScript, checking against undefined is a code smell. Is there a particular reason you are not testing to see if section is falsey instead?

When I see a variable set as "undefined"; that to me means "the variable has never been set to a value nor null." "section" refers to a key of an array; it is a code smell for that key to ever be undefined or even falsey.

src/encoded/static/components/glossary.js
```diff
@@ -1,24 +1,77 @@
...
+                                <h4>{section}</h4>
+                            : null}
+                            {glossaryBySection[section]
+                                .sort((a, b) => {
```
zoldello, line 66: I tend to use single-letter variable names like 'a', 'b' and 'i' in arrow functions.  Still, there is a strong argument against it as it makes the code a bit harder to understand.

src/encoded/static/components/glossary.js
```diff
@@ -1,24 +1,77 @@
...
+                                    if (a.term.toLowerCase() < b.term.toLowerCase()) { return -1; }
+                                    if (a.term.toLowerCase() > b.term.toLowerCase()) { return 1; }
+                                    return 0;
```
zoldello, line 69: This is a good idea. I don't think IE-11 supports it but that is no big deal. I am forward to August when all we need to worry about are evergreen browsers.

src/encoded/static/scss/encoded/modules/_faq_glossary.scss
```diff
@@ -38,7 +38,62 @@
...
+    /* sprites.less reset */
+    display: inline;
+    line-height: normal;
+    vertical-align: baseline;
+    background-image: none;
+    background-position: 0% 0%;
+    background-repeat: repeat;
+}
+
+.block.richtextblock .glossary-element {
```
zoldello, line 64: richtextblock? No dash?

zoldello (approved):
Changes look good. My feedback is cosmetic.

src/encoded/static/components/glossary.js
emro: This is the class name used on all the help pages so I'd like to leave it as-is though I agree with you!

forresttanaka: I would leave it as is. David tended to break a lot of JS and CSS practices, and the page stuff was originally all his years ago. Some page content might use CSS styles relying on richtextblock.

PR #3671, opened 2021-01-08
ENCD-5665-tie-eslint-to-circleci

zoldello:


.circleci/config.yml
```diff
@@ -141,6 +141,21 @@ jobs:
             cp junit.xml test-results/npm
       - store_test_results:
           path: test-results
+  eslint:
+    executor: encoded-executor
+    resource_class: medium
+    steps:
+      - buildout
```
paul-sud: For speed, maybe it is possible to get away with not running full buildout. Maybe just `npm` install would be enough. Even more minimal might be to skip npm install entirely, assuming `eslint` doesn't require access to dependencies.

.circleci/config.yml
```diff
@@ -141,6 +141,21 @@ jobs:
...
+      - run:
+          name: Install eslint
+          command: sudo npm install eslint
+      - run:
+          name: Lint the code with Eslint
+          command: |
+            mkdir -p test-results/eslint
+            node_modules/.bin/eslint src/encoded/static/components/
+      - store_test_results:
+          path: test-results
```
paul-sud: You don't need these two lines, since I don't think eslint is generating any results files.

.circleci/config.yml
```diff
@@ -141,6 +141,21 @@ jobs:
...
+    resource_class: medium
+    steps:
+      - buildout
+      - run:
+          name: Install eslint
+          command: sudo npm install eslint
+      - run:
+          name: Lint the code with Eslint
+          command: |
+            mkdir -p test-results/eslint
```
paul-sud: I don't think you need this line, it doesn't look like eslint has any results to report.

paul-sud (changes requested):
I think there is some unneeded config, and also a potential to speed up the linting.

.circleci/config.yml
zoldello: won't doing "npm install" be slower as it install all packages and not just want is needed?

paul-sud: buildout installs the python stuff, then runs `npm install`, among other things. So I'd expect that `npm install` on its own would be faster than full buildout.

zoldello: Done

.circleci/config.yml
zoldello: removed

.circleci/config.yml
zoldello: removed

.circleci/config.yml
```diff
@@ -141,6 +141,17 @@ jobs:
             cp junit.xml test-results/npm
       - store_test_results:
           path: test-results
+  eslint:
+    executor: encoded-executor
+    resource_class: medium
+    steps:
+      -  buildout
```
zoldello, line 148: I tried an image. I got issues with python needing to be installed. So, to keep things simple, I used "buildout". There does no seems to considerable lag

zoldello, line 148: @paul-sud  I tried an image. I got issues with python needing to be installed. So, to keep things simple, I used "buildout". There does no seems to considerable lag

paul-sud: Ok if you're fine with it then it's fine by me

paul-sud (dismissed):
LGTM

zoldello:
Note- The build will fail until the code base is rebased with ENCD-5661.

zoldello:
@paul-sud  After talking with Ben, I fixed ESlint errors rather than wait on ENCD-5661.

paul-sud (dismissed):
JS changes look fine.

paul-sud (approved):
LGTM

PR #3672, opened 2021-01-08
ENCD-5644-footprints-audit

yunhailuo:


src/encoded/audit/analysis.py
```diff
@@ -0,0 +1,61 @@
...
+        )
+        yield AuditFailure('missing footprints', detail, level='WARNING')
+        return
+
+    # Assume at least one qm['footprint_count'] exists?
+    for rep in zero_footprints_reps:
+        detail = f'Replicate {rep} has no significant footprints detected.'
+        yield AuditFailure(
+            'no significant footprints', detail, level='WARNING'
+        )
```
gabdank: would not the "missing footprints" be an ERROR level audit? the "no significant footprints" can remain WARNING. If I am correct on this assumption, I would suggest changing the first to be "ERROR", and would suggest having both Audits to be under the same category "missing footprints" (we don't gain much by adding two categories: "no significant footprints, missing footprints")

gabdank (approved):
looks good, re-running bdd

PR #3673, opened 2021-01-12
ENCD-5663-login-dropdown-misaligned

zoldello:


forresttanaka (approved):
Could also target the #useracxtions id for the CSS class, allowing this menu to not be the last in some future case, but I doubt that would ever happen.

PR #3676, opened 2021-01-12
ENCD-5615 Change cart search for more efficiency

forresttanaka:
The main goals of this ticket included changing the “Dataset search” button on the cart-view page to not build a big URL containing each `@id` of the datasets in the cart, but to instead use the new /cart-search/, /cart-matrix/, and /cart-report/ endpoints. Of these, this ticket uses /cart-report/ only and renames that button “Dataset report” on the cart-view page because it now leads to a report page instead of a search-list page.

The other main goal limited the size of carts to 8000 items to go with the /cart-report/ limit of returning a maximum of 8000 items. This brought up the old issue with the modal alerting the user that they have attempted to add more than the allowed number of items, and the modal not appearing if you navigate away from the page before the cart code could determine this. I solved this by making a new `<CartAlert />` component owned by `<App>` that can display any cart-related alert in a component (`<App>`) that doesn't get dismounted when you navigate. So if you attempt to add a lot of datasets to the cart and navigate to a new page before the cart code is done figuring out if it should display an alert, you still get the alert regardless of what page you navigated to.

I ran out of room in the header of the cart-view page panel, so I moved the Merge, Lock, and Clear buttons out of the panel header and put it just below the title, in the `<CartAccessories>` component.

### DIFF
--- a/src/encoded/static/components/cart/add_multiple.js
+++ b/src/encoded/static/components/cart/add_multiple.js
@@ -9,19 +9,30 @@ import url from 'url';
 import QueryString from '../../libs/query_string';
 import { atIdToType, truncateString } from '../globals';
 import { requestSearch } from '../objectutils';
-import { addMultipleToCartAndSave, cartOperationInProgress } from './actions';
+import { addMultipleToCartAndSave, cartOperationInProgress, triggerAlert } from './actions';
 import CartLoggedOutWarning, { useLoggedOutWarning } from './loggedout_warning';
+import CartMaxElementsWarning from './max_elements_warning';
 import {
     cartGetAllowedTypes,
     cartGetAllowedObjectPathTypes,
+    CART_MAX_ELEMENTS,
+    mergeCarts,
 } from './util';
 
 /**
  * Button to add all qualifying elements to the user's cart.
  */
-const CartAddAllSearchComponent = ({ savedCartObj, searchResults, inProgress, addAllResults, setInProgress, loggedIn }) => {
+const CartAddAllSearchComponent = ({
+    savedCartObj,
+    searchResults,
+    inProgress,
+    addAllResults,
+    setInProgress,
+    loggedIn,
+    showMaxElementsWarning,
+}) => {
     /** Get hooks for the logged-out warning modal */
-    const [warningStates, warningActions] = useLoggedOutWarning(false);
+    const [loggedOutWarningStates, loggedOutWarningActions] = useLoggedOutWarning(false);
 
     /**
      * Handle a click in the Add All button by doing a search of elements allowed in carts to get
@@ -46,16 +57,21 @@ const CartAddAllSearchComponent = ({ savedCartObj, searchResults, inProgress, ad
 
                     // Get all elements from results that qualify to exist in carts.
                     const elementsForCart = results['@graph'].filter(result => allowedTypes.includes(result['@type'][0])).map(result => result['@id']);
-                    if (elementsForCart.length > 0) {
-                        // We should always have elements qualified to exist in a cart because we
-                        // wouldn't have shown the "Add all items to cart" button if we didn't know we
-                        // had qualified elements in the search results, but just in case.
+
+                    // Check whether the final cart would have more elements than allowed by doing a
+                    // trial merge. If the merged cart fits under the limit, add the new elements
+                    // to the cart.
+                    const mergedElements = mergeCarts(savedCartObj.elements, elementsForCart);
+                    if (mergedElements.length > CART_MAX_ELEMENTS) {
+                        showMaxElementsWarning();
+                    } else if (elementsForCart.length > 0) {
                         addAllResults(elementsForCart);
                     }
                 }
             });
         } else {
-            warningActions.setIsWarningVisible(true);
+            // The user hasn't logged in, so show a modal that allows them to.
+            loggedOutWarningActions.setIsWarningVisible(true);
         }
     };
 
@@ -68,11 +84,11 @@ const CartAddAllSearchComponent = ({ savedCartObj, searchResults, inProgress, ad
                     disabled={inProgress || savedCartObj.locked}
                     className="btn btn-info btn-sm"
                     onClick={handleClick}
-                    title={`Add all experiments in search results to cart${cartName ? `: ${cartName}` : ''}`}
+                    title={`Add all datasets in search results to cart${cartName ? `: ${cartName}` : ''}`}
                 >
                     Add all items to cart
                 </button>
-                {warningStates.isWarningVisible ? <CartLoggedOutWarning closeModalHandler={warningActions.handleCloseWarning} /> : null}
+                {loggedOutWarningStates.isWarningVisible ? <CartLoggedOutWarning closeModalHandler={loggedOutWarningActions.handleCloseWarning} /> : null}
             </div>
         );
     }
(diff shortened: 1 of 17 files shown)


src/encoded/static/components/cart/add_multiple.js
```diff
@@ -46,16 +57,21 @@ const CartAddAllSearchComponent = ({ savedCartObj, searchResults, inProgress, ad
 
                     // Get all elements from results that qualify to exist in carts.
                     const elementsForCart = results['@graph'].filter(result => allowedTypes.includes(result['@type'][0])).map(result => result['@id']);
-                    if (elementsForCart.length > 0) {
-                        // We should always have elements qualified to exist in a cart because we
-                        // wouldn't have shown the "Add all items to cart" button if we didn't know we
-                        // had qualified elements in the search results, but just in case.
+
+                    // Check whether the final cart would have more elements than allowed by doing a
```
zoldello, line 61: should be "will" rather than "would"

src/encoded/static/components/cart/add_multiple.js
```diff
@@ -46,16 +57,21 @@ const CartAddAllSearchComponent = ({ savedCartObj, searchResults, inProgress, ad
...
+                    // trial merge. If the merged cart fits under the limit, add the new elements
```
zoldello, line 62: I typically see in software where an action is attempted. If it succeeds, great. If it fails, the code does what is necessary. I rarely don't see "trials"

src/encoded/static/components/cart/cart.js
```diff
@@ -842,66 +841,57 @@ FileFacets.defaultProps = {
...
-                        <ModalFooter
-                            closeModal={handleCloseClick}
-                        />
-                    </Modal>
-                : null}
-            </React.Fragment>
+            <DropdownButton.Immediate
+                label={<>Dataset report {svgIcon('chevronDown')}</>}
+                id="cart-dataset-report"
+                css="cart-dataset-report"
```
zoldello, line 869: Maybe called "cssClassName" (I have seen people do that in .NET)? I know this is a class as I searched the tool. But it would nicer if that need not be done.

src/encoded/static/components/cart/index.js
```diff
@@ -99,6 +108,8 @@ const cartModule = (state, action = { type: NO_ACTION }) => {
             });
         case CART_OPERATION_IN_PROGRESS:
             return Object.assign({}, state, { inProgress: action.inProgress });
+        case DISPLAY_ALERT:
+            return Object.assign({}, state, { alert: action.alert });
```
zoldello, line 112: Does this violate the rule of ENCD-5661 that says:

"You must clone objects with a spread operator only instead of using Object.assign, e.g. {...object}. You must clone arrays using spread operators also, e.g. [...array]"

or did I misread it?

src/encoded/static/components/cart/util.js
```diff
@@ -67,11 +76,34 @@ export const isAllowedElementsPossible = (resultFilters) => {
 
 /**
  * Return a new array containing the merged contents of two carts with no duplicates. Contains odd
- * ES6 syntax to merge, clone, and dedupe arrays in one operation.
+ * ES6 syntax to merge, clone, and de-dupe arrays in one operation.
```
zoldello, line 79: or maybe use "remove duplicate entries"?

src/encoded/static/components/cart/util.js
```diff
@@ -67,11 +76,34 @@ export const isAllowedElementsPossible = (resultFilters) => {
...
+
+
+/**
+ * Determine whether the given search results is for a cart search or not.
+ * @param {object} context Search-results object
+ *
+ * @return {boolean} True if search results are for a cart search.
+ */
+const cartSearchPaths = ['/cart-search/', '/cart-matrix/', '/cart-report/'];
+export const getIsCartSearch = (context) => {
```
zoldello, line 96: should this method be called "IsCartSearchable"? In theory you can even return "cartSearchPaths.includes" if you are into functional programming.

Do you need a null check to do defensive programming? This method is exported and there is no telling who passes what to it in the future.

src/encoded/static/components/facets/supressed.js
```diff
@@ -0,0 +1,9 @@
+import FacetRegistry from './registry';
+
+
+/**
+ * Used for all facets that need suppression unconditionally.
+ */
+const SuppressedFacet = () => null;
```
zoldello, line 7: This looks like a code smell. There is not a lot of utility in just returning null; or maybe I am missing the bigger picture.

src/encoded/static/components/globals.js
```diff
@@ -44,6 +44,13 @@ documentViews.preview = new Registry();
 documentViews.file = new Registry();
 documentViews.detail = new Registry();
 
+// UNICODE entity codes, needed in JSX string templates. Each property named after the equivalent
+// HTML entity. Add new entries to this object as needed.
+export const uc = {
```
zoldello, line 49: would it be better to call this "unicode" or something else more descriptive. 

Where is "uc" used? There are only two references and no involves extracting data from "uc"

src/encoded/static/scss/encoded/modules/_cart.scss
```diff
@@ -1267,3 +1260,41 @@ $file-type-colors: (
...
     }
 }
+
+// Contents of the type-selection modal for /cart-search/ page, Report view control button.
+.cart-type-selector {
+    display: flex;
+    justify-content: center;
+
+    .btn {
+        margin-left: 2.5px;
```
zoldello, line 1270: I have not seen someone set both left and right margin before.

src/encoded/static/scss/encoded/modules/_cart.scss
```diff
@@ -1267,3 +1260,41 @@ $file-type-colors: (
         width: auto;
     }
 }
+
+// Contents of the type-selection modal for /cart-search/ page, Report view control button.
+.cart-type-selector {
```
zoldello, line 1265: Where is "cart-type-selector"  used? I search for it and only saw one reference in this css file.

zoldello (approved):
The code looks good. I saw no serious problem that warrants a "Request Changes". My comments may or may not be important. I recommend you look at the, at least the ones about functionality I saw declared but not used.

src/encoded/static/components/cart/index.js
forresttanaka: The current version of Babel doesn't allow copying objects with spread operators. The new version does and the new version of AirBnB requires it.

src/encoded/static/components/cart/util.js
forresttanaka: Calling the method `isCartSearchable` makes it impossible to call its return value variable `isCartSearchable`.

src/encoded/static/components/facets/supressed.js
forresttanaka: That's how you prevent a facet from appearing under all circumstances. The new /cart-<something/ endpoints return a cart facet that we never want to see.

src/encoded/static/components/globals.js
forresttanaka: This needs to be used inline, embedded with a string, and spelling it out could generate complaints. It's used in line 59 of toggle.js.

src/encoded/static/scss/encoded/modules/_cart.scss
forresttanaka: Yeah, I think I stopped using it at some point.

PR #3677, opened 2021-01-12
ENCD-5661 Upgrade JS Packages

forresttanaka:
The vast majority of this will just be looking at rote conversions, but a few more complex changes exist, mostly involving new rules about React state with the old `setState` mechanism.

Since this branch won't be merged until the next sprint, I'll need to do probably multiple rebases to origin/dev after code review, so I think just before merge I'll do the last rebase and ask for code review again. The changes between them shouldn't be very large.


forresttanaka:
Though I brought back `import React from 'React';` you won’t see this in every file that includes JSX. You used to need this import any time any JSX appeared in the file even if you never used the `React` identifier in that file. Now you only need to import React if you actually use the `React` identifier in that file. If you import React without using the `React` identifier in that file, you get an error that React doesn’t get used in that file.

src/encoded/static/components/audit.js
```diff
@@ -218,26 +218,26 @@ class AuditGroup extends React.Component {
...
+                        <button type="button" onClick={this.detailSwitch} className="collapsing-title">
                             {collapseIcon(!detailOpen)}
                         </button>
                     </div>
                     <div className="audit-detail__info">
                         <i className={iconClass} />
                         <strong>&nbsp;{categoryName}</strong>
                         <div className="btn-info-audit">
-                            <a href={`/data-standards/audits/#${categoryName.toLowerCase().split(' ').join('_')}`} title={`View description of ${categoryName} in a new tab`} rel="noopener noreferrer" target="_blank"><i className="icon icon-info-circle" /></a>
+                            <a href={`/data-standards/audits/#${categoryName.toLowerCase().split(' ').join('_')}`} rel="noopener noreferrer" target="_blank"><span className="sr-only">{`View description of ${categoryName} in a new tab`}</span><i className="icon icon-info-circle" /></a>
```
emro, line 229: Which rule was it that led to this change?

src/encoded/static/components/award.js
```diff
@@ -1780,7 +1798,8 @@ export const ExperimentDate = (props) => {
         const sortedTerms = standardTerms.sort((termA, termB) => {
             if (termA.key < termB.key) {
                 return -1;
-            } else if (termB.key < termA.key) {
+            }
+            if (termB.key < termA.key) {
```
emro, line 1802: It is going to take a while to get used to not using "else if" statements!

src/encoded/static/components/genome_browser.js
```diff
@@ -57,10 +57,8 @@ const getDefaultCoordinates = (assemblyAnnotation, ignoreCache = false) => {
 
     if (gVState) {
         const savedState = gVState ? JSON.parse(gVState) : {};
-        contig = savedState.contig;
-        x0 = savedState.x0;
-        x1 = savedState.x1;
-        pinnedFiles = savedState.pinnedFiles;
+        ({ contig } = savedState);
+        ({ x0, x1, pinnedFiles } = savedState);
```
emro, line 61: Why are these parentheses here? I saw this in several places and I'm sure it's necessary but it seems odd to me and don't remember seeing it in the list of new conventions on the ticket.

src/encoded/static/libs/ui/tooltip.js
```diff
@@ -109,7 +109,7 @@ const Tooltip = (props) => {
     React.useEffect(() => {
         // Determine screen width and set mobile parameter and tooltip position
         const updateWidth = () => {
-            const screenWidth = Math.min(screen.width, window.innerWidth);
+            const screenWidth = Math.min(window.screen.width, window.innerWidth);
```
emro, line 112: See other comment about this.

src/encoded/static/libs/ui/button.js
```diff
@@ -291,10 +292,11 @@ export const Selected = ({ labels, execute, id, triggerVoice, css, inline, disab
     return (
         <div className={`dropdown-button${css ? ` ${css}` : ''}`} style={inline ? { display: 'inline-flex' } : null}>
             <div className="dropdown-button__composite" onMouseEnter={actions.handleMouseEnter} onMouseLeave={actions.houseMouseLeave}>
-                <button className="dropdown-button__composite-execute" onClick={handleExecute} onKeyUp={actions.handleKey} disabled={disabled}>
+                <button type="button" className="dropdown-button__composite-execute" onClick={handleExecute} onKeyUp={actions.handleKey} disabled={disabled}>
```
emro: I would probably have put all these attributes on separate lines for readability.

src/encoded/static/libs/svg-icons.js
```diff
@@ -142,15 +139,15 @@ const icons = {
...
+ * Render an SVG icon specified by `icon` which must match a property of the `icons` variable above.
  * Can optionally add css styles as a React style object to the SVG element.
  * @param {string} icon Specifies which icon to show; property of `icons`
  * @param {object} style React CSS styles (not classes) to add to svg
  */
 export const svgIcon = (icon, style) => icons[icon](style);
 
 // Render the icon used to collapse a panel from the title bar.
-//   collapsed: T if the icon should be rendered for the collapsed steate
+//   collapsed: T if the icon should be rendered for the collapsed state
```
emro, line 150: So many typos corrected in this ticket! I usually think that I catch them in code review but I guess a bunch get through periodically.

src/encoded/static/components/search.js
```diff
@@ -76,7 +76,7 @@ const datasetTypes = {
     GeneSilencingSeries: types.gene_silencing_series.title,
 };
 
-const getUniqueTreatments = treatments => _.uniq(treatments.map(treatment => singleTreatment(treatment)));
+const getUniqueTreatments = (treatments) => _.uniq(treatments.map((treatment) => singleTreatment(treatment)));
```
emro, line 79: There will likely be some merge conflicts in ENCD-5606 if this is merged first, which it probably will be.

src/encoded/static/components/app.js
```diff
@@ -501,119 +506,7 @@ class App extends React.Component {
...
-            if (!response.ok) {
-                throw new Error('Failed to create new account');
-            }
-            this.setState({ accountCreatedModalVisibility: true }); // tell user account was created
-            this.handleAuth0Login(authResult, false, false); // sign in after account creation
-        }).catch(() => {
-            this.setState({ accountCreationFailedVisibility: true });
-        });
-    }
+    /* eslint-enable react/no-did-update-set-state */
```
emro, line 509: I guess once we will be able to drop all these disable linting comments when everything is using React Hooks?

src/encoded/static/components/facets/defaults.js
```diff
@@ -231,6 +231,43 @@ export class DefaultDateSelectorFacet extends React.Component {
         });
     }
 
+    // Reset the dropdowns and state, and clear query
+    handleReset(resetString) {
+        this.setState({ activeFacet: 'date_released' }, () => {
+            this.setActiveFacetParameters();
+
+            // * Strip trailing & for the ENCD-4803 branch because it keeps the training ampersand.
```
emro: I know this was just moved up as-is but the `*` seems unnecessary to me.

src/encoded/static/components/filegallery.js
```diff
@@ -496,6 +495,11 @@ class RawSequencingTable extends React.Component {
         this.handleCollapse = this.handleCollapse.bind(this);
     }
 
+    handleCollapse() {
+        // Handle a click on a collapse button by toggling the corresponding tableCollapse state var
+        this.setState((state) => ({ collapsed: !state.collapsed }));
+    }
+
```
emro, line 501: Why is this rearranged? There were several places where functions were copy-pasted into a new order. Was there a new rule that meant that we needed these rearrangements?

src/encoded/static/components/genome_browser.js
```diff
@@ -596,15 +647,11 @@ class GenomeBrowser extends React.Component {
...
-        let x0 = this.state.x0;
-        let x1 = this.state.x1;
+        let { contig, x0, x1 } = this.state;
         if (this.chartdisplay) {
             const coordinates = readGenomeBrowserLabelCoordinates();
 
-            x0 = coordinates.x0;
-            x1 = coordinates.x1;
-            contig = coordinates.contig;
+            ({ x0, x1, contig } = coordinates.x0);
```
emro: I think it should just be `coordinates`.

src/encoded/static/components/graph.js
```diff
@@ -692,18 +693,18 @@ export class Graph extends React.Component {
...
                                 <tr>
-                                    <td className="zoom-indicator"><button onClick={() => this.changeZoom(-6)}><i className="icon icon-minus" /></button></td>
+                                    <td className="zoom-indicator"><button type="button" onClick={() => this.changeZoom(-6)}><span className="sr-only">Zoom out</span><i className="icon icon-minus" /></button></td>
                                     <td className="zomm-controller"><input type="range" className="zoom-slider" ref={this.slider} min={minZoom} max={maxZoom} value={this.state.zoomLevel === null ? 0 : this.state.zoomLevel} onChange={this.rangeChange} onInput={this.rangeChange} onDoubleClick={this.rangeDoubleClick} onMouseUp={this.rangeMouseUp} onMouseDown={this.rangeMouseDown} /></td>
-                                    <td className="zoom-indicator"><button onClick={() => this.changeZoom(6)}><i className="icon icon-plus" /></button></td>
+                                    <td className="zoom-indicator"><button type="button" onClick={() => this.changeZoom(6)}><span className="sr-only">Zoom in</span><i className="icon icon-plus" /></button></td>
```
emro, line 698: This is nit-picky but I didn't know what "sr-only" meant and had to look it up. Evidently it's a class name borrowed from Bootstrap, short for "screen reader only". It's fine but I might change it.

src/encoded/static/components/matrix_mouse_development.js
```diff
@@ -646,22 +648,51 @@ class MatrixPresentation extends React.Component {
...
     getQueryValues(key) {
         return this.query.getKeyValues(key);
     }
 
+    /* eslint-disable no-restricted-globals */
     updateWindowWidth() {
         this.setState({
             windowWidth: Math.min(screen.width, window.innerWidth),
         });
     }
```
emro, line 693: Here we are using `screen.width` but somewhere else you changed it to `window.screen.width`.

src/encoded/static/components/matrix_mouse_development.js
```diff
@@ -692,63 +723,46 @@ class MatrixPresentation extends React.Component {
...
     selectMouseDevelopmentStage(e) {
         const targetId = e.target.id.replace(/-/g, ' ');
         // selected something that was already selected so just de-select
         if (this.state.developmentStageClick.indexOf(targetId) > -1) {
-            const newSelections = this.state.developmentStageClick.filter(stage => stage !== targetId);
-            this.setState({ developmentStageClick: newSelections });
+            this.setState((state) => {
+                const newSelections = state.developmentStageClick.filter((stage) => stage !== targetId);
+                return { developmentStageClick: newSelections };
+            });
```
emro, line 732: I like all the changes like this a lot. It's much more intuitive.

emro (commented):
Well this was a monumental ticket! So many changes. I appreciate all the careful attention to detail that went into this update.
Everything looks great. I think I possibly found 2 things that should be fixed in thousands and thousands of lines of changes.

src/encoded/static/components/audit.js
forresttanaka: The `jsx-a11y/control-has-associated-label` rule dinged this one. The link only surrounds an icon, so a visually impaired person wouldn't know what the icon does. The `title` property doesn't help satisfy this rule, and in fact most a11y guides I've seen discourage the `title` attribute. Sometimes encoded specs require it, so we'll still have some of those.

src/encoded/static/components/genome_browser.js
forresttanaka: That's kind of an odd syntax rule, but if you use spread assignments for already-declared variables, you have to surround the expression with parentheses. If you declare it at the assignment (e.g. `const { contain } = savedState`) then you don't need (and probably can't have) the parens.

src/encoded/static/libs/ui/button.js
forresttanaka: Yeah, sometimes hard to decide at what point it's too much. We have far wider ones than this, but at some point they should get put onto separate lines.

src/encoded/static/components/app.js
forresttanaka: Probably, though in cases like this we might need to do pretty much the same refactoring as with class components.

src/encoded/static/components/filegallery.js
forresttanaka: The [`react/no-access-state-in-setstate`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-access-state-in-setstate.md) rule doesn't allow the simpler form. And this ESLint version is _very_ strict, and sometimes I would swear we weren’t breaking this rule, but it caught a very subtle indirect violation.

src/encoded/static/components/genome_browser.js
forresttanaka: Oh yeah. I missed a couple other ones like that that I already corrected, but missed this one.

src/encoded/static/components/graph.js
forresttanaka: I've used it for so long, I've just gotten used to it.

src/encoded/static/components/matrix_mouse_development.js
forresttanaka: I probably changed my mind on the best way to do this part way through. I’ll change this to use window.screen.width.

forresttanaka:
I’ll have to back out the change to config.yml before merge.

emro (dismissed):

forresttanaka:
Rebased on dev and returned old code in filegallery.js to handle raw sequencing files. For some reason, ESLint does not complain about modifying the file object this time. But I tested this update on https://encd-5661-upgrade-js-pkg-qa1-fytanaka.demo.encodedcc.org/experiments/ENCSR749SXH/ and it returns to properly pairing paired files.

emro (commented):
Since all the changes were collapsed into one commit - maybe because reverting helped you discover what the issue was with the tables? - I can't tell what was changed between my first code review and this one. GitHub told me that five files were changed since my first review - "button.js", "matrix_mouse_dev.js", "genome_browser.js", "filegallery.js", and "defaults.js". In your notes you only mention "filegallery.js" changing. In any case I looked at the five files all and they all look good. 

It looks like there are merge conflicts which should probably be fixed.

forresttanaka:
Just the Raw Sequencing table code in filegallery.js changed. The rest is from merging and I didn't touch those.

emro (approved):

PR #3682, opened 2021-01-21
ENCD-5672-update-schema-columns-amount

bonitalam:


gabdank (changes requested):
I think we should update the titles as well, otherwise, we can have a concentration of Kg or something similar.

gabdank (approved):

PR #3685, opened 2021-01-25
ENCD-5606 update series searches

emro:
Adds series specific information in the search list results and search results.
Detailed changes are described in slides linked on the Jira ticket.

### DIFF
--- a/src/encoded/static/components/dataset.js
+++ b/src/encoded/static/components/dataset.js
@@ -16,6 +16,7 @@ import { SortTablePanel, SortTable } from './sorttable';
 import { ProjectBadge } from './image';
 import { DocumentsPanelReq } from './doc';
 import { FileGallery, DatasetFiles } from './filegallery';
+import sortMouseArray from './matrix_mouse_development';
 import { AwardRef, ReplacementAccessions, ControllingExperiments, FileTablePaged, ExperimentTable } from './typeutils';
 
 // Return a summary of the given biosamples, ready to be displayed in a React component.
@@ -788,7 +789,7 @@ const ProjectComponent = (props, reactContext) => {
     const datasetDocuments = (context.documents && context.documents.length > 0) ? context.documents : [];
 
     // Collect organisms
-    const organisms = (context.organism && context.organism.length > 0) ? _.uniq(context.organism.map((organism) => organism.name)) : [];
+    const organisms = (context.organism && context.organism.length > 0) ? [...new Set(context.organism.map((organism) => organism.name))] : [];
 
     // Set up the breadcrumbs
     const datasetType = context['@type'][1];
@@ -854,14 +855,14 @@ const ProjectComponent = (props, reactContext) => {
                             {context.biosample_ontology && context.biosample_ontology.length > 0 ?
                                 <div data-test="biosampletermname">
                                     <dt>Biosample term name</dt>
-                                    <dd>{_.uniq(context.biosample_ontology.map((b) => b.term_name)).join(', ')}</dd>
+                                    <dd>{[...new Set(context.biosample_ontology.map((b) => b.term_name))]}</dd>
                                 </div>
                             : null}
 
                             {context.biosample_ontology && context.biosample_ontology.length > 0 ?
                                 <div data-test="biosampletype">
                                     <dt>Biosample type</dt>
-                                    <dd>{_.uniq(context.biosample_ontology.map((b) => b.classification)).join(', ')}</dd>
+                                    <dd>{[...new Set(context.biosample_ontology.map((b) => (b.classification ? b.classification : '')))].join(', ')}</dd>
                                 </div>
                             : null}
 
@@ -967,7 +968,7 @@ const UcscBrowserCompositeComponent = (props, reactContext) => {
     const datasetDocuments = (context.documents && context.documents.length > 0) ? context.documents : [];
 
     // Collect organisms
-    const organisms = (context.organism && context.organism.length > 0) ? _.uniq(context.organism.map((organism) => organism.name)) : [];
+    const organisms = (context.organism && context.organism.length > 0) ? [...new Set(context.organism.map((organism) => organism.name))] : [];
 
     // Set up the breadcrumbs
     const datasetType = context['@type'][1];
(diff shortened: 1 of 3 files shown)


src/encoded/static/components/dataset.js
```diff
@@ -1324,6 +1418,40 @@ const replicationTimingSeriesTableColumns = {
     },
 };
 
+// Sort the rows of the matrix by stage (embryo -> postnatal -> adult) and then by age
+// Age can be denoted in days or weeks
+// In the future there are likely to be additions to the data which will require updates to this function
+// For instance, ages measured by months will likely be added
+function sortMouseAge(a, b) {
```
zoldello: variables names 'a' and 'b' too generic?

src/encoded/static/components/dataset.js
```diff
@@ -1324,6 +1418,40 @@ const replicationTimingSeriesTableColumns = {
...
+    const aAge = a.split(/ (.+)/)[1];
+    const bAge = b.split(/ (.+)/)[1];
+    let aNumerical = 0;
+    let bNumerical = 0;
+    if (aAge.includes('days')) {
```
zoldello: This violates DRY principle. It is better to create a function and return the calculated values.

src/encoded/static/components/dataset.js
```diff
@@ -1342,59 +1470,155 @@ const organismDevelopmentSeriesTableColumns = {
...
+            experiments.forEach((experiment) => {
+                let biosamples;
+                if (experiment.replicates && experiment.replicates.length > 0) {
+                    biosamples = experiment.replicates.map(replicate => replicate.library && replicate.library.biosample);
+                }
+                if (biosamples && biosamples.length > 0) {
+                    ageUnits = [...ageUnits, ...biosamples.map(b => b.age_units)];
+                }
+            });
+            const uniqueAgeUnits = _.uniq(ageUnits);
```
zoldello: Is using Set() a better option than _.uniq? Javascript inbuilt function almost always is faster than third party equivalents. For example, see (I know lodash is not underscore and their are a few functional differences, but the principle still apply)- https://www.measurethat.net/Benchmarks/Show/2496/2/lodash-uniq-vs-javascript-set#latest_results_block

src/encoded/static/components/dataset.js
```diff
@@ -1342,59 +1470,155 @@ const organismDevelopmentSeriesTableColumns = {
...
                 biosamples = experiment.replicates.map(replicate => replicate.library && replicate.library.biosample);
             }
             if (biosamples && biosamples.length > 0) {
-                synchronizationBiosample = _(biosamples).find(biosample => biosample.synchronization);
-                if (!synchronizationBiosample) {
-                    ages = _.chain(biosamples.map(biosample => biosample.age_display)).compact().uniq().value();
-                }
+                biosamples.forEach((biosample) => {
+                    if (biosample.age_units === 'year') {
+                        lifeStageAge.push(`${biosample.age_display ? biosample.age_display : ''}`);
```
zoldello: Why are you pushing nothing to an array?

src/encoded/static/components/dataset.js
```diff
@@ -1342,59 +1470,155 @@ const organismDevelopmentSeriesTableColumns = {
...
+    let bBiosamples;
+    let aBiosamples;
+    let aSynchronizationBiosample;
+    let bSynchronizationBiosample;
+
+    if (a.replicates && a.replicates.length > 0) {
+        aBiosamples = a.replicates.map(replicate => replicate.library && replicate.library.biosample);
+    }
+    if (aBiosamples && aBiosamples.length > 0) {
+        aSynchronizationBiosample = _(aBiosamples).find(biosample => biosample.synchronization);
```
zoldello: You can do this via Javascript's inbuilt find function. It can wager one of my green shirts that it would be faster.

src/encoded/static/components/dataset.js
```diff
@@ -1342,59 +1470,155 @@ const organismDevelopmentSeriesTableColumns = {
...
+    }
+    return 0;
+}
+
+const organismDevelopmentSeriesWormFlyTableColumns = {
+    accession: {
+        title: 'Accession',
+        display: (experiment, meta) =>
+            <span>
+                {meta.adminUser || publicDataset(experiment) ?
```
zoldello, line 1584: If this evaluates to false, you will have a span in a span. Is that what you desire to do?

src/encoded/static/components/dataset.js
```diff
@@ -1451,14 +1676,39 @@ export const SeriesComponent = (props, reactContext) => {
...
-    const seriesType = context['@type'][0];
+    let seriesType = context['@type'][0];
     const crumbs = [
         { id: 'Datasets' },
-        { id: datasetType, uri: `/search/?type=${datasetType}`, wholeTip: `Search for ${datasetType}` },
-        { id: breakSetName(seriesType), uri: `/search/?type=${seriesType}`, wholeTip: `Search for ${seriesType}` },
+        { id: breakSetName(seriesType), uri: `/series-search/?type=${seriesType}`, wholeTip: `Search for ${seriesType}` },
     ];
 
+    if (seriesType === 'OrganismDevelopmentSeries' && context.organism && context.organism.length > 0) {
```
zoldello: This is a code smell. Why are there two if-statements side by side and no else or else if?

src/encoded/static/components/dataset.js
```diff
@@ -1451,14 +1676,39 @@ export const SeriesComponent = (props, reactContext) => {
...
+        if ((context.organism[0].scientific_name === 'Caenorhabditis elegans') || (context.organism[0].scientific_name === 'Drosophila melanogaster')) {
+            seriesType = 'OrganismDevelopmentSeriesWormFly';
+        }
+    }
+
+    let treatmentDuration = [];
```
zoldello, line 1702: This is a bit confusing. First, I was shocked that ESLint did not bark at you for not declaring an array const. Then I saw later you reassigned it. Then I saw it being used as a string. Type-switching is allowed in dynamic languages. But it is a code smell unless you are dealing with an esoteric/purposely-confusing situation like coding anything to do with politics. That is my personal take.

src/encoded/static/components/search.js
```diff
@@ -315,7 +315,9 @@ const ExperimentComponent = (props, reactContext) => {
             biosamples = _.compact(result.replicates.map(replicate => replicate.library && replicate.library.biosample));
         }
         // flatten treatment array of arrays
-        _.compact(biosamples.map(biosample => biosample.treatments)).forEach(treatment => treatment.forEach(t => treatments.push(t)));
+        _.compact(biosamples.map(biosample => biosample.treatments)).forEach(treatment => treatment.forEach((t) => {
```
zoldello: inbuilt filter() may be faster than compact()

zoldello: foreach in a foreach is a code smell; unless you are dealing with something like tree and hate recursion. Are you trying to do something like:

`[...[1,2,3], ...[4,5,6]]`

or something like reduce(https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce)

src/encoded/static/components/search.js
```diff
@@ -552,52 +554,128 @@ const DatasetComponent = (props, reactContext) => {
     let biosampleTerm;
     let organism;
     let lifeSpec;
+    let assays = [];
```
zoldello, line 557: My spider tingles go off when I see an [] and "let" used together. Maybe that is just me remembering the many battles ESLint I had with ESLint and lost badly

src/encoded/static/components/search.js
```diff
@@ -552,52 +554,128 @@ const DatasetComponent = (props, reactContext) => {
...
+                if (dataset.assay_term_name) {
+                    assays.push(dataset.assay_term_name);
+                }
                 if (dataset.replicates && dataset.replicates.length > 0) {
                     dataset.replicates.forEach((replicate) => {
                         if (replicate.library && replicate.library.biosample) {
                             const biosample = replicate.library.biosample;
                             const lifeStage = (biosample.life_stage && biosample.life_stage !== 'unknown') ? biosample.life_stage : '';
+                            if (biosample.life_stage !== 'unknown' && biosample.life_stage && biosample.age_display) {
+                                const fullStage = `${biosample.life_stage[0].toUpperCase()}${biosample.age_display.split(' ')[0]}`;
```
zoldello, line 600: I did this
`var i = []; i.push('a b'.split(' '))` // the use of var is long tangential story
Is the what you intended to do in line 602?

src/encoded/static/components/search.js
```diff
@@ -552,52 +554,128 @@ const DatasetComponent = (props, reactContext) => {
...
+                            }
+                            if (biosample.treatments && treatmentConcentration) {
+                                treatmentTerm = [...treatmentTerm, ...biosample.treatments.filter(t => t.treatment_term_name).map(t => t.treatment_term_name)];
+                                treatments = [...treatments, ...biosample.treatments.reduce((output, t) => {
+                                    if (t.amount) {
+                                        output.push(t.amount);
+                                    }
+                                    return output;
+                                }, [])];
+                                treatmentUnit = `${biosample.treatments[0].amount_units}s`;
```
zoldello: No check for value === 1?

zoldello (changes requested):
The only issue that need to be addressed is the violated of the DRY principle. For the others, I feel free to ignore if you feel that are irrelevant or needed for consistency (like the underscore.js comments).

src/encoded/static/components/dataset.js
emro: I could rename these to be `aAge` and `bAge` - does that seems better? I also asked @forresttanaka if he uses variable names like 'a' and 'b' for sort functions and he says he does (but only for sorting and loops). I don't really have strong opinions about this, but I guess I think we should be consistent.

src/encoded/static/components/dataset.js
emro: I think this is ok. This is how all the tables are.

src/encoded/static/components/dataset.js
emro: Good point, I should not be.

src/encoded/static/components/dataset.js
emro: I agree that I shouldn't be changing the type around, it's pretty confusing.

src/encoded/static/components/dataset.js
emro: This I just copied from the original tables but the inbuilt find seems just fine so I'm not sure why the lodash find was used.

src/encoded/static/components/search.js
emro: This was a good catch. I checked with Bonita and she says there should never be an "s" at the end for any cases.

src/encoded/static/components/search.js
emro: I don't think so. This is for constructing `Stages: E10.5, E11.5, E14.5` on the mouse development search result.

zoldello (dismissed):
Looks good.

zoldello (approved):
Looks good.

PR #3691, opened 2021-01-27
ENCD-5683-exp-series-hotfix

yunhailuo:


src/encoded/static/components/experiment_series.js
```diff
@@ -520,10 +522,16 @@ class ExperimentSeriesComponent extends React.Component {
         if (Object.keys(viewableDatasets).length > 0) {
             // Add the "Add all to cart" button and internal tags from all related datasets.
             const experimentIds = Object.values(viewableDatasets).map(experiment => experiment['@id']);
+            const hubUrl = `http://genome.ucsc.edu/cgi-bin/hgTracks?hubClear=${window.location.origin}/batch_hub/type=Experiment%26accession=${
+                Object.values(viewableDatasets).map(experiment => experiment.accession).join('%26accession=')
```
forresttanaka: I would just use question marks instead of %26 to make this easier to understand. Using question marks won’t cause a 301 redirect — our server just accepts it.

src/encoded/static/components/experiment_series.js
```diff
@@ -520,10 +522,16 @@ class ExperimentSeriesComponent extends React.Component {
...
+            }/hub.txt&db=hg38`;
             addAllToCartControl = (
                 <div className="experiment-table__header">
                     <h4 className="experiment-table__title">{`Experiments in experiment series ${context.accession}`}</h4>
-                    <CartAddAllElements elements={experimentIds} />
+                    <div className="experiment-table__header-controls">
+                        <CartAddAllElements elements={experimentIds} />
+                        <a className="btn btn-info btn-sm" href={hubUrl} type="button" target="_blank" rel="noopener noreferrer">Visualize</a>
```
forresttanaka, line 534: Did you decide the arrow after the “Visualize” label in the button is OK?

forresttanaka:
A comment and a question.

src/encoded/static/components/experiment_series.js
yunhailuo: Thanks for the help. I updated the code. Please take a look again.

src/encoded/static/components/experiment_series.js
yunhailuo: That's `&`. I put it as %26 to not confuse UCSC. I tested `&` and it won't work.

forresttanaka (dismissed):

src/encoded/static/components/experiment_series.js
```diff
@@ -520,10 +522,16 @@ class ExperimentSeriesComponent extends React.Component {
         if (Object.keys(viewableDatasets).length > 0) {
             // Add the "Add all to cart" button and internal tags from all related datasets.
             const experimentIds = Object.values(viewableDatasets).map(experiment => experiment['@id']);
```
forresttanaka: I missed this the first time, and this does work in this case by accident, but the `window` global is only available in the browser, so in theory when this code runs on the server it crashes. For these cases you should use the `location_href` property from `<App>` context. In this case it works because viewableDatasets doesn’t have a value on the server; only in the browser.

forresttanaka (dismissed):

forresttanaka (approved):

PR #3695, opened 2021-01-29
ENCD-5686-move-stimulation-to-purpose

jenjou:


src/encoded/tests/data/inserts/treatment.json
```diff
@@ -169,5 +169,15 @@
         "treatment_term_id": "CL:0000815",
         "treatment_term_name": "regulatory T cell",
         "treatment_type": "injection"
+    },
+        {
```
iyoungworth: It looks like this curly bracket is indented an extra tab inward

src/encoded/upgrade/treatment.py
```diff
@@ -97,3 +97,11 @@ def treatment_9_10(value, system):
 def treatment_10_11(value, system):
     # https://encodedcc.atlassian.net/browse/ENCD-4272
     value.pop('lab', None)
+
+
+@upgrade_step('treatment', '11', '12')
+def treatment_11_12(value, system):
+    # https://encodedcc.atlassian.net/browse/ENCD-5686
+    if value.get('treatment_type') == 'stimulation':
+        value.pop('treatment_type', None)
```
iyoungworth: It looks like this will default the upgraded objects to treatment_type:chemical. Is there a possibility some will need to be patched to other values? (If not, this is fine)

iyoungworth (changes requested):
Tests are passed, looks good other than 1 small indenting issue and a question about the upgrade

src/encoded/upgrade/treatment.py
jenjou: Hmm, does this mean the test I wrote isn't working? I tried to test that "treatment_type" is completely removed in the object after the upgrade and it seems like it passed.

iyoungworth: It gets removed, but immediately on removal I believe it gets automatically replaced with "chemical" because that's the default in the schema definition (https://github.com/ENCODE-DCC/encoded/blob/dev/src/encoded/schemas/mixins.json#L489)

src/encoded/tests/data/inserts/treatment.json
jenjou: Fixed

src/encoded/upgrade/treatment.py
jenjou: After discussing with Idan and Yunhai, I'm removing the default value, but will upgrade any "stimulation" to "chemical".

iyoungworth (approved):

src/encoded/upgrade/treatment.py
iyoungworth: Sounds great, thank you for clearing that up

PR #3697, opened 2021-02-01
Encd 5677 file details table breakup processed table

zoldello:


zoldello:
Merged into ENCD-5726

PR #3698, opened 2021-02-01
ENCD-5674-allow-submission-of-library-with-mixed-biosamples

iyoungworth:


src/encoded/audit/experiment.py
```diff
@@ -3670,35 +3670,50 @@ def audit_experiment_biosample_term(value, system, excluded_types):
...
             experiment_bs_type = value.get('biosample_ontology', {}).get('@id')
             experiment_bs_name = value.get('biosample_ontology', {}).get('name')
-            if bs_type != experiment_bs_type:
-                detail = ("Experiment {} contains a library {} linked to biosample "
-                    "type '{}', while experiment's biosample type is '{}'.".format(
-                        audit_link(path_to_text(value['@id']), value['@id']),
-                        audit_link(path_to_text(lib['@id']), lib['@id']),
-                        audit_link(path_to_text(bs_type), bs_type),
-                        audit_link(path_to_text(experiment_bs_type), experiment_bs_type)
+            technical_sample_type = '/biosample-types/technical_sample_NTR_0000637/'
```
gabdank: I think we should use UUID, rather than ID that could theoretically change if we get rid of NTR at some point. UUID will not change.

src/encoded/audit/experiment.py
```diff
@@ -4627,6 +4622,29 @@ def audit_experiment_mixed_strand_specific_libraries(value, system, excluded_typ
...
+        if 'biosample' in lib:
+            biosample += 1
+        elif 'mixed_biosamples' in lib:
+            pooled += 1
+    if biosample > 0 and pooled > 0:
+        detail = (
+            f'This dataset contains libraries with both standard and '
+            f'mixed biosamples.'
+            )
+        yield AuditFailure('inconsistent library structure', detail, level='ERROR')
```
gabdank: I think you should catch this under "inconsistent library biosample", I understand that it is more of "inconsistent libraries in the same experiment", but I don't think we should add a new facet audit entry for this. and "inconsistent library structure" is not self explanantory.

src/encoded/schemas/library.json
```diff
@@ -266,6 +267,21 @@
             "type": "string",
             "linkTo": "Biosample"
         },
+        "mixed_biosamples": {
+            "title": "Mixed biosamples",
+            "description": "Two or more biosamples mixed together before isolating nucleic acid for library generation.",
```
gabdank: I am not sure the samples are mixed prior to the extraction of nucleic acid. Unless you are confident that is what they are doing in LRGASP, I think you should make more general statement, like "biosamples nucleic acid of which was used to construct the library".

gabdank (changes requested):
comments in the PR

src/encoded/audit/experiment.py
iyoungworth: I think the only way to do this is to submit it to production first with an incorrect classification (since the 'technical sample' classification is being added in this ticket) to get the UUID to use here, and patch the BiosampleType record on production after v112 release. This is what I've done in this new commit, let me know if there's a better alternative. (I can have 'dummy' values in the term_name and term_id before release also, so that v112 doesn't invalidate it)

src/encoded/schemas/library.json
iyoungworth: That makes sense, so I reworded it to be more ambiguous

src/encoded/audit/experiment.py
iyoungworth: I reorganized this as suggested, thanks!

gabdank (approved):

PR #3704, opened 2021-02-03
ENCD-5247-experiment- series-table-update

yunhailuo:


src/encoded/static/components/experiment_series.js
```diff
@@ -520,6 +529,9 @@ class ExperimentSeriesComponent extends React.Component {
         let internalTags = [];
         const allRows = [];
         const { viewableDatasets } = this.state;
+        const handleAnalysisRFAChange = (e) => {
+            this.setState({ selectedAnalysisRFA: e.target.value });
+        };
```
forresttanaka: `setState` inside render() in a class component shouldn’t happen, as you can easily have infinite render loops. Should make this function its own method. You might find examples of this elsewhere, but they shouldn’t have happened.

src/encoded/static/components/experiment_series.js
```diff
@@ -543,26 +555,19 @@ class ExperimentSeriesComponent extends React.Component {
...
-                const filesByDesiredAssembly = [];
-                viewableDatasets[datasetAtId].files.forEach((f) => {
-                    if (selectedAssembly.includes(f.assembly)) {
-                        filesByDesiredAssembly.push(f['@id']);
-                    }
-                });
+                // Select one analysis based on analysis award
                 const analysisObjects = viewableDatasets[datasetAtId].analysis_objects || [];
-                const selectedAnalysis = analysisObjects.filter(
+                viewableDatasets[datasetAtId].selectedAnalysis = analysisObjects.filter(
```
forresttanaka, line 596: We’ll have to be careful going forward, as ESLint/AirBnB get more strict about mutating objects that don’t belong to us. I’m not sure if the datasets in viewableDatasets comprise copies of incoming datasets, or comprise the original datasets from React. If the latter, we shouldn’t mutate them. But in this case, I don’t remember anymore.

forresttanaka (commented):
As a first pass, I think conceptually it seems fine. You might want to add comments about the purpose of `selectedFiles` and `selectedAnalyses`. They’re key to this change, but I’d have to reverse engineer what their purpose is.

src/encoded/static/components/experiment_series.js
yunhailuo: Good point. Thank you! I will make it a method and use `this. handleAnalysisRFAChange` upon changes.

src/encoded/static/components/experiment_series.js
yunhailuo: I'm not sure if I understand "belong to us" correctly. So `viewableDatasets` is a state from this method and is build from a series of requests: https://github.com/ENCODE-DCC/encoded/blob/dev/src/encoded/static/components/experiment_series.js#L417-L449 . So maybe it belongs to the component thus belong to us?

forresttanaka: If the contents of viewableDatasets includes references to the datasets coming through props.context, then those objects belong to React and we shouldn't mutate them. Problem is, we've committed this mistake often in the past, and I don't remember if viewableDatasets makes copies of the datasets objects (so we can mutate those) or if it references props.context and we shouldn't, and it wasn't obvious by looking which was the case.

gabdank (dismissed):

src/encoded/static/components/experiment_series.js
```diff
@@ -385,8 +393,15 @@ class ExperimentSeriesComponent extends React.Component {
...
+             * there are two key prpoperies, selectedAnalysis and selectedFiles.
+             * selectedAnalysis is one analysis objects selected based on lab,
+             * award, etc. selectedFiles is a filtered list of experiment files
+             * belonging to the selectedAnalysis.
+             */
             viewableDatasets: {},
+            selectedAnalysisRFA: defaultAnalysisRFA,
         };
         this.statusFilteredDatasets = this.statusFilteredDatasets.bind(this);
         this.getViewableDatasets = this.getViewableDatasets.bind(this);
```
forresttanaka, line 435: You need to bind `this` for `handleAnalysisRFAChange` to work properly: `this.handleAnalysisRFAChange = this.handleAnalysisRFAChange.bind(this);` Surprised that doesn’t crash.

src/encoded/static/components/experiment_series.js
```diff
@@ -543,26 +562,19 @@ class ExperimentSeriesComponent extends React.Component {
...
-                    (f) => selectedAnalysis.files.includes(f['@id']),
+                viewableDatasets[datasetAtId].selectedFiles = viewableDatasets[datasetAtId].files.filter(
+                    (f) => viewableDatasets[datasetAtId].selectedAnalysis.files.includes(f['@id'])
                 );
                 // Number of subrows to be shown corresponds to the number of biological replicates. Show at least one by default.
                 let bioRepNums = _.uniq(
-                    viewableDatasets[datasetAtId].files.map((f) => f.biological_replicates).reduce(
+                    viewableDatasets[datasetAtId].selectedFiles.map((f) => f.biological_replicates).reduce(
                         (bioRepArray, reps) => bioRepArray.concat(reps), []
                     )
```
forresttanaka, line 608: Does this `reduce` do anything? I feel like it just outputs a copy of what was given from the output of `map`.

src/encoded/static/components/experiment_series.js
```diff
@@ -120,43 +125,30 @@ const experimentTableColumns = {
...
-                } else {
-                    qcClass = qcClass.concat(' qc-report--ideal');
+            if (experiment.selectedAnalysis && experiment.selectedAnalysis.quality_metrics_report) {
+                const metricObj = getQualityMetricsByReplicate(
+                    experiment.selectedAnalysis.quality_metrics_report,
+                    'Read depth',
+                    meta.bioRepNum
+                );
+                if (metricObj && metricObj.metric) {
+                    return <td key="readDepth" className={qcModifierMap[metricObj.quality] || qcBlock}>{(metricObj.metric / 1000000).toFixed(2).concat('M')}</td>;
```
forresttanaka: Do you need the `key` attribute here? You normally only need these within a `.map` loop. I’d be surprised if you needed them here and in the other table properties.

forresttanaka (changes requested):
Just a few things I can see. On the React keys on table column displays, I don’t _think_ you need that, but I’m not sure. You can tell if you remove them and you get red warnings in the Javascript console.

src/encoded/static/components/experiment_series.js
yunhailuo: Yes. You are right. I do get an error. Fixed.

src/encoded/static/components/experiment_series.js
yunhailuo: You are correct. I should just reduce as you suggested. Fixed.

src/encoded/static/components/experiment_series.js
yunhailuo: Removed keys and don't get errors on console. So fine to remove.

forresttanaka (dismissed):
Looks good! The Python looks fine to me too, but maybe Idan should take a look as well.

src/encoded/types/quality_metric.py
```diff
@@ -213,6 +229,18 @@ class HistoneChipSeqQualityMetric(QualityMetric, CalculatedAssayTermID):
...
 
+    @calculated_property(schema={
+        "title": "Best FRiP score",
+        "type": "number",
+        "description": "Best fraction reads in peaks (FRiP) from peaks",
+        "comment": "Do not submit. The value is extracted from submitted FRiP scores.",
+        "notSubmittable": True,
+    })
+    def frip(self, request, Fp=None, Ft=None, F1=None, F2=None):
+        frips = list(filter(lambda x: x is not None, [Fp, Ft, F1, F2]))
```
keenangraham: This is fine and concise though usually preferable to use list comprehension over filter/lambda.

```python
frips = [
    f
    for f in [Fp, Ft, F1, F2]
    if f is not None
]
```

https://stackoverflow.com/a/3013722

yunhailuo: 👍 Good reference. Thank you! Fixed.

forresttanaka (dismissed):
The commit since the last time I approved was in response to feedback from Keenan.

keenangraham (dismissed):

yunhailuo:
To clarify, pushed two commit for bug fix: one javascript and one python. So I requested review from @forresttanaka and @keenangraham

forresttanaka (approved):
Just looking at the changes since last time, this looks good.

gabdank (approved):

PR #3705, opened 2021-02-04
ENCD-5709 Remove component from select value

forresttanaka:
Not actually allowed ever, but previous versions of React let us get away with it.


emro (approved):
Looks good. I feel like there should have been a linting error that caught this.

forresttanaka:
> I feel like there should have been a linting error that caught this.

Yeah, that's a good point. I looked over other uses of `<option>` and at least that was the only one.

PR #3706, opened 2021-02-04
ENCD-4018-update-regulome-related-schema

yunhailuo:


src/encoded/schemas/annotation.json
```diff
@@ -76,11 +76,15 @@
             "type": "string",
             "enum": [
                 "binding sites",
+                "curated SNVs",
+                "dsQTLs",
```
hitz, line 80: should these be renamed caQTLS?

src/encoded/schemas/file.json
```diff
@@ -1443,11 +1445,13 @@
                         "DHS regions reference",
                         "differential expression quantifications",
                         "differential splicing quantifications",
+                        "dsQTLs",
```
hitz, line 1448: caQTLs?

src/encoded/schemas/file.json
```diff
@@ -2134,8 +2139,12 @@
         
         "HMM predicted chromatin state": "annotation",
         "semi-automated genome annotation": "annotation",
-        "capture targets": "annotation"
-        
+        "capture targets": "annotation",
+        "curated binding sites": "annotation",
+        "curated SNVs": "annotation",
+        "eQTLs": "annotation",
+        "dsQTLs": "annotation",
```
hitz, line 2146: caQTLs?

hitz (approved):
I think that the more generic name for dsQTLs is caQTLs, but since we don't have any yet, perhaps this isn't an issue.  Also we would have to patch regulome-encoded.

PR #3709, opened 2021-02-05
ENCD-5675 Display all children of a biosample including grandchildren

forresttanaka:
* I added a babel plugin allowing async/await and generators (though I don't use generators in this branch, we supposedly can now). I also disabled the no-await-in-loop ESLint rule because I needed to use the results of one await for the next, and even the [ESLint rule documentation](https://eslint.org/docs/rules/no-await-in-loop#when-not-to-use-it) says to disable the rule if you have this case.
* I rewrote `requestObjects` so it can search for properties aside from `@ids`, so we can search for experiments referencing specific biosamples instead of just specific objects. I removed the unused `fileFilters` parameter. While I was at it, I rewrote `requestObjects` to use async / await. This function remains compatible with all existing uses.


src/encoded/static/components/typeutils.js
```diff
@@ -373,6 +378,26 @@ const experimentTableColumns = {
...
+                            <a href={biosample['@id']}>{biosample.accession}</a>
+                        </span>
+                    ));
+                }
+                return null;
+            }
+            return null;
+        },
+        hide: (list, columns, meta) => !(meta && meta.showBiosamples),
+    },
```
emro, line 402: The biosamples aren't being concatenated quite right here.
Look at this example: https://encd-5675-biosample-hierarchy-flat-qa1-fytanaka.demo.encodedcc.org/biosamples/ENCBS056AAA/
Some of the `, `s between accessions are mixed up.

Screen Shot 2021-02-10 at 9 12 43 PM

emro: Furthermore, do we expect there to be repeated biosamples? We don't want to use `.uniq` or `Set` to get a list without repeated biosample links?

emro (changes requested):
I'm going to look at this again to make sure I didn't miss anything, but wanted to submit my review for now. It looks like there is a formatting issue on the functional genomics table which should be fixed. Overall the changes look good and I'm excited about the babel plugin allowing async/await.

src/encoded/static/components/typeutils.js
forresttanaka: Ah, I didn't realize you could have repeats here, and it caused a React keys problem I believe. I fixed that, but I found an issue that might cause me to return to doing an initial search request, or embed a new property. I'll find out Thursday.

emro (dismissed):

forresttanaka:
Had to take care of more duplicates, as I now see how that can validly happen. Most of the other issues involved me forgetting that `requestObjects` doesn’t automatically use limit=all (and I think it really should, but enough parts of the code call it that I don’t feel like fixing that in a branch update) so I wasn’t retrieving enough objects. I also had to request lab information on one request to fill in that column.

emro (dismissed):
Looks good, I'm glad the duplicates are sorted out now!

forresttanaka:
Some biosamples are embedded and some are just @ids that I need to request all the fields for, and I missed the summary field.

emro (approved):

PR #3710, opened 2021-02-06
SNO-117-top-hits-search

keenangraham:


keenangraham:
@forresttanaka this is not done but maybe you could take an initial quick look.

src/encoded/static/components/top_hits/form.js
```diff
@@ -0,0 +1,78 @@
...
+} from './input';
+import Results from './results';
+
+
+/** Renders the input and dropdown (if there are any results). */
+const Form = (props) => (
+    <form className="top-hits-search__form" action="/search/">
+        {props.children}
+        {
+            !!props.results.length &&
```
forresttanaka: [AirBnB rules](https://github.com/airbnb/javascript#comparison--shortcuts) say you have to have explicit length comparisons with zero.

src/encoded/static/components/top_hits/hit.js
```diff
@@ -0,0 +1,174 @@
...
+            this.item.term_name
+        );
+    }
+
+    formatTarget() {
+        return this.item.target && this.item.target.title;
+    }
+
+    formatName() {
+        return this.item.accession || this.item['@id'];
```
forresttanaka: Not sure if we should have a case for file.title? Some files don't have accesssions, but they all have titles, which often have the same value as their accessions if they have one. Otherwise they have the value of their external_accession.

src/encoded/static/components/top_hits/hit.js
```diff
@@ -0,0 +1,174 @@
...
+            ([field, value]) => (
+                [
+                    field,
+                    maybeConvertArrayToString(value),
+                ]
+            )
+        ).filter(
+            ({ 1: value }) => Boolean(value)
+        );
+    }
```
forresttanaka: Would this look simpler with no map function, and instead things like `return [['name', maybeConvertArrayToString(this.formatName())], etc.]`. I can see the argument for doing it the way you have it, but I feel like mapping an array onto effectively itself with a small modification seems the non-obvious solution.

src/encoded/static/components/top_hits/hit.js
```diff
@@ -0,0 +1,174 @@
...
+            ['description', this.formatDescription()],
+        ].map(
+            ([field, value]) => (
+                [
+                    field,
+                    maybeConvertArrayToString(value),
+                ]
+            )
+        ).filter(
+            ({ 1: value }) => Boolean(value)
```
forresttanaka: I see what this does, but I’m not clear how it works. I’ve used colons in destructuring for renaming, but not with a constant in front of the colon. Would this look clearer?:

```
        ).filter(
            ([, value]) => Boolean(value)
        );
```

Sort of odd in itself, but I feel I can wrap my head around a destructured array for array elements than a destructured object where we don’t have any objects. Having empty comma-separated elements in array destructuring constitutes defined Javascript behavior (e.g. [a,,,d,e,,g,]).

forresttanaka (changes requested):
This looks really good!

I and I think Phil have been using JSDoc on functions that aren’t React components because if you then hover over references to the function in VSCode, you get complete documentation right there including parameters. We don’t enforce JSDoc, but that’s why you’ll see it in many places. If we ever convert to TypeScript (which I think unlikely) then we can do that with React components too because we won’t need PropTypes.

All of my code-review comments are arguable, so if you change nothing, I’ll pass it.

I’m going to copy some of the way you format code because I find it much clearer and cleaner than what I’ve been doing.

src/encoded/static/components/top_hits/hit.js
keenangraham: Yeah, there was originally a third item in the list, and I was only trying to pull out the middle value. Couldn't seem to make eslint happy with array destructuring: 

`([, value, ]) => Boolean(value)` caused 
``` 
150:22  error  Unexpected trailing comma            comma-dangle
150:23  error  There should be no space before ']'  array-bracket-spacing
```

while removing the space `([, value,]) => Boolean(value)` caused:
```
150:22  error  Unexpected trailing comma      comma-dangle
150:22  error  A space is required after ','  comma-spacing
```
Found some suggestion about using this object destructuring syntax instead, but agree it's weird.

I think I can go back to `[, value]` now that there is only two values.

src/encoded/static/components/top_hits/form.js
keenangraham: Thanks, will update.

src/encoded/static/components/top_hits/hit.js
keenangraham: Yeah this might be good. Will see what it looks like.

src/encoded/static/components/top_hits/hit.js
keenangraham: I think there are a couple disadvantages for calling it in the list. One is that it will require duplicating the call for every item, and any future items that are added. The second is that it will make it less flexible if there is ever an intermediate formatting step that needs to be inserted between getting the values and converting arrays to strings.

src/encoded/static/components/top_hits/hit.js
forresttanaka: Ah, I see. yeah, with a third value you’d get that. Can probably avoid it with an awkward:

```
[
    ,
    value,
    ,
]
```

I haven't actually checked if ESLint is fine with that. Maybe we should turn off the comma-dangle rule permanently, because it doesn't take this case into account, and it's a valid case. If you want to turn off the command-dangle rule in .eslintrc.js, I wouldn’t mind.

src/encoded/static/components/top_hits/hit.js
forresttanaka: The duplication is more of a copy/paste cost though. The cost for execution or memory would be negligible since maybeConvertArrayToString isn’t inline. For inserting a new step, yeah, doing it all in one means more copy/paste, where doing it in a separate map keeps any future changes in one place. I’m fine either way. The way you have it involves another loop and another inline function which have their own costs, but probably close to negligible as well.

src/encoded/static/components/top_hits/hit.js
forresttanaka: Well, actually, if you had three values, does it complain if you just ignore the third by leaving out the trailing comma? I would think it just gets ignored, which is what you want anyway.

forresttanaka: Looks like ignoring works:

```
> a = [[1, 2, 3], [4, 5, 6], [7, 8, 9]];
[ [ 1, 2, 3 ], [ 4, 5, 6 ], [ 7, 8, 9 ] ]
> a.map(([,b]) => { console.log(b); })
2
5
8
```

keenangraham: Ah, yeah! I think that's the right way. Good to know.

src/encoded/static/components/top_hits/hit.js
keenangraham: Yeah, maybe a bit of opinion. If it was one or two items in a list that was never going to change I think your way is clearer. Since it's possible we'll want to parse other values in object types, and they might require additional formatting steps, I think this violates DRY less and makes it clearer that a uniform step is applied to all values. I think worst case you can imagine the visual clutter cost of unrolling something like: 

```
const results = [
    ['name', this.formatName()],
    ['title', this.formatTitle()],
    ['target', this.formatTarget()],
    ['details', this.formatDetails()],
    ['description', this.formatDescription()],
].map(
    ([field, value]) => (
        [
            field,
            maybeDoSomethingElseAfter(
                maybeConvertArrayToString(
                    maybeDoSomethingElseFirst(
                        value
                    ),
                ),
            ),
        ]
    )
);
```
to 
```
const results = [
    [
        'name',
        maybeDoSomethingElseAfter(
            maybeConvertArrayToString(
                maybeDoSomethingElseFirst(
                    this.formatName()
                ),
            ),
        ),
    ],
    [
        'title',
        maybeDoSomethingElseAfter(
            maybeConvertArrayToString(
                maybeDoSomethingElseFirst(
                    this.formatTitle()
                ),
            ),
        ),
    ],
    [
        'target',
        maybeDoSomethingElseAfter(
            maybeConvertArrayToString(
                maybeDoSomethingElseFirst(
                    this.formatTarget()
                ),
            ),
        ),
    ],
    [
        'details',
        maybeDoSomethingElseAfter(
            maybeConvertArrayToString(
                maybeDoSomethingElseFirst(
                    this.formatDetails()
                ),
            ),
        ),
    ],
    [
        'description',
        maybeDoSomethingElseAfter(
            maybeConvertArrayToString(
                maybeDoSomethingElseFirst(
                    this.formatDescription()
                ),
            ),
        ),
    ],
];
```
I think it's harder to determine if all the values go through the same steps in the second example, and also a pain if you change the name of one of the functions. I don't think the immutable functional folks would care much about the cost of creating a new list.

forresttanaka: I would definitely prefer your way if we needed anything remotely like that.

src/encoded/static/components/top_hits/hit.js
keenangraham: I ended up adding `external_accession` for files. Looks like this is the same thing as title for files (https://github.com/ENCODE-DCC/encoded/blob/e30528e81a1fd3dfae318fdb3de67aa5a8c6ebfb/src/encoded/types/file.py#L204) and avoids using title where we wouldn't really want it (e.g. for publications).

keenangraham:
@forresttanaka Tried to add some JSDoc param and return definitions for non-React functions.

forresttanaka (dismissed):
Very cool! Thanks for JSDoc comments. Again, we don’t enforce it, but it can help making development quicker. New changes look good.

keenangraham:
Noticed the homepage image borders were bleeding through results dropdown on mobile so bumped z-index by one. Seems to help.

keenangraham:
@forresttanaka helped get rid of unneeded button layer that was causing issues on IE11.

forresttanaka (dismissed):
Looks good!

keenangraham:
@forresttanaka Just so you know I rebased this branch and squashed everything already reviewed into one commit.

keenangraham:
Ready for rereview. The main change is we are now returning an item-specific `Hit` for certain item types (Experiment, Biosample, File). This allows for customized formatting of hits based on `@type`. Still have the base `Hit` class for default formatting of all other types.

src/encoded/static/components/top_hits/hits/base.js
```diff
@@ -0,0 +1,321 @@
...
+* @param {number} maxWords - Max number of words to return.
+* @return {string} Trimmed sentence.
+*/
+export const maybeTrimByWord = (value, maxWords) => {
+    const splitValue = value.split(WHITESPACE);
+    if (splitValue.length > maxWords) {
+        return `${splitValue.slice(0, maxWords).join(WHITESPACE)}...`;
+    }
+    return value;
+};
```
forresttanaka, line 23: Not sure if it’s appropriate here, but we have the `truncateString` function in globals.js that truncates a string to a number of characters, back to the previous word break before that character count. Embarrassingly, even _I_ didn’t comment that function.

src/encoded/static/components/top_hits/hits/base.js
```diff
@@ -0,0 +1,321 @@
...
+* unique values.
+* @param {array} values - Array to filter and make unique.
+* @return {array} Unique array with only truthy values.
+*/
+export const uniqueAndDefinedValues = (values) => (
+    _.uniq(
+        values.filter(
+            (value) => Boolean(value)
+        )
+    )
```
forresttanaka, line 51: Underscore.js has a _.compact() function that filters out falsy entries in an array, though it doesn’t really cut things down much over what you have.

src/encoded/static/components/top_hits/hits/biosample.js
```diff
@@ -0,0 +1,20 @@
...
+    */
+    getValues() {
+        return [
+            ['name', this.formatName()],
+            ['title', this.formatTitle()],
+            ['details', this.formatDetails()],
+            ['description', this.formatDescription()],
+        ];
+    }
+}
```
forresttanaka, line 15: Really nice use of classes.

src/encoded/static/scss/encoded/modules/_top_hits_search.scss
```diff
@@ -0,0 +1,155 @@
...
+            padding: 5px 10px;
+            width: 220px;
+            border-radius: 0;
+            border: none;
+            background: transparent;
+            box-shadow: none;
+            transition: width 0.1s ease-out;
+            -ms-transition: width 0.1s ease-out;
+            -moz-transition: width 0.1s ease-out;
+            -webkit-transition: width 0.1s ease-out;
```
forresttanaka, line 27: AFAIK don’t really need the browser-specific props, but they don’t hurt either.

forresttanaka (dismissed):
Beauty. I made a couple comments but I don’t even recommend making any changes for them; just things to note in case they help in the future.

src/encoded/static/components/top_hits/hits/base.js
keenangraham: Cool, updated to use.

src/encoded/static/scss/encoded/modules/_top_hits_search.scss
keenangraham: Good to know, pulled from old search box definition.

src/encoded/static/components/top_hits/hits/base.js
keenangraham: Ah all types of useful functions in there.

forresttanaka (dismissed):
Cool. I won't move the ticket forward in case you’re waiting for @hitz.

keenangraham:
Added a couple more boosts. Going to rebase this again to resolve css conflicts. Will collapse already reviewed commits again.

keenangraham:
I think maybe @hitz just needs to review Python.

hitz (approved):

PR #3715, opened 2021-02-10
ENCD-5348 WGBS audit

jenjou:


src/encoded/audit/experiment.py
```diff
@@ -1912,6 +2009,18 @@ def check_wgbs_lambda(bismark_metrics, threshold, pipeline_title):
                                    level='WARNING')
 
 
+def check_wgbs_lambda_ENCODE4(gembs_metrics, threshold, pipeline_title):
+    for metric in gembs_metrics:
+        conversion_rate = metric.get('conversion_rate')
+        if conversion_rate > threshold:
```
paul-sud: Think this should be reversed to check values under the 99% threshold:
```suggestion
        if conversion_rate < threshold:
```

src/encoded/audit/experiment.py
```diff
@@ -1912,6 +2009,18 @@ def check_wgbs_lambda(bismark_metrics, threshold, pipeline_title):
...
+            detail = (f'ENCODE experiment processed by {pipeline_title} '
+                f'pipeline has a lambda rate of {conversion_rate}. '
+                f'The lambda rate should be < 1.'
```
paul-sud: Would be clearer to read like this:
```suggestion
                f'The lambda conversion rate should be > 99%'
```

src/encoded/audit/experiment.py
```diff
@@ -1912,6 +2009,18 @@ def check_wgbs_lambda(bismark_metrics, threshold, pipeline_title):
...
+            )
+            yield AuditFailure('high lambda C methylation ratio', detail,
```
paul-sud: I think this message is a little confusing. I would propose something like this:
```suggestion
            yield AuditFailure('low lambda C conversion rate', detail,
```

src/encoded/audit/experiment.py
```diff
@@ -1077,6 +1085,49 @@ def check_experiment_wgbs_encode3_standards(experiment,
...
+        cpg_quantifications, 'CpgCorrelationQualityMetric', desired_assembly)
+    samtools_metrics = get_metrics(
+        cpg_quantifications, 'SamtoolsFlagstatsQualityMetric', desired_assembly)
+
+    yield from check_wgbs_coverage_ENCODE4(
+        gembs_metrics,
+        pipeline_title,
+        get_pipeline_objects(alignment_files))
+    yield from check_wgbs_pearson_ENCODE4(cpg_metrics, 0.8, pipeline_title)
+    yield from check_wgbs_lambda_ENCODE4(gembs_metrics, 1, pipeline_title)
```
paul-sud: Conversion rate is expressed as a float between 0 and 1. We want to audit values less than 0.99/99%:
```suggestion
    yield from check_wgbs_lambda_ENCODE4(gembs_metrics, 0.99, pipeline_title)
```

paul-sud (changes requested):
I think the lambda rate auditing needs some tweaks, looks fine otherwise.

paul-sud:
@jenjou could you update the tests too? Seems like a legit failure.

src/encoded/audit/experiment.py
```diff
@@ -1077,6 +1085,49 @@ def check_experiment_wgbs_encode3_standards(experiment,
...
+        cpg_quantifications, 'CpgCorrelationQualityMetric', desired_assembly)
+    samtools_metrics = get_metrics(
+        cpg_quantifications, 'SamtoolsFlagstatsQualityMetric', desired_assembly)
+
+    yield from check_wgbs_coverage_ENCODE4(
+        gembs_metrics,
+        pipeline_title,
+        get_pipeline_objects(alignment_files))
+    yield from check_wgbs_pearson_ENCODE4(cpg_metrics, 0.8, pipeline_title)
+    yield from check_wgbs_lambda_ENCODE4(gembs_metrics, 0.99, pipeline_title)
```
paul-sud: Sorry I didn't realize, this is supposed to be 0.98 according to the standards. So it should be 0.98:
```suggestion
    yield from check_wgbs_lambda_ENCODE4(gembs_metrics, 0.98, pipeline_title)
```

paul-sud (changes requested):
My bad, got the threshold for the conversion rate wrong. But everything else looks good now.

paul-sud (approved):

PR #3716, opened 2021-02-10
ENCD-5719 fix breadcrumbs and treatments

emro:
Two updates:
- Need special breadcrumbs for /series-search/ series vs other series
- Want to display different treatment entries for different cases: for Treatment Time, a list of amount + term name, for Treatment Concentration, term name + duration, for other cases, a list of treatment term names


zoldello (dismissed):
Code looks good. I will talk to Emma on slack about some UI questions.

zoldello (approved):
The code looks. Series link look okay. I have UI questions for Emma but will take up those on Slack.

PR #3717, opened 2021-02-11
ENCD-5695-highlight-file-detail-and-graph

zoldello:


zoldello:
Merging code into ENCD-5726

PR #3718, opened 2021-02-11
ENCD-5711-add-regulome-output-type-to-analysis-step

yunhailuo:


gabdank (approved):

gabdank (changes requested):
There is a typo in the changelog. Since version 15 and not 12
Could you also please include the new input types/output types in the inserts.

gabdank (dismissed):

gabdank (approved):

PR #3722, opened 2021-02-12
ENCD-5714-add-defect-mode-property

jenjou:


src/encoded/schemas/dnase_alignment_quality_metric.json
```diff
@@ -172,6 +172,11 @@
                     "title": "MD5sum"
                 }
             }
+        },
+        "defect_mode": {
+            "title": "Defect mode",
+            "description": "A flag to indicate whether the pipeline run that generated this quality metric was run in defect mode.",
+            "type": "boolean"
```
ottojolanki: Might make sense to add a default value `false` here?

ottojolanki: I would expect this to be undefined in 99% of the runs (in which case the pipeline falls into the default value, which is `false`).

ottojolanki (approved):
Looks Good To Me.

PR #3723, opened 2021-02-12
ENCD-5726-file-gallery-data-display-update

zoldello:


src/encoded/static/components/filegallery.js
```diff
@@ -178,7 +269,18 @@ export class FileTable extends React.Component {
...
+
+                let analysisObjectsAccession = null;
+
+                (analysisObjectsFiles || []).some((analysisObjectsFile) => {
+                    if (analysisObjectsFile.files.includes(file['@id'])) {
+                        analysisObjectsAccession = analysisObjectsFile.accession;
+                        return true; // break out
+                    }
+                    return false; // continue loop
+                });
```
forresttanaka: This looks more naturally like a `.find()` operation than a `.some()` operation. That way `analysisObjectsAccession` can be assigned the return value instead of ignoring the return value of `.some()`. And then you can just grab the `.accession` of whatever `.find()` returns.

src/encoded/static/components/filegallery.js
```diff
@@ -187,6 +289,17 @@ export class FileTable extends React.Component {
             // (2) move showReplicateNumber to meta
             FileTable.procTableColumns.biological_replicates.hide = () => !showReplicateNumber;
 
+            const analysisObjectKeys = Object.keys(files).filter((file) => file.includes('ENCAN'));
+            const otherKeys = Object.keys(files).filter((file) => file === nonAnalysisObjectPrefix); // all files not parts of an analysis object
+            const compileAnalysis = compileAnalyses(context, datasetFiles);
+
+            const getAnalysisName = (fileData) => {
+                const fileList = fileData.map((f) => f['@id']);
+                const selectedAnalysis = compileAnalysis.find((c) => c.files.some((file) => (fileList || []).includes(file)));
```
forresttanaka: Minor, but isn’t `fileList` guaranteed to contain a truthy value? If it does, you can simplify this line a bit by excluding the `|| []` part.

src/encoded/static/components/filegallery.js
```diff
@@ -365,6 +496,23 @@ FileTable.procTableColumns = {
         },
         objSorter: (a, b) => fileAccessionSort(a, b),
     },
+    Default: {
```
forresttanaka: Any specific reason this is uppercase? I don’t see an AirBnB rule about it, but I feel like this might be the _only_ uppercase object property in our entire code base.

src/encoded/static/components/filegallery.js
```diff
@@ -2861,16 +2945,25 @@ class FileGalleryRendererComponent extends React.Component {
             let allFiles = datasetFiles.concat(relatedFiles);
             allFiles = this.filterForInclusion(allFiles);
 
+            const compiledAnalysis = compileAnalyses(context, allFiles);
+            const dropdown = document.querySelector('.analyses-selector select');
```
forresttanaka: It’d be great if you could use a ref here instead of `querySelector`. I know you’d have to cross components, but the lowest common parent can store the ref, and `AnalysesSelector` can pass its `<select>` ref back up to this parent so it can pass it to its relevant children.

src/encoded/static/components/filegallery.js
```diff
@@ -2861,16 +2945,25 @@ class FileGalleryRendererComponent extends React.Component {
...
+            const selectedIndex = dropdown ? dropdown.selectedIndex : null;
+            const compliedAnalysisFiles = selectedIndex ? compiledAnalysis[selectedIndex].files : null;
```
forresttanaka: `compliedAnalysisFiles` or `compiledAnalysisFiles`?

src/encoded/static/components/filegallery.js
```diff
@@ -2861,16 +2945,25 @@ class FileGalleryRendererComponent extends React.Component {
...
+
+            const filterFilesByAnanlysis = (files) => (!compliedAnalysisFiles
```
forresttanaka: You probably meant `filterFilesByAnalysis`?

src/encoded/static/components/filegallery.js
```diff
@@ -3256,26 +3350,97 @@ FileGalleryRendererComponent.contextTypes = {
...
+        const fileTypes = filters.file_type && filters.file_type.length > 0
+            ? filters.file_type.map((fileType) => `&files.file_type=${encodedURIComponent(fileType)}`).join('')
+            : '';
+        const biologicalReplicates = filters.biological_replicates && filters.biological_replicates.length > 0
+            ? filters.biological_replicates.map((replicate) => `&files.biological_replicates=${encodedURIComponent(replicate)}`).join('')
+            : '';
+        const outputTypes = filters.output_type && filters.output_type.length > 0
+            ? filters.output_type.map((o) => `&files.output_type=${encodedURIComponent(o)}`).join('')
+            : '';
+        const fileStatus = document.querySelector('[name="filterIncArchive"]').checked
```
forresttanaka: Should pass the `inclusionOn` state here. Again, I realize this crosses components here, but this bypasses React’s state management, so I’d consider this non-optional.

src/encoded/static/components/filegallery.js
```diff
@@ -3256,26 +3350,97 @@ FileGalleryRendererComponent.contextTypes = {
...
+        const outputCategoryQueryString = outputCategory ? `&files.output_category=${outputCategory}` : '';
+        const outputTypeQueryString = outputType ? `&files.output_type=${outputType}` : '';
+
+        if (outputCategoryQueryString || outputType) {
+            reactContext.navigate(`/batch_download/?type=${type}&accession=${accession}${outputCategoryQueryString}${outputTypeQueryString}${fileTypes}${outputTypes}${biologicalReplicates}${fileStatus}`);
+        } else {
+            reactContext.navigate(`/batch_download/?type=Analysis&@id=${analysisObjectId}&${fileTypes}${outputTypes}${biologicalReplicates}${fileStatus}`);
+        }
+        setDownloadModalVisibility(false);
+    };
```
forresttanaka, line 3488: Seems like a very large function to get re-parsed on each render (not that we have a specific criterium for that). Could you put it in the global space and pass in `context`, `filters`, `outputCategory`, `outputType`, and `inclusionOn`?

src/encoded/static/components/status.js
```diff
@@ -108,6 +108,7 @@ const defaultObjectStatuses = {
             'released',
             'archived',
             'revoked',
+            'preferred default',
```
forresttanaka: This section is only for file status, but “preferred default” can’t be a file status.

src/encoded/static/scss/encoded/modules/_panels.scss
```diff
@@ -726,6 +731,11 @@ $file-gallery-controls-height: 34px; // Height of controls inside file-gallery-c
     }
 }
 
+.collapsing-title-file-count {
+    font-size: 0.7em;
```
forresttanaka: Did you mean `rem`? We haven’t used ems because of their cascading problem.

src/encoded/static/scss/encoded/modules/_status.scss
```diff
@@ -12,6 +12,7 @@ $status-colors: (
     exempt-from-standards: #fff712,
     archived: #a0a0a0,
     revoked: $brand-danger,
+    preferred-default: #004d00,
```
forresttanaka: Again, I wouldn’t corrupt the existing status system. Could you put a star in the file graph node instead of using a different color? I think it seems confusing to call “preferred default” a status when it’s not, and to use different symbology for the file graph and file table. You can add U+2605 for the star that you can pass to the file graph text.

forresttanaka (changes requested):

src/encoded/static/components/filegallery.js
zoldello: I needed to loop through an array, stop when a desired item is found and then and transform. I could not use forEach() as there is no means to break from it when desired. So I used some(). But it is a hack that can be confusing, so I changed it to find() as you recommended.

src/encoded/static/components/filegallery.js
zoldello: Thanks.

src/encoded/static/components/filegallery.js
zoldello: Good point. I don't recall why I used uppercase and see no need for it. I corrected this.

src/encoded/static/components/filegallery.js
zoldello: Oops. I took out document.querySelector

src/encoded/static/components/filegallery.js
zoldello: Thanks.

src/encoded/static/components/filegallery.js
zoldello: Thanks.

src/encoded/static/components/filegallery.js
zoldello: Done. Thanks again. for the catch. Using document.querySelector() is very addictive.

src/encoded/static/components/filegallery.js
forresttanaka: Yeah, I figured that was the reason, but .find() stops once it finds a match too.

src/encoded/static/components/filegallery.js
zoldello: This was not changed as although there is common code with another function, it is not common enough to warrant a function that is passed. The function is also not so large or resource-heavy enough to warrant a function of its own either. Let me know if you now beg to differ.

src/encoded/static/components/status.js
zoldello: Done.

src/encoded/static/scss/encoded/modules/_panels.scss
zoldello: Done.

src/encoded/static/scss/encoded/modules/_status.scss
zoldello: I added the the star to the file graph-node name.

src/encoded/static/components/filegallery.js
forresttanaka: Agreed.

src/encoded/static/components/filegallery.js
```diff
@@ -56,21 +56,98 @@ const inclusionStatuses = [
...
+                        assemblyAnnotationValue: computeAssemblyAnnotationValue(analysesByAssembly[assembly][0].assembly, analysesByAssembly[assembly][0].genome_annotation),
+                        files: _.uniq(assemblyFiles),
+                    });
+                });
+            });
+        }
+    }
+    return _(compiledAnalyses).sortBy((compiledAnalysis) => -compiledAnalysis.assemblyAnnotationValue);
+};
+
```
forresttanaka, line 155: Do we still need all that logic to re-form the title for certain cases? Why doesn’t analysis.title work on its own? Does the back end need an update for its analysis.title creation so you don’t have to massage it?

FYI In ENCD-5701 I moved this function to its own file so that cart can use it too.

src/encoded/static/components/filegallery.js
```diff
@@ -220,19 +331,25 @@ export class FileTable extends React.Component {
                                 adminUser,
                             }}
                         />
-                        <SortTable
+                        {[analysisObjectKeys, otherKeys].map((keys) => keys.map((key) => <SortTable
```
forresttanaka: Very minor, but seems odd to see `<SortTable` way over on the right when it encapsulates everything to its left.

forresttanaka (changes requested):
Mostly a question or comment on compileAnalyses. I hadn’t heard analysis_object.title wasn’t good enough to use unmodified. I’ll approve without changes if this is needed.

src/encoded/static/components/filegallery.js
zoldello: I think the front-end has to massage the title given the requirements and the need to not tie the backend to specific UI needs but maintain flexibility. Let me know if you think different and I will adjust it.

src/encoded/static/components/filegallery.js
zoldello: I am not sure I understand this. The SortTable (line 223) is indeed on the same level as the code in line 334; they are siblings.

src/encoded/static/components/filegallery.js
forresttanaka: It sounds like you do have to massage the versions for the file gallery, but the cart doesn’t, so you do need to duplicate some of the work the back end does to form the `title` property.

src/encoded/static/components/filegallery.js
forresttanaka: Your line 334 has `<SortTable` after the arrow on the right side, but its props are way over on the left.

forresttanaka: I think it’d be less awkward to have:
```
{[analysisObjectKeys, otherKeys].map((keys) => keys.map((key) => (
    <SortTable
        prop1={prop1}
        prop2={props}
        ...
     >
```

zoldello: Done.  Thanks. Are you okay with the many commits thats part of this PR or do you prefer a rebase to just. get one commit to focus on?

forresttanaka: Many commits is fine, but whatever works for you.

zoldello:
@forresttanaka What change needs to be made?

forresttanaka:
> What change needs to be made?

Nothing I’m aware of.

zoldello:
> > What change needs to be made?
> 
> Nothing I’m aware of.

I meant code review-wise. The PR says you want something changed. But I do not see what.

forresttanaka:
You’ve already addressed them, the issue with `compileAnalysis` and `<SortTable>` formatting.

src/encoded/static/components/filegallery.js
```diff
@@ -115,20 +101,50 @@ export const compileAnalyses = (experiment, files) => {
                     // annotation and add each to the compiled list. Filter out any not included in
                     // the experiment's files.
                     const assemblyFiles = _.uniq(analysesByAssembly[assembly].reduce((accFiles, analysis) => accFiles.concat(analysis.files), []).filter((file) => fileIds.includes(file)));
-                    const { status } = analysesByAssembly[assembly][0];
+                    // eslint-disable-next-line camelcase
```
forresttanaka, line 107: Oh, I guess there's no choice than to disable that rule here. I can't think of an alternative either.

src/encoded/static/components/filegallery.js
```diff
@@ -71,7 +71,7 @@ const inclusionStatuses = [
  *      files: Array of files selected with the pipeline lab and assembly
  * }
  */
-export const compileAnalyses = (experiment, files) => {
+export const compileAnalyses = (experiment, files, dataFormat = null) => {
```
forresttanaka, line 77: `dataFormat` not documented. I’m not sure what its purpose is.

forresttanaka (dismissed):
I think this is OK. As I mentioned I don’t understand the purpose of the `dataFormat` parameter.

forresttanaka (dismissed):
Thanks for the documentation! I get it now.

src/encoded/static/components/filegallery.js
```diff
@@ -2525,38 +2528,36 @@ AnalysesSelector.propTypes = {
...
+    const selectedIndex = dropdown ? dropdown.selectedIndex : null;
+    const analysis = compiledAnalysis[selectedIndex];
+    const compileAnalysisFiles = analysis ? analysis.files : null;
+    let files = compileAnalysisFiles
+        ? allFiles.filter((f) => compileAnalysisFiles && compileAnalysisFiles.includes(f['@id']))
+        : allFiles;
+
+    if (currentTab === 'browser') {
+        files = filterForVisualizableFiles(files);
+    }
```
forresttanaka: It’s OK, but it seems like a shame to possibly filter `allFiles` and assign to `files`, just to throw that away if the browser tab is current. `allFiles` can potentially contain hundreds of thousands of files.

forresttanaka (dismissed):
OK, I think this is good to go. I feel like someday we should get the `FIleTable` props under control somehow.

forresttanaka (dismissed):

forresttanaka (approved):
I still have a question of why `compiledAnalyses` gets just one object with https://old-encd-5726-a3641cc-phil.demo.encodedcc.org/experiments/ENCSR000AKJ/ initially. It later gets updated to nine objects, when things seem to stabilize.

PR #3740, opened 2021-02-22
ENCD-5717-convert-cCRE-annotation-biochemical-inputs-property

iyoungworth:


gabdank (changes requested):
@iyoungworth can you please add a test that makes sure the calculated property is a list as expected?

iyoungworth:
No problem, there's now a new test (in new file test_types_dataset.py, since there weren't any others yet)

gabdank (approved):

PR #3741, opened 2021-02-22
ENCD-5749-make-bigbed9-collapsible

zoldello:


emro (dismissed):

src/encoded/static/components/genome_browser.js
```diff
@@ -733,9 +733,19 @@ class GenomeBrowser extends React.Component {
...
                 (['bedrnaelements', 'peptidemapping', 'bedexonscore', 'bed12', 'bed9'].indexOf(file.file_format_type.toLowerCase()) > -1)) {
+                const openedOutputTypes = [
+                    'predicted transcription start sites',
+                    'contigs',
+                    'splice junctions',
+                    'clusters',
+                    'exon quantifications',
+                    'unfiltered peptide quantification',
+                    'filtered peptide quantification',
+                ];
```
forresttanaka, line 744: I like to set static arrays in the global space so they only get assigned at page load instead of every render. These arrays aren’t _huge_ though, so I doubt it makes too much difference.

forresttanaka (approved):

PR #3745, opened 2021-02-25
ENCD-5743 add glossary link

emro:
Adds a link to GO annotations on the target categorization page to the glossary header "Target categories". It could be extended if we want to append links to other headers as well -- not really sure if we expect to want to do that.

### DIFF
--- a/src/encoded/static/components/glossary.js
+++ b/src/encoded/static/components/glossary.js
@@ -18,9 +18,10 @@ const GlossaryTerm = (props) => {
             {props.glossaryEntry.additional_information ?
                 <p className="glossary-extra-info">
                     <span className="glossary-label">Additional information: </span>
-                    {props.glossaryEntry.additional_information.map((href) => (
+                    {props.glossaryEntry.additional_information.map((href, hrefIndex) => (
                         <span className="glossary-field" key={href.url}>
                             <a href={href.url}>{href.title}</a>
+                            <span>{((props.glossaryEntry.additional_information.length > 1) && hrefIndex < (props.glossaryEntry.additional_information.length - 1)) ? ', ' : ''}</span>
                         </span>
                     ))}
                 </p>
--- a/src/encoded/static/components/glossary.js
+++ b/src/encoded/static/components/glossary.js
@@ -33,6 +33,14 @@ GlossaryTerm.propTypes = {
     glossaryEntry: PropTypes.object.isRequired,
 };
 
+// Optional links appended to section headers
+const sectionHeaderLinks = {
+    'Target categories': {
+        link: 'https://www.encodeproject.org/target-categorization/',
+        text: 'Category assignment details',
+    },
+};
+
 const Glossary = (props) => {
     const { context } = props;
     const glossary = require('./glossary/glossary.json');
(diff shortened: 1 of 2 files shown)


src/encoded/static/components/glossary.js
```diff
@@ -33,6 +33,14 @@ GlossaryTerm.propTypes = {
     glossaryEntry: PropTypes.object.isRequired,
 };
 
+// Optional links appended to section headers
+const sectionHeaderLinks = {
+    'Target categories': {
+        link: 'https://www.encodeproject.org/target-categorization/',
```
forresttanaka, line 40: Is this supposed to be an internal link or external link? With this full URL you’ll get the external-link icon with the arrow after the link, and I wasn’t sure that’s the intention or not.

forresttanaka (dismissed):
Looks like a good way to do this. Just one question on internal/external links, which your CSS change implies is intentional, but I didn’t see anything in the ticket about it either way.

src/encoded/static/components/glossary.js
emro: Well, it is an internal link so perhaps we should not be using that icon. We're using this icon for all the links on the glossary page, which are all internal links, so maybe they should all be changed. Looking around on Encode I think we're very inconsistent about our use of the external link icon, sometimes we use it, sometimes we don't. Is there another icon we use for internal links?

forresttanaka: Internal links get no icon, like the blue text in the search results. Anything in the code with a path and no protocol display no icon (e.g. '/target-categorization/'), but any with a full URL including protocol get the icon automatically, which is why this one displays that icon.

forresttanaka: Hmm, I see one inconsistency so far: the SCREEN link on the home page should have that icon because the link has a full URL, but the .site-search__screen a:after style deletes it.

emro: There are a lot of inconsistencies on the help pages.

forresttanaka: Oh really? People who edit the pages probably sometimes use paths and sometimes full URLs. Should probably tell everyone to make sure to use paths for on-site targets.

forresttanaka: I’ll write a WRAN ticket for that, and bring it up in their channel.

src/encoded/static/components/glossary.js
```diff
@@ -33,6 +34,14 @@ GlossaryTerm.propTypes = {
...
```
forresttanaka, line 40: Oh, I get it now. We have more than one reason these little icons are inconsistent. You fully specify the URL above so in most cases that would cause the external-link icon to appear, but because this page isn’t under a `<div class=".panel">`, that icon style doesn’t apply. But doc pages don’t have `<div class=".panel">` either, but still that icon shows. That’s because they appear inside a `<table>`, which enables that icon again.

This needs multiple levels of clean-up, in the doc pages, in the CSS, in our content. We’ll have to worry about that another day.

forresttanaka (approved):

PR #3749, opened 2021-02-28
ENCD-5767-fix-file-details-facets-viz-files

zoldello:


forresttanaka (approved):
Sounds good for an RC2 IMO.

PR #3753, opened 2021-03-02
ENCD-5744 DOI

forresttanaka:
While this ticket concerned itself with dataset pages, the Copy button it uses overlapped the functionality of the Copy button in the Share Cart modal which used an older method of copying. I decided to modify the Share Cart modal to share the new Copy button on the dataset pages.

The .btn-svgicon CSS class wasn’t used, so I deleted it.

I had a leftover Jest test experiment now in v113rc1. It affects nothing, but I’ll remove it here.


forresttanaka:
Updating based on feedback from Ben and Bonita. Will open a new PR when I have the update ready.

PR #3755, opened 2021-03-03
ENCD-5742-fix-valis-track-sticking

zoldello:


emro (approved):
Looks great. This is a very good solution to the issue.

PR #3756, opened 2021-03-03
ENCD-5780 Fix cart analysis crash

forresttanaka:
Was trying to dereference files array when no files array exists. This is for v113rc2.


zoldello (approved):
Change looks good.

PR #3757, opened 2021-03-04
ENCD-5782 Restore assay_term_name to browser label

forresttanaka:
I had changed the configuration object to retrieve the assay title instead, which left me feeding no assay term name to the genome browser. I now retrieve the assay term name though I don’t display it on the cart page, so I can pass it to the genome browser.


zoldello (approved):
Looks good.

PR #3759, opened 2021-03-04
ENCD-5758 Add biological replicate to cart facets

forresttanaka:
I added a method to convert the biological replicate array into a displayable, comma-separated string, so I added a new `getValue` transform that any facet can use if needed. So far, we only need it for the biological replicates.

### DIFF
--- a/src/encoded/static/components/cart/cart.js
+++ b/src/encoded/static/components/cart/cart.js
@@ -71,7 +71,7 @@ const assemblySorter = (facetTerms) => (
  *
  * @return {array} Same as `facetTerms`
  */
-const analysisSorter = (facetTerms, files, analyses) => (
+const analysisSorter = (facetTerms, analyses) => (
     _(facetTerms).sortBy((facetTerm) => (
         analyses.findIndex((analysis) => analysis.title === facetTerm.term)
     ))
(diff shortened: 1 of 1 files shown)


src/encoded/static/components/cart/cart.js
```diff
@@ -66,10 +66,12 @@ const assemblySorter = (facetTerms) => (
  * Sorter function for analyses facet terms matches the order from the given analyses which have
  * already been sorted.
  * @param {array} facetTerms Analysis facet terms to sort
+ * @param {array} files Files contributing to facets; unused here
+ * @param {array} analyses Analyses from selected compiled analyses objects
  *
  * @return {array} Same as `facetTerms`
  */
-const analysisSorter = (facetTerms, analyses) => (
+const analysisSorter = (facetTerms, files, analyses) => (
```
zoldello: Why is there "files" as a parameter when it is not used? Your jsdoc comment does not explain why.

src/encoded/static/components/cart/cart.js
```diff
@@ -1428,7 +1454,7 @@ const assembleFacets = (selectedTerms, files, analyses, usedFacetFields) => {
             // built from it, so no not-found condition needs checking.
             const facetDisplay = usedFacetFields.find((facetField) => facetField.field === facet.field);
             facet.title = facetDisplay.title;
-            facet.terms = facetDisplay.sorter ? facetDisplay.sorter(facet.terms, analyses) : _(facet.terms).sortBy((facetTerm) => facetTerm.term.toLowerCase());
+            facet.terms = facetDisplay.sorter ? facetDisplay.sorter(facet.terms, files, analyses) : _(facet.terms).sortBy((facetTerm) => facetTerm.term.toLowerCase());
```
zoldello: Image 3-5-21 at 3 09 PM
I looked at sorter function and it does not take in files...?

zoldello (commented):
My comments are questions. They center on "files" as a parameter.

src/encoded/static/components/cart/cart.js
forresttanaka: Hard to know what future sorting functions need, and I figured files was an overwhelmingly plausible one. It’s like SortTable, where most of the optional functions don’t use all the given parameters.

zoldello (changes requested):
It would be better to have an alternative to having parameters not used...

forresttanaka:
Closing to address a crash.

PR #3760, opened 2021-03-04
ENCD-5744 Display DOI of all datasets

forresttanaka:
The dbxref code had gotten a little bit messy in cases where you only needed a subset of functionality, so I refactored the dbxref functions to reduce duplication, and allow clients to pick what functionality they want — a whole list of dbxrefs, just one, or just the dbxref link.

While this ticket concerns itself with dataset pages, the Copy button it now uses overlapped the functionality of the Copy button in the Share Cart modal which used an older method of copying. I decided to modify the Share Cart modal to share the new Copy button on the dataset pages. Since I was modifying that area, I also fixed the lack of focus on the modal’s submit button when using a screen reader, and converted that component to a functional component since it’s not too large.

The .btn-svgicon CSS class wasn’t used, so I deleted it.

I had a leftover Jest test experiment now in v113rc2. That experiment affects nothing, but I’ll remove it here.

### DIFF
--- a/src/encoded/static/components/cart/share.js
+++ b/src/encoded/static/components/cart/share.js
@@ -6,74 +6,53 @@ import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import url from 'url';
 import { Modal, ModalHeader, ModalBody, ModalFooter } from '../../libs/ui/modal';
+import { CopyButton, useMount } from '../objectutils';
 
 
 /**
  * Internal component to display and process the modal used to share a cart URL. The copyable cart
  * URL gets placed into a read-only <input> element.
  */
-class CartShareComponent extends React.Component {
-    constructor() {
-        super();
-        this.copyUrl = this.copyUrl.bind(this);
-    }
+const CartShareComponent = ({ userCart, locationHref, closeShareCart, inProgress }) => {
+    /** Modal Visit Sharable Cart button */
+    const submitRef = React.useRef(null);
 
-    /**
-     * Called when the user clicks the Copy button to copy the URL to the clipboard.
-     */
-    copyUrl() {
-        // Get the URL text <input> element in the DOM and select all of the text in it to copy to
-        // the user's clipboard.
-        // https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Interact_with_the_clipboard#Using_execCommand()
-        this.cartUrlBox.select();
+    // Generate the shared cart URL.
+    const parsedUrl = url.parse(locationHref);
+    Object.assign(parsedUrl, {
+        pathname: userCart['@id'],
+        search: '',
+        query: '',
+    });
+    const sharableUrl = url.format(parsedUrl);
 
-        // Execute copy command. Firefox can throw errors on rare occasion. As this is so unusual,
-        // we just show a console warning in that case.
-        try {
-            document.execCommand('copy');
-        } catch (err) {
-            console.warn('Text copy failed.');
-        }
+    useMount(() => {
+        // Focus on the Visit Sharable Cart button on mount.
+        submitRef.current.focus();
+    });
 
-        // Remove the selection after copying.
-        this.cartUrlBox.setSelectionRange(0, 0);
-    }
-
-    render() {
-        const { userCart, locationHref, closeShareCart, inProgress } = this.props;
-
-        // Generate the shared cart URL.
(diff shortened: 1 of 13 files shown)


src/encoded/static/components/cart/share.js
```diff
@@ -6,74 +6,53 @@ import PropTypes from 'prop-types';
...
-                    cancelTitle="Close"
-                    submitBtn={<a data-bypass="true" disabled={inProgress} target="_self" className="btn btn-info" href={sharableUrl}>Visit sharable cart</a>}
-                    closeId="share-cart-close"
-                />
-            </Modal>
-        );
-    }
-}
+const CartShareComponent = ({ userCart, locationHref, closeShareCart, inProgress }) => {
+    /** Modal Visit Sharable Cart button */
```
emro: I think "sharable" should be spelled "shareable" throughout this component.

src/encoded/static/scss/encoded/modules/_common_item.scss
```diff
@@ -15,3 +15,38 @@
         flex: 0 1 auto;
     }
 }
+
+// doi: (Digital Object Identifier) references, including Copy button.
+$doi-ref-border-color: #a0a0a0;
```
emro, line 30: Just a couple of thoughts on the style here -- I think it would look better with a little more padding/margin below the title / above the doi reference. Also, is there a reason that you have square borders? I think most of our buttons and borders are slightly rounded.

src/encoded/static/libs/svg-icons.js
```diff
@@ -115,6 +115,15 @@ const chevronRight = () => (
...
 );
 
+const clipboard = (style) => (
+    <svg version="1.1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 249 249" style={style} className="svg-icon svg-icon-clipboard">
+        <path d="M164,24.5h26.54c6.31,0,11.46,5.05,11.46,11.23v190.05c0,6.17-5.16,11.23-11.46,11.23H57.46
+            C51.16,237,46,231.95,46,225.77V35.73c0-6.17,5.16-11.23,11.46-11.23H84"
+        />
+        <rect x="84" y="11" width="80" height="46" />
+    </svg>
+);
```
emro, line 125: I like this new icon.

emro (commented):
Looks good! Just very minor comments.

src/encoded/static/components/cart/share.js
forresttanaka: I’ve always used the older US spelling, though apparently the one with the “e” [appears more](https://grammarist.com/spelling/sharable-shareable/) now. The macOS spell checker doesn’t ding “sharable.” I should, perhaps, start trying to get used to the one with an “e” going forward.

src/encoded/static/scss/encoded/modules/_common_item.scss
forresttanaka: I did the square corners deliberately to avoid people thinking the whole thing was a button. I originally had no border around the doi text, but that made the button looks weird to me, standing on its own after the text. Putting the button to the left of the doi text looked better but it felt strange to have the Copy button before the text it refers to. And I couldn’t make the whole thing a button because people might want to copy the doi text. Any other suggestions? I didn’t know what else to do, and this seemed not that confusing to me.

I could add more space below the title. I minimized the space because I thought it likely Ben would want less space.

src/encoded/static/components/cart/share.js
emro: Very interesting, I didn't know that this is a spelling that has changed over time and is particular to the US. I wasn't actually sure what the correct spelling was so I googled it and saw "shareable". Seems like either way is good but maybe an "e" is slightly better.

forresttanaka:
I find the Github Copy button annoying in that when you click it you get the barest of indications that it noticed your click, which I had solved in this ticket by having “Copy” change to “Copied.” Now without any text, I decided to have a flash of green when you click, so you should see that here with the DOI Copy button and the Share Cart Copy button. I set up a test experiment at:

https://encd-5744-doi-fytanaka.demo.encodedcc.org/experiments/ENCSR492FKD/

I also deployed a clean demo.

I changed all “sharable” to “shareable” which still looks very UK to me but I’ll get used to it.

emro (approved):
I think this looks great! I like the extra space and I think the icon is self-explanatory. The flash of green works really well, it's an excellent visual way to show that something (copying) was executed.

PR #3763, opened 2021-03-05
ENCD-5783-deprecrated-files-crashing-file-details

zoldello:


src/encoded/static/components/filegallery.js
```diff
@@ -3034,14 +3044,19 @@ class FileGalleryRendererComponent extends React.Component {
...
-                ? this.getBrowserSelectedIndex()
-                : dropdown ? dropdown.selectedIndex : null;
+            let selectedIndex = 0;
+            let compiledAnalysis = compileAnalyses(context, allFiles, 'choose analysis');
+
+            if (currentTab === 'browser') {
+                compiledAnalysis = compiledAnalysis.filter((c) => {
+                    const analysisFiles = filterForVisualizableFiles(c.files.map((f) => allFiles.find((aF) => aF['@id'] === f)));
+                    return analysisFiles.length > 0;
+                });
```
forresttanaka: Seems like an exact copy of `this.getBrowserAnalysis()`. You couldn’t call that here?

forresttanaka (changes requested):
Looks like the functionality change is `compileAnalysis` gets called regardless of the tab, but the rest is the same?

src/encoded/static/components/filegallery.js
zoldello: Corrected.

forresttanaka (approved):
Looks much simpler; thanks!

PR #3766, opened 2021-03-08
ENCD-5740 fix mouse colors

emro:
Updates in this ticket:
- colors are adjusted for higher contrast
- blood vessel and vein are added as organ slims and to circulatory system
- testis organ is moved away from mouse skin of body
- paths that were incorrectly assigned to organs are fixed
- the entire mouse body does not highlight on “skin of body”


src/encoded/static/scss/encoded/modules/_mouse_map.scss
```diff
@@ -1,20 +1,40 @@
...
             fill: #ccbdb4;
         }
     }
 
+    .cls-mousebody {
+        stroke: #85a2b9;
+        fill: #d8e2ea;
+
+        &.active {
+            // fill: #ccbdb4;
```
zoldello: Why not remove this comment as it serves no purpose?

zoldello (dismissed):
Code looks good. My comment is not a show-stopper.

src/encoded/static/scss/encoded/modules/_mouse_map.scss
emro: I guess I'm a little torn on this one. I shouldn't be storing colors in comments, but I think there's a chance at some point we'll want to highlight the mouse body and this is the color for it.

emro:
Meenakshi pointed out a bug with disabling paths that were ellipses so I fixed that, and the commented out style.

zoldello (dismissed):
Code looks good. The changed for the pointer-events is a good catch.

zoldello (approved):
Looks good.

PR #3767, opened 2021-03-08
ENCD-5793 Fix cart downloads for selected analysis and default files

forresttanaka:


forresttanaka:
The problem with downloading files with analysis selected is that I expected an array of @ids in file.analyses — I’m assuming that’s how they appeared when I started on this. Later they became embedded and I wasn’t aware of this. I now request files.analyses.@id.

The other was that the “preferred_default=true” prevents the All and Raw downloads on the cart-view page from downloading enough files. I now remove that from the query string if the user selects either of these download types.

zoldello (dismissed):
Code looks okay.

forresttanaka:
@zoldello Yunhai found another issue, and also a much easier way to do this now that file.analyses is embedded.

src/encoded/static/components/cart/batch_download.js
```diff
@@ -76,7 +76,7 @@ const batchDownload = (
     setInProgress(true);
     const visualizableOption = `${visualizable ? '&option=visualizable' : ''}`;
     const rawOption = `${raw ? '&option=raw' : ''}`;
-    const preferredDefaultQuery = preferredDefault ? '&files.preferred_default=true' : '';
+    const preferredDefaultQuery = preferredDefault && !raw && !all ? '&files.preferred_default=true' : '';
```
zoldello, line 79: This is odd. I thought "!raw && !all" this was added in the last PR.

zoldello (approved):
Code looks good.  I checked the UI for the download mismatch and the download count is as expected now.

PR #3773, opened 2021-03-10
ENCD-5775-update-biosample-summary

jenjou:


jnau (changes requested):
for a cleaner way to switch to the use of f-string (which is definitely good), we should change this entire for loop to use f-string instead of format.

jnau (dismissed):

jnau (approved):

PR #3776, opened 2021-03-12
ENCD-5718-add-rna-get-ui

pedrohr:
https://encodedcc.atlassian.net/browse/ENCD-5718

demo available on: https://encd-5718-1a48d3f51-pedroassis.demo.encodedcc.org/rnaget


ini-templates/production-template.ini
```diff
@@ -29,6 +30,8 @@ local_storage_port = 6379
 local_storage_redis_index = 1
 local_storage_timeout = 5
 
+genomic_data_service = http://data-service.demo.regulomedb.org
```
zoldello, line 33: How will this get started? Is it acceptable that test and demo use this too?

ini-templates/development-template.ini
```diff
@@ -29,6 +29,8 @@ local_storage_port = 6378
 local_storage_redis_index = 1
 local_storage_timeout = 5
 
+genomic_data_service = http://localhost:5000
```
zoldello, line 32: How is this started? Like, is there documentation?

src/encoded/genomic_data_service.py
```diff
@@ -0,0 +1,40 @@
+import requests
+
+
+class GenomicDataService():
+    REGISTRY_DATA_SERVICE = 'genomic_data_service'
```
zoldello: Why not put these constants outside of the class?

src/encoded/genomic_data_service.py
```diff
@@ -0,0 +1,40 @@
...
+
+    RNA_GET_FILTERS = ['assayType', 'annotation']
+    RNA_GET_ENDPOINT = '/expressions/bytes'
+
+    def __init__(self, registry, default_search):
+        self.path = registry.settings.get(GenomicDataService.REGISTRY_DATA_SERVICE)
+        self.default_search = default_search
+
+    def rna_get(self, genes, sort, units='tpm', page=1, filters=[]):
```
zoldello: It is normally a bad idea to initial a parameter with [] unless you know what you are doing and must. It is best to set it to None. See- https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument

src/encoded/genomic_data_service.py
```diff
@@ -0,0 +1,40 @@
...
+        params = []
+
+        search = self.default_search
+        if genes:
+            search = ",".join([gene.strip() for gene in genes.split(",")])
```
zoldello: What if genes is None? The method is public so you need to at consider bad data.

zoldello: The standard is to use single quotes rather than double. It would be a good idea to add pylint to CircleCI to catch things like this.

src/encoded/genomic_data_service.py
```diff
@@ -0,0 +1,40 @@
...
+            params.append(f"sort={sort}")
+
+        if page:
+            params.append(f"page={page}")
+
+        for filter_ in GenomicDataService.RNA_GET_FILTERS:
+            if filters.get(filter_):
+                params.append(f"{filter_}={filters.get(filter_)}")
+
+        url = self.path + GenomicDataService.RNA_GET_ENDPOINT + '?format=json&' + '&'.join(params)
```
zoldello: Why not use f-string here too?

src/encoded/genomic_data_service.py
```diff
@@ -0,0 +1,40 @@
...
+
+        results = requests.get(url, timeout=3).json()
```
zoldello, line 92: Why 3? A comment can help readers know why.

src/encoded/search_views.py
```diff
@@ -711,3 +714,94 @@ def top_hits(context, request):
...
+    page  = request.params.get('page')
+
+    filters = {
+        'assayType': request.params.get('assayType'),
+        'annotation': request.params.get('annotation')
+    }
+
+    filters_data = [{'field': 'type', 'term': 'Rna Get', 'remove': '/rnaget'}]
+
+    for f in filters:
```
zoldello: 'filter_' instead of 'f'?

src/encoded/search_views.py
```diff
@@ -711,3 +714,94 @@ def top_hits(context, request):
...
+        if filters.get(f):
+            query_string = []
+
+            for param in request.params:
```
zoldello: This is a bit confusing. It searches for the last instance of f !=param and use that. Why not directly search for it and stopped when found rather than loop for it through the entire list?

src/encoded/search_views.py
```diff
@@ -711,3 +714,94 @@ def top_hits(context, request):
...
+                if param != f:
+                    query_string.append(f'{param}={request.params.get(param)}')
+
+            filters_data.append({
+                'field': f,
+                'term': filters[f],
+                'remove': f'{request.path}?{"&".join(query_string)}'
```
zoldello: Will there be a problem is query_string is empty?

src/encoded/search_views.py
```diff
@@ -711,3 +714,94 @@ def top_hits(context, request):
...
+        'featureID': {'title': 'Feature ID'},
+        'tpm': {'title': 'Counts (TPM)'},
+        'fpkm': {'title': 'Counts (FPKM)'},
+        'assayType': {'title': 'Assay (RNA SubType)'},
+        'libraryPrepProtocol': {'title': 'Experiment'},
+        'expressionID': {'title': 'File'},
+        'annotation': {'title': 'Genome Annotation'}
+    }
+
+    if units == 'tpm':
```
zoldello: what is neither case is satisfied? In theory it can. In cases like this I normally like testing one case and default if it is false, like

`column.pop('fpkm' if units == 'tmp' else 'tpm')`

But this is just me.

src/encoded/search_views.py
```diff
@@ -711,3 +714,94 @@ def top_hits(context, request):
...
+        columns.pop('fpkm')
+    elif units == 'fpkm':
+        columns.pop('tpm')
+
+    data_service = GenomicDataService(context.registry, default_search='ENSG00000088320.3')
```
zoldello: Why default_search='ENSG00000088320.3'?

src/encoded/search_views.py
```diff
@@ -711,3 +714,94 @@ def top_hits(context, request):
...
+        'non_sortable': ['annotation'],
+        'columns': columns,
+        'filters': filters_data,
+        'facets': facets_data,
+        '@graph': expressions
+    }
+
+    if sort:
+        response['sort'] = {}
+
```
zoldello: Why this space?

src/encoded/static/scss/encoded/modules/_matrix.scss
```diff
@@ -1091,6 +1091,24 @@ table.matrix th.group-all-groups-cell {
     margin: 10px;
 }
 
+/* stylelint-disable-next-line selector-class-pattern */
```
zoldello, line 1094: Is there a reason why you are ignoring the class name pattern (https://stylelint.io/user-guide/rules/selector-class-pattern)? 

The css naming style ENCODE uses is BEM (https://scalablecss.com/bem-quickstart-guide/). The naming style you use does not match the style... It seems to me.

src/encoded/static/components/rnaget.js
```diff
@@ -0,0 +1,889 @@
...
+import { encodedURIComponent } from '../libs/query_encoding';
+import * as Pager from '../libs/ui/pager';
+import QueryString from '../libs/query_string';
+import * as globals from './globals';
+import { TableItemCount } from './objectutils';
+import { FacetList, TextFilter } from './search';
+
+
+/**
+ * Default number of results when no "limit=x" specified in the query string. Determined by our
```
zoldello, line 15: This should be: 
 
Default number of results when no "limit=x" **is** specified in the query string. It is determined by the rna ui client search code.

Something along these. lines. The backend is huge so just mentioning it is like telling someone their car key is somewhere in North America...

src/encoded/static/components/rnaget.js
```diff
@@ -0,0 +1,889 @@
...
+    static _generateKey(field, type) {
+        return `${field}-${type || '*'}`;
+    }
+
+    /**
+     * Register a React component to render a type-field or *-field pair.
+     * @param {object} {field, type} Field and type to register component for; type optional
+     * @param {array} component Rendering component to call for this field-type value
+     */
+    register({ field, type }, component) {
```
zoldello, line 54: Let's talk about this.

src/encoded/static/components/rnaget.js
```diff
@@ -0,0 +1,889 @@
...
+        const key = RNAGetDataRegistryCore._generateKey(field, type);
+        if (this._registry[key]) {
+            // Registering key already exists.
```
zoldello, line 57: The code makes the comment unnecessary.

src/encoded/static/components/rnaget.js
```diff
@@ -0,0 +1,889 @@
...
+            console.warn('RNAGetDataRegistry collision %s:%s', field, type);
```
zoldello, line 58: We tend not to push console messages to github. But there is no rule against it.

src/encoded/static/components/rnaget.js
```diff
@@ -0,0 +1,889 @@
...
+const MAX_VIEWABLE_RESULTS = 99999;
+
+
+/**
+ * A single instance of this class gets used to register special-case displays of data in a cell in
+ * cases where the default renderer won't do. You can register rendering components for a specific
+ * field regardless of the "type=x" type, or you can register one for a specific field only when
+ * a specific "type=x" type gets displayed.
+ */
+class RNAGetDataRegistryCore {
```
zoldello, line 32: Do you need a class for this when an object would suffice? Classes are computationally expensive and an object can have functions.

src/encoded/static/components/rnaget.js
```diff
@@ -0,0 +1,889 @@
...
+/**
+ * @id cell renderer that displays the accession and links to the object.
+ */
+const AtIdRenderer = ({ value }) => (
+    <a href={value}>{value.split('/').pop()}</a>
+);
+
+AtIdRenderer.propTypes = {
+    /** @id value to render */
+    value: PropTypes.any.isRequired,
```
zoldello, line 90: Let's talk about this. The type seems to be a string always. Using pop is valid but something I have never seen anyone do on it. People normally do things like: a[1].

src/encoded/static/components/rnaget.js
```diff
@@ -0,0 +1,889 @@
...
+
+RNAGetDataRegistry.register({ field: 'accession' }, AccessionRenderer);
+RNAGetDataRegistry.register({ field: 'title', type: 'File' }, AccessionRenderer);
+
+
+/**
+ * `related_series` renderer for type=Experiment. Displays a comma-separated list of their @ids.
+ */
+const ExperimentRelatedSeriesRenderer = ({ value }) => {
+    if (value && typeof value === 'object' && Array.isArray(value) && value.length > 0) {
```
zoldello, line 124: Let's talk about this. This is a code smell.

src/encoded/static/components/rnaget.js
```diff
@@ -0,0 +1,889 @@
...
+/**
+ * Get the property specified by a field from the search results and return a string from that.
+ */
+const lookupColumn = (result, column) => {
+    let nodes = [result];
+    const names = column.split('.');
+
+    for (let i = 0, len = names.length; i < len && nodes.length > 0; i += 1) {
+        let nextnodes = [];
+        _.each(nodes.map((node) => node[names[i]]), (v) => {
```
zoldello, line 216: It is okay to use underscore to loop. But it's less common in the ES2015+ world as its best to use in-built functionality which are faster and IE11 is as good as gone.  People normally use something like every() or some().

src/encoded/static/components/rnaget.js
```diff
@@ -0,0 +1,889 @@
...
+            if (v === undefined) return;
```
zoldello, line 217: Let's talk about this. Doing v === undefined is a code smell. You can just do "!v". However,  it is valid if you want to check if v has never been defined. I don't think that is your intent. 

It is also bad practice in Javascript to put return on the same line as an if-statement.

src/encoded/static/components/rnaget.js
```diff
@@ -0,0 +1,889 @@
...
+            if (Array.isArray(v)) {
```
zoldello, line 218: Code smell. In Javascript, you search for functionality like length. You treat object and arrays as collections and like they act the same way. When someone looks specifically for types, my spider senses start tingling.

src/encoded/static/components/rnaget.js
```diff
@@ -0,0 +1,889 @@
...
+            // can't get it that way. Just return the length of the array.
+            nodes = [nextnodes.length];
+            break;
+        } else {
+            // Moving on to the next node defined by the `names` array.
+            nodes = nextnodes;
+        }
+    }
+    // if we ended with an embedded object, show the @id
+    if (nodes.length > 0 && nodes[0]['@id'] !== undefined) {
```
zoldello, line 235: Let's talk about this.

src/encoded/static/components/rnaget.js
```diff
@@ -0,0 +1,889 @@
...
+        nodes = nodes.map((node) => node['@id']);
+    }
+
+    // Stringify any nodes that are objects or arrays. Objects and arrays have typeof `object`.
```
zoldello, line 239: "Stringify any nodes that are **collections**."

Everyone knows most things in Javascript are objects. The CEO of Brave-browser who designed Javascript was obsessed with them. So there is no need to tell developers.

src/encoded/static/components/rnaget.js
```diff
@@ -0,0 +1,889 @@
...
+    if (nodes.length > 0) {
+        nodes = nodes.map((item) => (typeof item === 'object' ? JSON.stringify(item) : item));
+    }
+
+    return _.uniq(nodes).join(', ');
```
zoldello, line 244: _.uniq is okay and Forrest likes it. But I personally prefer to use set. Set is inbuilt so its runs faster.

src/encoded/static/components/rnaget.js
```diff
@@ -0,0 +1,889 @@
...
+};
+
+
+/**
+ * Render each cell of the table header. This exists as a component mostly to reduce code
```
zoldello, line 249: No need to tell developers you are trying to reduce code duplication. It is implied.

src/encoded/static/components/rnaget.js
```diff
@@ -0,0 +1,889 @@
...
+    const extraFields = getExtraFields(allColumns, visibleFields);
+
+    const toggleColumn = (clickedField) => {
+        let updatedColumns;
+
+        // Called every time a column's checkbox gets clicked on or off in the modal.
+        if (selectedFields.includes(clickedField)) {
+            // The clicked column is currently visible, so before we make it invisible, make sure
+            // at least one other column is also visible.
+            if (Object.keys(selectedFields).length === 1) {
```
zoldello, line 569: === 1 or > 0?

src/encoded/static/components/rnaget.js
```diff
@@ -0,0 +1,889 @@
...
+                } else {
+                    limitQuery.replaceKeyValue('limit', limit);
+                }
+
+                return (
+                    <a
+                        key={limit}
+                        href={`?${limitQuery.format()}`}
+                        className={`page-limit-selector__option${limit === pageLimit ? ' page-limit-selector__option--selected' : ''}`}
+                        aria-label={`Show ${limit} items per page`}
```
zoldello, line 731: This need to be updated handle the case where there is 1. In that case the 's' in the end should be removed.

src/encoded/static/components/rnaget.js
```diff
@@ -0,0 +1,889 @@
...
+    /** Current page's QueryString query */
+    query: PropTypes.object.isRequired,
+};
+
+/**
+ * Renderer for the entire /report/ page.
+ */
+const RNAGet = ({ context }, reactContext) => {
+    /** All possible columns for the current type */
+    const [allColumns] = React.useState(null);
```
zoldello, line 753: Does this ever need updating? It's rhetorical as the code tells me know albeit it seems odd.

zoldello (changes requested):
Let's talk about some things when you are available. See my comments for what they are.

src/encoded/search_views.py
```diff
@@ -711,3 +714,94 @@ def top_hits(context, request):
         ]
     )
     return fr.render()
+
+
+@view_config(route_name='rnaget', request_method='GET', permission='search')
```
keenangraham, line 719: Can we move logic to its own module/file and just expose entry point here? Will allow for cleaner development in the future if we have encapsulated space for helper functions/tests and can modify functionality without editing this file.

keenangraham (changes requested):

ini-templates/development-template.ini
pedrohr: yes, there is: https://github.com/ENCODE-DCC/genomic-data-service

it should be only used for the /rnaget page though.

ini-templates/production-template.ini
pedrohr: the /rnaget is an alpha version so people can start seeing the data we have. We still don't have an official stable Data Services server. It's ok for test and demo to use this.

src/encoded/genomic_data_service.py
pedrohr: interesting! I didn't know about this weird behavior, I will push a change

src/encoded/search_views.py
pedrohr: yes, this was on my to-do list, it just got de-prioritized as I was rushing to get this into this sprint. I will push a refactor

src/encoded/search_views.py
pedrohr: this is a very ugly temporary workaround regarding the interface. We need to add gene name autocomplete to the interface so it's easier to search for genes. We decided to populate with a random value so people can at least start using it. I tried to move this to the frontend but it was not simple with the current architecture.

src/encoded/genomic_data_service.py
pedrohr: there's no methodical choice for the timeout value, usually, 3-5 seconds of timeout is the default choice and then adjusted based on the behaviour of users + production architecture

src/encoded/genomic_data_service.py
pedrohr: if `genes` is `None` we will use the `self.default_search` value for searching. Related to a workaround for the alpha version of this interface

src/encoded/search_views.py
pedrohr: we are looking to remove the query string and its value from the url, it can be placed in any position and also multiple times. This reproduces the snovault behavior.

src/encoded/search_views.py
pedrohr: the default value is []
"&".join([]) => ''
it should be ok

src/encoded/search_views.py
pedrohr: the interface should only try `tpm` and `fpkm`. If users decide to submit another value, the endpoint should fail. As we don't have a final interface for this and this won't be advertised, we just skipped the treatment for invalid parameters for now.

I like the one-line solution, I will replace it

src/encoded/static/scss/encoded/modules/_matrix.scss
pedrohr: ok, I will take a look

src/encoded/static/components/rnaget.js
pedrohr: Yeah, this comes from the report.js. I just duplicated the object and edited it.

zoldello (approved):
The issues Pedro could address were addressed. I will talk to Forrest about code Pedro got from report.js.

src/encoded/genomic_data_service.py
```diff
@@ -0,0 +1,136 @@
...
+RNA_GET_COLUMNS = {
+    'featureID': {'title': 'Feature ID'},
+    'tpm': {'title': 'Counts (TPM)'},
+    'fpkm': {'title': 'Counts (FPKM)'},
+    'assayType': {'title': 'Assay (RNA SubType)'},
+    'libraryPrepProtocol': {'title': 'Experiment'},
+    'expressionID': {'title': 'File'},
+    'annotation': {'title': 'Genome Annotation'}
+}
+
```
keenangraham, line 17: Some formatting nits, not a big deal. Usually try to do two lines between function, one line between methods, and remove extra lines.

https://www.python.org/dev/peps/pep-0008/#blank-lines

src/encoded/genomic_data_service.py
```diff
@@ -0,0 +1,136 @@
...
+        self.parse_params()
+
+
+    def parse_params(self):
+        params = self.request.params
+
+        self.genes   = params.get('genes')
+        self.units   = params.get('units', 'tpm')
+        self.page    = params.get('page', 1)
+        self.sort    = params.get('sort')
```
keenangraham, line 34: Nit! https://www.python.org/dev/peps/pep-0008/#whitespace-in-expressions-and-statements

src/encoded/genomic_data_service.py
```diff
@@ -0,0 +1,136 @@
...
+            self.sort_order = 'desc' if desc else 'asc'
+
+        self.filter_params = {}
+        for filter_ in RNA_GET_FILTERS:
+            if params.get(filter_):
+                self.filter_params[filter_] = params.get(filter_)
+
+        self.filters = [{'field': 'type', 'term': 'Rna Get', 'remove': '/rnaget'}]
+
+        for filter_ in self.filter_params:
```
keenangraham, line 50: Suggestion for the future. I've been using https://github.com/ENCODE-DCC/snovault/blob/9fb33ae20ac5b7de775d0455838b2a6b3906be60/src/snovault/elasticsearch/searches/parsers.py#L441 for most of these common param parsing/querystring tasks. Would have to look at interface to see how it works but maybe some use here.

keenangraham (approved):
Python looks good for now. A couple of formatting nits and suggestions for future.

src/encoded/genomic_data_service.py
pedrohr: that was me still with a Rails brain haha

src/encoded/genomic_data_service.py
pedrohr: oh good to know! I will take a look

PR #3779, opened 2021-03-13
ENCD-5799 Make BDD cart tests more reliable

forresttanaka:
I found out that Firefox-based local BDD testing failed when clicking the “Show default data only” because my HTML didn’t allow Firefox to detect clicks on the label, though that worked in other browsers. So I redesigned that checkbox and use it both for that and the “Show visualizable data only” checkbox. My hope is that this allows the CircleCI BDD test, which is Chrome based, so fail less often, though we’ll still have to see. I haven’t yet seen a failure of that test with this branch.


zoldello (approved):
Code looks fine. It is interesting that you did not try to do a hack like add add a div with an event as opposed to re-writing the check box.

PR #3780, opened 2021-03-13
ENCD-5167-fix-gene-search

keenangraham:
I think this is more or less final structure. Still need to add more tests.

### DIFF
--- a/src/encoded/static/components/gene_search/gene.js
+++ b/src/encoded/static/components/gene_search/gene.js
@@ -1,0 +1,158 @@
+import { partitionStringByMatchingOrNot } from './match';
+import { WHITESPACE } from '../top_hits/constants';
+import {
+    maybeConvertArrayToString,
+    uniqueAndDefinedValues,
+} from '../top_hits/hits/base';
+
+
+/**
+* Used to add square brackets around a string value that might not exist.
+* @param {string} value - String value to put in brackets.
+* @return {string} Either the empty string or a string with square brackets.
+*/
+export const maybeFormatStringInSquareBrackets = (value) => {
+    if (value.length > 0) {
+        return `[${value}]`;
+    }
+    return value;
+};
+
+
+/**
+* Used to normalize a value for improved highlighting of matching term.
+* Removes single and double quotes (helps highlight terms even if query
+* is in quotes), makes lowercase (helps for case insenstive matching),
+* and trims whitespace.
+* @param {string} value - String value to normalize.
+* @return {string} Normalized string.
+*/
+export const normalizeStringForMatching = (value) => (
+    value.replace(/['"]+/g, '').toLowerCase().trim()
+);
+
+
+/**
+* Encapsulates all of the parsing and highlighting for a gene search result
+* as well the calculation for getting the gene location by assembly for
+* visualization on a browser.
+*/
+class Gene {
+    constructor(item, searchTerm) {
+        this.item = item;
+        this.searchTerm = searchTerm;
+        this.normalizedSearchTerm = normalizeStringForMatching(
+            searchTerm
+        );
+    }
+
+    /**
+    * We only want to show synonyms or dbxrefs if they match
+    * the user's searchTerm.
+    */
+    getFilteredGeneSynonymsAndRefs() {
+        const filteredSynonymsAndRefs = (
+            this.item.synonyms || []
+        ).concat(
+            this.item.dbxrefs || []
+        ).filter(
+            (alias) => alias.toLowerCase().includes(
+                this.normalizedSearchTerm
+            )
+        );
+        return maybeFormatStringInSquareBrackets(
+            maybeConvertArrayToString(
+                filteredSynonymsAndRefs
+            )
+        );
+    }
+
+    /**
+    * We check to see if the search result has the user's
+    * searchTerm in the title of the gene.
+    */
+    searchTermMatchesGeneTitle() {
+        return this.item.title.toLowerCase().includes(
+            this.normalizedSearchTerm
+        );
+    }
+
+    /**
+    * We show the title if the user's searchTerm has a match
+    * in the title. Otherwise we show the title and matching
+    * synonyms or dbxrefs, if any.
+    */
+    formatGeneTitleAndMaybeSynonyms() {
+        if (this.searchTermMatchesGeneTitle()) {
+            return this.item.title;
+        }
+        return maybeConvertArrayToString(
+            uniqueAndDefinedValues(
+                [
+                    this.item.title,
+                    this.getFilteredGeneSynonymsAndRefs(),
+                ],
+            ),
+            WHITESPACE
+        );
+    }
+
+    /**
+    * Used for display in input box.
+    */
+    asString() {
+        return this.formatGeneTitleAndMaybeSynonyms();
+    }
+
+    /**
+    * Used for mapping the matching and mismatching parts
+    * to formatting components.
+    */
+    asMatchingOrNot() {
+        return partitionStringByMatchingOrNot(
+            this.normalizedSearchTerm,
+            this.asString(),
+        );
+    }
+
+    /**
+    * Sometimes nice to have the plain title for display.
+    */
+    title() {
+        return this.item.title;
+    }
+
+    /**
+    * This returns the appropriate gene location (which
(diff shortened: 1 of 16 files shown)


keenangraham:
Summary of changes:
- Use `/search/?type=Gene` to return gene search results instead of the annotation suggestion endpoint.
- Add more robust term highlighting mechanism to make all matching terms bold instead of just the first one.
- Add logic to display matching synonyms or dbxrefs if the match isn’t in the gene title.
- Debounce user input to reduce number of requests hitting ES, and fix weird delay bug if user types too fast.
- Hide browser autosuggest of user’s old search terms.
- Disentangle the gene search from genome browser and region search. Encapsulate into smaller components and separate module.
- Expand to show gene search for model organism assemblies, and add browser gene tracks for ce11/dm6.

src/encoded/static/components/genome_browser.js
```diff
@@ -300,6 +308,8 @@ export function getCoordinateData(geneLink, fetch) {
     });
 }
 
+const getAssemblyWithoutGenomeAnnotationOrMinimal = (assembly) => assembly.split(' ')[0].replace('-minimal', '');
```
forresttanaka, line 311: Not your issue but I wish we pass `GenomeBrowser` the assembly and annotation separately so we don’t have to split it inside `GenomeBrowser`.

src/encoded/static/components/genome_browser.js
```diff
@@ -405,8 +415,7 @@ class GenomeBrowser extends React.Component {
         this.state = {
             trackList: [],
             visualizer: null,
-            showAutoSuggest: true,
-            searchTerm: '',
+            assemblyWithoutAnnotationOrMinimal: '',
```
forresttanaka: We’ve been trying to reduce the number of React states we use, keeping down the number of ineffective rerenders. Looks like `assemblyWithoutAnnotationOrMinimal` gets directly calculated from props and never changes after getting calculated, so you can just calculate it each time you need it. State _generally_ should, when possible, hold [changes from user actions](https://reactjs.org/docs/react-component.html#state) or server responses.

src/encoded/static/components/gene_search/gene.js
```diff
@@ -0,0 +1,156 @@
...
+import {
+    maybeConvertArrayToString,
+    uniqueAndDefinedValues,
+} from '../top_hits/hits/base';
+
+
+/**
+* Used to add square brackets around a string value that might not exist.
+* @param {string} value - String value to put in brackets.
+* @param {string} Either the empty string or a string with square brackets.
```
forresttanaka: I see you use the `@returns` tag in other places.

src/encoded/static/components/gene_search/gene.js
```diff
@@ -0,0 +1,156 @@
...
+* visualization on a browser.
+*/
+class Gene {
+    constructor(item, searchTerm) {
+        this.item = item;
+        this.searchTerm = searchTerm;
+        this.normalizedSearchTerm = normalizeStringForMatching(
+            searchTerm
+        );
+    }
```
forresttanaka, line 47: Babel apparently supports [private methods and fields](https://babeljs.io/docs/en/babel-plugin-proposal-private-methods) with a plugin. I’ve never tried that, but maybe we should test this soon.

src/encoded/static/components/gene_search/input.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+
+/**
+* Controlled input box that accepts user's query
+* and triggers onChange callback.
+*/
+const Input = React.forwardRef((props, ref) => (
```
forresttanaka, line 9: I’ve never heard of `forwardRef` before. I could have used that in a couple places.

src/encoded/static/components/gene_search/match.js
```diff
@@ -0,0 +1,108 @@
...
+* @return {array} The start indices of the pattern in the text.
+*/
+export const getIndicesOf = (pattern, text) => {
+    const indices = [];
+    let currentIndex = 0;
+    const patternLength = pattern.length;
+    // We check patternLength to avoid infinite loop if user passes in
+    // empty string. Otherwise we set the currentIndex while checking
+    // the condition on every loop.
+    /* eslint no-cond-assign: ["error", "except-parens"] */
```
forresttanaka: Seems pretty easy to avoid the eslint exception, and we’ve been trying to avoid them except in cases where we just have no choice.

src/encoded/static/components/gene_search/match.js
```diff
@@ -0,0 +1,108 @@
...
+* @return {array} Array with partitioned text.
+* >> partitionStringByMatchingOrNot(
+*        'ABC',
+*        'xyzABCzzzzzABC'
+*    ) === [
+*        ['mismatch', 'xyz'],
+*        ['match', 'ABC'],
+*        ['mismatch', 'zzzzz'],
+*        ['match', 'ABC'],
+*    ];
```
forresttanaka, line 60: That’s super helpful.

src/encoded/static/components/gene_search/results.js
```diff
@@ -0,0 +1,125 @@
...
+        <li>
+            <button
+                type="button"
+                tabIndex="0"
+                onClick={() => props.handleClick(gene)}
+            >
+                {
+                    gene.asMatchingOrNot().map(
+                        ([key, value], i) => {
+                            const Component = matchingOrNotToComponent[key];
```
forresttanaka, line 53: Cool way to do that.

forresttanaka (changes requested):
Nice and clean as usual. On my comment about using state, I might have missed an important reason to use state, so let me know if I did.

src/encoded/static/components/gene_search/gene.js
keenangraham: Good catch, thanks! Fixed.

src/encoded/static/components/gene_search/input.js
keenangraham: It also seemed to work when I passed the `ref` in `props` under a different name, but was concerned something would break if I didn't use `forwardRef`.

src/encoded/static/components/gene_search/match.js
keenangraham: https://eslint.org/docs/rules/no-cond-assign#options

Maybe I'm using this wrong, the rule seems to have a default string option that `allows assignments in test conditions only if they are enclosed in parentheses (for example, to allow reassigning a variable in the test of a while or do...while loop)`. And the assignment is in parenthesis: `(currentIndex = text.indexOf(pattern, currentIndex))`.

It seems to work if I don't include the `except-parens` in the `eslint no-cond-assign` list, since that's the default, but I was trying to make it explicit.

In this case I think it's one of the `valid reasons to use assignment operators in conditional statements` since it avoids either checking the same indexOf twice per loop or having a duplicate indexOf call and extra variable before the loop. 

Is there a reason the `except-parens` option isn't the default for our eslint rules?

src/encoded/static/components/genome_browser.js
keenangraham: Calling the function explicitly now when needed instead of storing in state. Seems to work.

src/encoded/static/components/gene_search/match.js
forresttanaka: I think you make a reasonable argument. I’d have no issue if you allowed except-parens in .eslintrc.js so you don’t need the eslint comment.

keenangraham: Okay! Think I've updated the rules and removed this line.

keenangraham:
Pushed tests.

forresttanaka (dismissed):
Looks cool.

keenangraham:
Rebased and squashed to already reviewed commit.

src/encoded/static/components/gene_search/gene.js
forresttanaka: FYI I tried to add this to one of my branches, and babel seems ready (the proposal for private methods and fields still lives in committee, but it seems unlikely to change), the ESLint config — while possible — feels very duct-taped together, so I think I’ll wait until things get on more stable ground on their end.

PR #3781, opened 2021-03-15
ENCD-5561-organize-pyramid-config

paul-sud:


.circleci/config.yml
```diff
@@ -51,17 +62,27 @@ commands:
             sudo chown -R circleci /etc/elasticsearch
             sed -i '1s;^;export PATH=/usr/share/elasticsearch/bin:/usr/lib/postgresql/10/bin:$PATH\n;' $BASH_ENV
       - run:
-          name: Install application and run buildout
+          name: Install application
+          command: |
+            sudo pip3 install --upgrade pip
+            make install
+  install_jsonnet:
```
ottojolanki: super NIT: use same word separator in all the commands.

conf/pyramid/sections.libsonnet
```diff
@@ -0,0 +1,267 @@
...
+bar = baz
+```
+
+Each function's arguments can change the resulting section returned. The default values
+are not particularly special, they usually represent the most common configuration
+values to make assembling the sections into complete INI files less verbose by reducing
+the need to supply overriding keyword arguments.
+
+Jsonnet requires quoting any dotted keys or keys with spaces. If you need to use
+PasteDeploy `set` keyword then enclose it in quotes (e.g. `set timeout`).
```
ottojolanki: Do you mean quotes, or grave accent (Unicode Character “`” (U+0060)) as is written in the comment?

conf/pyramid/demo.ini
```diff
@@ -0,0 +1,144 @@
...
+download_proxy = https://download.encodeproject.org/
+elasticsearch.server = localhost:9201
+embed_cache.capacity = 25000
+external_aws_s3_transfer_allow = false
+external_aws_s3_transfer_buckets = /srv/encoded/.aws/direct-external-s3-list
+file_upload_bucket = encoded-files-dev
+file_upload_profile_name = encoded-files-upload
+hostname_command = command -v ec2metadata > /dev/null && ec2metadata --public-hostname || hostname
+indexer.chunk_size = 1024
+indexer.processes = 16
```
ottojolanki, line 18: Does these get overridden if I run `deploy --index-procs xyz --index-chunk-size abc`?

ottojolanki (dismissed):
One nit and couple of questions. I mean this is hard to review exhaustively, but everything looks correct, and makes sense to me. Ready for some QA.

conf/pyramid/sections.libsonnet
paul-sud: Yeah this documentation is unclear, I mean quotes. The backticks are just there to indicate the text is code. I can put the code sample on its own line, hopefully then it will make more sense.

conf/pyramid/demo.ini
paul-sud: No, those two deploy options were removed. The canonical way to change those values is to edit the Jsonnet, generate config, commit the changes, push, and then deploy.

.circleci/config.yml
paul-sud: You're saying they should all use the pipe? No `command: foo bar`?

ottojolanki: No. `install-app` and `install-jsonnet` or `install_app` and `install_jsonnet`. When I say super NIT, I mean it :)

conf/pyramid/demo.ini
ottojolanki: Got it.

.circleci/config.yml
paul-sud: Gotcha, changed to dashes for consistency with the existing command.

paul-sud:
Rerequested review since I needed to update a little code after rebasing (ontology URL and add setting for `genomic_data_service`

ottojolanki (dismissed):
LGTM

PR #3785, opened 2021-03-15
ENCD-5803-update-wgbs-vis-def

paul-sud:


src/encoded/static/vis_defs/WGBS_vis_def.json
```diff
@@ -17,7 +17,7 @@
                     "tag": "aCpG",
                     "visibility": "dense",
                     "type": "bigBed",
-                    "file_format_type": [ "bedMethyl" ],
+                    "file_format_type": ["bedMethyl", "bed 9+"],
```
gabdank: I think it should be "bed9+" rather than "bed 9+"

gabdank (changes requested):

src/encoded/static/vis_defs/WGBS_vis_def.json
paul-sud: thanks, fixed

gabdank (dismissed):

gabdank (dismissed):

gabdank (approved):

PR #3786, opened 2021-03-15
ENCD-5789 Cart pseudo-default files

forresttanaka:
* I changed the name of `filterForPreferredFiles` to `filterForDefaultFiles` because Ben seems to use the term “default” instead of “preferred.”
* I also fixed a very long-existing bug in `requestDatasets` that apparently causes no/few issues.
* I changed a BDD test for forms to try to make that more reliable. Works locally, and seems to succeed on CircleCI too.

### DIFF
--- a/src/encoded/static/components/__tests__/cart-test.js
+++ b/src/encoded/static/components/__tests__/cart-test.js
@@ -14,6 +14,7 @@ import {
     CART_OPERATION_IN_PROGRESS,
 } from '../cart/actions';
 import { isAllowedElementsPossible } from '../cart/util';
+import { processPseudoDefaultFiles } from '../cart/cart';
 
 // Temporary use of adapter until Enzyme is compatible with React 17.
 Enzyme.configure({ adapter: new Adapter() });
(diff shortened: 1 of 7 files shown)


src/encoded/static/components/__tests__/cart-test.js
```diff
@@ -436,3 +437,263 @@ describe('Cart manager while logged in as admin', () => {
         expect(tableRows.at(2).find('.cart-manager-table__current-row')).toHaveLength(1);
     });
 });
+
+describe('Process pseudo-default files', () => {
+    let nonVisuzalizablefiles;
```
zoldello, line 442: Spelling error. It is a unit test though so no big deal

src/encoded/static/components/__tests__/cart-test.js
```diff
@@ -436,3 +437,263 @@ describe('Cart manager while logged in as admin', () => {
...
+
+    const getDefaultFiles = (datasets) => (
+        datasets.reduce((accFiles, dataset) => {
+            const defaultFiles = dataset.files.filter((file) => file.preferred_default || file.pseudo_default);
+            return accFiles.concat(defaultFiles);
+        }, [])
+    );
+
+    beforeEach(() => {
+        nonVisuzalizablefiles = [
```
zoldello, line 455: spelling error

src/encoded/static/components/cart/cart.js
```diff
@@ -309,16 +309,25 @@ CartBrowser.defaultProps = {
...
+const CartFiles = ({ files, currentPage, defaultOnly }) => {
     if (files.length > 0) {
         const pageStartIndex = currentPage * PAGE_FILE_COUNT;
         const currentPageFiles = files.slice(pageStartIndex, pageStartIndex + PAGE_ELEMENT_COUNT);
+        const pseudoDefaultFiles = files.filter((file) => file.pseudo_default);
         return (
             <div className="cart-list cart-list--file">
+                {defaultOnly && pseudoDefaultFiles.length > 0
+                    ? (
+                        <div className="cart-list__no-dl">
```
zoldello, line 321: Does the class name supposed to be something like "cart-list--file___no-dl"?

src/encoded/static/components/cart/cart.js
```diff
@@ -598,7 +613,7 @@ const requestDatasets = (elements, fetch, session) => {
         sessionPromise = fetch('/session');
     } else {
         // We have a session CSRF token, so retrieve it immediately.
-        sessionPromise = Promise.resolve(session._csrft);
```
zoldello, line 601: How did **session_csrft** ever work? The backend seems to need have it set as **session_csrft_**

src/encoded/static/components/cart/cart.js
```diff
@@ -1556,6 +1571,67 @@ const processFilesAnalyses = (files, analyses) => {
...
 
+/**
+ * Mutate the files of the given datasets that have no files with preferred_default set so that
+ * they appear as though they had preferred_default set. Once wranglers have patched all
+ * appropriate files' preferred_default properties, this function serves no purpose. A number of
+ * criteria determine how and whether any files within a dataset get this modification. Exported
+ * for Jest test.
+ * @param {array} datasets Datasets containing files to mutate with preferred_default
+ */
+export const processPseudoDefaultFiles = (datasets = []) => {
```
zoldello, line 1596: "Pseudo" and your comments imply to me that the function **processPseudoDefaultFiles**  
 will get a name and functionality change. If that is true, would it be a good idea to name it as if the functionality will not change? Plus, sometimes, temporary last as long as permanent.

zoldello (dismissed):

forresttanaka:
@zoldello Ben wanted to adjust the way we choose the pseudo-default files.

forresttanaka:
BDD test passed locally, but I’m still restarting and restarting and restarting and restarting Circle.

src/encoded/static/components/cart/cart.js
```diff
@@ -1571,9 +1571,17 @@ const processFilesAnalyses = (files, analyses) => {
...
+/**
+ * Sort an array of files by their statuses according to their order in
+ * pseudoDefaultFileStatusOrder. Anything not in pseudoDefaultFileStatusOrder gets sorted randomly
+ * at the end.
+ */
 const pseudoDefaultFileStatusOrder = ['released', 'archived', 'in progress'];
 const sortPseudoDefaultFilesByStatus = (files) => (
-    _(files).sortBy((file) => pseudoDefaultFileStatusOrder.findIndex((status) => status === file.status))
+    _(files).sortBy((file) => {
+        const foundIndex = pseudoDefaultFileStatusOrder.indexOf(file.status);
```
zoldello, line 1582: In Javascript sort in evergreen browsers, 0 has a special meaning (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort). The length is +1 larger than the range of values. There may be some odd corner cases that may make this wanky. But I do not think it is a big deal.

zoldello (approved):
Code look okay. I added a comment but it is more so me thinking aloud.

PR #3789, opened 2021-03-16
ENCD-5773-exp-series-tables-dnase-rna

khzlin:


yunhailuo (approved):

forresttanaka (approved):
Looks very good. One thing to think about if you like: many of the `display` methods seem _very_ close to the same but for a few details. These could be condensed to one shared display function with any differences parameterized.

khzlin:
thanks forrest. We're planning to add 6 more assays in ENCD-5811. I'll keep in mind on condensing the display methods there.

PR #3790, opened 2021-03-17
ENCD-4612-add-Valis-tracks-GRCm39-M26-V33-dm3-genes-ce10-genes

yunhailuo:


src/encoded/static/components/cart/cart.js
```diff
@@ -296,14 +296,12 @@ CartBrowser.propTypes = {
...
-    assembly: PropTypes.string,
+    assembly: PropTypes.string.isRequired,
     /** Page of files to display */
     pageNumber: PropTypes.number.isRequired,
 };
 
-CartBrowser.defaultProps = {
-    assembly: '',
-};
+CartBrowser.defaultProps = {};
```
forresttanaka: You can actually delete this completely, though it doesn’t hurt anything.

src/encoded/static/components/genome_browser.js
```diff
@@ -49,17 +50,19 @@ const readGenomeBrowserLabelCoordinates = () => {
  * @param {boolean} [ignoreCache=false] True to not look into cache, false to use cache
  * @returns Default coordinates
  */
-const getDefaultCoordinates = (assemblyAnnotation, ignoreCache = false) => {
+const getDefaultCoordinates = (assemblyAnnotation, geneAnnotation, ignoreCache = false) => {
     const assembly = assemblyAnnotation.split(' ')[0];
+    const annotation = geneAnnotation.split(' ')[0];
     // Files to be displayed on all genome browser results
```
forresttanaka, line 52: Do you need to split these? I think these are already split now?

src/encoded/static/components/genome_browser.js
```diff
@@ -554,10 +596,11 @@ class GenomeBrowser extends React.Component {
 
     setGenomeAndTracks() {
         const genome = mapGenome(this.props.assembly);
+        const annotation = mapGenome(this.props.annotation);
```
forresttanaka: Does `mapGenome` do anything with the annotation? Is this in case we need it in the future? Also, the first line of `mapGenome` splits the parameter, which I don’t think you need to do anymore.

forresttanaka (changes requested):
Thank you for keeping assembly and annotation separate — it had been a problem that we kept them combined before.

src/encoded/static/components/genome_browser.js
yunhailuo: You are correct. No need. I didn't think carefully and just copied the behavior from assembly.

src/encoded/static/components/genome_browser.js
yunhailuo: You are right. No need. I didn't fully understand what it is.

forresttanaka (dismissed):
OK, I see nothing wrong.

forresttanaka (approved):

PR #3796, opened 2021-03-18
ENCD-5768-download-btn-appearing-unneededly

zoldello:


src/encoded/static/components/filegallery.js
```diff
@@ -56,6 +56,9 @@ const inclusionStatuses = [
     'replaced',
 ];
 
+/** Get data set */
```
forresttanaka, line 59: The comment doesn’t seem applicable.

forresttanaka (dismissed):
Yeah, this seems like a good change.

PR #3802, opened 2021-03-22
ENCD-5819 Fix missing related-files table

forresttanaka:
* The main crux of this ticket involves the call to `requestFiles` which passes an array of files to filter out — something we don’t need anymore, and which `requestFiles` itself now interprets as a query-string key.
* Converted some boolean properties to `FileTable` to get passed in a new `options` property to try to keep the number of props under a little control. I added a `collapsedNone` boolean property which indicates no file subtables should appear collapsed by default, which I only use with Reference, Project, and UcscBrowserComposite pages. Experiment, Annotation, and FunctionalCharacteriationExperiment pages still appear with some tables collapsed by default.
* RawSequencingTable and CollapsingTitle were configured to take an object in `outputType` and `analysisObjectKey ` but these had recently become strings. This clears the “Failed prop type: Invalid prop `outputType` of type `string` supplied to `CollapsingTitle`, expected `object`” warnings seen on v114 developer builds.
* Removed `altFilterDefault` from most datasets that use FileGallery because this only affects the browser dropdown which only appears with a `visualize` property, which only appears for Experiment, Annotation, and FunctionalCharacterizationExperiment.
* Removed `encodevers` which a few years ago had been used to display extra statuses on experiments, but not since. This also involves removing the `encodeVersion` function from globals.js.

The patched demo https://encd-5819-related-files-fytanaka.demo.encodedcc.org/ has the up-to-date changes until the QA demo indexes.

### DIFF
--- a/src/encoded/static/components/filegallery.js
+++ b/src/encoded/static/components/filegallery.js
@@ -178,7 +178,7 @@ export class FileTable extends React.Component {
             collapsed: {
                 // Keeps track of which tables are collapsed
                 ref: false,
-                Other: true,
+                Other: !props.options.collapseNone,
                 ...analysisObjects,
             },
         };
@@ -237,7 +237,6 @@ export class FileTable extends React.Component {
             items,
             graphedFiles,
             filePanelHeader,
-            encodevers,
             showFileCount,
             setInfoNodeId,
             setInfoNodeVisible,
@@ -245,8 +244,7 @@ export class FileTable extends React.Component {
             session,
             filters,
             adminUser,
-            showReplicateNumber,
-            inclusionOn,
+            options,
         } = this.props;
         const sessionProperties = this.context.session_properties;
         const loggedIn = !!(session && session['auth.userid']);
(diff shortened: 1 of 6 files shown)


src/encoded/static/components/filegallery.js
```diff
@@ -1201,25 +1226,35 @@ DatasetFiles.defaultProps = {
...
+            showReplicateNumber={showReplicateNumber}
+        />
     </FetchedData>
 );
 
 FileGallery.propTypes = {
     context: PropTypes.object.isRequired, // Dataset object whose files we're rendering
-    encodevers: PropTypes.string, // ENCODE version number
-    hideGraph: PropTypes.bool, // T to hide graph display
+    hideGraph: PropTypes.bool, // T to only render file-table pane
```
emro, line 1247: I think it's a little funny "true" is sometimes written out ("True") here and sometimes not ("T") but this is a minor detail.

emro (approved):
All the code looks good, also the demo looks good!

src/encoded/static/components/filegallery.js
forresttanaka: Yeah I used to use “T” to save space, then felt like that was needlessly obscure so I switched.

PR #3804, opened 2021-03-23
ENCD-5832 Consider assembly when choosing pseudo-default files

forresttanaka:


src/encoded/static/components/cart/cart.js
```diff
@@ -1581,6 +1583,19 @@ const sortPseudoDefaultFilesByStatus = (files) => (
 );
 
 
+/**
+ * Sort an array of assemblies, newer assemblies first. Carts don’t track genome_annotation, so those
+ * don't enter into this sorting.
+ * @param {array} files Files to sort
```
zoldello: the parameter is "assemblies" not "files"

src/encoded/static/components/cart/cart.js
```diff
@@ -1571,6 +1571,8 @@ const processFilesAnalyses = (files, analyses) => {
  * Sort an array of files by their statuses according to their order in
  * pseudoDefaultFileStatusOrder. Anything not in pseudoDefaultFileStatusOrder gets sorted randomly
  * at the end.
+ * @param {array} files Files to sort
+ * @return {array} Files sorted by the three following statuses
  */
 const pseudoDefaultFileStatusOrder = ['released', 'archived', 'in progress'];
 const sortPseudoDefaultFilesByStatus = (files) => (
```
zoldello, line 1583: Lines 1577 and 1578 should be swapped to make align sortPseudoDefaultFilesByStatus with the jsdoc comment

src/encoded/static/components/cart/cart.js
```diff
@@ -2022,7 +2041,9 @@ const CartComponent = ({ context, savedCartObj, inProgress, fetch, session }) =>
                                         totalPageCount={totalPageCount.browser}
                                         updateCurrentPage={updateDisplayedPage}
                                     />
-                                    {Object.keys(selectedTerms).length > 0 ? <CartBrowser files={selectedVisualizableFiles} assembly={selectedTerms.assembly[0]} pageNumber={pageNumbers.browser} /> : null}
+                                    {Object.keys(selectedTerms).length > 0 && selectedTerms.assembly && selectedTerms.assembly.length > 0
```
zoldello: Why determine if selectedTerms have keys if all you need is selectedTerms.assembly; which you added a check for?

src/encoded/static/components/cart/cart.js
```diff
@@ -1606,17 +1621,21 @@ export const processPseudoDefaultFiles = (datasets = []) => {
...
-                        const replicateFiles = _(visualizableFiles).groupBy((file) => file.biological_replicates[0]);
-                        Object.keys(replicateFiles).forEach((replicate) => {
-                            const outputTypeFiles = _(replicateFiles[replicate]).groupBy((file) => file.output_type);
-                            Object.keys(outputTypeFiles).forEach((outputType) => {
-                                // Select the first bigWig and first bigBed within each
-                                // output_type within each replicate.
-                                const firstBigWig = outputTypeFiles[outputType].find((file) => file.file_format === 'bigWig');
-                                const firstBigBed = outputTypeFiles[outputType].find((file) => file.file_format === 'bigBed');
-                                pseudoDefaultFiles.push(..._.compact([firstBigWig, firstBigBed]));
+                        // Multiple output types among the files. Group first by sorted assembly,
```
zoldello, line 1629: This looks fine. Philosophically, since you have one part feeding another part, you can use underscore.js chain. But it may be overkill to some.

zoldello (changes requested):
The code looks good. The only needed change is to correct the first comment in cart.js where you have the jsdoc comment applied to the wrong object. You took the time to add jsdoc; it might as well be put to work.

src/encoded/static/components/cart/cart.js
forresttanaka: 👍

src/encoded/static/components/cart/cart.js
forresttanaka: I moved `pseudoDefaultFileStatusOrder` above all this to avoid doing a static assignment inside a function.

src/encoded/static/components/cart/cart.js
forresttanaka: Yeah, that’s true.

zoldello (approved):
Look good.

PR #3805, opened 2021-03-24
VALIS-36-remove-expander

emro:
Includes the following Valis updates on a Valis branch:
- an update from George, which disables the dragging expandability of the tracks when expandable = false:
VALIS-software@c5b5286
- 0 spacing between tracks
Note: I think the 0 spacing between tracks is important if we want to have some tracks as expandable with the carets but don't want them to be expandable by dragging.

I've reduced the heights of the tracks now that the heights do not include pixels of empty space.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7401,8 +7401,8 @@
       }
     },
     "genome-visualizer": {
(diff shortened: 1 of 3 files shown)


zoldello (commented):
Looks good.

zoldello (dismissed):
Looks good.

PR #3808, opened 2021-03-25
ENCD-5833 Display audits on analysis file tables

forresttanaka:
* Neither `FileGraph` nor `Graph` uses the audit HOCs anymore, so I removed them from their props.
* Removed `{...this.props}` from call to `FileTable` as this only added one property (schemas) not explicitly passed in before, and it caused many unused props to be passed in.


forresttanaka:
Patched demo with the latest code for this branch at https://encd-5833-analysis-audits-fytanaka.demo.encodedcc.org.

forresttanaka:
Will update to include statuses in the table headers.

PR #3809, opened 2021-03-26
ENCD-5820-select-better-analysis-by-default

zoldello:


src/encoded/static/components/filegallery.js
```diff
@@ -56,6 +56,77 @@ const inclusionStatuses = [
...
+ *  In the keys matching the key-argument, get all values that match any entry in the filters-argument.
+ *  Stop after getting any result and return the result.
+ *
+ *  For example, if -
+ *      analyis = [{'a': 1, 'b': 2}, {'a': 3, 'c': '5'}, {'c': 1, 'a': 2}]
+ *      filters = ['c', 'b']
+ *      key = 'c'
+ *
+ *  The result will be {'a': 3, 'c': '5'} as the first key-argument 'c' is in the filter-argument 'c'
+ *  and it matches the entry {'a': 3, 'c': '5'} first.
```
forresttanaka, line 77: Really helpful comment for a function that’s pretty difficult to understand at first.

src/encoded/static/components/filegallery.js
```diff
@@ -56,6 +56,77 @@ const inclusionStatuses = [
...
+ * Sort analyses by version
+ *
+ * @param {array} analysesGroup
+ * @returns Sorted analyses group by version
+*/
+const sortAnalysisByVersion = (analysesGroup) => {
+    const sortedAnalysesAwarded = analysesGroup.sort((a, b) => {
+        // remove text
+        const numbersInTitle1 = a.title.replace(/\D/g, '');
+        const numbersInTitle2 = b.title.replace(/\D/g, '');
```
forresttanaka: I feel like this might be cleaner if `compileAnalyses` included `pipeline_version` in the compiled analyses objects so you don’t have to extract them in a potentially unreliable way.

src/encoded/static/components/filegallery.js
```diff
@@ -56,6 +56,77 @@ const inclusionStatuses = [
...
+        // add numbers so ones with version have the same number of digits as non-versioned ones
+        const number1 = numbersInTitle1.padEnd ? numbersInTitle1.padEnd(8, 0) : numbersInTitle1;
+        const number2 = numbersInTitle2.padEnd ? numbersInTitle2.padEnd(8, 0) : numbersInTitle2;
+
+        // sort
+        return parseInt(number2, 10) - parseInt(number1, 10);
+    });
+
+    return sortedAnalysesAwarded;
+};
```
forresttanaka, line 139: Should we instead put `compareAnalysisVersions` from cart/analysis.js into objectutils.js and use that?

src/encoded/static/components/filegallery.js
```diff
@@ -2500,40 +2571,20 @@ const AnalysesSelector = ({ analyses, selectedAnalysesIndex, handleAnalysesSelec
...
-        const awardSuffix = ['ENCODE4', 'ENCODE3', 'ENCODE2', 'Mixed', 'Lab custom']; // ordered in decreasing precedence
-        const analysesAwarded = analyses.filter((analysis) => awardSuffix.some((award) => (analysis.title || '').includes(award)));
-
-        // a bit hacky, removes text and use number to determine sorting
-        const sortedAnalysesAwarded = analysesAwarded.sort((a, b) => {
-            // remove text
-            const numbersInTitle1 = a.title.replace(/\D/g, '');
-            const numbersInTitle2 = b.title.replace(/\D/g, '');
+        const awardSuffixes = ['ENCODE4', 'ENCODE3', 'ENCODE2', 'Mixed', 'Lab custom']; // ordered in decreasing precedence
+        const statuses = ['released', 'in progress', 'archived', 'revoked', 'deleted']; // ordered in decreasing precedence
```
forresttanaka, line 2684: You can get the statuses of any defined object type in a certain order in `getObjectStatuses`, and filtered to only include those available at the user’s current access level. You’d have to add an entry for “Analysis” in `defaultObjectStatuses` as we update this object as needed.

forresttanaka (changes requested):
Operationally this looks good. I have a couple suggestions, but if you prefer not to use those, let me know and I’ll pass it.

src/encoded/static/components/filegallery.js
zoldello: @forresttanaka pipeline_version won't help as I need more than just the ENCODE number. I need all number-versions. For example pipeline_version can be "ENCODE4". But I need to know "ENCODE4 v3.0.0-alpha.2 mm10"; the "vXYZ".

src/encoded/static/components/filegallery.js
zoldello: @forresttanaka This function has no need for **compareAnalysisVersions**. We can  **compareAnalysisVersions** in objectutils.js but I doubt if many other functions will call on it as we don't tend to do a lot of calculations. Please, let me know if I am missing something; my spider senses are tingling on this.

src/encoded/static/components/filegallery.js
zoldello: @forresttanaka I tried this. It seems to only give user's current access level. It does not always correlate o the analysis_object status. Like in **<url>//experiments/ENCSR000EIY/**, admin has access levels "released”, “in progress”, “deleted". But analysis_object has these and "archived"

src/encoded/static/components/filegallery.js
forresttanaka: `pipeline_version` has the version number. `pipeline_award_rfas` has the ENCODE number. For example, https://test.encodedcc.org/analyses/ENCAN627PJS/ has `pipeline_version` 1.7.1.

src/encoded/static/components/filegallery.js
forresttanaka: Doesn’t `sortAnalysisByVersion` sort by version number? `compareAnalysisVersions` is specifically to help sort by version number.

src/encoded/static/components/filegallery.js
forresttanaka: `getObjectStatuses` takes the user’s current access level in the `accessLevel` parameter — it doesn’t return the user’s current access level. It “Returns array of possible statuses for given object.” For example, if you pass `getObjectStatuses` an “item” of “AntibodyCharacterization” and `accessLevel` “external,” it returns the array `["compliant", "exempt from standards", "not compliant", "not reviewed", "not submitted for review by lab"]`.

src/encoded/static/components/filegallery.js
```diff
@@ -2500,40 +2576,20 @@ const AnalysesSelector = ({ analyses, selectedAnalysesIndex, handleAnalysesSelec
...
-        const sortedAnalysesAwarded = analysesAwarded.sort((a, b) => {
-            // remove text
-            const numbersInTitle1 = a.title.replace(/\D/g, '');
-            const numbersInTitle2 = b.title.replace(/\D/g, '');
-
-            // add numbers so ones with version have the same number of digits as non-versioned ones
-            const number1 = numbersInTitle1.padEnd ? numbersInTitle1.padEnd(8, 0) : numbersInTitle1;
-            const number2 = numbersInTitle2.padEnd ? numbersInTitle2.padEnd(8, 0) : numbersInTitle2;
+        const awardSuffixes = ['ENCODE4', 'ENCODE3', 'ENCODE2', 'Mixed', 'Lab custom']; // ordered in decreasing precedence
+        const statuses = ['released', 'in progress', 'archived', 'deleted', 'revoked']; // ordered in decreasing precedence
```
zoldello: @forresttanaka  I tried doing

```
     const accessLevel = sessionToAccessLevel(context.session, context.session_properties);
        const statuses = getObjectStatuses('Analysis', accessLevel); // ordered in decreasing precedence
```

The issue is that I did not get all needed. Like for [http://encd-5820-5283310b1-phil.demo.encodedcc.org/experiments/ENCSR000CNC/](http://encd-5820-5283310b1-phil.demo.encodedcc.org/experiments/ENCSR000CNC/), I, did not get back archived even though an analyses record has status of archived.

forresttanaka: You’re using `standardStatuses` which doesn’t include archived. Analysis statuses don’t match any existing status scheme, so you need to make a custom one, though in my example I don’t know exactly what access levels each one has in Analysis objects, nor do I know what order they should be, but this should come close to correct:

```
    AccessKey: {
        external: [
            'current',
        ],
        administrator: [
            'deleted',
        ],
    },
    Analysis: {
        external: [
            'released',
            'archived',
            'revoked',
        ],
        administrator: [
            'in progress',
            'deleted',
        ].
    },
    AnalysisStepRun: standardStatuses,
    AnalysisStepVersion: standardStatuses,

File: src/encoded/static/components/filegallery.js [hunk 6]
```diff
@@ -56,6 +56,81 @@ const inclusionStatuses = [
...
+*/
+const sortAnalysisByVersion = (analysesGroup) => {
+    const sortedAnalysesAwarded = analysesGroup.sort((a, b) => {
+        if (!a || !b) {
+            return 0;
+        }
+
+        // extract numbers relevant for versioning
+        const aNumber = `${(a.pipelineAwardRfa || '').replace(/\D/g, '')}${(a.pipeline_version || '').replace(/\D/g, '')}`;
+        const bNumber = `${(b.pipelineAwardRfa || '').replace(/\D/g, '')}${(b.pipeline_version || '').replace(/\D/g, '')}`;
```
forresttanaka, line 128: Will this handle things like 1.11.4 sorting after 1.2.0?

src/encoded/static/components/filegallery.js
```diff
@@ -114,6 +189,7 @@ export const compileAnalyses = (experiment, files, dataFormat = null) => {
                             assembly,
                             status,
                             accession,
+                            pipeline_version,
```
forresttanaka, line 199: Glad you could do that.

forresttanaka (dismissed):

forresttanaka (dismissed):

forresttanaka:
Need to resolve conflicts now, probably from ENCD-5825.

forresttanaka (dismissed):
Post rebase looks fine.

forresttanaka (approved):

PR #3812, opened 2021-03-29
SNO-197-add-range-queries-and-min-max-aggs

keenangraham:


hitz (dismissed):

keenangraham:
Some added complexity here for batch_download/metadata to respect inequality syntax when filtering files (e.g. `files.file_size=gt:50000`). 

At a certain point we may just want to pull File metadata directly from ES (using all of the specified filters) and join it to Experiments similar to how PublicationData metadata works (though we would have to pull bigger file batches, and it might get expensive to store a cache of experiments). This would allow us to avoid reimplementing ES filtering features in slow Python, as well as tricky comparison behavior between string, int, float, and boolean values when all of the the query string parameters come in as strings.

src/encoded/reports/inequalities.py
```diff
@@ -0,0 +1,53 @@
...
+        inequalities.append(
+            make_inequality_from_relation_and_operand(
+                relation,
+                map_string_to_boolean_and_int(operand)
+            )
+        )
+    return inequalities
+
+
+def try_to_evaluate_inequality(inequality, LHS):
```
pedrohr: should we rename this method to `evaluate_inequality`? it sounds uncertain... In these cases that I miss the `!` in methods in Ruby to indicate that it can throw an exception.

this is just a suggestion, not a big deal

src/encoded/reports/inequalities.py
```diff
@@ -0,0 +1,53 @@
...
+    'lt': operator.lt,
+    'lte': operator.le,
+}
+
+
+def parse_inequality_param_value(value):
+    return value.split(COLON, 1)
+
+
+def partial_inequality(operator, RHS):
```
pedrohr: can we rename RHS and LHS to something like `RIGHT_OPERAND` and `LEFT_OPERAND`? or `RIGHT_OP` and `LEFT_OP`?

it is a bit hard to read RHS and LHS

src/encoded/reports/inequalities.py
```diff
@@ -0,0 +1,53 @@
+import operator
+
+from encoded.reports.serializers import map_string_to_boolean_and_int
+from snovault.elasticsearch.searches.interfaces import COLON
```
pedrohr, line 4: just for discussion... we are following the pattern that is already in place in Snovault by turning tokens into constants. But many of those are very unlikely to change. Should we continue to use follow it?

keenangraham: Yeah, I mainly do this to avoid misspelling strings, probably a little much in this case. The `parse_inequality_param_value` logic actually sort of exists in the snovault PR already (https://github.com/ENCODE-DCC/snovault/blob/240e121f4ec6a2a37700c4d98d3f6dc63cb00fcb/src/snovault/elasticsearch/searches/queries.py#L564) but it gets dumped directly into a dict which wouldn't work here and it didn't seem worth the effort to split it out.

src/encoded/reports/inequalities.py
keenangraham: Yeah, will do this. Variable names in mathy functions always seem pretty bad.

src/encoded/reports/inequalities.py
keenangraham: You could call the the inequality directly if you want to evaluate it. The intention here is to suggest that the inequality is being evaluated in a try/except block.

src/encoded/reports/inequalities.py
keenangraham: Updated, thanks for the suggestion!

keenangraham:
I think I'm going to try to make one more change here to potentially help @forresttanaka customize file prefixes in `ENCD-5825-series-files`. Basically to allow overriding the assumption that all file filters and file columns are prefixed with `files.` path.

pedrohr (dismissed):
👍

PR #3813, opened 2021-03-29
ENCD-5827-migrate-audits-to-analysis

jenjou:


src/encoded/tests/test_audit_analysis.py
```diff
@@ -301,3 +301,1024 @@ def test_audit_dnase_missing_read_depth(
...
+    # And reproducible peaks/replicate concordance audits are not reported
+    testapp.patch_json(
+        atac_replication_quality_metric_high_peaks['@id'],
+        {'quality_metric_of': [file_bed_IDR_peaks_2_atac['@id']]})
+    res2 = testapp.get(base_analysis['@id'] + '@@index-data')
+    audit_errors = res2.json['audit']
+    assert any(error['category'] == 'duplicate QC metrics' for error in audit_errors.get('ERROR', []))
+
+
+def test_audit_analysis_chia_encode4_qc_standards(
```
gabdank: since you are moving all these functions into Analysis audits, I think you can follow one function naming convention, it looks that some function names use capital letters, some don't, some have analysis in the name some don't. I am not saying there is one correct way, I am just saying since you are creating analysis audits file -  it would be a good idea to give consistent function names

src/encoded/audit/analysis.py
```diff
@@ -115,13 +376,23 @@ def audit_experiment_standards_dispatcher(value, system, files_structure):
         'ENCODE2-Mouse',
```
yunhailuo, line 392: I'm wondering for `len(value['pipeline_award_rfas']) != 1` above, should analysis having mixed rfas be audited?

src/encoded/audit/analysis.py
```diff
@@ -115,13 +376,23 @@ def audit_experiment_standards_dispatcher(value, system, files_structure):
...
+    # Allow White lab for Transcription factor ChIP-seq pipeline (modERN) and
+    # Ruan lab for ChIA-PIPE
     if len(value['pipeline_labs']) != 1 or \
-            value['pipeline_labs'][0] != '/labs/encode-processing-pipeline/':
+            value['pipeline_labs'][0] not in [
+                '/labs/encode-processing-pipeline/',
+                '/labs/kevin-white/',
+                '/labs/yijun-ruan/'
+    ]:
         return
```
yunhailuo, line 407: I'm wondering if this will exclude something we do want? I've seen this in the old version but it doesn't seem to be for all assays. Also, maybe the pipeline title check below will ensure proper auditing so we don't need this check here?

src/encoded/audit/analysis.py
```diff
@@ -146,6 +417,180 @@ def audit_experiment_standards_dispatcher(value, system, files_structure):
...
+                'ATAC-seq (replicated)',
+                'ATAC-seq (unreplicated)'
+            ],
+            '/atac-seq/')
+        return
+
+    '''
+    ChIA-PET analyses
+    '''
+    if any(pipeline['title'] == 'Ruan Lab ChIA-PIPE Pipeline' for pipeline in value['pipelines']):
```
yunhailuo, line 562: Is this pipeline title check new? The old version checks on dataset assay_term_name? I guess we are sure all Ruan lab processed data are submitted with proper pipeline?

src/encoded/audit/analysis.py
```diff
@@ -291,24 +736,1430 @@ def check_analysis_dnase_seq_standards(
...
+
+
+def check_analysis_bulk_rna_standards(
+    value,
+    files_structure,
+    expected_pipeline_titles,
+    link_to_standards,
+):
+    if value['assembly'] not in ['GRCh38', 'mm10']:
+        return
```
yunhailuo, line 1276: Is this check universal? ChIP-seq really doesn't check this?

yunhailuo:
This is really great! I mostly focus on `audit_experiment_standards_dispatcher`. Overall, this changes should need a good QA.

Conflicts need to be resolved.

src/encoded/audit/analysis.py
jenjou: It is new. I think it's OK to keep this pipeline check, because all current ChIA-PET datasets with analysis objects, are also processed with this specific Ruan pipeline (https://www.encodeproject.org/report/?type=Experiment&assay_title=ChIA-PET&field=%40id&field=analysis_objects.pipelines&sort=analysis_objects.pipelines). Idan mentioned Leslie lab may also submit their own ChIA-PET data using a different pipeline, which shouldn't get audited under these standards.

yunhailuo: Oh. That's very good to know this is indeed specific to that pipeline. The check is then good to have.

src/encoded/audit/analysis.py
jenjou: Not universal. I tried to stay consistent with whatever was in the original Experiment audit, and some exclude hg19 and others do not.

src/encoded/audit/analysis.py
jenjou: I think having a mixed rfas audit sounds good, but maybe I can add that in a separate ticket to avoid complicating the QA.
After reviewing what we have on the portal, I did find 472 cases of > 1 rfas, but all of them were Graveley lab analyses. So at least, I don't think we are wrongly excluding any ENCODE uniformly processed pipelines with this check.

src/encoded/audit/analysis.py
yunhailuo: Thank you for clarifying. Good to stay this way then.

src/encoded/audit/analysis.py
gabdank: agree this probably should be. separate ticket

src/encoded/audit/analysis.py
gabdank: I think explicit spec of the labs we do check is OK

gabdank (dismissed):

gabdank (approved):

PR #3816, opened 2021-03-30
SNO-198-allow-configuration-of-reserved-search-params

keenangraham:
Moved all of the search defaults to `default.py` and added `RESERVED_KEYS` definition that can be passed to query builder. This adds a little bit more upfront configuration for every view but I think it makes sense to be able to specify these from `encoded`.


forresttanaka (dismissed):
Cool. What was `DEFAULT_ITEM_TYPES` for? Or maybe, under what situations would you modify its contents?

keenangraham:
> Cool. What was `DEFAULT_ITEM_TYPES` for? Or maybe, under what situations would you modify its contents?

It really controls which indices Elasticsearch will look in if you don't explicitly define a `?type=Foo` in the query string. So you'll only see results from types specified in `DEFAULT_ITEM_TYPES` if you do a plain search box search (`/search/?searchTerm=abc`).

I suppose there's some efficiency in only searching over a limited number of indices by default. It also makes sense to inject it from `encoded` since `snovault` really shouldn't know about any of these object types.

The one place we have actually customized it so far is in the top hits search, where grouping by concrete types by default was more useful than abstract types (e.g. `Dataset`). You could still override this behavior by explicitly specifying a type in the URL, e.g: https://www.encodeproject.org/top-hits-raw/?searchTerm=abc&type=QualityMetric

hitz (dismissed):

hitz (approved):

PR #3819, opened 2021-03-31
ENCD-5840-move-analysis-objects-to-dataset

bonitalam:


hitz (approved):

keenangraham:
Not a big deal and probably a lot of work now but shouldn't these be called `analyses` instead of `analysis_objects`? @yunhailuo @hitz

hitz:
> Not a big deal and probably a lot of work now but shouldn't these be called `analyses` instead of `analysis_objects`? @yunhailuo @hitz

analyses was the old name, but they transitions to objects.  I guess this might be our last chance to rename it forever, before people start writing API tests.

keenangraham:
> > Not a big deal and probably a lot of work now but shouldn't these be called `analyses` instead of `analysis_objects`? @yunhailuo @hitz
> 
> analyses was the old name, but they transitions to objects. I guess this might be our last chance to rename it forever, before people start writing API tests.

Yeah, might be worth trying to change while we still can. Not sure exactly all the code that already relies on this naming but `_objects` is redundant since all of our `linkTos` point to objects.

forresttanaka:
I agree with @keenangraham, and partly because @zoldello on a code review mentioned that variable names shouldn’t contain the type of the variable, and I generally agree. I think an exception could be made when two variables have the same contents but only differ in type (e.g. a number version of a string variable), but that doesn’t apply here.

yunhailuo:
> I agree with @keenangraham, and partly because @zoldello on a code review mentioned that variable names shouldn’t contain the type of the variable, and I generally agree. I think an exception could be made when two variables have the same contents but only differ in type (e.g. a number version of a string variable), but that doesn’t apply here.

I actually totally agree with @keenangraham and @forresttanaka . `analysis_objects` is necessary during the transfer but it should be temporal. I would think it will be good if it can be named as `analyses`.

bonitalam:
Ok, sounds like consensus is that `analysis_objects` should be changed to `analyses`.

hitz:
> Ok, sounds like consensus is that `analysis_objects` should be changed to `analyses`.

We need @gabdank to sign off.  But maybe someone could work up a s/analysis_objects/objects/g demo.  Also have to patch all the accessioning code.   I feel like maybe in snovault 2.0 we could have property aliiases or something.

bonitalam:
Closing to work off of my branch that changes `analysis_objects` should be changed to `analyses`

PR #3820, opened 2021-03-31
ENCD-5834-default-analysis-calculation

yunhailuo:


hitz (changes requested):
If we are making these enums ORDER DEPENDENT we need to put this in the comment string in the JSON file.

yunhailuo:
> If we are making these enums ORDER DEPENDENT we need to put this in the comment string in the JSON file.

Added comment and also mentioned in change log: https://github.com/ENCODE-DCC/encoded/pull/3820/commits/8ccc7bc23ae0eac4311f04a52fa12d5d9b85f6fa

hitz (approved):

PR #3825, opened 2021-03-31
ENCD-5831-add-metadata-columns-and-small-improvements-to-rnaget-ui

pedrohr:
* Small adjustments to RNA Get UI
* Adding two extra metadata columns and one facet

Demo: https://encd-5831-12f28d4cd-pedroassis.demo.encodedcc.org/rnaget

Screen Shot 2021-03-31 at 1 41 45 PM

### DIFF
--- a/src/encoded/genomic_data_service.py
+++ b/src/encoded/genomic_data_service.py
@@ -1,7 +1,7 @@
 import requests
 
 REGISTRY_DATA_SERVICE = 'genomic_data_service'
-RNA_GET_FILTERS = ['assayType', 'annotation']
+RNA_GET_FACETS = ['assayType', 'annotation', 'biosample_sex']
 RNA_GET_ENDPOINT = '/expressions/bytes'
 
 # react component orders columns by "the position" in the hash map
@@ -12,7 +12,9 @@ RNA_GET_COLUMNS = {
     'assayType': {'title': 'Assay (RNA SubType)'},
     'libraryPrepProtocol': {'title': 'Experiment'},
     'expressionID': {'title': 'File'},
-    'annotation': {'title': 'Genome Annotation'}
+    'annotation': {'title': 'Genome Annotation'},
+    'biosample_sex': {'title': 'Biosample Sex'},
+    'analysis': {'title': 'Analysis'}
 }
 
 class GenomicDataService():
@@ -41,7 +43,7 @@ class GenomicDataService():
             self.sort_order = 'desc' if desc else 'asc'
 
         self.filter_params = {}
-        for filter_ in RNA_GET_FILTERS:
+        for filter_ in RNA_GET_FACETS:
             if params.get(filter_):
                 self.filter_params[filter_] = params.get(filter_)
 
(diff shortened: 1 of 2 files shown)


src/encoded/genomic_data_service.py
```diff
@@ -114,6 +120,19 @@ def rna_get_request(self, format_='json'):
     def rna_get(self):
         self.rna_get_request()
 
+        for expression in self.expressions:
+            expression['featureID'] += f"/{expression['encodeID'].split('/')[-2]}"
+
+            for key in expression:
+                if expression[key] == None:
```
zoldello, line 127: Do you mean to do a falsy test (if Not expression[key]) or specific need to chek for None?

src/encoded/static/components/rnaget.js
```diff
@@ -93,6 +93,22 @@ AtIdRenderer.propTypes = {
...
+
+/**
+ * @id cell renderer that displays links to genes from the RNAGet API.
+ */
+const RNAGetIdRenderer = ({ value }) => (
+    <a href={`/genes/${value.split('/')[1]}`}>{value.split('/')[0]}</a>
+);
+
+RNAGetIdRenderer.propTypes = {
+    /** @id value to render */
```
zoldello, line 107: You cannot use a term to define the same term.

src/encoded/genomic_data_service.py
```diff
@@ -41,7 +47,7 @@ def parse_params(self):
             self.sort_order = 'desc' if desc else 'asc'
 
         self.filter_params = {}
-        for filter_ in RNA_GET_FILTERS:
+        for filter_ in RNA_GET_FACETS:
```
zoldello: Do you mean 
`for facet in RNA_GET_FACETS:`

zoldello (dismissed):
My comments are nothing serious. Code looks okay.

src/encoded/genomic_data_service.py
pedrohr: Specific check for None. It comes back as None from the API. If we have a falsy test it might assign an empty string if the key is not present.

src/encoded/genomic_data_service.py
pedrohr: good catch, I will change the variable name

src/encoded/genomic_data_service.py
keenangraham: I think `if expression[key] is None`? https://stackoverflow.com/a/14247419

src/encoded/static/components/rnaget.js
```diff
@@ -93,6 +93,22 @@ AtIdRenderer.propTypes = {
...
+    value: PropTypes.any.isRequired,
```
keenangraham, line 108: Is there reason this isn't `PropTypes.string.isRequired`?

src/encoded/genomic_data_service.py
```diff
@@ -12,7 +12,13 @@
     'assayType': {'title': 'Assay (RNA SubType)'},
     'libraryPrepProtocol': {'title': 'Experiment'},
     'expressionID': {'title': 'File'},
```
keenangraham, line 14: These should probably be snake case eventually to be consistent with other fields.

src/encoded/static/components/rnaget.js
keenangraham: The splitting in the renderer a little confusing. Doesn't actually seem like a normal `@id`.

keenangraham:
Maybe a different ticket but seems weird these facet terms aren't sorted by count: 

Screen Shot 2021-04-05 at 11 42 55 AM

zoldello (approved):
Looks good.

src/encoded/static/components/rnaget.js
pedrohr: good catch! yes, it should be a string

src/encoded/genomic_data_service.py
pedrohr: the behavior should be the same, but yes, `is None` is what should be used here

pedrohr:
> Maybe a different ticket but seems weird these facet terms aren't sorted by count:
> 
> Screen Shot 2021-04-05 at 11 42 55 AM

I need to make a change in the Data Service API, I will create a ticket on the DSERV project

PR #3827, opened 2021-04-01
ENCD-5829-allow-batch-download-from-more-dataset-pages

forresttanaka:
This involves both back- and front-end changes; back to allow those page types to download files, and front to enable those controls on those pages.

### DIFF
--- a/src/encoded/reports/constants.py
+++ b/src/encoded/reports/constants.py
@@ -5,6 +5,9 @@ METADATA_ALLOWED_TYPES = [
     'Experiment',
     'Annotation',
     'FunctionalCharacterizationExperiment',
+    'Reference',
+    'Project',
+    'UcscBrowserComposite',
     'PublicationData',
     'Analysis',
 ]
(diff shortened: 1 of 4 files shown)


emro (dismissed):
The code here looks good and it looks like it does what it's supposed to do. I have some feedback about the <FileGallery> component, and how the buttons look - I think the design could be a little more straightforward - but I'm not sure if this ticket is the place to address these thoughts, which I could put in Jira if that seems useful.

forresttanaka:
@emro The design has issues specifically for these three new pages, or for the File Gallery panel in general?

forresttanaka:
I now use the same criteria for the section download buttons as we’ve used for including files in dataset.files in dataset.py. This has also affected the experiment file galleries the same way but no one noticed, so hopefully this cleans it up for everything.

Also I fixed the bug with making too-long URLs to get contributing files.

forresttanaka:
Updated https://encd-5829-datasets-dl-fytanaka.demo.encodedcc.org/ to this latest version and will launch a new clean demo for tomorrow.

emro (approved):
Everything looks good, code and demo.

PR #3829, opened 2021-04-02
ENCD-5770 add legend to browser

emro:
- Adds legend to genome browser with 4 possible sets of defined colors: genome, gene, ccres, chromatin
- Genome and gene colors are displayed all the time, but ccre and chromatin colors are only displayed for tracks with relevant output types
- It is not optimized for mobile but should work ok
- States 16 and 17 have different colors for mouse and human files so the color blocks for those states are split diagonally and display both colors

### DIFF
--- a/src/encoded/static/components/genome_browser.js
+++ b/src/encoded/static/components/genome_browser.js
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import _ from 'underscore';
 import { BrowserFeat } from './browserfeat';
 import { filterForVisualizableFiles } from './objectutils';
+import Tooltip from '../libs/ui/tooltip';
 import GeneSearch from './gene_search';
 
 /**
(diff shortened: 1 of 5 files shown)


src/encoded/static/components/genome_browser.js
```diff
@@ -20,6 +21,55 @@ const GV_COORDINATES_KEY = 'ENCODE-GV-coordinates';
...
+    'Promoter-like': '#ff0000',
+    'Proximal enhancer-like': '#ffa700',
+    'Distal enhancer-like': '#ffcd00',
+    'DNase-H3K4me3': '#ffaaaa',
+    'CTCF-only': '#00b0f0',
+    'DNase-only': '#06da93',
+    'Low-DNase': '#ffffff',
+};
+
+const colorGenome = {
```
zoldello, line 55: Is there a reason why you use passive voice like "colorGenome" rather than active voice like "GenomeColors"?

src/encoded/static/components/genome_browser.js
```diff
@@ -382,6 +432,83 @@ const sortLookUp = (obj, param) => {
...
+                    <div className="legend-element" key={gene}>
+                        <div className={`legend-swatch ${colorGenes[gene] === '#ffffff' ? 'with-border' : ''}`} style={{ background: `${colorGenes[gene]}` }} />
+                        <div className="legend-label">{gene}</div>
+                    </div>
+                ))}
+            </div>
+            {(props.colorBlock.indexOf('ccres') > -1) ?
+                <div className="legend-block">
+                    <h5>CCREs</h5>
+                    {Object.keys(colorCCREs).map((ccre) => (
```
zoldello, line 478: Technically, its color and not ccre.

src/encoded/static/libs/ui/tooltip.js
```diff
@@ -67,16 +68,32 @@ const Tooltip = (props) => {
     const positionTooltip = React.useCallback(() => {
         // Determining the position of the tooltip trigger and accounting for vertical scroll
         const scrollTop = (window.pageYOffset !== undefined) ? window.pageYOffset : (document.documentElement || document.body.parentNode || document.body).scrollTop;
-        const viewportWidth = window.innerWidth || document.documentElement.clientWidth;
+        let viewportWidth = 0;
+        if (document.getElementById('content')) {
```
zoldello, line 72: useRef instead?

src/encoded/static/components/genome_browser.js
```diff
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import _ from 'underscore';
 import { BrowserFeat } from './browserfeat';
 import { filterForVisualizableFiles } from './objectutils';
+import Tooltip from '../libs/ui/tooltip';
```
zoldello, line 6: Is it safe to assume mobile in not a concern? The swash is distorted on mobile.

zoldello (dismissed):
Code looks good. My comments are very minor.

src/encoded/static/components/genome_browser.js
emro: This should be fixed, thank you for pointing it out!

src/encoded/static/libs/ui/tooltip.js
emro: Phil and I discussed, and decided it is not worth re-writing the component in React Hooks at this time, though this would definitely be a better coding solution.

zoldello (dismissed):
Looks good.

zoldello (approved):

PR #3836, opened 2021-04-06
ENCD-5825-series-files

forresttanaka:
`FileGalleryRendererComponent` can now optionally receive an array of file objects instead of retrieving them, allowing this component to get called much earlier, before `session` has a chance to get set, so I made the changes to allow a null session. We currently use both `context.session` as well as `this.props.session` in `FileGalleryRendererComponent`. We should just use one or the other, but it gets used in a way that makes me think that change shouldn’t happen in this ticket.

The `anisogenic` property to `FileGallery` hasn’t been used in a while, so I removed it. The Isogenic/Anisogenic label in the file table header gets determined a different way.

Added the `requestUri` function to avoid repeating that same pattern all over the place.


src/encoded/static/components/dataset.js
```diff
@@ -1690,6 +1690,33 @@ export const SeriesComponent = (props, reactContext) => {
...
+                (datasetFiles, dataset) => (
+                    datasetFiles.concat(
+                        dataset.files.filter(
+                            (file) => file.preferred_default
+                        )
+                    )
+                ), []
+            )
+        ) : null;
+    const analyses = context.related_datasets
```
zoldello: missing gap above:
`const analyses = context.related_datasets`
?

src/encoded/static/components/filegallery.js
```diff
@@ -1223,41 +1236,103 @@ DatasetFiles.defaultProps = {
...
+    hideGraph,
+    hideControls,
+    collapseNone,
+    showReplicateNumber,
+    showDetailedTracks,
+}, reactContext) => {
+    // Holds all schemas loaded from a request; used by the file graph.
+    const [schemas, setSchemas] = React.useState(null);
+    // Holds file search results from a request.
+    const [data, setData] = React.useState([]);
```
zoldello, line 1287: name does not seem to match the functionality. Use of variable name "data" is used for general multi-purpose variables/constants

src/encoded/static/libs/ui/panel.js
```diff
@@ -163,7 +163,7 @@ class TabPanel extends React.Component {
         // here too so that each pane knows whether it's the active one or not. ### React14
         if (this.props.children) {
             children = React.Children.map(this.props.children, (child, i) => {
-                if (child.type === TabPanelPane) {
+                if (child !== null && child.type === TabPanelPane) {
```
zoldello, line 166: What if child is empty string?

src/encoded/types/dataset.py
```diff
@@ -804,7 +804,9 @@ def control_type(self, request, related_datasets):
 class MatchedSet(Series):
     item_type = 'matched_set'
     schema = load_schema('encoded:schemas/matched_set.json')
-    embedded = Series.embedded
+    embedded = Series.embedded + [
```
zoldello, line 966: The data is small, so it won't make a big difference. But in general,  .extends() or something else my be better. '+' on list is very slow. See- https://stackoverflow.com/a/12088747/178550

zoldello (dismissed):
Code looks good. My comments are cosmetic or non-deal breakers.

src/encoded/tests/data/inserts/reference_epigenome.json
```diff
@@ -1,5 +1,5 @@
 [
-	{
+    {
```
zoldello, line 2: Is this needed?

zoldello (dismissed):
Looks good.

forresttanaka:
Updated demo: https://encd-5825-series-files-qa2-fytanaka.demo.encodedcc.org/. Will launch a clean demo shortly.

zoldello (approved):

PR #3838, opened 2021-04-07
ENCD-5806-cCRE-annotation-subtypes-and-other-file-output-types

iyoungworth:
continuing-work

schema-dependency-and-vis-changes

new-test-file

adjusted-calc-property

test-adjustment

reverted-insert-edit

fixing-embedded-inputs-on-file

test-small-change

revert-small-change

adding-facetfields-to-cart-js

required-dataset

ENCD-5806-cCRE-annotation-subtypes

requestedFacetFields now included in datasetFacets so that annotation_subtype gets included in facet objects. Also *actually* copying file array when switching to the browser tab.

adding-properties-to-long-labels

shifted-non-cart-additions

undid-move

reverted-annotation-calc-added-file-column-properties

added-element-barcode-mapping-output-type


gabdank (changes requested):
can you please add a test that checks the correct calculation of the new calculated properties in file? I think the demo you already created can be checked as is, if you only add test code, in case of any doubt probably worth spinning up another demo

iyoungworth:
New tests added, only other code change was rebasing after 5834 was reverted on dev. I deployed another demo that can be checked just in case.

gabdank (approved):

PR #3839, opened 2021-04-07
ENCD-5830-autocomplete-rnaget-ui

pedrohr:
Adding autocomplete feature to the RNA Get UI:

Screen Shot 2021-04-07 at 9 42 57 AM

### DIFF
--- a/src/encoded/genomic_data_service.py
+++ b/src/encoded/genomic_data_service.py
@@ -2,7 +2,8 @@ import requests
 
 REGISTRY_DATA_SERVICE = 'genomic_data_service'
 RNA_GET_FACETS = ['assayType', 'annotation', 'biosample_term_name', 'biosample_classification', 'biosample_sex', 'biosample_organ',  'biosample_system']
-RNA_GET_ENDPOINT = '/expressions/bytes'
+RNA_GET_EXPRESSIONS = '/expressions/bytes'
+RNA_GET_AUTOCOMPLETE = '/autocomplete'
 
 # react component orders columns by "the position" in the hash map
 RNA_GET_COLUMNS = {
@@ -40,6 +41,7 @@ class GenomicDataService():
         self.sort    = params.get('sort')
         self.columns = RNA_GET_COLUMNS.copy()
         self.columns.pop('tpm' if self.units == 'fpkm' else 'fpkm')
+        self.autocomplete_query = params.get('q', '')
 
         if self.sort:
             desc = self.sort[0] == '-'
@@ -92,8 +94,8 @@ class GenomicDataService():
         return '&'.join(params)
 
 
-    def rna_get_request(self, format_='json'):
-        url = f'{self.path}{RNA_GET_ENDPOINT}?{self.rna_get_request_query_string(format_)}'
+    def rna_get_request(self, endpoint=RNA_GET_EXPRESSIONS, format_='json'):
+        url = f'{self.path}{endpoint}?{self.rna_get_request_query_string(format_)}'
 
         results = requests.get(url, timeout=3).json()
 
@@ -115,6 +117,28 @@ class GenomicDataService():
         if self.total == 0:
             for facet in self.facets:
                 facet['total'] += 1
+
+
+    def rna_get_autocomplete(self):
+        suggestions = []
+
+        if self.autocomplete_query:
+            try:
+                url = f'{self.path}{RNA_GET_AUTOCOMPLETE}?gene={self.autocomplete_query}'
+
+                results = requests.get(url, timeout=3).json()
+
+                if 'genes' in results:
+                    suggestions = [{'text': result} for result in results['genes']]
+            except:
+                pass
+
+        return {
+            '@graph': suggestions,
+            '@type': ['suggest'],
+            '@id': f'/rnaget_autocomplete?q={self.autocomplete_query}',
+            'title': 'Suggest'
+        }
 
 
     def rna_get(self):
--- a/src/encoded/genomic_data_service.py
+++ b/src/encoded/genomic_data_service.py
@@ -136,7 +136,7 @@ class GenomicDataService():
         return {
             '@graph': suggestions,
             '@type': ['suggest'],
-            '@id': f'/rnaget_autocomplete?q={self.autocomplete_query}',
+            '@id': f'/rnaget-autocomplete?q={self.autocomplete_query}',
             'title': 'Suggest'
         }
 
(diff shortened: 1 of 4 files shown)


src/encoded/search_views.py
```diff
@@ -68,6 +68,7 @@ def includeme(config):
     config.add_route('top-hits-raw', '/top-hits-raw{slash:/?}')
     config.add_route('top-hits', '/top-hits{slash:/?}')
     config.add_route('rnaget', '/rnaget{slash:/?}')
+    config.add_route('rnaget_autocomplete', '/rnaget_autocomplete{slash:/?}')
```
keenangraham: Can we use dash instead of underscore for URL paths? https://developers.google.com/search/docs/advanced/guidelines/url-structure

pedrohr: yes, I didn't know this affects SEO, I will make the change

src/encoded/genomic_data_service.py
```diff
@@ -111,6 +113,28 @@ def rna_get_request(self, format_='json'):
...
+
+        if self.autocomplete_query:
+            try:
+                url = f'{self.path}{RNA_GET_AUTOCOMPLETE}?gene={self.autocomplete_query}'
+
+                results = requests.get(url, timeout=3).json()
+
+                if 'genes' in results:
+                    suggestions = [{'text': result} for result in results['genes']]
+            except:
```
zoldello, line 134: Code-smell. Why catch an exception but not use it? It makes debugging harder.

src/encoded/static/components/rnaget.js
```diff
@@ -9,6 +9,43 @@ import QueryString from '../libs/query_string';
...
 import { TableItemCount } from './objectutils';
 import { FacetList, TextFilter } from './search';
+import { FetchedData, Param } from './fetched';
+
+
+const AutocompleteBox = (props) => {
+    const terms = props.suggestions['@graph'];
+    const { handleClick } = props;
+
+    /* eslint-disable jsx-a11y/no-noninteractive-element-interactions, jsx-a11y/no-noninteractive-tabindex, jsx-a11y/click-events-have-key-events, arrow-body-style */
```
zoldello, line 19: Why are you disabling eslint-events on a function that is not an if-statement?

src/encoded/static/components/rnaget.js
```diff
@@ -9,6 +9,43 @@ import QueryString from '../libs/query_string';
...
+
+    if (terms && terms.length > 0) {
```
zoldello: This is not the style we normally use. We normally use  something like- 
x == 1 ? y : z

src/encoded/static/components/rnaget.js
```diff
@@ -657,42 +694,81 @@ class RNASeqMatrixSearch extends TextFilter {
...
+    onUnitsChange(e) {
         this.setState({ unitsOption: e.target.value });
     }
 
-    handleInputChange(e) {
-        this.setState({ genes: e.target.value });
+    lastGene() {
+        const inputNode = this.geneInput;
+
+        if (inputNode === undefined) {
```
zoldello: Why look for undefined exactly and not falsey?

src/encoded/static/components/rnaget.js
```diff
@@ -657,42 +694,81 @@ class RNASeqMatrixSearch extends TextFilter {
...
+            return '';
+        }
+
+        const geneTerms = inputNode.value.split(',');
+        return (geneTerms[geneTerms.length - 1].trim());
+    }
+
+    handleInputChange() {
+        this.setState({ showAutoSuggest: (this.lastGene() !== ''), genes: this.geneInput.value });
```
zoldello, line 740: can lastGene be null?

src/encoded/static/components/rnaget.js
```diff
@@ -657,42 +694,81 @@ class RNASeqMatrixSearch extends TextFilter {
...
 
     handleSubmit(e) {
         e.preventDefault();
         window.location.href = `/rnaget?genes=${this.state.genes}&units=${this.state.unitsOption}`;
     }
 
+    handleAutocompleteClick(term) {
+        const inputNode = this.geneInput;
+
+        let geneTerms = inputNode.value.split(',');
```
zoldello, line 755: This looks hacky. You declare a variable and then change it with itself...

src/encoded/static/scss/encoded/modules/_matrix.scss
```diff
@@ -1107,6 +1107,36 @@ table.matrix th.group-all-groups-cell {
     select, button {
         margin-left: 5px;
     }
+
+    .error {
+        display: none;
+    }
+
+    .rnaseq-search-autocomplete {
```
zoldello, line 1115: This is not really the BEM naming style. We enforce the format but not the principle so technically it is okay as per for code review.

zoldello (commented):
@pedrohr Let's talk about my comments.

src/encoded/genomic_data_service.py
pedrohr: for this scenario, in case of any of the multiple possible exceptions from the `requests` call, we should just ignore and return an empty result back to the client. We won't just show any autocomplete results in case of failure

src/encoded/static/components/rnaget.js
pedrohr: oh true, I will push a change

src/encoded/static/components/rnaget.js
pedrohr: falsey is indeed better, pushing a change...

src/encoded/static/components/rnaget.js
pedrohr: it shouldn't... it only reads the input of the box which is '' by default, for safety we also return '' if this is executed and the element is not yet there

src/encoded/static/components/rnaget.js
pedrohr: this is mostly for readability, I can change it if it breaks a JS pattern it is used

src/encoded/static/scss/encoded/modules/_matrix.scss
pedrohr: oh my bad, I will push a change

src/encoded/static/components/rnaget.js
pedrohr: I modified an existing Autocomplete element we have. I didn't reuse it because it was way too complex for this very simple scenario. The original Autocomplete component also turns off these eslint rules.

The render ended up very simple and easy to read, but still breaking ESlint with rules such as:
`onClick={() => handleClick(term.text)}` not being allowed. But I don't see a simpler way to fix it the way we have it. As we are going to rewrite the interface, I just decided to move on with simplicity and turn off the rules just for these lines

src/encoded/genomic_data_service.py
zoldello: Fair enough.

zoldello (dismissed):
Looks okay. Let's have QA have a take.

zoldello (approved):
Looks good

PR #3858, opened 2021-04-19
ENCD-5858-add-single-cell-units-to-series

bonitalam:


iyoungworth (dismissed):
Looks good, changelogs updated for all Series objects, new property added with test and inserts, change to related_datasets to allow SingleCellUnit on SingleCellRnaSeries

iyoungworth (dismissed):

iyoungworth (approved):
I think the code looks good, not sure if removing most the embedded objects/properties will break anything unexpectedly during QA

PR #3859, opened 2021-04-20
ENCD-5614-cart-dataset-filters

forresttanaka:
* Instead of tracking only the `@ids` of datasets, and retrieving a page of dataset objects every time the user wants to view a new page of datasets, I now load and keep all datasets (just a subset of their properties) just as file objects have worked. So functions like `CartSearchResults` no longer have to fetch dataset objects.
* We now load dataset descriptions on the cart-view page so that the selected annotations appear with full descriptions, just as they do on the annotation search page.
* `FileFacets` no longer resets facets when switching between showing default files only and all files, so we no longer need the useUpdate callback.
* I have a lot of duplication between file-facet code and dataset-facet code. I looked into trying to consolidate some of that, but it doesn’t look straightforward. Maybe someday.

### DIFF
--- a/src/encoded/static/components/cart/cart.js
+++ b/src/encoded/static/components/cart/cart.js
@@ -12,14 +12,14 @@ import * as Pager from '../../libs/ui/pager';
 import { Panel, PanelBody, PanelHeading, TabPanel, TabPanelPane } from '../../libs/ui/panel';
 import { tintColor, isLight } from '../datacolors';
 import GenomeBrowser, { annotationTypeMap } from '../genome_browser';
-import { itemClass, atIdToType } from '../globals';
+import { itemClass, atIdToType, keyCode } from '../globals';
 import {
-    requestObjects,
     ItemAccessories,
     isFileVisualizable,
     computeAssemblyAnnotationValue,
     filterForVisualizableFiles,
     filterForDefaultFiles,
+    filterForDatasetFiles,
     Checkbox,
 } from '../objectutils';
 import { ResultTableList } from '../search';
@@ -29,6 +29,7 @@ import CartClearButton from './clear';
 import CartLockTrigger from './lock';
 import CartMergeShared from './merge_shared';
 import Status from '../status';
+import CartRemoveElements from './remove_multiple';
 import { allowedDatasetTypes, defaultDatasetType } from './util';
 
 /**
@@ -92,16 +93,47 @@ const analysisFieldMap = (analysisTitles) => (
 
 
 /**
- * List of allowed dataset types. Generally maps from collection names (e.g. '/experiments/' to a
- * displayable title.
+ * Field `getValue` function so that files without a `biosample_ontology` can have a "None" facet
+ * term for those files lacking a biosample ontology.
+ * @param {object} file ...from which to extract the `biosample_ontology` property
+ * @return {object} file's biosample_ontology property, or one containing "None"
  */
-const datasetTypes = {
-    all: { title: 'All dataset types', type: '' }, // Default always first
-    experiments: { title: 'Experiments', type: 'Experiment' },
-    annotations: { title: 'Annotations', type: 'Annotation' },
-    'functional-characterization-experiments': { title: 'Functional characterizations', type: 'FunctionalCharacterizationExperiment' },
+const fileBiosampleValue = (file) => {
+    if (file.biosample_ontology) {
+        return file.biosample_ontology;
+    }
+    return { term_name: 'None' };
 };
-const DEFAULT_DATASET_TYPE = Object.keys(datasetTypes)[0];
+
+
+/**
+ * Field `getValue` function to extract the top @type from the given dataset.
+ * @param {object} dataset ...from which to extract its @type
+ * @return {string} Top-level @type of given dataset
+ */
+const datasetTypeValue = (dataset) => dataset['@type'][0];
+
+
+/**
+ * Display a term of the Dataset Term facets mapped to the English-language equivalent from the
+ * profile titles from <App> page load, e.g. "Functional Characterization Experiment" instead of
+ * "FunctionalCharacterizationExperiment."
+ */
+const DatasetTypeTermDisplay = ({ term }, { profilesTitles }) => {
+    if (profilesTitles && profilesTitles[term]) {
+        return <div className="cart-facet-term__text">{profilesTitles[term]}</div>;
+    }
+    return <div className="cart-facet-term__text">{term}</div>;
+};
+
+DatasetTypeTermDisplay.propTypes = {
+    /** Displayed facet term */
+    term: PropTypes.string.isRequired,
+};
+
+DatasetTypeTermDisplay.contextTypes = {
+    profilesTitles: PropTypes.object,
+};
 
 
 /**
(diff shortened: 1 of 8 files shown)


src/encoded/static/components/cart/cart.js
```diff
@@ -92,16 +93,47 @@ const analysisFieldMap = (analysisTitles) => (
 
 
 /**
- * List of allowed dataset types. Generally maps from collection names (e.g. '/experiments/' to a
- * displayable title.
+ * Field `getValue` function so that files without a `biosample_ontology` can have a "None" facet
```
zoldello: This comment is okay. Still, "function" in code to me means "JavaScript object" rather than "purpose/role" as you intended. So it took me a second reading to understand. Also, since "getValue" is singular, "function" which is a verb, should be plural (functions).

src/encoded/static/components/cart/cart.js
```diff
@@ -184,79 +217,96 @@ const displayedFacetFields = [
...
-            return <div className="nav result-table cart__empty-message">No visible datasets on this page.</div>;
-        }
-        return <ResultTableList results={this.state.elementsForDisplay || []} cartControls={this.props.cartControls} mode="cart-view" />;
+/**
+ * Display a page of cart contents within the cart display.
+ */
+const CartSearchResults = ({ elements, currentPage, cartControls }) => {
+    if (elements.length > 0) {
+        const pageStartIndex = currentPage * PAGE_ELEMENT_COUNT;
+        const currentPageElements = elements.slice(pageStartIndex, pageStartIndex + PAGE_ELEMENT_COUNT);
```
zoldello, line 286: pageStartIndex + PAGE_ELEMENT_COUNT can exceed elements.length and there is no check for this. But JavaScript fails gracefully.

src/encoded/static/components/cart/cart.js
```diff
@@ -382,67 +432,53 @@ CartFiles.defaultProps = {
...
-            );
-        }
+    React.useEffect(() => {
+        // Inserts the <div> to show the yellow highlight when the file count changes.
+        setTriggerEnabled(true);
+    }, [count]);
 
-        return <progress value={facetLoadProgress} max="100" />;
+    if (facetLoadProgress === -1) {
+        const fileCountFormatted = count.toLocaleString ? count.toLocaleString() : count.toString();
```
zoldello, line 443: Good idea. Some numbers differ from America. Like in parts of Europe, "1,000" is written as "1.000"

src/encoded/static/components/cart/cart.js
```diff
@@ -1186,6 +1349,66 @@ CartPager.propTypes = {
...
+
+
+/**
+ * Add the datasets search results to the given array of datasets.
+ * @param {array} datasets Add new search results to this array
+ * @param {object} currentResults Dataset search results
+ * @return {array} `datasets` with search-result datasets added
+ */
+const addToAccumulatingDatasets = (datasets, currentResults) => {
+    if (currentResults['@graph'] && currentResults['@graph'].length > 0) {
```
zoldello, line 1389: Storing currentResults['@graph'] may make the code better, albeit computationally it won't make a difference since access to array entry is O(1)

src/encoded/static/components/objectutils.js
```diff
@@ -870,6 +870,20 @@ export const filterForDefaultFiles = (fileList) => (
 );
 
 
+/**
+ * Filter the given files to only include those that appear in the given datasets.
+ * @param {array} fileList Files to filter
+ *
```
zoldello, line 884: Why the gap?

zoldello (dismissed):
Looks okay and UI looks okay too. My comments on the PR have nothing significant that needs changing.

src/encoded/static/components/objectutils.js
forresttanaka: There was something I saw that seemed authoritative that there should be a gap before @returns, but now I don’t see it and wonder if I imagined it. I’ll not do that anymore.

forresttanaka:
I’m going to have to rebase to get a demo started.

zoldello (dismissed):

src/encoded/tests/features/cart.feature
```diff
@@ -8,7 +8,7 @@ Feature: Cart
         When I press "/experiments/ENCSR611ZAL/"
         And I press "/experiments/ENCSR000INT/"
         And I press "/experiments/ENCSR000DZQ/"
-        And I wait for the content to load
+        And I wait for 5 seconds
```
zoldello, line 11: Why did you use 5 seconds rather than wait until the DOM-object is loaded?

src/encoded/static/components/cart/cart.js
forresttanaka: I agree it reads awkwardly. Not sure if I improved it or not.

src/encoded/tests/features/cart.feature
forresttanaka: I found that waiting for the content to load didn’t help — the test just barged on even if the cart was still loading.

forresttanaka:
Redid how the raw and genome browser file lists get calculated so that they take the currently selected set of datasets into account. I also enabled the progress bar in the dataset facet as I had originally thought it wasn’t necessary, but using it with large carts proved me wrong.

Updated demo: https://encd-5614-cart-dataset-filters-fytanaka.demo.encodedcc.org/

I’ll start a new clean demo for tomorrow.

src/encoded/static/components/cart/cart.js
```diff
@@ -1113,6 +1114,8 @@ DatasetFacets.propTypes = {
     selectedDatasetCount: PropTypes.number,
     /** Callback when the user clicks on a file format facet item */
     termClickHandler: PropTypes.func.isRequired,
+    /** Facet-loading progress for progress bar, or null if not displayed */
+    facetLoadProgress: PropTypes.number,
```
zoldello, line 1106: Should you use PropTypes.oneOfType instead of PropTypes.number to cover the null case?

zoldello (dismissed):
Looks good.

src/encoded/static/components/cart/cart.js
```diff
@@ -184,79 +185,116 @@ const displayedFacetFields = [
...
-    { field: 'analyses', dataset: true },
-    { field: 'preferred_default' },
-    { field: 'annotation_subtype', dataset: true },
-    { field: 'biochemical_inputs', dataset: true },
-]);
-
-/** Facet `field` values for properties from dataset instead of files */
-const datasetFacets = displayedFacetFields.concat(requestedFacetFields).filter((facetField) => facetField.dataset).map((facetField) => facetField.field);
-
+// const requestedFacetFields = displayedFacetFields.filter((field) => !field.calculated).concat([
```
zoldello, line 190: Why is this not removed?

zoldello (dismissed):
Looks good.

src/encoded/static/components/cart/cart.js
forresttanaka: Oh dang; I forgot to remove it.

forresttanaka:
Closing to implement a completely different design.

PR #3860, opened 2021-04-20
ENCD-5891-exclude-rrbs-from-audits

jenjou:


gabdank (changes requested):
please add a test making sure the RRBS are not audited

gabdank (approved):

PR #3867, opened 2021-04-22
ENCD-5894-restrict-strand-specificity-requirement-to-RNA-libraries

iyoungworth:
updated-fixtures-and-inserts

adjusted-expected-metadata

corrected-other-expected-metadata


src/encoded/schemas/library.json
```diff
@@ -67,6 +67,33 @@
                 }
             ]
         },
+        "nucleic_acid_term_name": {
+            "oneOf": [
+                {
+                    "required": ["strand_specificity"],
+                        "properties": {
```
gabdank: I think the indentation here is off?

gabdank (changes requested):
indentation is off

src/encoded/schemas/library.json
gabdank: properties should be same indentation as required

iyoungworth: Ah you're right- the section below has an extra indent also (L87-92), fixed both

gabdank (approved):

PR #3868, opened 2021-04-23
ENCD-5863-speed-up-chip-matrix

zoldello:


src/encoded/static/components/datatable.js
```diff
@@ -178,4 +178,38 @@ DataTable.propTypes = {
...
+/**
+ * This generates a table but built with div-tags rather than table-tag
+ *
+ * @param {tableData} Data for the table. It should be an array of arrays of object. For example-
+ * [
+ *  [{className: id: 'an-id', 'a-class', style: 'a-style', content: <span>text</span>}],
+ *  [{className: id: 'an-id2', 'a-class-2', style: 'a-style-2', content: <span>text 2</span>}],
+ * ]
+ *
+ * For entires, id is the key for react. className is the class or classes associated with entry. The style should be an object (key/value pair)
```
forresttanaka, line 190: “entries”?

src/encoded/static/components/datatable.js
```diff
@@ -178,4 +178,38 @@ DataTable.propTypes = {
...
+
+DivTable.propTypes = {
+    /** Whole table data */
+    tableData: PropTypes.array.isRequired,
+};
+
+export {
+    DataTable,
+    DivTable,
+};
```
forresttanaka, line 215: An alternative is to leave `DataTable` as default export and `DivTable` an alternate export, since these two components don’t really have equivalent functionality, and then other files don’t have to change their import.

src/encoded/static/components/matrix_chip_seq.js
```diff
@@ -118,50 +117,37 @@ const getChIPSeqData = (context, assayTitle, organismName) => {
...
-    const dataTable = [];
-    const headerRow = chIPSeqData.headerRow.map((x) => ({
-        header: <a href={`/search/?type=Experiment&status=released&replicates.library.biosample.donor.organism.scientific_name=${chIPSeqData.organismName}&biosample_ontology.term_name=${x}&assay_title=${chIPSeqData.assayTitle}${isAssayTitleHistone ? '&assay_title=Mint-ChIP-seq' : ''}`} title={x}>{x}</a>,
-    }));
-
-    dataTable.push({
-        rowContent: [{ header: null }, ...headerRow],
-        css: 'matrix__col-category-header',
-    });
+    const removeSpecialCharacters = (name) => (!name ? name : name.replace(/\s/g, ''));
```
forresttanaka, line 134: I feel like we should add a function to globals.js to convert arbitrary strings to something acceptable for ids and maybe CSS class names.

forresttanaka (approved):
OK, from what I can tell, this seems fine. I’m not sure how much better the speed is in general? I just wonder if we should consider whether it’s worth it to put in something with what sounds like minor speed benefits, and instead focus on some big speed improvement over multiple sprints, and which we can apply to multiple matrices?

PR #3869, opened 2021-04-23
ENCD-5808 facet for biochemical inputs

emro:
Adds a new "biochemical_inputs" facet for annotations with the following properties:
- Facet applies to annotations but is only present on the encyclopedia page, not the search page for annotations
- Terms work like regular facet links but have no negation and look like buttons
- An additional "All" button is present which allows you to select all biochemical inputs or de-select all biochemical inputs, and it automatically highlights/unhighlights if all terms are selected/unselected

A couple thoughts:
- I originally called the new facet a "multiselect" which made sense to me at the time but now I'm not sure. However I didn't think of anything better so I am leaving it for now.
- The facet jumps position-wise in the sidebar based on whether or not "Target" is an available facet so maybe we will want to rearrange or maybe this is ok.

### DIFF
--- a/src/encoded/schemas/annotation.json
+++ b/src/encoded/schemas/annotation.json
@@ -254,8 +254,8 @@
             "length": "long",
             "open_on_load": true
         },
-        "annotation_type": {
-            "title": "Annotation type",
+        "annotation_subtype": {
+            "title": "Annotation subtype",
             "length": "long",
             "open_on_load": true
         },
@@ -269,6 +269,9 @@
         "organism.scientific_name": {
             "title": "Organism",
             "open_on_load": true
+        },
+        "biochemical_inputs": {
+            "title": "Biochemical inputs"
         },
         "targets.investigated_as": {
             "title": "Target category"
--- a/src/encoded/static/components/encyclopedia.js
+++ b/src/encoded/static/components/encyclopedia.js
@@ -1,0 +1,53 @@
+import PropTypes from 'prop-types';
+import url from 'url';
+import { Panel, PanelBody } from '../libs/ui/panel';
+import { ResultTable } from './search';
+import * as globals from './globals';
+
+const Encyclopedia = (props, context) => {
+    const searchBase = url.parse(context.location_href).search || '';
+
+    const currentRegion = (assembly, region) => {
+        let lastRegion = {};
+        if (assembly && region) {
+            lastRegion = {
+                assembly,
+                region,
+            };
+        }
+        return lastRegion;
+    };
+
+    return (
+        <div className="layout">
+            <div className="layout__block layout__block--100">
+                <div className="block encyclopedia" data-pos="0,0,0">
+                    <h1>{props.context.title}</h1>
+                    <Panel>
+                        <PanelBody>
+                            <ResultTable
+                                {...props}
+                                searchBase={searchBase}
+                                onChange={context.navigate}
+                                currentRegion={currentRegion}
+                                hideDocType
+                            />
+                        </PanelBody>
+                    </Panel>
+                </div>
+            </div>
+        </div>
+    );
+};
+
+Encyclopedia.propTypes = {
+    context: PropTypes.object.isRequired,
+};
+
+Encyclopedia.contextTypes = {
+    location_href: PropTypes.string,
+    navigate: PropTypes.func,
+    fetch: PropTypes.func,
+};
+
+globals.contentViews.register(Encyclopedia, 'Encyclopedia');
--- a/src/encoded/static/components/facets/biochemical_inputs.js
+++ b/src/encoded/static/components/facets/biochemical_inputs.js
@@ -1,0 +1,204 @@
+import PropTypes from 'prop-types';
+import FacetRegistry from './registry';
+import QueryString from '../../libs/query_string';
+
+
+/**
+ * Biochemical input facet works just like a regular facet but negating is not allowed, links look more like buttons, and there is an "Any" button to select any possible terms
+ */
+const BiochemicalInputsFacet = ({ results, facet, queryString }, reactContext) => {
+    // Get query terms from url
+    const originalQuery = new QueryString(queryString);
+    const originalInputs = originalQuery.getKeyValues('advancedQuery');
+
+    // Possible query entries must be hard-coded because the facet object does not always include all terms, plus we want to exclude rDHS
+    const facetTerms = ['CTCF', 'DNase-seq', 'H3K27ac', 'H3K4me3'];
+
+    // URl should indicate toggle position, but if not, default is "only"
+    let switchState = 'only';
+    if (reactContext.location_href.indexOf('#or') > -1) {
+        switchState = 'or';
+    }
+    // If no toggle position is indicated, "or" is the position when an "AND NOT" is present or when search is not biochemical_inputs:*
+    if (originalInputs && originalInputs[0]) {
+        const defaultOnly = (originalInputs[0] === 'biochemical_inputs:*') || (originalInputs[0].toString().indexOf('AND NOT') > -1);
+        if ((reactContext.location_href.indexOf('#') === -1) && !(defaultOnly)) {
+            switchState = 'or';
+        }
+    }
+
+    // Generate array of search parameters, ones selected by AND and de-selected by AND NOT
+    let originalInputsArray = [];
+    let originalNonincludedArray = [];
+    if (originalInputs.length > 0) {
+        originalInputsArray = originalInputs[0].toString().replace('biochemical_inputs:(', '').replace(')', '').split(' AND ');
+        originalInputsArray = originalInputsArray.filter((input) => input.indexOf('NOT') === -1);
+    }
+    if (originalInputs[0] === 'biochemical_inputs:*') {
+        originalInputsArray = facetTerms;
+    }
+    originalNonincludedArray = facetTerms.filter((key) => originalInputsArray.indexOf(key) === -1);
+
+    // Generate object which includes url for each button (each button is essentially a link)
+    // The object also indicates whether the button is currently selected or not (whether the url will be adding or removing that term)
+    const facetObj = {};
+    facetTerms.forEach((term) => {
+        const query = new QueryString(queryString);
+        facetObj[term] = {};
+        // Add new term if it does not exist
+        if (originalInputsArray.indexOf(term) === -1) {
+            const newInputsArray = originalInputsArray.concat([term]);
+            const nonincludedArray = originalNonincludedArray.filter((elem) => elem !== term);
+            if (switchState === 'only') {
+                if (nonincludedArray.length === 0) {
+                    query.replaceKeyValue('advancedQuery', 'biochemical_inputs:*');
+                } else {
+                    query.replaceKeyValue('advancedQuery', `biochemical_inputs:(${(Array.isArray(newInputsArray) && (newInputsArray.length > 0)) ? newInputsArray.join(' AND ') : newInputsArray}${(Array.isArray(nonincludedArray) && (nonincludedArray.length > 0)) ? ` AND NOT ${nonincludedArray.join(' AND NOT ')}` : ''})`);
+                }
+            } else {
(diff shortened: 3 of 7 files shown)


src/encoded/schemas/annotation.json
```diff
@@ -289,6 +289,9 @@
             "type": "typeahead",
             "open_on_load": true
         },
+        "biochemical_inputs": {
+            "title": "Biochemical inputs"
+        },
```
forresttanaka: Nothing to do with your changes, but sort of odd we have two "annotation_type" facets defined. The second overrides the first. If you feel like deleting the first one in this branch, maybe ask Ingrid if she did that deliberately.

src/encoded/static/components/facets/defaults.js
```diff
@@ -763,6 +763,88 @@ DefaultDateSelectorFacet.contextTypes = {
...
+ */
+export const DefaultMultiselectFacet = ({ facet, relevantFilters, queryString }, reactContext) => {
+    let initializeBiochemicalInput = relevantFilters.map((filter) => filter.term);
+    if (relevantFilters.length === facet.terms.length) {
+        initializeBiochemicalInput = initializeBiochemicalInput.concat(['All']);
+    }
+    const [biochemicalInput, setBiochemicalInput] = React.useState(initializeBiochemicalInput);
+
+    // Select or de-select biochemical input terms
+    const handleClick = React.useCallback((event) => {
```
forresttanaka: In this case, the cache created by `useCallback` never gets used because you pass `handleClick` to a `<button>` element which gets rerendered regardless of whether its properties have changed since the last render or not. This doesn’t affect operation, but this cache just takes up memory without preventing the `<button>` from rerendering.

src/encoded/static/components/facets/defaults.js
```diff
@@ -763,6 +763,88 @@ DefaultDateSelectorFacet.contextTypes = {
...
+        const href = `?${query.format()}`;
+        reactContext.navigate(href);
+    });
+
+    return (
+        <div className="facet">
+            <h5>{facet.title}</h5>
+            <div className="facet__multiselect">
+                {facet.terms.sort((a, b) => (a.key).localeCompare(b.key)).map((term) => (
+                    <button className={(biochemicalInput.indexOf(term.key) > -1) ? 'selected' : ''} key={term.key} id={term.key} onClick={handleClick} type="button">
```
forresttanaka: You can also do this without the `id` by having an inline function in `onClick` that passes `term.key` to `handleClick`, which can then get the id from its parameter instead of extracting it from the event. What you have works too of course, but you see the inline function method a lot, and is a little bit simpler.

src/encoded/static/components/facets/defaults.js
```diff
@@ -763,6 +763,88 @@ DefaultDateSelectorFacet.contextTypes = {
...
+                        {term.key}
+                    </button>
+                ))}
+                <button className={(biochemicalInput.indexOf('All') > -1) ? 'selected' : ''} key="All" id="All" onClick={handleClick} type="button">
+                    All
+                </button>
+            </div>
+        </div>
+    );
+};
```
forresttanaka: I think `DefaultMultiselectFacet` should move to biochemical_inputs.js as the `BiochemicalInputsFacet` implementation, because this seems much too specific for defaults.js. If we needed another multi-select facet in the future, it would probably not use "biochemical_inputs" and might not have an "All" option, so it couldn’t use this default component.

I think you can simplify this by not having not having the `biochemicalInput` state. Instead, calculate all the button states on each render entirely based on the query string, and you can avoid the click handler by using`<a>` tags for buttons. Each button’s href would take some calculation, but probably not too bad. I’d guess you can make this component about half its current size with this method.

src/encoded/static/components/search.js
```diff
@@ -1047,7 +1047,12 @@ export const FacetList = (props) => {
     }
 
     // Combine facets from search results with special facets, and treat them mostly the same.
-    const allFacets = SpecialFacetRegistry.Facet.getFacets().concat(facets);
+    let allFacets = SpecialFacetRegistry.Facet.getFacets().concat(facets);
+
+    // Biochemical inputs facet exists for annotations but we only want to display it for the Encyclopedia page
+    if (context['@type'].indexOf('Encyclopedia') === -1) {
+        allFacets = allFacets.filter((facet) => facet.field !== 'biochemical_inputs');
+    }
```
forresttanaka: Since we got rid of all the special cases in the generic facet code, I’d like to avoid putting special cases back. You can have `BiochemicalInputsFacet` supress its own display if its `results` property doesn’t have “Encyclopedia” in its `@type[0]`. Then you don’t need this special-case code.

forresttanaka (changes requested):
OK, the buttons look nice! I have changes to simplify the facet code and to take care of the new special case in search.js. It looks like Safari still has buttons that grow in height when highlighted, unless the demo I tried doesn’t have that fix in it.

src/encoded/static/components/facets/defaults.js
emro: I'm not sure why I coded it this way, but I agree it wasn't needed!

src/encoded/static/components/facets/defaults.js
emro: Another excellent point.

src/encoded/static/components/facets/defaults.js
emro: I've been thinking about using hrefs instead of the click handler and I'm not sure if it will be simpler, but maybe I'm missing something. I think I would have to check the url to see if the term is already present, and add it if not, and then check the url for all of the terms for the "All" link/button and construct the url based on that. This seems like I would be to some extent duplicating code / logic that already exists in the QueryString library? Maybe I'm over-thinking this or missing something though. I was thinking you were suggesting using location_href and doing string parsing.

src/encoded/static/components/search.js
emro: I also fixed this.

src/encoded/static/components/facets/defaults.js
forresttanaka: > I think I would have to check the url to see if the term is already present, and add it if not, and then check the url for all of the terms for the "All" link/button and construct the url based on that

That’s what I was thinking.

> This seems like I would be to some extent duplicating code / logic that already exists in the QueryString library?

I think you could _use_ the QueryString library to parse and modify the query string for each button. The `getKeyValues` method can tell you all the values in every “biochemical_input=” query-string parameter, and using that to determine which buttons get highlighted, along with (probably cloning and) updating the query-string object to generate the new links for each button.

src/encoded/schemas/annotation.json
emro: I checked with Ingrid about this and she says that the second facet should be "annotation_subtype". I will update this.

src/encoded/static/components/facets/defaults.js
emro: Ok, well I re-wrote it, hopefully this is sort of along the lines of what you were thinking!

Since all the buttons are all now links but we wanted to keep the button styles, I had to add some styles and I think I resolved the Safari border issue which I had forgotten to address previously.

src/encoded/static/components/facets/biochemical_inputs.js
```diff
@@ -0,0 +1,81 @@
...
+            if (initializeBiochemicalInput.indexOf(term.key) === -1) {
+                query.addKeyValue('biochemical_inputs', term.key);
+            }
+        });
+        facetObj.All.selected = false;
+        facetObj.All.link = `?${query.format()}`;
+    }
+
+    return (
+        <div className={`facet ${reactContext.location_href.indexOf('encyclopedia') === -1 ? 'hide' : ''}`}>
```
forresttanaka: Probably easier to use the rarely used `results['@type']` property that you don’t pull in here, just to avoid the highly unusual situation where someone tricks your code by putting a spurious “encyclopedia” in the query string somewhere.

src/encoded/schemas/annotation.json
```diff
@@ -265,8 +265,8 @@
             "length": "long",
             "open_on_load": true
         },
-        "annotation_type": {
-            "title": "Annotation type",
+        "annotation_subtype": {
+            "title": "Annotation subtype",
```
forresttanaka, line 258: Thanks! I didn’t think about the case where both were needed, just with a different property name. VSCode is really strict about JSON properties, and had marked this one as duplication error.

forresttanaka (dismissed):
Looks very cool!

src/encoded/static/components/facets/biochemical_inputs.js
emro: I fixed this with the QA update. I think maybe "encyclopedia" could be in the url if an encyclopedia version is chosen from the encyclopedia version facet so this was a good catch.

forresttanaka (dismissed):
Looks cool.

src/encoded/static/components/facets/biochemical_inputs.js
```diff
@@ -0,0 +1,203 @@
...
+    return (
+        <div className={`facet ${results['@type'].indexOf('Encyclopedia') === -1 ? 'hide' : ''}`}>
+            <h5>{facet.title}</h5>
+            <label className="boolean-switch">
+                <div style={switchStyles} className="boolean-switch__frame">
+                    <div style={actuatorStyles} className="boolean-switch__actuator" />
+                </div>
+                <input type="checkbox" checked={switchState === 'only'} onChange={toggleSwitchState} />
+                <div className="boolean-switch__title">Only</div>
+            </label>
```
forresttanaka, line 163: Someday it might be useful to put this switch into its own component so multiple modules can use it with one `<Component />` call.

forresttanaka (dismissed):
Looks nice! The logic for determining the switches became very complex, but looks like it just has very complex conditions.

src/encoded/static/components/facets/biochemical_inputs.js
emro: That's a good idea. There were a lot of cases to account for with the conditions. I hope the comments were somewhat clear, at least?

emro:
I should probably rebase and start a new demo, but since building demos isn't working right now, I thought it might make more sense to hold off on those things. I'm not sure if you want to code review now or later when demos are working, but the old demo is updated.

forresttanaka (dismissed):
I like how the description looks. Should it appear _below_ the controls? That way people who already know how it works can just get to the controls right away.

If you wanted slightly shorter text, one suggestion:
All search results match your selections exactly when “Only” is enabled. When disabled, your search results must include your selections and might include others.

I’m not sure if that exactly describes how it works though.

src/encoded/static/components/encyclopedia.js
```diff
@@ -0,0 +1,53 @@
...
+                region,
+            };
+        }
+        return lastRegion;
+    };
+
+    return (
+        <div className="layout">
+            <div className="layout__block layout__block--100">
+                <div className="block encyclopedia" data-pos="0,0,0">
```
forresttanaka, line 24: What’s `data-pos` for?

src/encoded/static/components/facets/biochemical_inputs.js
```diff
@@ -0,0 +1,204 @@
...
+    // Possible query entries must be hard-coded because the facet object does not always include all terms, plus we want to exclude rDHS
+    const facetTerms = ['CTCF', 'DNase-seq', 'H3K27ac', 'H3K4me3'];
+
+    // URl should indicate toggle position, but if not, default is "only"
+    let switchState = 'only';
+    if (reactContext.location_href.indexOf('#or') > -1) {
+        switchState = 'or';
+    }
+    // If no toggle position is indicated, "or" is the position when an "AND NOT" is present or when search is not biochemical_inputs:*
+    if (originalInputs && originalInputs[0]) {
```
forresttanaka, line 23: `getKeyValues` always returns an array, which might contain nothing. So `originalInputs` should always exist, though it might contain an empty array. Not that it makes much difference; just a technicality.

forresttanaka (approved):

PR #3878, opened 2021-04-27
ENCD-5854-series-download

forresttanaka:
@keenangraham Could you check the batch download and metadata code? This branch allows all series-derived pages to perform a batch download. I’m spinning up a demo which will be ready Tuesday evening.


src/encoded/reports/metadata.py
```diff
@@ -478,6 +480,53 @@ def _generate_rows(self):
...
+
+    def _get_column_to_fields_mapping(self):
+        return SERIES_METADATA_COLUMN_TO_FIELDS_MAPPING
+
+    def _should_not_report_file(self, file_):
+        """Only include preferred_default files in addition to existing
+        criteria.
+        """
+        is_file_not_reportable = super()._should_not_report_file(file_)
+        return is_file_not_reportable or not file_.get('preferred_default', False)
```
keenangraham, line 508: I wonder if it makes sense to just always add `&related_datasets.files.preferred_default=true` to the frontend batch_download link? Putting it here works, but might reduce flexibility in the future if someone was interested in series files that weren't preferred defaults.

keenangraham (approved):
Nice. Looks like a good addition. The only thing I would recommend is adding a few series-specific tests for both batch_download and metadata. You can see following for some examples:

https://github.com/ENCODE-DCC/encoded/blob/0a4868a8f1602f4f6d242493379d01838afddef6/src/encoded/tests/test_reports_metadata.py#L1723

https://github.com/ENCODE-DCC/encoded/blob/0a4868a8f1602f4f6d242493379d01838afddef6/src/encoded/tests/test_reports_metadata.py#L1807

https://github.com/ENCODE-DCC/encoded/blob/0a4868a8f1602f4f6d242493379d01838afddef6/src/encoded/tests/test_reports_metadata.py#L2261

https://github.com/ENCODE-DCC/encoded/blob/0a4868a8f1602f4f6d242493379d01838afddef6/src/encoded/tests/test_reports_batch_download.py#L44

src/encoded/reports/metadata.py
forresttanaka: Thanks! I’ll give that preferred_default idea a little thought, and I’ll add the tests.

PR #3879, opened 2021-04-27
ENCD-5898-separate-ami-demo-prod

ottojolanki:


src/encoded/commands/deploy.py
```diff
@@ -1077,7 +1077,14 @@ def hostname(value):
...
     if not args.image_id:
-        current_ami_id = _fetch_ami_id()
+        if args.profile_name == "default":
+            print("Fetching development AMI.")
+            current_ami_id = _fetch_ami_id()
+        elif args.profile_name == "production":
+            print("Fetching production AMI.")
+            current_ami_id = _fetch_ami_id(key="ami-id/production/current_ami_id.txt")
+        else:
+            raise ValueError("--profile-name must be default or production")
```
paul-sud: If it's not hard, maybe we could make this into `choices` in the argument parser?

paul-sud (changes requested):
One potential change requested.

src/encoded/commands/deploy.py
ottojolanki: Yea, good point. That is the way.

src/encoded/commands/deploy.py
```diff
@@ -1024,7 +1024,7 @@ def hostname(value):
     parser.add_argument('--wale-s3-prefix', default='s3://encoded-backups-prod/production-pg11')
 
     # AWS
-    parser.add_argument('--profile-name', default='default', help="AWS creds profile")
+    parser.add_argument('--profile-name', default='default', choices = ("default", "production"), help="AWS creds profile")
```
paul-sud: Formatting nit, there shouldn't be spaces around the `=` for kwargs:
```suggestion
    parser.add_argument('--profile-name', default='default', choices=("default", "production"), help="AWS creds profile")
```

paul-sud (changes requested):
One formatting change requested

paul-sud (dismissed):

src/encoded/commands/deploy.py
ottojolanki: correct. man, we really need to get some automatic formatting going in this repo.

paul-sud: For sure, the current formatting is a mess.

paul-sud (approved):

PR #3883, opened 2021-04-30
ENCD-5854-support-series-page-downloads

forresttanaka:
The key change was to allow a dataset page to pass an array of file objects to the file gallery, which stops the file gallery from retrieving the file objects itself. So Series pages grab the already-embedded files from `related_datasets.files` and pass them directly to the file gallery to use immediately.


forresttanaka:
Keenan reviewed the Python, but I have a small update to that, and added indexer tests.

keenangraham (approved):

emro (approved):
Looks good!

PR #3893, opened 2021-05-08
ENCD-5689-update-chip-matrix

zoldello:


zoldello:
Changed UX and will. create a new PR.

PR #3896, opened 2021-05-10
ENCD-5909-fix-single-cell-set-ui

zoldello:


bonitalam:
Going to work off other branch

PR #3897, opened 2021-05-11
ENCD-5909-single-cell-series-UI

bonitalam:


src/encoded/static/components/dataset.js
```diff
@@ -1794,7 +1794,7 @@ export const SeriesComponent = (props, reactContext) => {
         if (d.replicates && d.replicates.length > 0) {
             biosamples = d.replicates.map((replicate) => replicate.library && replicate.library.biosample);
         }
-        if (biosamples && biosamples.length > 0) {
+        if (biosamples && biosamples.treatments && biosamples.length > 0) {
```
forresttanaka: This if statement never passes because biosamples is an array, so it can’t have any properties itself, such as `treatments`. I think you intended “If we have any biosamples, and if any of those biosamples have treatments”?

If that’s the case, I’d simply revert this line and I think it should all work OK. All biosamples have `treatments` even if empty, so the `biosample.treatments.forEach` line below works fine — it just skips the loop if `biosample.treatments` has no elements in it.

forresttanaka (changes requested):

forresttanaka (dismissed):
That should work! Another way is to check if the replicates are strings or objects, and skip if they’re strings. That has the advantage of adapting automatically if we de-embed replicates from another series type. But that seems unlikely so this should work fine.

src/encoded/static/components/dataset.js
```diff
@@ -1894,10 +1894,12 @@ export const SeriesComponent = (props, reactContext) => {
...
-                            <div data-test="donordiversity">
-                                <dt>Donor diversity</dt>
-                                <dd>{diversity}</dd>
-                            </div>
+                            {seriesType !== 'SingleCellRnaSeries' ?
+                                <div data-test="donordiversity">
+                                    <dt>Donor diversity</dt>
+                                    <dd>{diversity}</dd>
+                                </div>
+                            : null}
```
forresttanaka, line 1902: While this is correct, I think it would be less prone to future breakage if you used the `diversity` variable as the condition for displaying itself so that you test the very thing you display, e.g.:

```
{diversity ?
    <div data-test="donordiversity">
        <dt>Donor diversity</dt>
        <dd>{diversity}</dd>
     </div>
: null}
```
…kind of thing. Then above on line 1836 you can do something like:
```
    const diversity = seriesType !== 'SingleCellRnaSeries' ? donorDiversity(context) : null;
```
That way if some future condition includes no diversity, you can put that logic where we calculate the diversity, and the part where you display `diversity` remains unchanged. Plus the `donorDiversity` calculation is a little bit involved, so it’s nice to skip it completely if you don’t need it.

forresttanaka (changes requested):

src/encoded/static/components/dataset.js
bonitalam: Thank you @forresttanaka! That makes a lot of sense (especially avoiding the calculation altogether if you don't need to do it), I have a lot to learn when it comes to JS. Will try implementing your suggestion.

forresttanaka (approved):
😎

PR #3903, opened 2021-05-12
ENCD-4945-status-color-key

khzlin:


zoldello (commented):
@khzlin  Let's work on this together. Slack me when you see this so we can arrange a time.

zoldello (changes requested):
@khzlin  Let's work on this together. Slack me when you see this so we can arrange a time.

zoldello (approved):
Looks good.

PR #3908, opened 2021-05-17
ENCD-5912-FCC-series-search-update

khzlin:


src/encoded/static/components/search.js
```diff
@@ -708,6 +728,7 @@ const DatasetComponent = (props, reactContext) => {
                                         )
                                     </span>
                                 : null}
+                                {examineLociGene && examineLociGene.length > 0 && fccSeries ? <span>{` targeting ${examineLociGene.join(', ')}`} </span> : null}
```
emro: I feel like this line and the one on line 716 would be easier to read if the ternary was spread over separate lines. I think most of the code in Encode follows that convention.

src/encoded/static/components/search.js
```diff
@@ -605,6 +610,17 @@ const DatasetComponent = (props, reactContext) => {
                 if (dataset.assay_term_name) {
                     assays.push(dataset.assay_term_name);
                 }
+                if (dataset.crispr_screen_readout) {
+                    crisprReadout.push(dataset.crispr_screen_readout);
+                }
+                if (dataset.examined_loci && dataset.examined_loci.length > 0) {
+                    dataset.examined_loci.forEach((loci) => {
+                        examineLociGene.push(loci.gene.symbol);
+                    });
```
emro, line 618: Can you do: `let examineLociGene = dataset.examined_loci.map((loci) => loci.gene.symbol);` here?
I haven't tried this, I'm not sure if it would work depending on what the data looks like, but it could be shorter if it does.

emro (changes requested):
Minor style changes, one possible change that could make the code shorter / more readable but also might not work.

src/encoded/static/components/search.js
khzlin: ok, fixed

src/encoded/static/components/search.js
khzlin: Oh I tried it but it only got the first examined_loci gene from each dataset. I think we would want all examine_loci gene (if more than one) from each dataset in the series.

src/encoded/static/components/search.js
```diff
@@ -591,6 +595,7 @@ const DatasetComponent = (props, reactContext) => {
     const treatmentConcentration = result['@type'].indexOf('TreatmentConcentrationSeries') >= 0;
     const showTreatment = treatmentTime || treatmentConcentration;
     const organismSeries = result['@type'].indexOf('OrganismDevelopmentSeries') >= 0;
+    const fccSeries = result['@type'].indexOf('FunctionalCharacterizationSeries') >= 0;
```
forresttanaka, line 597: Not the purpose of this ticket, but I think we should eventually break all these different dataset types into their own components so we don't need to keep checking for each type, which is becoming a pain. Also probably make new sub components to display things in common between them.

src/encoded/static/components/search.js
```diff
@@ -696,9 +713,24 @@ const DatasetComponent = (props, reactContext) => {
...
+                                {(assays && assays.length > 0 && fccSeries) ?
+                                    <span>
+                                        {` ${assays.join(', ')} series`}
+                                    </span>
+                                : null}
+                                {(crisprReadout && crisprReadout.length > 0 && fccSeries) ?
+                                    <span>
+                                        {` (${crisprReadout.join(', ')})`}
+                                    </span>
+                                : null}
```
forresttanaka, line 734: Technically, you only need to check for the length of `perturbationType` and the other arrays, because you already declared their existence above. This has a minor performance penalty in this case — mostly for cleanliness.

src/encoded/static/components/search.js
```diff
@@ -696,9 +713,24 @@ const DatasetComponent = (props, reactContext) => {
             <div className="result-item">
                 <div className="result-item__data">
                     <a href={result['@id']} className="result-item__link">
-                        {datasetTypes[result['@type'][0]]}
+                        {fccSeries ? '' : <span>{`${datasetTypes[result['@type'][0]]}`}</span>}
                         {seriesDataset ?
                             <span>
+                                {(perturbationType && perturbationType.length > 0 && fccSeries) ?
```
forresttanaka: Generally you should check the easy cases first, like if `fccSeries` is true, before more complicated things like the length of `perturbationType`. That way if `fccSeries` is false, JS just moves on without checking the length of `perturbationType`. In this case, it has nearly no performance penalty.

forresttanaka (dismissed):
I had a couple comments that don’t need a change — just for future reference. We really need to clean `DatasetComponent` up, but that’s not for this ticket, and not sure it would pass Ben’s criteria for a worthwhile ticket.

src/encoded/static/components/search.js
khzlin: ok, fixed

forresttanaka (approved):
😎

emro (approved):

PR #3911, opened 2021-05-18
ENCD-5614-cart-dataset-filters

forresttanaka:
* Instead of tracking only the @ids of datasets, and retrieving a page of dataset objects every time the user wants to view a new page of datasets, I now load and keep all datasets (just a subset of their properties) just as file objects have worked. So functions like `CartSearchResults` no longer have to fetch dataset objects.
* We now load dataset descriptions on the cart-view page so that the selected annotations appear with full descriptions, just as they do on the annotation search page.
* `FileFacets` no longer resets facets when switching between showing default files only and all files, so we no longer need the useUpdate callback.
* I have a lot of duplication between file-facet code and dataset-facet code. I looked into trying to consolidate some of that, but it doesn’t look straightforward. Maybe someday.
* All facet-specific code has moved out of cart.js to a new facet.js file.
* When retrieving datasets, I now use a 100-dataset chunk size instead of 500 so that the progress bar makes more sense, at a small cost of overall performance.

### DIFF
--- a/src/encoded/static/components/cart/batch_download.js
+++ b/src/encoded/static/components/cart/batch_download.js
@@ -26,8 +26,8 @@ const batchDownload = (
     cartType,
     elements,
     analyses,
-    selectedTerms,
-    selectedType,
+    selectedFileTerms,
+    selectedDatasetTerms,
     facetFields,
     savedCartObj,
     sharedCart,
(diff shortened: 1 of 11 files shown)


src/encoded/static/components/cart/batch_download.js
```diff
@@ -48,27 +48,33 @@ const batchDownload = (
...
-                    const matchingFacetField = facetFields.find((facetField) => facetField.field === field);
-                    if (matchingFacetField && matchingFacetField.fieldMapper) {
-                        mappedQuery = matchingFacetField.fieldMapper(selectedTerms[field], analyses);
-                    } else {
-                        // Build the query string from `files` properties in the dataset, or from the
-                        // dataset properties itself for fields marked in `facets`.
-                        subQueryString = selectedTerms[field].map((term) => (
-                            `${datasetFacets.includes(field) ? '' : 'files.'}${field}=${encoding.encodedURIComponent(term)}`
-                        )).join('&');
+            _.compact(Object.keys(combinedTerms).map((field) => {
```
zoldello, line 56: _.compact is perfectly fine to use; but some may argue it is best to use third-party libraries only if inbuilt functionality does not exist as inbuilt ones usually are faster. So Javascript .filter would work as well.

src/encoded/static/components/cart/batch_download.js
```diff
@@ -77,7 +83,9 @@ const batchDownload = (
     const visualizableOption = `${visualizable ? '&option=visualizable' : ''}`;
     const rawOption = `${raw ? '&option=raw' : ''}`;
     const preferredDefaultQuery = preferredDefault && !raw && !all ? '&files.preferred_default=true' : '';
-    fetch(`/batch_download/?type=${selectedType}${cartId ? `&cart=${encoding.encodedURIComponent(cartId)}` : ''}${fileFormatSelections.length > 0 ? `&${fileFormatSelections.join('&')}` : ''}${visualizableOption}${rawOption}${preferredDefaultQuery}`, {
+    const query = `${formatSelections.length > 0 ? `&${formatSelections.join('&')}` : ''}`;
```
zoldello, line 86: [].join implicitly handles cases where the array is empty or has only one entry.

src/encoded/static/components/cart/cart.js
```diff
@@ -52,216 +60,72 @@ const PAGE_FILE_COUNT = 25;
...
 
-    render() {
-        if (this.state.elementsForDisplay && this.state.elementsForDisplay.length === 0) {
-            return <div className="nav result-table cart__empty-message">No visible datasets on this page.</div>;
-        }
-        return <ResultTableList results={this.state.elementsForDisplay || []} cartControls={this.props.cartControls} mode="cart-view" />;
-    }
-}
+    // Page is currently loading, so don't display anything for now.
+    return <div className="nav result-table cart__empty-message">Page currently loading&hellip;</div>;
```
zoldello, line 117: A spinner..?

src/encoded/static/components/cart/cart.js
```diff
@@ -981,13 +338,9 @@ const CartDatasetReport = ({ savedCartObj, sharedCartObj, usedDatasetTypes, cart
                 css="cart-dataset-report"
             >
                 {usedDatasetTypes.map((type) => (
```
zoldello, line 340: Good idea to not wrap a hyperlink around a parent element like a span?

src/encoded/static/components/cart/cart.js
```diff
@@ -1570,7 +749,7 @@ const processFilesAnalyses = (files, analyses) => {
 
 
 /**
- * Use the following order when sortnig files by status for evaluating pseudo-default files.
+ * Use the following order when sorting files by status for evaluating pseudo-default files.
```
zoldello, line 754: pseudo-default?

src/encoded/static/components/cart/cart.js
```diff
@@ -1803,92 +982,137 @@ const CartComponent = ({ context, savedCartObj, inProgress, fetch, session }) =>
...
-                selectedDatasetType === 'all'
-                    ? cartDatasets
-                    : cartDatasets.filter((datasetAtId) => atIdToType(datasetAtId) === selectedDatasetType)
-            );
-        }
-        return [];
-    }, [selectedDatasetType, cartDatasets, cartType]);
+    // Determine if the cart contents have changed.
+    const { cartType, cartName, cartDatasets } = getCartInfo(context, savedCartObj);
+    const isCartDatasetsChanged = !_.isEqual(cartDatasetsRef.current, cartDatasets);
```
zoldello, line 993: This is okay; but it is usually better to compare by an id rather than a deep equality

src/encoded/static/components/cart/facet.js
```diff
@@ -0,0 +1,1418 @@
+import React from 'react';
```
zoldello, line 1: cut and paste dump to refactor out facet... Now the previous mass code deletion makes sense.

zoldello (dismissed):

forresttanaka:
@zoldello I did one small change to disable a button if you’re not looking at an active cart.

zoldello (approved):

PR #3914, opened 2021-05-19
ENCD-5903-audit-mixed-expression-measurement-methods

iyoungworth:


src/encoded/audit/functional_characterization_experiment.py
```diff
@@ -126,6 +126,29 @@ def audit_experiment_replicate_with_no_files(value, system, excluded_statuses):
...
+            methods.append('none')
+        else:
+            methods.append(locus.get('expression_measurement_method'))
+    if len(set(methods)) == 1:
+        return
+    else:
+        methods_detail = ', '.join(methods)
+        detail = f'This experiment uses multiple expression measurement methods, {methods_detail}.'
+        yield AuditFailure(
+            'mixed expression_measurement_method', detail, level='NOT_COMPLIANT'
```
gabdank: I think it should be WARNING

gabdank (changes requested):

src/encoded/audit/functional_characterization_experiment.py
iyoungworth: Got it, adjusted

gabdank (approved):

PR #3917, opened 2021-05-20
ENCD-5757-multisearch-within-collections

keenangraham:
Think this more or less final structure but would still like to add some tests and comments. 

This avoids any changes to current code and adds a new `MultiSearch` component that renders a list of queries to sections in the search-as-you-type dropdown. Should allow for future additions of independent sections.

A little bit of duplication in new `MultiSearch` but maybe okay since it allows it to be independent from normal `Search`.

### DIFF
--- a/src/encoded/static/components/top_hits/constants.js
+++ b/src/encoded/static/components/top_hits/constants.js
@@ -59,33 +59,33 @@ export const TERM_NAME = 'term_name';
 
 export const SCIENTIFIC_NAME = 'scientific_name';
 
-export const COLLECTION_ENDPOINT_URLS = [
+export const COLLECTIONS = [
     {
         title: 'ENTEx',
         details: 'Epigenomes from four individuals',
-        searchURL: '/search/?type=Experiment&status=released&internal_tags=ENTEx&limit=0',
-        linkURL: '/entex-matrix/?type=Experiment&status=released&internal_tags=ENTEx'
+        searchUrl: '/search/?type=Experiment&status=released&internal_tags=ENTEx&limit=0',
+        '@id': '/entex-matrix/?type=Experiment&status=released&internal_tags=ENTEx'
     },
     {
         title: 'SESCC',
         details: 'Stem Cell Development Matrix',
-        searchURL: '/search/?type=Experiment&internal_tags=SESCC&limit=0',
-        linkURL: '/sescc-stem-cell-matrix/?type=Experiment&internal_tags=SESCC',
+        searchUrl: '/search/?type=Experiment&internal_tags=SESCC&limit=0',
+        '@id': '/sescc-stem-cell-matrix/?type=Experiment&internal_tags=SESCC',
     },
     {
         title: 'ENCORE',
         details: 'RNA-protein interactions',
-        searchURL: '/search/?type=Experiment&status=released&internal_tags=ENCORE&limit=0',
-        linkURL: '/encore-matrix/?type=Experiment&status=released&internal_tags=ENCORE',
+        searchUrl: '/search/?type=Experiment&status=released&internal_tags=ENCORE&limit=0',
+        '@id': '/encore-matrix/?type=Experiment&status=released&internal_tags=ENCORE',
     },
     {
         title: 'Human reference epigenomes',
         details: null,
-        searchURL: (
+        searchUrl: (
             '/search/?type=Experiment&related_series.@type=ReferenceEpigenome' +
             '&replicates.library.biosample.donor.organism.scientific_name=Homo+sapiens'
         )
-        linkURL: (
+        '@id': (
             '/reference-epigenome-matrix/?type=Experiment&related_series.@type=ReferenceEpigenome' +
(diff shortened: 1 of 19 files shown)


src/encoded/static/components/top_hits/multi/constants.js
```diff
@@ -0,0 +1,14 @@
+import TopHitsQuery, {
+    CollectionsQuery,
```
forresttanaka, line 2: I’m not a big fan of renaming default exports, as I think it makes examining code more confusing and difficult to trace (what is this function _really_ called?), along with VSCode not being able to find the actual definition when you right click on the call to the function, class, etc. This doesn’t go against AirBnB rules, but I’m not sure why this improves anything.

src/encoded/static/components/top_hits/query.js
```diff
@@ -95,4 +97,76 @@ class Query {
...
+    fetch(
+        makeCollectionUrl(collection.searchUrl, searchTerm),
+        {
+            method: 'GET',
+            headers: {
+                'Content-Type': 'application/json',
+            },
+        }
+    ).then(
+        (response) => (response.ok ? response.json() : { total: 0 })
```
forresttanaka, line 146: Good way to handle the no-results case.

src/encoded/static/scss/encoded/modules/_top_hits_search.scss
```diff
@@ -95,13 +95,15 @@
             &.section-item {
                 padding-top: 3px;
                 padding-bottom: 3px;
+                overflow-x: scroll;
```
forresttanaka, line 98: These can scroll horizontally? I couldn't come up with a search result to see that.

forresttanaka (dismissed):
Seems pretty clean, if a bit complex. Would have been nice to see a description for each function and property. Maintaining this code would seem needlessly difficult. And it’s nice to see something as clear as this on hover:

Screen Shot 2021-05-20 at 08 59 30

…instead of:

Screen Shot 2021-05-20 at 08 32 26

Why the empty index.js under multi?

src/encoded/static/scss/encoded/modules/_top_hits_search.scss
keenangraham: Screen Shot 2021-05-20 at 10 04 32 AM

Think it comes up if there is a long filename displayed as block.

src/encoded/static/components/top_hits/multi/constants.js
keenangraham: I was hoping it would make it more apparent which query is which in this list:

Screen Shot 2021-05-20 at 10 12 21 AM


I didn't want to change the actual name since other things still rely on it outside of the multisearch context.

keenangraham:
> Would have been nice to see a description for each function and property.

Still plan to add.

> Why the empty index.js under multi?

I thought maybe it was like `__init__.py` in Python modules. Will try to remove.

forresttanaka:
> I thought maybe it was like __init__.py in Python modules. Will try to remove

Oh, I see; it’s _sort_ of like that. index.js is handy so you can import the directory name from other modules, but it has to export stuff for others to use. At least, I always thought so; maybe it has a capability I haven’t been aware of.

keenangraham:
@forresttanaka Added comments and tests.

src/encoded/static/components/top_hits/multi/constants.js
forresttanaka: I was thinking if you changed the name within '../query' to `TopHitsResults` that’s still accurate regarldess of who relies on it; it’s just a more specific name than it used to be, but no less accurate. I’m not sure I’m covering all cases of this though, but that was my impression.

src/encoded/static/components/top_hits/constants.js
```diff
@@ -58,3 +58,53 @@ export const DONOR = 'Donor';
 export const TERM_NAME = 'term_name';
 
 export const SCIENTIFIC_NAME = 'scientific_name';
+
+// Specifies all of the details needed for searchinig over data collections with a
```
forresttanaka: “searching”

forresttanaka (dismissed):
This is way easier to understand. I didn't realize you had planned to add comments, so sorry about lecturing.

src/encoded/static/components/top_hits/multi/constants.js
keenangraham: >I was thinking if you changed the name within '../query' to TopHitsResults

Ah, yeah that makes sense. This also brings up other issue of aliasing the top hits results component and the data collection results component in same way.

I think maybe the problem in general is that the organizational structure is not quite right:

Screen Shot 2021-05-20 at 2 18 48 PM

Originally it made sense that everything was in `top_hits` folder, and the `query` was simply the top hits query. But with new multisearch the top hits is just one possible section.

I've stuck most of the functionality in the `multi` folder mainly for lack of anywhere else to put it. I agree this is not ideal.

forresttanaka (approved):
Thought I _had_ approved it. Trying again…

PR #3920, opened 2021-05-21
ENCD-5945-index-annotations-logging

ottojolanki:
Increases initial timeout to 45s (was 30s), which seems to be enough. Logs possible exception and retries with 10s more.

### DIFF
--- a/src/encoded/commands/index_annotations.py
+++ b/src/encoded/commands/index_annotations.py
@@ -59,19 +59,17 @@ def run(app):
     # bulk index of annotations
     annotations = json_from_path(registry.settings.get('annotations_path'), {})
 
-    base_timeout = 45 
+    base_timeout = 45
     additional_timeout = 0
     annotation_indexing_success = False
 
     while not annotation_indexing_success:
         try:
-            print(f"Sending annotation indexing request with timeout {base_timeout + additional_timeout}")
             es.bulk(index=index, body=annotations, refresh=True, request_timeout=base_timeout+additional_timeout)
             annotation_indexing_success = True
         except:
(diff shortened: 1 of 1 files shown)


src/encoded/commands/index_annotations.py
```diff
@@ -58,10 +58,19 @@ def run(app):
...
+    additional_timeout = 0
+    annotation_indexing_success = False
+
+    while not annotation_indexing_success:
+        try:
+            es.bulk(index=index, body=annotations, refresh=True, request_timeout=base_timeout+additional_timeout)
+            annotation_indexing_success = True
+        except:
+            log.error("Unable to index the annotations.", exc_info=True)
+            additional_timeout += 10
```
paul-sud: Could eliminate the boolean, would be a little easier to read IMO:
```suggestion
    while True:
        try:
            es.bulk(index=index, body=annotations, refresh=True, request_timeout=base_timeout+additional_timeout)
            break
        except:
            log.error("Unable to index the annotations.", exc_info=True)
            additional_timeout += 10
```

paul-sud: This code could loop forever, say if ES never comes up. Probably want to limit the number of retries then exit when it is exceeded.

paul-sud (changes requested):

src/encoded/commands/index_annotations.py
ottojolanki: yeah. Probably if it won't succeed in, say 5 times it will never succeed.

paul-sud (dismissed):

hitz (approved):

PR #3925, opened 2021-05-25
ENCD-5957-wale-broken

ottojolanki:


cloud-config/run-scripts/app-pg.sh
```diff
@@ -159,7 +159,8 @@ if [ ! -f "$WALE_BIN/pip" ]; then
     touch "$encd_failed_flag"
     exit 1
 fi
-sudo -H -u postgres "$WALE_BIN/pip" install pip==20.2.4 setuptools --upgrade
+sudo -H -u postgres "$WALE_BIN/pip" install pip setuptools --upgrade
+sudo -H -u postgres "$WALE_BIN/pip" install wheel
```
paul-sud: I'm a little hesitant to remove the pins since it's already caused us problems in the past. Could you pin to the versions that you confirmed were working? If you think we shouldn't pin I'd be happy to hear your reasoning.

paul-sud (changes requested):

cloud-config/run-scripts/app-pg.sh
ottojolanki: yes, I agree. Will pin to the current working versions.

paul-sud (approved):

PR #3926, opened 2021-05-25
ENCD-5942-display-disease-name-on-series

zoldello:


src/encoded/static/components/dataset.js
```diff
@@ -1879,6 +1879,15 @@ export const SeriesComponent = (props, reactContext) => {
         combinedTreatmentAmounts = treatmentAmounts.join(', ');
     }
 
+    // As of code-time, there is a chance disease term namess across biosamples could be different. So
+    // the code takes this into account
+    const diseases = [
+        ...new Set(context.related_datasets.reduce((accumulatedDiseases, currentRelatedDataset) => {
+            const diseasesTermNames = currentRelatedDataset.replicates.map((replicate) => replicate.library.biosample.disease_term_name);
```
forresttanaka, line 1886: replicates don’t require a library, and libraries don’t require a biosample, so should you test for their existence here? I’m not sure for disease series objects if maybe these properties are basically guaranteed even if not required.

src/encoded/static/components/dataset.js
```diff
@@ -1939,6 +1948,12 @@ export const SeriesComponent = (props, reactContext) => {
                                 </div>
                             : null}
 
+
+                            <div data-test="diseasetermname">
+                                <dt>Disease{diseases && diseases.length !== 1 ? 's' : ''}</dt>
+                                <dd>{diseases && diseases.length > 0 ? diseases.join(', ') : 'Not reported'}</dd>
```
forresttanaka, line 1954: I think technically `diseases` exists no matter what happens in the loops above, though it could have a zero length, so you don’t have to test for its existence here.

forresttanaka (approved):

PR #3939, opened 2021-06-01
ENCD-5690-add-azure-uri-to-file-metadata

keenangraham:
This relies on `s3_uri` to avoid redundant calculations and calls to external sheet.

### DIFF
--- a/src/encoded/tests/test_types_file.py
+++ b/src/encoded/tests/test_types_file.py
@@ -376,3 +376,108 @@ def test_file_embedded_annotation_properties(testapp, file_ccre,
     res = testapp.get(file_ccre['@id'] + '@@index-data')
     assert res.json['object']['annotation_subtype'] == 'CTCF-only'
     assert res.json['object']['biochemical_inputs'] == ['cDHS', 'rDHS']
+
+
+def test_file_show_azure_uri():
+    from encoded.types.file import show_azure_uri
+    for status in File.public_s3_statuses:
+        assert show_azure_uri(
+            's3://encode-public/2020/09/26/71e9637f-bbcb-4e4b-9d92-d3acdda4bb7b/ENCFF841DAP.bigWig',
+            status,
+        )
+        assert not show_azure_uri(
+            None,
+            status,
+        )
+    for status in File.private_s3_statuses:
+        assert not show_azure_uri(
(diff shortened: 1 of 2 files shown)


src/encoded/types/file.py
```diff
@@ -92,6 +107,10 @@ def property_closure(request, propname, root_uuid):
 ]
 
 
+AZURE_URI_PREFIX = 'https://datasetencode.blob.core.windows.net/dataset'
+AZURE_PUBLIC_TOKEN = '?sv=2019-10-10&si=prod&sr=c&sig=9qSQZo4ggrCNpybBExU8SypuUZV33igI11xw0P7rB3c%3D'
```
hitz, line 111: this is generic for all files? weird

src/encoded/tests/test_types_file.py
```diff
@@ -376,3 +376,108 @@ def test_file_embedded_annotation_properties(testapp, file_ccre,
...
+        get_key_from_s3_uri(s3_uri) == '2020/09/26/71e9637f-bbcb-4e4b-9d92-d3acdda4bb7b/ENCFF841DAP.bigWig'
+    )
+
+
+def test_convert_s3_uri_to_azure_uri():
+    from encoded.types.file import convert_s3_uri_to_azure_uri
+    s3_uri = 's3://encode-public/2020/09/26/71e9637f-bbcb-4e4b-9d92-d3acdda4bb7b/ENCFF841DAP.bigWig'
+    azure_uri = convert_s3_uri_to_azure_uri(s3_uri)
+    assert azure_uri == (
+        'https://datasetencode.blob.core.windows.net/dataset/'
```
hitz, line 416: these links don't seem to actually work....

hitz (approved):
The code looks fine but did the URLs actually get tested?  I'm wondering if we need new tokens or they are otherwise blocked.

src/encoded/tests/test_types_file.py
keenangraham: Sounds like bucket is synced first week of every month so only works on files released before last sync:
https://azure.microsoft.com/en-us/services/open-datasets/catalog/genomics-encode/

I thought about adding some condition to show URI based on date released being less than current month, and we still can, but wasn't sure it was worth the additional indexing time. Maybe it's too confusing to show links that are only eventually consistent though.

keenangraham: I guess it would help if files had `date_released` property.

PR #3941, opened 2021-06-03
ENCD-5776-new-batch-download-ui

forresttanaka:
* I added a “batch_download” directory containing the new module in its entirety. Within that, a “docs” directory contains an overview that describes how to use and customize the module.
* I think the `filters` state in `FileGalleryRendererComponent` can no longer have any values for `assembly` than “All assemblies,” so I removed the download code to process assemblies.


forresttanaka:
Preview demo: ~https://encd-5776-batch-dl-options-pv1-fytanaka.demo.encodedcc.org/~

forresttanaka:
Taking this out of code review momentarily. I thought I had fixed something, but I hadn't.

forresttanaka:
I changed my mind — I think it’s fine the way it is. I’m writing up QA notes now and will have Meenakshi give feedback on whether any of the options aren’t appropriate.

src/encoded/static/components/search.js
```diff
@@ -1274,7 +1263,9 @@ export const SearchControls = ({ context, visualizeDisabledTitle, showResultsTog
             <div className="results-table-control__main">
                 <ViewControls results={context} additionalFilters={additionalFilters} />
                 {resultsToggle}
-                {showDownloadButton ? <BatchDownloadControls results={context} additionalFilters={additionalFilters} modalText={modalText} canDownload={canDownload} /> : ''}
+                {showDownloadButton
+                    ? <BatchDownloadControls results={context} additionalFilters={additionalFilters} modalText={modalText} canDownload={canDownload} />
```
emro, line 1267: I always put the ? on the top line, I'm not sure why.

src/encoded/static/components/search.js
```diff
@@ -1215,29 +1215,18 @@ export const SearchControls = ({ context, visualizeDisabledTitle, showResultsTog
...
+        : (
+            <>
+                <p>
+                    This search is too large (&gt;{MAX_DOWNLOADABLE_RESULT} datasets) to automatically generate a manifest or metadata file.
+                </p>
+                <p>
+                    You can directly access the files in AWS: <a href="https://registry.opendata.aws/encode-project/" target="_blank" rel="noopener noreferrer">https://registry.opendata.aws/encode-project/</a>
+                </p>
+            </>
+        );
```
emro, line 1229: I usually write these in the other order, I'd probably say `!canDownload ? () : null;`, but I don't think this matters and it might make more sense the way you wrote it.

emro (dismissed):
There are a lot of changes here! The code and the demo both look good.

I really appreciate the detail descriptions and examples about how to use the new interface, those will make it much easier when I'm trying to figure out how to update the Encyclopedia page to use this interface.

src/encoded/static/components/search.js
forresttanaka: I just do this because the AirBnB examples do it this way. I don’t feel strongly about it, and without guidance I never really decided how I like to do it.

src/encoded/static/components/search.js
forresttanaka: I’m not very consistent on this kind of thing. Sometimes I prefer not to test the negative of something; other times I prefer the code that does something first and nothing second.

emro (dismissed):

emro:
Code looks good but it looks like there are some conflicts that need to be resolved.

forresttanaka:
Self-inflicted merge conflicts resolved.

emro (approved):
Looks good!

PR #3943, opened 2021-06-03
ENCD-5962-avoid-results-from-stale-request

keenangraham:


src/encoded/static/components/top_hits/multi/search.js
```diff
@@ -13,7 +13,11 @@ import QUERIES from './constants';
 * dropdown using the associated custom results component.
 */
 const Search = ({ children }) => {
+    // User input.
     const [input, setInput] = useState('');
+    // Reference to latest input for use in callback.
+    const inputRef = useRef();
+    inputRef.current = input;
```
forresttanaka: I didn’t think about this before, but I think these two lines can be combined into:
```
const inputRef = useRef(input);
```
I’m about 80% sure that’s equivalent to what you have.

src/encoded/static/components/top_hits/multi/search.js
```diff
@@ -22,6 +26,18 @@ const Search = ({ children }) => {
     // Wait this long after last user input making queries.
     const debounceTime = 200;
 
+    // Comparse searchTerm from query with latest user input.
```
forresttanaka: “Compare”

src/encoded/static/components/top_hits/__tests__/top-hits-search-test.js
```diff
@@ -1,23 +1,369 @@
...
+            },
+        },
+    },
+    num_reduce_phases: 2,
+    timed_out: false,
+    took: 58,
+};
+
+
+// Search for H taks 1000ms to return, search for HBB takes 500ms.
```
forresttanaka: “takes”

forresttanaka (dismissed):
Thanks for working out the complex Jest tests. I don’t think we’ve ever had to work something like that out before. Just a couple feedback items that make (probably) no technical difference.

src/encoded/static/components/top_hits/multi/search.js
keenangraham: Hmm, this doesn't seem to work (tests break). I wonder if I should wrap the assignment in useEffect?
```javascript
useEffect(() => {
        inputRef.current = input;
    }, [input]);
```

forresttanaka (approved):

PR #3949, opened 2021-06-04
ENCD-5965-enable-association-graph-in-series

zoldello:


src/encoded/static/components/dataset.js
```diff
@@ -1971,6 +1971,8 @@ export const SeriesComponent = (props, reactContext) => {
         }, [])),
     ].filter((disease) => disease);
 
+    const experimentType = context['@type'][0];
```
forresttanaka: The `seriesType` variable above already has this.

src/encoded/static/components/filegallery.js
```diff
@@ -118,6 +118,15 @@ const filterAssembly = (analysesGroup) => {
...
 };
 
+/**
+ * Get the analyses' files from context
+ *
+ * #param {object} context
+ *
+ * The result can be either a list of file ids of a list of files, depending on the file object
+*/
+const getAnalysesFilesFromContext = (context) => [...new Set([].concat(...(context.analyses || []).map((analysis) => analysis.files)))];
```
forresttanaka, line 128: Why concat from an empty array? Doesn’t `context.analyses || []` supply its own starting array?

src/encoded/static/components/dataset.js
```diff
@@ -2134,7 +2137,7 @@ export const SeriesComponent = (props, reactContext) => {
                 showReplicateNumber={false}
                 hideControls={!METADATA_SERIES_TYPES.includes(context['@type'][0])}
                 collapseNone
-                hideGraph
+                hideGraph={experimentType !== 'FunctionalCharacterizationSeries'}
```
forresttanaka: FYI In ENCD-5971 I’m working on eliminating type checking in `SeriesComponent`.

src/encoded/static/components/filegallery.js
```diff
@@ -479,6 +497,54 @@ export class FileTable extends React.Component {
                                 adminUser,
                             }}
                         />
+                        {analysesSeries.map((key) => {
+                            const seriefiles = files.series && files.series.length > 0 ? files.series.filter((s) => s.analyses[0] === key) : [];
```
forresttanaka: If `files.series` doesn’t exist, then `analysesSeries` never enters this loop. Also I think if `analysesSeries` has any entries, then `files.series` must have a length > 0, though I’m not 100% sure on that one.

src/encoded/static/components/filegallery.js
```diff
@@ -419,6 +434,9 @@ export class FileTable extends React.Component {
                 return analysisObjectsAccession || nonAnalysisObjectPrefix;
             });
 
+            // Get unique analyses for series object
+            const analysesSeries = files.series ? [...new Set(files.series.map((a) => (a.analyses && a.analyses.length > 0 ? a.analyses[0] : '')).filter((a) => a !== ''))] : [];
```
forresttanaka, line 466: Is there a way to add these to `analysisObjectKeys` to avoid the duplicate `<SortTable>` below?

src/encoded/static/components/filegallery.js
```diff
@@ -2481,6 +2551,8 @@ FileGraph.propTypes = {
     handleNodeClick: PropTypes.func.isRequired, // Parent function to call when a graph node is clicked
     colorize: PropTypes.bool, // True to enable node colorization based on status
     loggedIn: PropTypes.bool, // True if current user has logged in
+    nonSeriesFiles: PropTypes.array, // Series  files
```
forresttanaka: `nonSeriesFiles` are Series files?

src/encoded/static/components/filegallery.js
```diff
@@ -2916,6 +3002,10 @@ class FileGalleryRendererComponent extends React.Component {
             compiledAnalyses: compileAnalyses(props.analyses, datasetFiles, 'choose analysis'),
             /** Index of currently/last selected `compiledAnalyses`. */
             selectedAnalysesIndex: 0,
+            /** True if object is of type Series */
+            isSeriesType,
+            /** Series Files */
+            seriesFiles,
```
forresttanaka: I don’t think `isSeriesType` and `seriesFiles` shouldn’t exist in state because state shouldn’t serve as a cache. I do realize why you do this, but I hope we have a simpler solution. One is I think I made a mistake in `FileGallery` and shouldn’t have passed `data` in as its own state, but should pass it in directly, that way `data` in this constructor gets the series files passed in instead of extracting them from analyses (you can still use context.analyses for further filtering).

src/encoded/static/components/genome_browser.js
```diff
@@ -879,7 +879,7 @@ class GenomeBrowser extends React.Component {
             // Here we do some approximate math to try to figure out how many lines the labels extend to assuming that ~30 characters fit on one line
             // Labels on the experiment pages are short enough to fit on one line (they contain less information) so we can bypass these calculations for those pages
             if (label === 'cart') {
-                labelLength += file.target ? file.target.label.length + 2 : 0;
+                labelLength += file.target && file.target.label && file.target.label ? file.target.label.length + 2 : 0;
```
forresttanaka: Repeated `file.target.label`?

forresttanaka (changes requested):
I’d like to see if you can vastly simplify this, as I think this has become more complicated than needed. My comment on the `isSeriesType` and `seriesFiles` states expresses some of that. I think correcting my mistake in `FileGallery` solves a lot of this, the goal (if possible) being that you don’t have to alter the incoming files and keep track of series and non-series files — just deal with incoming files in the `data` prop.

src/encoded/static/components/dataset.js
zoldello: @forresttanaka The issue is that seriesType changes and I need something that is constant. Is my reason acceptable to you?

src/encoded/static/components/dataset.js
zoldello: Thanks

src/encoded/static/components/filegallery.js
zoldello: Thanks. That is how it is done in Python and I did not context-switch :)

src/encoded/static/components/genome_browser.js
zoldello: Corrected. Thanks.

src/encoded/static/components/filegallery.js
zoldello: Corrected. Thanks.

src/encoded/static/components/filegallery.js
zoldello: Done. Let me know if my way is unacceptably hacky.

src/encoded/static/components/filegallery.js
zoldello: Done. How okay are you with going a little overboard with defensive programming? I worry about edge cases so I tend to assume the data I have could be invalid even if it is data from the method in question itself.

src/encoded/static/components/filegallery.js
zoldello: I removed these.

src/encoded/static/components/dataset.js
forresttanaka: I’m working on a branch where `seriesType` doesn’t change. It doesn’t change in the data so it should never have changed in the code.

zoldello: corrected.  removed ExperimentType

src/encoded/static/components/genome_browser.js
zoldello: @forresttanaka I am not sure if I satisfied your comment "just deal with incoming files in the data prop". It is tricky. In series, the association graph is drawn with series files. But non-series files are used in other types. But in series files I also need non-series files to satisfy a requirement to have accession ids draw on the nodes.  If I combine it with the data I get into an issue we both wrestled with where you cannot alter props (to remove non-series files) or else react explodes. That is why I hack-ly added nonseries argument that is only used by series files in assembleGraph().

src/encoded/static/components/filegallery.js
```diff
@@ -145,6 +154,33 @@ const sortAnalysisByVersion = (analysesGroup) => {
     return sortedAnalysesAwarded;
 };
 
+/**
+ * True is context is series type
```
forresttanaka: "if"

forresttanaka (dismissed):

forresttanaka (approved):

PR #3950, opened 2021-06-04
ENCD-5980-fix-download-button

zoldello:


forresttanaka (approved):
Glad there was a reasonable solution to this.

PR #3951, opened 2021-06-04
ENCD-5929-analysis-audits-reappear-in-file-details-table

forresttanaka:
* I now request all analyses in a dataset from the server so that I can get their audit data. That's what now appears next to the analysis names in the file details table titles.
* I modified the analysis test data to set one up to have a non-default analysis with audits.

### DIFF
--- a/src/encoded/static/components/filegallery.js
+++ b/src/encoded/static/components/filegallery.js
@@ -6,7 +6,7 @@ import utc from 'dayjs/plugin/utc';
 import { Panel, PanelHeading, TabPanel, TabPanelPane } from '../libs/ui/panel';
 import { Modal, ModalHeader, ModalBody, ModalFooter } from '../libs/ui/modal';
 import { collapseIcon } from '../libs/svg-icons';
-import { auditDecor, auditsDisplayed, filterAuditByPath } from './audit';
+import { auditDecor, auditsDisplayed } from './audit';
 import GenomeBrowser from './genome_browser';
 import * as globals from './globals';
 import { Graph, JsonGraph, GraphException } from './graph';
@@ -18,6 +18,7 @@ import {
     useMount,
     requestSearch,
     requestUri,
+    requestObjects,
 } from './objectutils';
 import { qcIdToDisplay } from './quality_metric';
 import { softwareVersionList } from './software';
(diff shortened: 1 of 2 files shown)


src/encoded/static/components/filegallery.js
```diff
@@ -1384,11 +1392,29 @@ export const FileGallery = ({
         }
     }, [fileQuery, files]);
 
+    useMount(() => {
+        // Request the audits for all given analyses.
+        if (usedAnalyses && usedAnalyses.length > 0) {
```
zoldello, line 1397: August will be great because then we can use Optional chaining to stop having to do this; not be stuck in having to appease IE11.

src/encoded/static/components/filegallery.js
```diff
@@ -1363,6 +1367,10 @@ export const FileGallery = ({
     const [schemas, setSchemas] = React.useState(null);
     // Holds file search results from a request.
     const [data, setData] = React.useState([]);
+    // Holds analysis audit objects loaded by server request, keyed by analysis @id.
+    const [analysisAudits, setAnalysisAudits] = React.useState({});
+
+    const usedAnalyses = analyses || context.analyses;
```
zoldello, line 1373: In August we can then use Nullish coalescing operator and life will be better.

src/encoded/static/components/filegallery.js
```diff
@@ -3769,8 +3799,8 @@ const CollapsingTitleComponent = ({
 
     const canDownload = downloadEnabled(files);
 
-    // Filter to audits relevant to the current analysis.
-    const filteredAudit = analysis ? filterAuditByPath(audit, analysis['@id']) : {};
+    // Get the audit object relevant to the current analysis
+    const analysisAudit = analysis && analysisAudits ? analysisAudits[analysis['@id']] : null;
```
zoldello, line 3803: Line 3803 means if **analysis** is _not falsey_ use the result o **analysisAudits ? analysisAudits[analysis['@id']] : null;** else, **analysisAudit** is set to a falsey value. It is the opposite of use **||** on non-booleans outside of if(). That may be confusing to anyone not aware of that Javascript quirk and may think **analysisAudit** would be a boolean.

zoldello (approved):
Code looks good. My comments are not important.

PR #3955, opened 2021-06-08
ENCD-5955-file-info-in-analysis-titles

forresttanaka:
Adds a two-row title to the analysis subsections in file details tables.


forresttanaka:
Found a React warning I need to fix.

PR #3958, opened 2021-06-09
ENCD-5926-adjust-biosample-characterizations-audit-to-check-parents

iyoungworth:
continuing-work

continuing-work

branch-switch

last-tests

### DIFF
--- a/src/encoded/audit/experiment.py
+++ b/src/encoded/audit/experiment.py
@@ -1754,13 +1754,15 @@ def audit_experiment_biosample_characterization(value, system, excluded_types):
             child_mods = sorted(mods)
             child_treatments = sorted(biosample.get('treatments', []))
             relationship = ''
+            part_of_biosample = biosample.get('part_of', [])
+            originated_from_biosample = biosample.get('originated_from', [])
             if (
-                biosample.get('part_of')
-                and biosample.get('part_of').get('characterizations')
+                part_of_biosample
+                and part_of_biosample.get('characterizations')
             ):
-                part_ontology = biosample.get('part_of').get('biosample_ontology')
-                part_mods = sorted(biosample.get('part_of').get('applied_modifications', []))
-                part_treatments = sorted(biosample.get('part_of').get('treatments', []))
+                part_ontology = part_of_biosample.get('biosample_ontology')
+                part_mods = sorted(part_of_biosample.get('applied_modifications', []))
+                part_treatments = sorted(part_of_biosample.get('treatments', []))
                 if (
                     part_ontology == ontology
                     and part_mods == child_mods
@@ -1768,12 +1770,12 @@ def audit_experiment_biosample_characterization(value, system, excluded_types):
                 ):
                     relationship = 'part_of'
             elif (
-                biosample.get('originated_from')
-                and biosample.get('originated_from').get('characterizations')
+                originated_from_biosample
+                and originated_from_biosample.get('characterizations')
             ):
-                orig_ontology = biosample.get('originated_from').get('biosample_ontology')
-                orig_mods = sorted(biosample.get('originated_from').get('applied_modifications', []))
-                orig_treatments = sorted(biosample.get('originated_from').get('treatments', []))
+                orig_ontology = originated_from_biosample.get('biosample_ontology')
+                orig_mods = sorted(originated_from_biosample.get('applied_modifications', []))
+                orig_treatments = sorted(originated_from_biosample.get('treatments', []))
                 if (
                     orig_ontology == ontology
                     and orig_mods == child_mods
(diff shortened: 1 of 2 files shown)


src/encoded/audit/experiment.py
```diff
@@ -1746,6 +1748,63 @@ def audit_experiment_biosample_characterization(value, system, excluded_types):
...
+                    and part_treatments == child_treatments
+                ):
+                    relationship = 'part_of'
+            elif (
+                biosample.get('originated_from')
+                and biosample.get('originated_from').get('characterizations')
+            ):
+                orig_ontology = biosample.get('originated_from').get('biosample_ontology')
+                orig_mods = sorted(biosample.get('originated_from').get('applied_modifications', []))
+                orig_treatments = sorted(biosample.get('originated_from').get('treatments', []))
```
gabdank: I think this code could be rewritten to be something like:
part_of_biosample = biosample.get('part_of')
if (part_of_biosample and par_of_biosample.get('characterizations'):
    part_of_biosample.get('treatments')....
    ....
originated_from_biosample = biosample.get('originated_from')
if (originated_from_biosample and originated_From_biosample.get'characterizations'):
     originated_from_biosample.get('treatments').....
    ....

this way you are not "getting" the parent biosamples only once

src/encoded/audit/experiment.py
```diff
@@ -3730,6 +3789,7 @@ def check_award_condition(experiment, awards):
         'replicates.library.biosample.applied_modifications',
         'replicates.library.biosample.applied_modifications.modified_site_by_target_id',
         'replicates.library.biosample.characterizations',
+        'replicates.library.biosample.part_of.characterizations',
```
gabdank, line 3795: don't you need also "replicats.library.biosample.originated_from.characterizations' ?

src/encoded/tests/test_audit_experiment.py
```diff
@@ -2194,6 +2194,139 @@ def test_audit_experiment_pooled_biosample_characterization(
...
+                       {'genetic_modifications': [interference_genetic_modification['@id']],
+                        'biosample_ontology': k562['uuid'],
+                        'donor': donor_1['@id'],
+                        'part_of': base_biosample['@id']})
+    testapp.patch_json(biosample_2['@id'],
+                       {'genetic_modifications': [interference_genetic_modification['@id']],
+                        'biosample_ontology': k562['uuid'],
+                        'donor': donor_1['@id'],
+                        'part_of': base_biosample['@id']})
+    testapp.patch_json(base_biosample['@id'],
```
gabdank: I guess same test needed for originated_from?

gabdank (dismissed):

src/encoded/audit/experiment.py
iyoungworth: Thank you for pointing this out, I've added that line

src/encoded/tests/test_audit_experiment.py
iyoungworth: I trimmed down to the most useful tests and parametrized so they'd be run for both part_of and originated_from

src/encoded/audit/experiment.py
iyoungworth: Thank you for the suggestion, I applied this

src/encoded/audit/experiment.py
```diff
@@ -1730,6 +1731,7 @@ def audit_experiment_biosample_characterization(value, system, excluded_types):
         ):
             continue
         biosample = rep['library']['biosample']
+        ontology = biosample['biosample_ontology']['@id']
```
hitz, line 1734: this is actually a little dangerous unless you want a 500 error if the keys don't exist.  I think they are strictly require here so it's fine.

hitz: could wrap that in a try... except KeyError.

src/encoded/audit/experiment.py
```diff
@@ -1746,6 +1748,65 @@ def audit_experiment_biosample_characterization(value, system, excluded_types):
...
+                if (
+                    orig_ontology == ontology
+                    and orig_mods == child_mods
+                    and orig_treatments == child_treatments
+                ):
+                    relationship = 'originated_from'
+            if relationship != '':
+                parent_characterization_status = {
+                    c.get('review', {}).get('status')
+                    for c in biosample.get(relationship).get('characterizations')
```
hitz, line 1788: this is going to fail if biosample['relationship']['characterizations'] is not an iterable.  Maybe you secured that already?

hitz (commented):
I am not sure if I need changes here or not @iyoungworth will slack you.

hitz (approved):

PR #3962, opened 2021-06-10
ENCD-5955-analysis-titles-in-file-details-table-headers

forresttanaka:


src/encoded/static/scss/encoded/modules/_panels.scss
```diff
@@ -767,6 +769,25 @@ $file-gallery-controls-height: 34px; // Height of controls inside file-gallery-c
     color: #606060;
 }
 
+.title-main-content {
```
zoldello, line 772: I am surprised you did not use '&' to capture the "title-main-content" part of the name.

src/encoded/static/components/filegallery.js
```diff
@@ -3748,6 +3744,226 @@ FileGalleryRendererComponent.contextTypes = {
 const FileGalleryRenderer = auditDecor(FileGalleryRendererComponent);
 
 
+/**
+ * Maps each Series-object subtype to a function that generates a subtitle in the file details
+ * table section title bars.
+ */
```
zoldello, line 3831: can this be on top of the file?

src/encoded/static/components/filegallery.js
```diff
@@ -503,7 +503,6 @@ export class FileTable extends React.Component {
                                             && filterDownloadableFilesByStatus(context, files[key]).length > 0
                                         }
                                         inclusionOn={options.inclusionOn}
-                                        files={files[key]}
```
zoldello: Not sure why you removed this. I do not see functionality-replacement (for non-series objects).

src/encoded/static/components/filegallery.js
```diff
@@ -3748,6 +3744,226 @@ FileGalleryRendererComponent.contextTypes = {
...
+}) => {
+    // Based on the @type of the dataset, get a composer function appropriate for it to generate
+    // a subtitle, passing it the related datasets if it exists. Any dataset objects without a
+    // composer function get rendered normally.
+    let subtitle;
+    const type = context['@type'][0];
+    if (context.related_datasets && context.related_datasets.length > 0 && analysis) {
+        const relatedDataset = context.related_datasets.find((dataset) => analysis.datasets.includes(dataset['@id']));
+        if (relatedDataset) {
+            const composer = subtitleComposers[type];
```
zoldello, line 3988: Where is the logic for reading series without a composer?

zoldello (dismissed):
I have some questions ...

src/encoded/static/components/filegallery.js
forresttanaka: Without a composer, you get an empty `subtitle`, and the title gets renders like it does in production today. That applies to all non-Series objects too.

zoldello: thanks.

forresttanaka:
Updating with Idan’s latest spec.

PR #3964, opened 2021-06-11
ENCD-5992-correct-cart-batch-download-analysis-query-string

forresttanaka:
I also corrected a couple other small unrelated problems that caused React warnings.

### DIFF
--- a/src/encoded/static/components/cart/facet.js
+++ b/src/encoded/static/components/cart/facet.js
@@ -2,7 +2,6 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import _ from 'underscore';
 import { Modal, ModalHeader, ModalBody } from '../../libs/ui/modal';
-import { encodedURIComponent } from '../../libs/query_encoding';
 import { svgIcon } from '../../libs/svg-icons';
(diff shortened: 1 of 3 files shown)


src/encoded/static/components/cart/facet.js
```diff
@@ -50,8 +49,8 @@ const analysisSorter = (facetTerms, analyses) => (
  *
  * @return {string} Combined query string selecting file.analyses titles.
  */
-const analysisFieldMap = (analysisTitles) => (
-    analysisTitles.map((title) => `files.analyses.title=${encodedURIComponent(title)}`).join('&')
+const analysisFieldMap = () => (
```
zoldello, line 52: Why. does this function just return the same string always?

zoldello (approved):
I cannot test this to see if it works but the code looks okay. That said, I do have a question about why a function returns just a string that never changes.

src/encoded/static/components/cart/facet.js
forresttanaka: In this case it's a static string. In a different case it could be different.

PR #3968, opened 2021-06-15
ENCD-6013-bump-queue-worker-get-size

keenangraham:


keenangraham:
Presumably there's some way to set this once in `sections.libsonnet` and build all of the INIs @paul-sud? I just updated them individually.

paul-sud (approved):
LGTM

paul-sud:
> Presumably there's some way to set this once in `sections.libsonnet` and build all of the INIs @paul-sud? I just updated them individually.

I think you did it right in this PR, I did see you updated the sections.libsonnet. The workflow I used is 1) edit jsonnet 2) generate new configs locally with `make config` 3) commit and push . Circle checks that the jsonnet is in sync with generated config so manually editing can work too like it did here.

PR #3970, opened 2021-06-17
ENCD-5967-use-biosample-summary-series-tables

bonitalam:


forresttanaka (approved):
Yup, that’s the correct way to do this.

PR #3971, opened 2021-06-17
ENCD-5720-add-target-facet-to-cart-view-page

forresttanaka:
This adds a calculated property to datasets for their targets; one in the case of Experiment and FunctionalCharacterizationExperiment, or an array of targets in the case of Annotations. Must of the changes in this branch involve populating facets when a property for the term can be an array.


forresttanaka:
Preview demo at: https://encd-5720-target-cart-facet-pv1-fytanaka.demo.encodedcc.org/

paul-sud (approved):
LGTM

PR #3980, opened 2021-06-21
ENCD-5995-update-rnaget-ui

keenangraham:


keenangraham:
@pedrohr Not the cleanest but minimal modification of what's there to achieve requested tweaks.

src/encoded/genomic_data_service.py
```diff
@@ -12,34 +22,93 @@
...
 }
 
+
+FACET_TERM_MAP = {
+    'GRCh38': 'Homo sapiens',
+    'mm10': 'Mus musculus',
+}
+
+
+REVERSE_FACET_TERM_MAP = {
```
pedrohr, line 43: for a small number of values this is ok, but it can be generated automatically:

`{v: k for k, v in FACET_TERM_MAP.items()}`

same for FACET_FIELD_MAP

pedrohr (dismissed):

src/encoded/genomic_data_service.py
keenangraham: Ha, yeah. This occurred to me after I PRed yesterday.

keenangraham:
@pedrohr Can you rereview? Just added a default units to frontend to avoid adding empty `units=` to query string.

pedrohr (approved):
👍

PR #3984, opened 2021-06-22
ENCD-5971-split-diseaseseries-experiment-tables-between-disease-and-non-disease-experiments

forresttanaka:
* Much of the changes involves removing the type checking being done in the `Series` component, which caused a lot of complication. I also removed the `seriesComponents` object that was my early attempt at preventing type checking, and replaced it with wrapper components around the `Series` component. Each wrapper component registers with the content-view registry and then does the prep work for its specific kind of series object. The `Series` component no longer gets called directly from the content-view registry. Some wrapper components handle multiple types of series objects, and some handle specific ones.
* Most of the test data changes involve making a DiseaseSeries object that demonstrates a split table, in this case for /disease-series/ENCSR200DIS/.
* A couple unrelated changes removed duplicate properties in the test data.
* I now get the titles of the different series objects from the `profilesTitles` property from `<App>` context, meaning it takes a bit of time to display the titles when reloading a series page, but navigating to the series page has no delay because `<App>` has already loaded them before.
* ~Preview demo at https://encd-5971-split-experiments-pv1-fytanaka.demo.encodedcc.org/search/?type=DiseaseSeries~
* QA demo at https://encd-5971-split-experiments-qa1-fytanaka.demo.encodedcc.org/search/?type=DiseaseSeries
* Note: I don’t believe any existing DiseaseSeries objects demonstrate a split table. Only the local instance /disease-series/ENCSR200DIS/ does.


forresttanaka:
Github says the BDD test hasn’t finished, but CircleCI says it passed a long time ago.

forresttanaka:
Closing this PR again for updates after talking to Ben, and noticing I had forgotten to place the “Add all to cart” button.

PR #3987, opened 2021-06-24
ENCD-5968-add-MOI-guide-type-to-GM

iyoungworth:


gabdank (dismissed):

src/encoded/schemas/genetic_modification.json
```diff
@@ -986,6 +1017,15 @@
                 "homologous recombination"
             ]
         },
+        "MOI": {
+            "title": "Multiplicity of infection",
+            "description": "The ratio or frequency of viral particles introduced to cells.",
```
cricketsloan: Should there be something that indicates the meaning of high and low.  The description says that it is a ratio, but we are not providing a ratio but rather a classification of that ratio.

cricketsloan (commented):
I seem to remember that we have a checklist of things that need to be modified when changing a schema. I checked
 x changelogs
 x schemas (properties, dependencies, boosts, facets, embeddings)
 x summaries
 x static
 x tests
 x upgrade

Maybe I am missing some bits.  Here are my questions.

A) It seems that we will need upgrades for the existing data.  
B) I would think that one or both of these would want to be on the genetic modification facet

Additionally, I have an inline comment about high/low.

Nice job on the biosample summary.  That has gotten wild yet is so very useful.

src/encoded/schemas/genetic_modification.json
cricketsloan: This conversation is on the jira ticket

src/encoded/upgrade/genetic_modification.py
```diff
@@ -294,3 +294,12 @@ def genetic_modification_9_10(value, system):
     if 'donor' in value:
         value['introduced_elements_donor'] = value['donor']
         value.pop('donor')
+
+
+@upgrade_step('genetic_modification', '10', '11')
+def genetic_modification_10_11(value, system):
+    # https://encodedcc.atlassian.net/browse/ENCD-5968
+    notes = value.get('notes', '')
+    if value.get('introduced_elements', '') == 'gRNAs and CRISPR machinery':
```
cricketsloan, line 303: For this to be a safe upgrade that could be completed again after there was new data, the if statement must also check that guide_type is not present.  This will not cause a problem in the current upgrade, but if there was ever a need to go backwards, i would be concerned that you would overwrite the corrected data.

src/encoded/schemas/genetic_modification.json
```diff
@@ -986,6 +1057,15 @@
                 "homologous recombination"
             ]
         },
+        "MOI": {
+            "title": "Multiplicity of infection",
+            "description": "A proxy for whether this modification was intended to result in single (low MOI) or multiple (high MOI) transduction events per cell.",
```
cricketsloan, line 1062: I love this

src/encoded/upgrade/genetic_modification.py
```diff
@@ -294,3 +294,13 @@ def genetic_modification_9_10(value, system):
...
+        if 'guide_type' not in value:
+            value['guide_type'] = 'sgRNA'
+            value['notes'] = (notes + ' guide_type on this GM was defaulted to sgRNA in an upgrade.').strip()
```
cricketsloan, line 306: not sure the .strip is needed in this addition, but it cannot hurt.

src/encoded/upgrade/genetic_modification.py
cricketsloan: Did we discover that everything that we have is sgRNA, or is that just the assumed defualt?

src/encoded/schemas/genetic_modification.json
```diff
@@ -293,6 +296,9 @@
                                 },
                                 "introduced_elements": {
                                     "enum": ["gRNAs and CRISPR machinery"]
+                                },
+                                "guide_type": {
+                                    "enum": ["sgRNA", "pgRNA"]
```
cricketsloan, line 301: I think if you are allowing any value in the enum (i.e. not a subset) you don't need to specify in the dependancy.  But it doesn't hurt

cricketsloan (approved):
First, I have just done a code review.  I did not test the actual behavior yet.  I also think that one of the associates might be doing that. 

I have left some comments, but nothing that needs to be changed.  There are small things you could consider if you need to go back in the code for other reasons like something fails in a test.

src/encoded/upgrade/genetic_modification.py
iyoungworth: sgRNA is just the assumed default, we expect a few pgRNA but need confirmation before patching those (will be addressed in WRAN-2003)

src/encoded/upgrade/genetic_modification.py
iyoungworth: It should remove the space at the beginning of the string if 'notes' is empty before the upgrade

src/encoded/upgrade/genetic_modification.py
iyoungworth: sgRNA is just the assumed default, we expect a few pgRNA but need confirmation before patching those (will be addressed in WRAN-2003)

PR #3989, opened 2021-06-24
ENCD-5971-split-diseaseseries-experiment-tables-and-updated-series-and-search-result-implementation

forresttanaka:
* Much of the changes involves removing the type checking being done in the `Series` component, which caused a lot of complication. I also removed the `seriesComponents` object that was my early attempt at preventing type checking, and replaced it with wrapper components around the `Series` component. Each wrapper component registers with the content-view registry and then does the prep work for its specific kind of series object. The `Series` component no longer gets called directly from the content-view registry. Some wrapper components handle multiple types of series objects, and some handle specific ones.
* I did a similar but much less ambitious refactoring of the Dataset search results, just splitting off the Series search results to simplify the base Dataset search result component. The new Series component should eventually get further refactored into (more or less) each kind of Series object, and just share whatever code they can.
* I now get the titles of the different series objects from the `profilesTitles` property from `<App>` context, meaning it takes a bit of time to display the titles when reloading a series page, but navigating to the series page has no delay because `<App>` has already loaded them before.
* Most of the test data changes involve making a DiseaseSeries object that demonstrates a split table, in this case for /disease-series/ENCSR200DIS/.
* A couple unrelated changes removed duplicate properties in the test data.
* Demo at https://encd-5971-split-experiments-qa3-fytanaka.demo.encodedcc.org//search/?type=DiseaseSeries.


emro (approved):
This ticket looks great, the code all makes sense and I think the way that the pages are constructed makes more sense now, though there is still some inevitable hard-coding of components.

I do feel like these changes would have made more sense as two tickets, one for the code refactoring and then the table update. I feel like the majority of the changes do not address the main ticket issue, though I know they make fixing the main ticket issue easier.

The demo looks good, I see the changes to the tables and the other series pages look unchanged, except for the flickering of the tab titles, which I see you made a QA note about. There's no fix for that, I guess, which is too bad, but it's ok.

PR #3990, opened 2021-06-24
ENCD-6022-add-filter-type-and-filter-value-properties

jenjou:


gabdank (approved):

gabdank (changes requested):
(can we have the enums change to be without underscores and add units?)
bin size (bp)
bin size (kb)
bin size (mb)
mapping quality
false discovery rate percentage

gabdank (dismissed):

hitz (approved):

PR #3992, opened 2021-06-25
ENCD-6025-display-targets-in-cart-valis-display

forresttanaka:


forresttanaka:
Demo: https://encd-6025-cart-valis-targets-qa1-fytanaka.demo.encodedcc.org/

src/encoded/static/components/cart/cart.js
```diff
@@ -595,6 +601,14 @@ const addToAccumulatingFiles = (files, currentResults) => {
                                 file[fileProp] = facetField.getValue(file);
                             }
                         });
+
+                        // Special case: mutate the files to copy the target/targets of the parent
```
zoldello, line 605: It is a code smell to change an iterator. But I guess this is the best option.

src/encoded/static/components/genome_browser.js
```diff
@@ -553,6 +553,21 @@ GenomeLegend.propTypes = {
...
 
 
+/**
+ * Retrieve the target from the given file. Some files have a `targets` array, in which case return
+ *  the first target from this array.
+ * @param {object} file File included in browser
+ * @returns {object} `target` or first `targets` object from file.
+ */
+const getFileTarget = (file) => {
+    let fileTarget = file.targets && file.targets[0];
```
zoldello, line 563: I look forward to August when we don't have to do use 20 year-old hacks like this; that make me pause. In August, we can use "Optional chaining" and not worry about IE11 self-destruction.

src/encoded/static/components/genome_browser.js
```diff
@@ -879,7 +895,8 @@ class GenomeBrowser extends React.Component {
             // Here we do some approximate math to try to figure out how many lines the labels extend to assuming that ~30 characters fit on one line
             // Labels on the experiment pages are short enough to fit on one line (they contain less information) so we can bypass these calculations for those pages
             if (label === 'cart') {
-                labelLength += file.target && file.target.label ? file.target.label.length + 2 : 0;
+                const fileTarget = getFileTarget(file);
+                labelLength += fileTarget && fileTarget.label ? fileTarget.label.length + 2 : 0;
```
zoldello, line 899: += on strings is code smell and it is better to use a join given that strings are immutable so += is doing extra work to concatenate strings. But I guess since the work is small it is no big deal.

zoldello (approved):

PR #3995, opened 2021-06-26
ENCD-5984-add-title-for-entex-file-details

zoldello:


src/encoded/static/components/objectutils.js
```diff
@@ -862,7 +862,7 @@ export function computeAssemblyAnnotationValue(assembly, annotation) {
     // First level of sorting: most recent assemblies are ordered first (represented by numerical component of assembly)
     // Second level of sorting: assemblies without '-minimal' are sorted before assemblies with '-minimal' at the end (represented by tenths place value which is 5 if there is no '-minimal')
     // Third level of sorting: Annotations within an assembly are ordered with most recent first, with more recent annotations having a higher annotation number (with the exception of "ENSEMBL V65") (represented by the annotation number divided by 10,000, or, the three decimal places after the tenths place)
-    let assemblyNumber = +assembly.match(/[0-9]+/g)[0];
+    let assemblyNumber = Number.isNaN(Number(assembly)) ? 0 : +assembly.match(/[0-9]+/g)[0];
```
forresttanaka: When would Number(assembly) ever have any value other than NaN? I think `assemblyNumber` always gets the value 0.

The cart uses this function for sorting assemblies. With dev I get:

dev

…and with this branch I get:

branch

…which seems wrong.

forresttanaka (changes requested):
I think the BDD failure might be real, though it’s kind of hard to say.

PR #3998, opened 2021-06-29
ENCD-5618 encyclopedia page

emro:
Adds a new encyclopedia landing page

### DIFF
--- a/src/encoded/static/components/encyclopedia.js
+++ b/src/encoded/static/components/encyclopedia.js
@@ -1,47 +1,490 @@
+import React from 'react';
 import PropTypes from 'prop-types';
 import url from 'url';
-import { Panel, PanelBody } from '../libs/ui/panel';
-import { ResultTable } from './search';
+import { Panel, TabPanel } from '../libs/ui/panel';
+import GenomeBrowser from './genome_browser';
 import * as globals from './globals';
+import getSeriesData from './series_search.js';
+import { FacetList } from './search';
+import { ASSEMBLY_DETAILS } from './vis_defines';
+import {
+    SearchBatchDownloadController,
+    BatchDownloadActuator,
+} from './batch_download';
+import { svgIcon } from '../libs/svg-icons';
+import QueryString from '../libs/query_string';
 
-const Encyclopedia = (props, context) => {
-    const searchBase = url.parse(context.location_href).search || '';
+// Generate tabs for available organisms
+const organismTabs = {};
+const organismTerms = ['Homo sapiens', 'Mus musculus'];
+organismTerms.forEach((organismName) => {
+    organismTabs[organismName] = <div id={organismName} className={`organism-button ${organismName.replace(' ', '-')}`}><img src={`/static/img/bodyMap/organisms/${organismName.replace(' ', '-')}.svg`} alt={organismName} /><span>{organismName}</span></div>;
+});
 
-    const currentRegion = (assembly, region) => {
-        let lastRegion = {};
-        if (assembly && region) {
-            lastRegion = {
-                assembly,
-                region,
-            };
+// Not all annotation facets are displayed, these are the ones we want to keep
+const keepFacets = [
+    'assembly', // we actually will not display the assembly facet, but we need it in order to determine which assemblies correspond to results
+    'biochemical_inputs', // likely to remove and add back when toggle is working correctly
+    'biosample_ontology.term_name',
+    'biosample_ontology.organ_slims',
+    'targets.label',
+    'assay_title',
+    'encyclopedia_version',
+];
+
+// Annotation types to display
+const annotationTypes = [
+    'candidate Cis-Regulatory Elements',
+    'chromatin state',
+    // 'imputation', // These will be added in later
+];
+
+// Find element or parent of element with matching tag
+function findUpTag(el, tag) {
+    while (el.parentNode) {
+        el = el.parentNode;
+        if (el.tagName === tag) {
+            return el;
         }
-        return lastRegion;
+    }
+    return null;
+}
+
+// Find element or parent of element with matching class name
+function findUpClassName(el, tag) {
+    while (el.parentNode) {
+        el = el.parentNode;
+        if (el.className === tag) {
+            return el;
+        }
+    }
+    return null;
+}
+
+// Default assembly for each organism
+const defaultAssemblyByOrganism = {
+    'Homo sapiens': 'GRCh38',
+    'Mus musculus': 'mm10',
+};
+
+// Hide facets that we don't want to display
+const filterFacet = (response) => {
+    const newResponse = response;
+    const filteredFacets = response.facets.filter((facet) => (keepFacets.indexOf(facet.field) > -1));
+    newResponse.facets = filteredFacets;
+    return newResponse;
+};
+
+// The encyclopedia page displays a table of results corresponding to a selected annotation type
+const Encyclopedia = (props, context) => {
+    // Set defaults
+    const defaultOrganism = 'Homo sapiens';
+    const defaultAnnotation = 'candidate Cis-Regulatory Elements';
+    const defaultAssembly = 'GRCh38';
+    const defaultFileDownload = 'all';
+    const encyclopediaVersion = 'ENCODE v5'; // 'ENCODE v6';
+    const searchBase = url.parse(context.location_href).search || '';
+    const fileOptions = ['CTCF-only', 'proximal enhancer-like', 'DNase-H3K4me3', 'distal enhancer-like', 'promoter-like', 'rDHS', 'all']; // Download options for cell-type agnostic cCREs
+
+    // Set variables based on defaults
+    const [selectedOrganism, setSelectedOrganism] = React.useState(defaultOrganism);
+    const [annotationType, setAnnotationType] = React.useState([defaultAnnotation]);
+    const [selectedAssembly, setAssembly] = React.useState(defaultAssembly);
+    const [assemblyList, setAssemblyList] = React.useState([defaultAssembly]);
+    const [selectedFiles, setSelectedFiles] = React.useState([defaultFileDownload]);
+
+    // Data which populates the browser
+    const [facetData, setFacetData] = React.useState([]);
+    const [vizFiles, setVizFiles] = React.useState([]);
+
+    // Links are used to generate batch download objects
+    const [downloadHref, setdownloadHref] = React.useState(`type=Annotation&annotation_subtype=${defaultFileDownload}&status=released&encyclopedia_version=${encyclopediaVersion}&assembly=${defaultAssembly}`);
+    const [downloadController, setDownloadController] = React.useState(null);
+    const [browserHref, setBrowserHref] = React.useState('');
+    const [browserController, setBrowserController] = React.useState(null);
+
+    // vizFilesError is true when there are no results for selected filters
+    const [vizFilesError, setVizFilesError] = React.useState(false);
+    // Number of files available is displayed
+    const [totalFiles, setTotalFiles] = React.useState(0);
+
+    // Reset button resets to cCREs and the assembly matching the organism tab
+    const resetPage = () => {
+        const assembly = defaultAssemblyByOrganism[selectedOrganism];
+        setAssembly(assembly);
+        const newBrowserHref = `?type=File&annotation_type=${defaultAnnotation}&status=released&encyclopedia_version=${encyclopediaVersion}&file_format=bigBed&file_format=bigWig&assembly=${assembly}`;
+        setAnnotationType(['candidate Cis-Regulatory Elements']);
+        setBrowserHref(newBrowserHref);
+        getSeriesData(newBrowserHref, context.fetch).then((response) => {
+            if (response) {
+                const newResponse = filterFacet(response, selectedOrganism);
+                setFacetData(newResponse);
+                setTotalFiles(newResponse.total);
+                setVizFilesError(false);
+            } else {
+                setFacetData([]);
+                setTotalFiles(0);
+                setVizFilesError(true);
+            }
+        });
+    };
+
+    // Compile list of available assemblies from facet
+    const findAvailableAssemblies = (facets) => {
+        const newAssemblyList = [];
+        facets.forEach((facet) => {
+            if (facet.field === 'assembly') {
+                facet.terms.forEach((term) => {
+                    const termName = term.key.toString();
+                    if (ASSEMBLY_DETAILS[termName] && ASSEMBLY_DETAILS[termName].species === selectedOrganism) {
+                        newAssemblyList.push(termName);
+                    }
+                });
+                setAssemblyList(newAssemblyList);
+            }
+        });
+    };
+
+    // Update browser link and facet data when organism, assembly, or annotation type changes
+    React.useEffect(() => {
+        let newBrowserHref;
+        // If "browserHref" is already set, update assembly and annotation
+        if (browserHref) {
+            const query = new QueryString(browserHref);
+            query.deleteKeyValue('annotation_type');
+            annotationType.forEach((type) => {
+                query.addKeyValue('annotation_type', type);
+            });
+            query.deleteKeyValue('assembly');
+            query.addKeyValue('assembly', selectedAssembly);
+            newBrowserHref = query.format();
+        // Set browserHref if it is not set
+        } else {
+            newBrowserHref = `?type=File&annotation_type=${annotationType}&status=released&encyclopedia_version=${encyclopediaVersion}&file_format=bigBed&file_format=bigWig&assembly=${selectedAssembly}`;
+        }
+        setBrowserHref(newBrowserHref);
+        getSeriesData(newBrowserHref, context.fetch).then((response) => {
+            if (response) {
+                const newResponse = filterFacet(response, selectedOrganism);
+                findAvailableAssemblies(newResponse.facets);
+                setFacetData(newResponse);
+                setTotalFiles(newResponse.total);
+                setVizFilesError(false);
+            } else {
+                setFacetData([]);
+                setTotalFiles(0);
+                setVizFilesError(true);
+            }
+        });
+    }, [selectedOrganism, selectedAssembly, annotationType, context.fetch]);
+
+    // Update visualized files when facet data is updated
+    // Filter out cCRE files that are not the "cCRE, all" file
+    React.useEffect(() => {
+        let newFiles = [];
+        if (facetData && facetData['@graph']) {
+            newFiles = facetData['@graph'];
+            const filteredFiles = newFiles.filter(((file) => {
+                let hideFile = false;
+                hideFile = file.annotation_type === 'candidate Cis-Regulatory Elements' && file.annotation_subtype && file.annotation_subtype !== 'all';
+                return (!hideFile);
+            }));
+            setVizFiles(filteredFiles);
+        } else {
+            setVizFiles([]);
+        }
+    }, [facetData]);
+
+    // Update cell-type agnostic batch download button when download link changes
+    React.useEffect(() => {
+        const query = new QueryString(downloadHref);
+        const controller = new SearchBatchDownloadController('Annotation', query);
+        setDownloadController(controller);
+    }, [downloadHref]);
+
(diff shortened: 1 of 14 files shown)


src/encoded/static/components/encyclopedia.js
```diff
@@ -1,49 +1,492 @@
...
+// Generate tabs for available organisms
+const organismTabs = {};
+const organismTerms = ['Homo sapiens', 'Mus musculus'];
+organismTerms.forEach((organismName) => {
+    organismTabs[organismName] = <div id={organismName} className={`organism-button ${organismName.replace(' ', '-')}`}><img src={`/static/img/bodyMap/organisms/${organismName.replace(' ', '-')}.svg`} alt={organismName} /><span>{organismName}</span></div>;
+});
+
+// Not all annotation facets are displayed, these are the ones we want to keep
+const keepFacets = [
+    'assembly', // we actually will not display the assembly facet, but we need it in order to determine which assemblies correspond to results
```
zoldello: I don't understand the comment at first glance. Maybe it is just me though.

src/encoded/static/components/encyclopedia.js
```diff
@@ -1,49 +1,492 @@
...
+    'biochemical_inputs', // likely to remove and add back when toggle is working correctly
```
zoldello: likely to remove**d** ~~and~~ **but** **re-** add ~~back~~ when toggle is working correctly"

src/encoded/static/components/encyclopedia.js
```diff
@@ -1,49 +1,492 @@
...
+    'targets.label',
+    'assay_title',
+    'encyclopedia_version',
+];
+
+// Annotation types to display
+const annotationTypes = [
+    'candidate Cis-Regulatory Elements',
+    'chromatin state',
+    // 'imputation', // These will be added in later
```
zoldello: I recommend you delete until it is needed.

src/encoded/static/components/encyclopedia.js
```diff
@@ -1,49 +1,492 @@
...
+];
+
+// Find element or parent of element with matching tag
+function findUpTag(el, tag) {
```
zoldello: The code is fine. Still, is there a reason you are not using arrow function? Nowadays I see being use the reserved keyword "function" when dealing with function-context (aka "this") and need it not to be the same as the function's parent. Long story short- Something fancy.

src/encoded/static/components/encyclopedia.js
```diff
@@ -1,49 +1,492 @@
...
+    while (el.parentNode) {
+        el = el.parentNode;
```
zoldello, line 45: I assume you are guaranteed to never have cycles. If you do, you will get an infinite loop.

src/encoded/static/components/encyclopedia.js
```diff
@@ -1,49 +1,492 @@
...
+        if (el.tagName === tag) {
+            return el;
+        }
+    }
+    return null;
+}
+
+// Find element or parent of element with matching class name
+function findUpClassName(el, tag) {
```
zoldello: `findUpTag` and `findUpClassName` can be combined as the codes are essentially the same. Like, instead of -
```
el.className === tag
```

You can do -
```
el[key] === tag
```
Where ```key``` is either a class or tagName.

src/encoded/static/components/encyclopedia.js
```diff
@@ -1,49 +1,492 @@
...
+    while (el.parentNode) {
+        el = el.parentNode;
+        if (el.className === tag) {
+            return el;
+        }
+    }
+    return null;
```
zoldello: Do you want to return `null` or `undefined`? `null` to me means "you intentionally sent something to empty-reference." `undefined` to me means "object has never been set or does not exist." That is just me and I have not seen this set in stone.

src/encoded/static/components/encyclopedia.js
```diff
@@ -1,49 +1,492 @@
...
+
+// Default assembly for each organism
+const defaultAssemblyByOrganism = {
+    'Homo sapiens': 'GRCh38',
+    'Mus musculus': 'mm10',
+};
+
+// Hide facets that we don't want to display
+const filterFacet = (response) => {
+    const newResponse = response;
```
zoldello, line 61: [code smell](https://en.wikipedia.org/wiki/Code_smell). Why not just calculate `filteredFacets` and let the caller worry about using it to modify something.

src/encoded/static/components/encyclopedia.js
```diff
@@ -1,49 +1,492 @@
...
+};
+
+// The encyclopedia page displays a table of results corresponding to a selected annotation type
 const Encyclopedia = (props, context) => {
+    // Set defaults
+    const defaultOrganism = 'Homo sapiens';
+    const defaultAnnotation = 'candidate Cis-Regulatory Elements';
+    const defaultAssembly = 'GRCh38';
+    const defaultFileDownload = 'all';
+    const encyclopediaVersion = 'ENCODE v5'; // 'ENCODE v6';
```
zoldello: ENCODE v5? Really? Anyways, plus; why do you hav the v6 comment?

src/encoded/static/components/encyclopedia.js
```diff
@@ -1,49 +1,492 @@
...
     const searchBase = url.parse(context.location_href).search || '';
+    const fileOptions = ['CTCF-only', 'proximal enhancer-like', 'DNase-H3K4me3', 'distal enhancer-like', 'promoter-like', 'rDHS', 'all']; // Download options for cell-type agnostic cCREs
+
+    // Set variables based on defaults
```
zoldello: People reading the code should know you are setting defaults or can look up hooks and figure that out. You may want to consider just saying - "Default states". Like your comment one line 96 (Data which populates the browser) is informative.

src/encoded/static/components/encyclopedia.js
```diff
@@ -1,49 +1,492 @@
...
+    const [annotationType, setAnnotationType] = React.useState([defaultAnnotation]);
+    const [selectedAssembly, setAssembly] = React.useState(defaultAssembly);
+    const [assemblyList, setAssemblyList] = React.useState([defaultAssembly]);
+    const [selectedFiles, setSelectedFiles] = React.useState([defaultFileDownload]);
+
+    // Data which populates the browser
+    const [facetData, setFacetData] = React.useState([]);
+    const [vizFiles, setVizFiles] = React.useState([]);
+
+    // Links are used to generate batch download objects
```
zoldello: Should say - Links ~~are~~ used to generate batch download objects

src/encoded/static/components/encyclopedia.js
```diff
@@ -1,49 +1,492 @@
...
+    };
+
+    // Compile list of available assemblies from facet
+    const findAvailableAssemblies = (facets) => {
+        const newAssemblyList = [];
+        facets.forEach((facet) => {
+            if (facet.field === 'assembly') {
+                facet.terms.forEach((term) => {
+                    const termName = term.key.toString();
+                    if (ASSEMBLY_DETAILS[termName] && ASSEMBLY_DETAILS[termName].species === selectedOrganism) {
```
zoldello, line 126: When IE11 support is dropped, we can do [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) and don't have to resort to 20-year old hacks anymore :)

src/encoded/static/components/encyclopedia.js
```diff
@@ -1,49 +1,492 @@
...
+    // Update visualized files when facet data is updated
+    // Filter out cCRE files that are not the "cCRE, all" file
+    React.useEffect(() => {
+        let newFiles = [];
+        if (facetData && facetData['@graph']) {
+            newFiles = facetData['@graph'];
+            const filteredFiles = newFiles.filter(((file) => {
+                let hideFile = false;
+                hideFile = file.annotation_type === 'candidate Cis-Regulatory Elements' && file.annotation_subtype && file.annotation_subtype !== 'all';
+                return (!hideFile);
```
zoldello: Why is this in parenthesis (or "bracket" if you are like me and keep slipping into Queen's English)? The code is fine, but I am just curious if you did it for a special reason.

src/encoded/static/components/encyclopedia.js
```diff
@@ -1,49 +1,492 @@
...
+        const query = new QueryString(downloadHref);
+        let newFiles;
+        if (selectedFiles.indexOf(input) === -1) {
+            newFiles = [...selectedFiles, input];
+        } else {
+            newFiles = selectedFiles.filter((file) => file !== input);
+        }
+        // Do not allow the user to de-select all checkboxes
+        if (newFiles.length === 0) {
+            return;
```
zoldello, line 212: This returns `undefined`. Is that intended or you mean `null` or it does not matter either way..?

src/encoded/static/components/encyclopedia.js
```diff
@@ -1,49 +1,492 @@
...
+        }
+        // Clear current selections
+        query.deleteKeyValue('annotation_type');
+        query.deleteKeyValue('annotation_subtype');
+        query.deleteKeyValue('biosample_ontology');
+        // All file options are selected
```
zoldello: You might want to add a space above this.

src/encoded/static/components/encyclopedia.js
```diff
@@ -1,49 +1,492 @@
...
+        if (newFiles.length === fileOptions.length) {
+            query.addKeyValue('annotation_type', 'representative DNase hypersensitivity sites');
+            query.addKeyValue('annotation_type', 'candidate Cis-Regulatory Elements');
+            query.addKeyValue('biosample_ontology', '*', true);
+        // Only rDHS is selected
+        } else if (newFiles.length === 1 && newFiles[0] === 'rDHS') {
+            query.addKeyValue('annotation_type', 'representative DNase hypersensitivity sites');
+        // rDHS and other annotation subtypes are selected
+        // This is the option that does not currently work *****
```
zoldello: It is standard to write something like -
`// TODO: Get this option working`

"// TODO" has special meaning to IDE's and code readers. Of course, this should be either fixed or functionality removed before merging the code to dev.

src/encoded/static/components/encyclopedia.js
```diff
@@ -1,49 +1,492 @@
...
+        const query = new QueryString(downloadHref);
+        query.replaceKeyValue('assembly', defaultAssemblyByOrganism[tab]);
+        const newHref = query.format();
+        setdownloadHref(newHref);
+    };
+
+    // Annotation type facet is not a normal facet, at least one annotation must be selected
+    // Facet filters look like radiobuttons
+    const chooseAnnotationType = (type) => {
+        let newAnnotationType = annotationType;
```
zoldello, line 244: Why do you `newAnnotationType` with an initial value only to later never use the initial value?

src/encoded/static/components/encyclopedia.js
```diff
@@ -1,49 +1,492 @@
...
+        if (annotationType.indexOf(type) > -1 && annotationType.length > 1) {
+            newAnnotationType = annotationType.filter((annotation) => annotation !== type);
+        } else if (annotationType.indexOf(type) === -1) {
+            newAnnotationType = [...annotationType, type];
+        }
+        setAnnotationType(newAnnotationType);
+    };
+
+    // Facet filters are redirected to update the urls that populate the browser
```
zoldello: Maybe say - "Update Facet filters urls" ..?

src/encoded/static/components/encyclopedia.js
```diff
@@ -1,49 +1,492 @@
...
+    const handleFacetClick = (e) => {
+        // If clicked filter belongs to the annotation facet, clear filters, or boolean switch, we do not want to prevent default
+        const isButton = findUpTag(e.target, 'BUTTON');
```
zoldello: The code is fine to me. Although some may argue it is best to use a ref to reference HTML elements.

src/encoded/static/components/encyclopedia.js
```diff
@@ -1,49 +1,492 @@
...
+            newAnnotationType = annotationType.filter((annotation) => annotation !== type);
+        } else if (annotationType.indexOf(type) === -1) {
+            newAnnotationType = [...annotationType, type];
+        }
+        setAnnotationType(newAnnotationType);
+    };
+
+    // Facet filters are redirected to update the urls that populate the browser
+    const handleFacetClick = (e) => {
+        // If clicked filter belongs to the annotation facet, clear filters, or boolean switch, we do not want to prevent default
```
zoldello, line 255: preventDefault to me generally deals with stopping elements from doing default behavior like making url calls and/or bubbling up an event. Is that your intent?

src/encoded/static/components/encyclopedia.js
zoldello: Element-sniffing (checking for element type) seems like a code smells and makes the code brittle. Like how about if someone uses -

`<input type"button">`

or adds some other element you did not cover..? I may be wrong on this.

src/encoded/static/components/encyclopedia.js
```diff
@@ -1,49 +1,492 @@
...
-                            <ResultTable
-                                {...props}
-                                searchBase={searchBase}
-                                onChange={context.navigate}
-                                currentRegion={currentRegion}
-                                hideDocType
-                            />
-                        </PanelBody>
-                    </Panel>
+                <div className="block series-search">
```
zoldello, line 289: Is it a good idea to follow BEM..? This is me nitpicking but I am just curious.

src/encoded/static/components/encyclopedia.js
```diff
@@ -1,49 +1,492 @@
...
+                                        </div>
+                                    : null}
+                                    <hr />
+                                </div>
+                                <div className="series-wrapper">
+                                    <div>There are {vizFiles.length} file{vizFiles.length === 1 ? '' : 's'} displayed out of {totalFiles} file{totalFiles.length > 1 ? '' : 's'} that match the selected filters.</div>
+                                    <Panel>
+                                        <div className="file-gallery-container encyclopedia-file-gallery">
+                                            {(facetData && facetData.facets && facetData.facets.length > 0) ?
+                                                <div className="file-gallery-facet-redirect" onClick={(e) => handleFacetClick(e)}>{/* eslint-disable-line */}
```
zoldello, line 354: Why are you disabling eslint on this line? I see no reason to.

src/encoded/static/components/genome_browser.js
```diff
@@ -839,7 +840,8 @@ class GenomeBrowser extends React.Component {
         newFiles = _.uniq(newFiles, (file) => file.href);
         let tracks = [];
         if (files.length > 0) {
-            tracks = this.filesToTracks(newFiles, this.props.label, domain);
+            const maxCharPerLine = this.props.maxCharPerLine || 26;
```
zoldello: maybe make 26 into a variable with upper case letters..?

zoldello (commented):
I added comment for you to consider on github.

On the UI, here are things I noticed -
- Reset filters does not only check box
- Do filters need to tied to the URL?
- I played with download and the count does not seem to match. Like I clicked on “Download cell-type agnostic cCREs”, “CTCF-only”; the count does not match and the download button is disabled until I close and reopen the download box.

src/encoded/static/components/encyclopedia.js
emro: I think I'm just going to delete this comment. The facet is ok though not perfectly intuitive/as-planned but I think we're keeping it. I'll make a ticket for an update.

src/encoded/static/components/encyclopedia.js
emro: Let me try again. There are two "fake" facets at the top which don't function quite like normal facets. They're multi-selects where one selection is required. I'm filtering out most of the facets since we only want a few, but I use the assembly facet data to figure out which assemblies are available and then I use that to generate the fake facet, so I don't want to filter it out. For instance, for Homo sapiens, GRCh38 and hg19 might give results for the selected filters, or maybe only hg19 gives results, so I find that out by looking at the assembly facet, but then I don't actually display that facet, I display my fake assembly facet.

src/encoded/static/components/encyclopedia.js
emro: This is a great idea

src/encoded/static/components/encyclopedia.js
emro: I fixed this, should be arrow function

src/encoded/static/components/encyclopedia.js
emro: Deleted

src/encoded/static/components/encyclopedia.js
emro: I sometimes put in extra parentheses because it's clearer in my head for whatever reason - they're not needed

src/encoded/static/components/encyclopedia.js
emro: Yes. I just don't want the rest of the function to execute in this case. It doesn't really matter what is returned, we just want to exit the function. There's no point to having a download link that doesn't download anything so it's necessary to have at least one checkbox selected. If the user tries to de-select the last option, we prevent that by not allowing the query / selectedFiles to be modified.

src/encoded/static/components/encyclopedia.js
emro: Well, this is all not that ideal - redirecting links is not the greatest - but we're bypassing the API so there's not much else we can do. If someone adds a new element I haven't accounted for, this may need an update, but as it is, the buttons are from the special facets that I've added and the links are from the default facets, so it should be ok.

src/encoded/static/components/encyclopedia.js
emro: Yes, this goes back to the redirecting the links on the default facets. We're using the API to generate the data, but we're not using the links as the url, it's hidden. This all started because we were going to have one link generate one part of the page and another link generate the other part of the page and the API doesn't allow that.

src/encoded/static/components/encyclopedia.js
emro: There's an error here. For a11y compliance, click events need to have a role and keyboard listeners. However, this isn't a real button, it's a redirect for the buttons/links in the facet wrapper, so it's not a real button and I don't think we want it to be a thing people actually click on, if that makes sense. Maybe I'm wrong.

src/encoded/static/components/encyclopedia.js
emro: I'm not sure what you mean. How should this be changed?

src/encoded/static/components/encyclopedia.js
emro: I'm a little confused by this question. The initial value can be updated using the "fake" annotations facet, and this function executes when a new annotation is selected.

emro:
Re the three UI comments:
- I would like to fix the rest on the "Only" toggle when I fix the biochemical inputs facet in another follow-up ticket. It does reset any selection for that facet.
- Filters are not tied to the UI, which is unusual, but that is what we decided to do for this ticket
- Downloads should be fixed, the queries were not correct for all cases

src/encoded/static/components/encyclopedia.js
```diff
@@ -1,49 +1,460 @@
...
+];
+
+// Annotation types to display
+const annotationTypes = [
+    'candidate Cis-Regulatory Elements',
+    'chromatin state',
+];
+
+// Find element or parent of element with matching tag or class name
+const findElementOrParent = (el, name, key) => {
```
zoldello: `findElement`. A parent is an element.

src/encoded/static/components/encyclopedia.js
zoldello: Pardon me for the typos. Nevermind, the comment. `newAnnotationType` uses the default value if neither the if-statement nor the else if-statement are true.

src/encoded/static/components/encyclopedia.js
zoldello: We can talk about this offline if you like.

zoldello (dismissed):
My comments are not significant. Code looks okay.

zoldello (approved):
Looks fine.

PR #4002, opened 2021-06-30
ENCD-5955-series-analysis-title-updates

forresttanaka:
Preview demo: https://encd-5955-analysis-titles-pv3-fytanaka.demo.encodedcc.org/
A demo is coming up and I will update its Javascript with the latest changes once it finishes indexing, to save time at this stage.

### DIFF
--- a/src/encoded/static/components/objectutils.js
+++ b/src/encoded/static/components/objectutils.js
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import _ from 'underscore';
 import url from 'url';
+import getNumberWithOrdinal from '../libs/ordinal_suffix';
 import * as encoding from '../libs/query_encoding';
 import { svgIcon } from '../libs/svg-icons';
 import { CartToggle, cartGetAllowedTypes } from './cart';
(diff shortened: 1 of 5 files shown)


src/encoded/static/components/objectutils.js
```diff
@@ -1079,3 +1080,29 @@ Checkbox.propTypes = {
...
+/**
+ * Examines the `examined_loci` property, if any, of the given dataset and formats it into
+ * displayable text.
+ * @param {object} experiment Dataset containing examined_loci to display
+ * @param {boolean} includeMethod True to include expression method following the expression data
+ * @returns {string} Formatted examined_loci display
+ */
+export function computeExaminedLoci(experiment, includeMethod) {
+    const examinedLoci = [];
+    if (experiment.examined_loci && experiment.examined_loci.length > 0) {
```
zoldello, line 1094: `computeExaminedLoci` need cannot trust its arguments as the function can be called by anyone rather than a fixed group in a class. So I think it would be a good idea to validate arguments before using them.

zoldello (approved):
I did not test the demo; but the code looks fine. My comment is in the PR is not significant.

PR #4009, opened 2021-07-12
ENCD-5724 differentiation series

emro:
Adds differentiation series to series search page. There are two cases, series with one biosample_ontology type, and series with more than one. In the case of one type, the list view and the tables include the post-differentiation times, and in the case of more than one type, biosample ontology types and treatments and biosample summaries are included in the list view and tables.


src/encoded/static/components/dataset.js
```diff
@@ -1481,6 +1481,116 @@ const replicationTimingSeriesTableColumns = {
     },
 };
 
+
+function computeDifferentiation(experiment) {
+    let biosamples;
+    let postDifferentiationTime;
+
+    if (experiment.replicates && experiment.replicates.length > 0) {
```
forresttanaka, line 1489: `replicates` is calculated in an experiment, so it always exists, though it might contain no elements.

src/encoded/static/components/series_search.js
```diff
@@ -28,6 +28,10 @@ const seriesList = {
         title: 'Gene silencing',
         schema: 'gene_silencing_series',
     },
+    DifferentiationSeries: {
+        title: 'Differentiation',
+        schema: 'differentiation_series',
+    },
```
forresttanaka, line 34: Nice that this is the only thing you need to do to add a new section.

forresttanaka (dismissed):
Everything looks good. I have one technicality comment, but it hardly matters at all.

src/encoded/static/components/dataset.js
```diff
@@ -2506,6 +2620,43 @@ ReplicationTimingSeries.contextTypes = {
...
 
 
+/**
+ * Wrapper component for differentiation series pages.
+ */
+const DifferentiationSeries = ({ context }, reactContext) => {
+    const seriesType = context['@type'][0];
+    const seriesTitle = reactContext.profilesTitles[seriesType] || '';
+    const findDifferentiations = context.related_datasets.map((dataset) => (computeDifferentiation(dataset) ? 1 : 0)).reduce((a, b) => a + b, 0);
+    const differentiationTableColumns = findDifferentiations > 0 ? differentiationTableColumnsWithTime : differentiationTableColumnsWithoutTime;
```
forresttanaka, line 2630: I was thinking another way to handle this is to use the `hide` property for SortTable. Whether that’s easier or harder than what you have varies by case, but it does reduce the code duplication for the other columns.

forresttanaka (approved):
Looks like this addresses the QA feedback. I included one suggestion I hadn’t thought of before, but as I mention, whether that suggestion helps or hurts depends.

src/encoded/static/components/dataset.js
emro: I like this idea. I did think it was quite a bit of extra code to define two sets of columns for when one variable existed vs not.

PR #4012, opened 2021-07-14
ENCD-6042-add-selected-filters-to-search-download-options

forresttanaka:
Adds a fourth download option on search-list pages — “Selected files” — that downloads only what the user has selected in the facets, allowing for downloading raw files.

Demo: https://encd-6042-search-dl-options-qa1-fytanaka.demo.encodedcc.org/


emro (approved):
Looks good, works well on the demo

PR #4014, opened 2021-07-19
ENCD-6008-convert-nuc-acid-start-quant-to-number

iyoungworth:
fixture-correction

### DIFF
--- a/src/encoded/tests/fixtures/schemas/library.py
+++ b/src/encoded/tests/fixtures/schemas/library.py
@@ -71,7 +71,7 @@ def library_url(testapp, lab, award, biosample):
 def library_starting_quantity(library):
     item = library.copy()
     item.update({
-        'nucleic_acid_starting_quantity': '10',
+        'nucleic_acid_starting_quantity': 10,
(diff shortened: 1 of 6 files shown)


src/encoded/tests/fixtures/schemas/library.py
```diff
@@ -71,7 +71,7 @@ def library_url(testapp, lab, award, biosample):
 def library_starting_quantity(library):
     item = library.copy()
     item.update({
-        'nucleic_acid_starting_quantity': '10',
+        'nucleic_acid_starting_quantity': 10,
```
hitz, line 74: is there a difference between 10 and 10.0 here once we get to python?

hitz (approved):

src/encoded/tests/fixtures/schemas/library.py
iyoungworth: This one won't change because it's a fixture, but if it were being upgraded, it would become 10.0. It looks like JSON number allows submission of automatically drops trailing zeroes though, so it would get stripped off again after the upgrade

PR #4015, opened 2021-07-19
ENCD-6024-make-search-results-targets-into-links

forresttanaka:
The vast majority of file changes involve adding `target.@id` to various dataset schemas’ `columns` property so that search results can retrieve a link to the target objects, as all dataset targets in search results now link to the target page. Links with factorbook dbxrefs now link to both our own target page as well as the factorbook link.

### DIFF
--- a/src/encoded/static/components/search.js
+++ b/src/encoded/static/components/search.js
@@ -157,6 +157,40 @@ PickerActions.contextTypes = {
 };
 
 
+/**
+ * Displays the search result data line for target objects when the parent object's targets
+ * comprise an array. The parent object's schema must include the target array's @id in its
+ * `columns` property.
+ */
+export const TargetsDataLine = ({ result, targetPropertyName }) => {
+    if (result[targetPropertyName] && result[targetPropertyName].length > 0) {
+        return (
+            <>
+                <span className="result-item__property-title">Target{result[targetPropertyName].length > 1 ? 's' : ''}: </span>
+                {result[targetPropertyName].map((target, i) => (
+                    <React.Fragment key={target['@id']}>
+                        {i > 0 ? ', ' : null}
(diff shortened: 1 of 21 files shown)


forresttanaka:
Because of the schema updates, I can’t patch an old demo to demonstrate this ticket. A new demo is coming up which might begin to index enough by maybe 4:00 p.m. Local instances demonstrate this ticket.

src/encoded/static/components/search.js
```diff
@@ -157,6 +157,40 @@ PickerActions.contextTypes = {
...
+                    <React.Fragment key={target['@id']}>
+                        {i > 0 ? ', ' : null}
+                        <a href={target['@id']} aria-label={`Target ${target.label}`}>{target.label}</a>
+                    </React.Fragment>
+                ))}
+            </>
+        );
+    }
+    return null;
+};
```
emro, line 180: It's too bad you can't use .join for cases like this. I guess you could also use .reduce. But this is a good solution, very readable.

emro (dismissed):
Looks good!

src/encoded/static/components/search.js
forresttanaka: Yeah, there’s another [solution](https://stackoverflow.com/questions/34034038/how-to-render-react-components-by-using-map-and-join/35840806#answer-40788571) that I think Phil found, but I didn’t see any advantage over this method, and it was more difficult to understand, I thought.

forresttanaka:
I keep forgetting about ExperimentSeries since it’s so different from other series objects. This change just adds target.@id to ExperimentSeries search results so the link to the target works. I’m bringing up a new demo for tomorrow.

emro (approved):

src/encoded/static/components/search.js
emro: I actually found that exact link when I was going down the rabbit hole of why can't you use .join!

PR #4017, opened 2021-07-20
ENCD-6038-biosample-summary-series-header

bonitalam:


forresttanaka (dismissed):
For me, I see no issues. You might want to also pass it by someone better at Python though. I looked over it and see no issues, but someone else (Keenan or Phil perhaps) might. For example, I’m not sure if what you have is better, or something like:
```
replicates = datasetObject.get('replicates', None)
if replicates is not None:
```
...is better, or if it makes any difference, or if you already know.

forresttanaka (dismissed):
I didn’t match up code changes with QA feedback, but from what I see, this seems correct.

forresttanaka (approved):

PR #4018, opened 2021-07-20
encd-6039 software cleanup

hitz:
Will need a big patch on demo for full testing.


src/encoded/schemas/changelogs/software.md
```diff
@@ -1,4 +1,7 @@
 ## Changelog for software.json
+### Schema version 11
```
bonitalam: As the code is now, this doesn't require a whole new schema version and is just a minor change. If we are making the "software_type" and "used_by" properties required that would warrant a new schema version and corresponding upgrade.

src/encoded/schemas/software.json
```diff
@@ -43,7 +43,14 @@
                     "quantification",
                     "transcript identification",
                     "transcriptome assembly",
-                    "variant annotation"
+                    "variant annotation",
```
bonitalam: Can we put the new enums in alphabetical order with the rest of the existing enums? Just makes it easier to navigate through the form.

src/encoded/schemas/changelogs/software.md
```diff
@@ -1,4 +1,7 @@
...
+
+* Added several items to software_type enum; deprecated purpose field but did not remove.
```
bonitalam: We should probably just list all the enums here for consistency.

bonitalam (changes requested):
See comments. Also wondering if we are making any properties required are not?

src/encoded/schemas/changelogs/software.md
hitz: do you mean the CHANGES to the enums or just the additions.

bonitalam: List just the new additions

src/encoded/schemas/software.json
```diff
@@ -20,7 +20,7 @@
     ],
     "properties": {
         "schema_version": {
-            "default": "10"
+            "default": "11"
```
bonitalam: This should be 10.

bonitalam (changes requested):

src/encoded/schemas/changelogs/software.md
```diff
@@ -1,4 +1,7 @@
 ## Changelog for software.json
+### Minor changes since chema version 10
+
+* Added utility, proteomics, statistics, visualization, machine learning, integrated analysis, genetics to software_type.  Deprecated but did not remove purpose.
```
bonitalam: Enums and properties should be italicized

bonitalam (changes requested):

src/encoded/schemas/changelogs/software.md
bonitalam: Also should intregated analysis be integrative analysis? If so, need update here and in json file.

bonitalam (approved):

PR #4021, opened 2021-07-20
ENCD-5851 single cell landing page

emro:
Creates a page with three tabs:
- High Throughput
- Perturbed High Throughput
- Low Throughput

### DIFF
--- a/src/encoded/static/components/dataset.js
+++ b/src/encoded/static/components/dataset.js
@@ -2036,7 +2036,12 @@ export const SeriesComponent = ({ context, title, tableColumns, breadcrumbs, opt
     // the code takes this into account
     const diseases = [
         ...new Set(context.related_datasets.reduce((accumulatedDiseases, currentRelatedDataset) => {
-            const diseasesTermNames = currentRelatedDataset.replicates.map((replicate) => replicate.library.biosample.disease_term_name);
+            const diseasesTermNames = currentRelatedDataset.replicates.map((replicate) => {
+                if (typeof replicate !== 'string') {
+                    return replicate.library.biosample.disease_term_name;
+                }
+                return null;
+            });
             return accumulatedDiseases.concat(...diseasesTermNames);
         }, [])),
     ].filter((disease) => disease);
--- a/src/encoded/static/components/search.js
+++ b/src/encoded/static/components/search.js
@@ -295,6 +295,9 @@ globals.listingViews.register(Biosample, 'Biosample');
 const ExperimentComponent = (props, reactContext) => {
     const { context: result, cartControls, mode } = props;
     let synchronizations;
+    let constructionPlatforms;
+    let constructionMethods;
+    let cellularComponents;
 
     // Determine if search result is allowed in carts.
     const isResultAllowedInCart = cartGetAllowedTypes().includes(result['@type'][0]);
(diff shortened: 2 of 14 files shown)


src/encoded/static/components/dataset.js
```diff
@@ -2036,7 +2036,12 @@ export const SeriesComponent = ({ context, title, tableColumns, breadcrumbs, opt
     // the code takes this into account
     const diseases = [
         ...new Set(context.related_datasets.reduce((accumulatedDiseases, currentRelatedDataset) => {
-            const diseasesTermNames = currentRelatedDataset.replicates.map((replicate) => replicate.library.biosample.disease_term_name);
+            const diseasesTermNames = currentRelatedDataset.replicates.map((replicate) => {
+                if (typeof replicate !== 'string') {
```
zoldello, line 2040: Let's talk about this. Why are you checking for type rather than property the object has or states like falsey

src/encoded/static/components/search.js
```diff
@@ -353,6 +356,21 @@ const ExperimentComponent = (props, reactContext) => {
         }));
     }
 
+    // Collect library construction platforms / methods / cellular components
+    if (result.replicates && result.replicates.length > 0) {
```
zoldello, line 360: use ?. instead. We are in a post IE11 world.

src/encoded/static/components/search.js
```diff
@@ -353,6 +356,21 @@ const ExperimentComponent = (props, reactContext) => {
...
+        constructionPlatforms = _.uniq(result.replicates.filter((replicate) => (
+            replicate.library && replicate.library.construction_platform && replicate.library.construction_platform.term_name
+        )).map((replicate) => replicate.library.construction_platform.term_name));
+
+        constructionMethods = _.uniq(result.replicates.filter((replicate) => (
```
zoldello, line 365: code is fine albeit using Set is better as it is faster because it is inbuilt.

src/encoded/static/components/search.js
```diff
@@ -449,6 +467,15 @@ const ExperimentComponent = (props, reactContext) => {
                                 {screenLink ?
                                     <div><span className="result-item__property-title">candidate Cis-Regulatory Elements (cCREs): </span><a href={`https://screen.encodeproject.org/search?q=${screenSearch}&assembly=${screenAssembly}`}>SCREEN</a></div>
                                 : null}
+                                {cellularComponents && cellularComponents.length > 0 ?
```
zoldello, line 470: "?."?

src/encoded/static/components/search.js
```diff
@@ -734,6 +764,17 @@ const SeriesComponent = ({ context: result, auditDetail, auditIndicators }, reac
                             phases.push(biosample.phase);
                         }
                     }
+                    if (replicate.library) {
```
zoldello, line 767: Would it be better if you do -
`if (!replicate.library) { return; }`
and if this above results in falsey, continue on with the multiple if-statements...

src/encoded/static/components/single_cell.js
```diff
@@ -0,0 +1,152 @@
...
+import { Panel, PanelBody, TabPanel } from '../libs/ui/panel';
+import { ResultTable } from './search';
+import QueryString from '../libs/query_string';
+import * as globals from './globals';
+
+// The single cell page includes the following three groups of experiments
+const singleCellList = {
+    highThroughput: {
+        title: 'High throughput',
+        search: '?type=Experiment&assay_slims=Single+cell&status!=replaced',
```
zoldello, line 13: can you use context.base_search?

src/encoded/static/components/single_cell.js
```diff
@@ -0,0 +1,152 @@
...
+    newProps.context = filterFacet(props.context);
+
+    const handleTabClick = React.useCallback((series) => {
+        const href = `/single-cell/${singleCellList[series].search}`;
+        context.navigate(href);
+        setDescriptionData(singleCellList[series].description);
+    }, [context]);
+
+    const seriesTabs = {};
+    Object.keys(singleCellList).forEach((s) => {
```
zoldello, line 81: Object.entrices?

zoldello (approved):
I have some comments about the code and they may be insignificant. I will leaving testing to Bointa.

PR #4023, opened 2021-07-21
ENCD-5876-series-updates

forresttanaka:
I have a preview demo at: https://encd-5876-series-updates-pv2-fytanaka.demo.encodedcc.org/

I also have a QA demo coming up for tomorrow.

### DIFF
--- a/src/encoded/static/components/dataset.js
+++ b/src/encoded/static/components/dataset.js
@@ -8,9 +8,9 @@ import * as globals from './globals';
 import { DbxrefList } from './dbxref';
 import { FetchedItems } from './fetched';
 import { auditDecor } from './audit';
-import Status from './status';
 import pubReferenceList from './reference';
 import {
+    collectDatasetBiosamples,
     computeExaminedLoci,
     donorDiversity,
     publicDataset,
@@ -25,6 +25,7 @@ import { ProjectBadge } from './image';
 import { DocumentsPanelReq } from './doc';
 import { FileGallery } from './filegallery';
 import sortMouseArray from './matrix_mouse_development';
+import Status, { getObjectStatuses, sessionToAccessLevel } from './status';
 import { AwardRef, ReplacementAccessions, ControllingExperiments, FileTablePaged, ExperimentTable, DoiRef } from './typeutils';
 
 /**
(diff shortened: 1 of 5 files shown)


src/encoded/static/components/objectutils.js
```diff
@@ -1109,3 +1109,20 @@ export function computeExaminedLoci(experiment, includeMethod) {
...
+
+
+/**
+ * Collects the biosamples contained in all replicate libraries within the given dataset;
+ * deduplicated by accession.
+ * @param {object} dataset Dataset from which to extract all biosamples.
+ * @returns {array} Array of all biosamples in all replicates of `dataset`.
+ */
+export const collectDatasetBiosamples = (dataset) => {
+    if (dataset.replicates && dataset.replicates.length > 0) {
```
zoldello, line 1121: You can use "?." now. Something like - >
```
if (dataset?.replicates?.length > 0) {
 // some code
}
```

src/encoded/static/components/genome_browser.js
```diff
@@ -913,6 +919,7 @@ class GenomeBrowser extends React.Component {
                 labelLength += file.assay_term_name ? file.assay_term_name.length + 2 : 0;
                 labelLength += file.biosample_ontology && file.biosample_ontology.term_name ? file.biosample_ontology.term_name.length + 2 : 0;
                 labelLength += file.annotation_subtype ? file.annotation_subtype.length : 0;
+                labelLength += supplementalShortLabel ? supplementalShortLabel.length : 0;
```
zoldello, line 922: You can also do something like - 
`labelLength += supplementalShortLabel?.length || 0;`

src/encoded/static/components/filegallery.js
```diff
@@ -484,7 +484,7 @@ export class FileTable extends React.Component {
 
             return (
                 <div>
-                    {showFileCount ? <div className="file-gallery-counts">Displaying {filteredCount} of {unfilteredCount} files</div> : null}
+                    {showFileCount ? <div className="table-counts">Displaying {filteredCount} of {unfilteredCount} files</div> : null}
```
zoldello, line 487: Is `table-counts` descriptive enough? `file-gallery-counts` tells me more, IMHO.

src/encoded/static/components/dataset.js
```diff
@@ -1206,28 +1207,92 @@ FilePanelHeader.propTypes = {
 };
 
 
-function displayPossibleControls(item, adminUser) {
-    if (item.possible_controls && item.possible_controls.length > 0) {
-        return (
+/**
+ * Series experiment table target column definition.
+ */
+const seriesTableTargetColumn = {
```
zoldello, line 1213: This reminds me of Javascript code in a pre-Class word where objects were used to encapsulate functionality like a Class could. Unless you are concerned with performance, would as class suffice?

src/encoded/static/components/dataset.js
```diff
@@ -1270,7 +1356,20 @@ const basicTableColumns = {
...
+
+/**
+ * Collect all donor objects from all the given biosamples. Any duplicate donors get removed.
+ * @param {array} biosamples List of biosamples from which to extract donors.
+ * @returns {array} List of all distinct donors.
+ */
+const collectDonorsFromBiosamples = (biosamples) => (
+    _(biosamples.reduce((donors, biosample) => (
+        biosample.donor ? donors.concat(biosample.donor) : donors
+    ), [])).uniq((donor) => donor.accession)
```
zoldello, line 1368: Minor detail. Would it be a good idea to use Set or _.uniq and not both interchangably?

src/encoded/static/components/dataset.js
```diff
@@ -8,9 +8,9 @@ import * as globals from './globals';
 import { DbxrefList } from './dbxref';
 import { FetchedItems } from './fetched';
```
zoldello, line 9: Sometimes I start code reviews from the smaller files then progress to large files; because the smaller files gives me context related to the larger ones... I made some comments  in other files that may apply but I won't repeat them.

zoldello (approved):
None of my comments are significant. Did you see Ben's comment he made an hour after you posted the demo?

src/encoded/static/components/dataset.js
forresttanaka: I don’t know of a way to use sets with objects, only strings.

src/encoded/static/components/filegallery.js
forresttanaka: It applied to tables that aren’t the file gallery.

PR #4025, opened 2021-07-22
ENCD-6036 reset chr range in cart

emro:
"sortAndRefresh" is getting called multiple times when the genome assembly/annotation are updated - first for the assembly/annotation state update and then for the files state update - however the first call doesn't finish executing before the second commences, causing incorrect files to be displayed. Deferring "sortAndRefresh" till the call stack is clear resolves this.

### DIFF
--- a/src/encoded/static/components/genome_browser.js
+++ b/src/encoded/static/components/genome_browser.js
@@ -703,23 +703,28 @@ class GenomeBrowser extends React.Component {
     /* eslint-disable react/no-did-update-set-state */
     componentDidUpdate(prevProps, prevState) {
         if (!(this.state.disableBrowserForIE) && this.GV) {
-            if (this.state.contig !== prevState.contig) {
+            const contigUpdate = this.state.contig !== prevState.contig;
+            const assemblyUpdate = this.props.assembly !== prevProps.assembly || this.props.annotation !== prevProps.annotation;
+            const filesUpdate = !(_.isEqual(this.props.files, prevProps.files));
+            const sizeUpdate = this.props.expanded !== prevProps.expanded;
+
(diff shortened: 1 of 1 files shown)


src/encoded/static/components/genome_browser.js
```diff
@@ -703,23 +703,28 @@ class GenomeBrowser extends React.Component {
     /* eslint-disable react/no-did-update-set-state */
     componentDidUpdate(prevProps, prevState) {
         if (!(this.state.disableBrowserForIE) && this.GV) {
-            if (this.state.contig !== prevState.contig) {
+            const contigUpdate = this.state.contig !== prevState.contig;
```
zoldello, line 706: `contigUpdate` follows react-naming style of booleans so it is fine. But personally I prefer `IsContigUpdated`. WIth that, I know it is a boolean and would not be expected a value like a number or string.

src/encoded/static/components/genome_browser.js
```diff
@@ -703,23 +703,28 @@ class GenomeBrowser extends React.Component {
...
+            const assemblyUpdate = this.props.assembly !== prevProps.assembly || this.props.annotation !== prevProps.annotation;
+            const filesUpdate = !(_.isEqual(this.props.files, prevProps.files));
```
zoldello, line 708: _.isEqual does a deep copy. Is that needed as a shallow copy is faster if it is possible.

zoldello (approved):
Code looks good. My comments are insignificant.

src/encoded/static/components/genome_browser.js
emro: I think so, though I wrote this code a while ago.

PR #4027, opened 2021-07-27
ENCD-6065 fix post diff table crash

emro:
We needed to account for experiments without a post-differentiation time.
I thought about updating the computeDifferentiation() function instead of the fix below, but the result of computeDifferentiation() is used for the display so I didn't want to change that.

### DIFF
--- a/src/encoded/static/components/dataset.js
+++ b/src/encoded/static/components/dataset.js
@@ -1744,8 +1744,8 @@ const differentiationTableColumnsWithTime = {
             );
         },
         objSorter: (a, b) => {
-            const aTime = computeDifferentiation(a) ? computeDifferentiation(a).split(' ')[0] : 0;
-            const bTime = computeDifferentiation(b) ? computeDifferentiation(b).split(' ')[0] : 0;
+            const aTime = computeDifferentiation(a)?.split(' ')[0] || 0;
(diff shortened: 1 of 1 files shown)


src/encoded/static/components/dataset.js
```diff
@@ -1744,8 +1744,8 @@ const differentiationTableColumnsWithTime = {
             );
         },
         objSorter: (a, b) => {
-            const aTime = computeDifferentiation(a).split(' ')[0];
-            const bTime = computeDifferentiation(b).split(' ')[0];
+            const aTime = computeDifferentiation(a) ? computeDifferentiation(a).split(' ')[0] : 0;
```
zoldello: You could of had done -

```
const aTime = computeDifferentiation(a)?.split(' ')[0] || 0;
```

zoldello (dismissed):
Code look fine. I recommend you DON'T implement the way I suggested you could of had done it; as we are in crunch time. I showed it for future reference.

src/encoded/static/components/dataset.js
emro: Oh, I like that better, it's more concise

zoldello (approved):
Looks fine.

PR #4028, opened 2021-07-28
ENCD-6041-remove-ie11-from-babel

zoldello:


src/encoded/static/libs/compat.js
```diff
@@ -1,56 +1,39 @@
...
-require("regenerator-runtime/runtime");
+/* jshint strict:false */
+require('core-js/stable');
+require('regenerator-runtime/runtime');
 
 // Chrome 42 fetch does not have abort.
 window.fetch = undefined;
 require('whatwg-fetch');
 
-(function () {
```
zoldello, line 9: @forresttanaka Immediately invoked function expression (IIFE) no longer needed in ES2015 and higher. Like see - https://stackoverflow.com/a/33534601/178550

src/encoded/static/components/browserfeat.js
```diff
@@ -55,12 +55,6 @@ module.exports.BrowserFeat = {
                 document.body.removeChild(scrollDiv);
                 return hiddenscroll;
             })();
-
-            // UA checks; should be retired as soon as possible
-            this.feat.uaEdge = (() => navigator.userAgent.indexOf('Edge') > 0)();
-
-            // UA checks; should be retired as soon as possible
-            this.feat.uaTrident = (() => navigator.userAgent.indexOf('Trident') > 0)();
```
forresttanaka, line 63: I guess we could get rid of a lot of these checks (we can guarantee we have canvas, svg, flexbox, etc.), but I’m happy to keep those for now in case there’s some weird case.

src/encoded/static/components/graph.js
```diff
@@ -417,11 +417,6 @@ export class Graph extends React.Component {
                 el = this.dlButton;
                 el.setAttribute('disabled', 'disabled');
             }
```
forresttanaka: Can get rid of the `dlDisabled` state too.

src/encoded/static/libs/compat.js
```diff
@@ -1,56 +1,39 @@
...
+        'table',
+        'assert',
+    ];
+    for (let i = 0, l = consoleMethods.length; i < l; i += 1) {
+        const name = consoleMethods[i];
+        if (window.console[name] === undefined) {
+            window.console[name] = window.console.log;
+        }
     }
 }
```
forresttanaka: Not even sure what any of this does.

forresttanaka (changes requested):
Yeah, looks really good; nice coverage. I’ll request one minor deletion for an unneeded state, and whether you wanted to get more aggressive about browserfeat, I’ll leave up to you.

forresttanaka (approved):

PR #4029, opened 2021-07-28
ENCD-6053-file-view-update

forresttanaka:
A couple of changes aren’t directly related to this ticket:
* The restyling of `SelectionItem` so that it appears green with an “x” on the right end of the buttons, to make them stand out more and be clearer what they do.
* The addition of “File format” in the facets that appear while “Show default data only” enabled.


zoldello (approved):
Code looks good.  I have some UI comment I placed on the ticket. But none are significant.

PR #4031, opened 2021-07-29
ENCD-6066-show-series-experiment-tables-controls

forresttanaka:
Preview demo: ~https://encd-6066-series-controls-pv1-fytanaka.demo.encodedcc.org/~


emro (dismissed):
Looks great!

forresttanaka:
For the second PR, added a new table for the controls portion of the table to show control_type instead of assay. Also, for adding all experiments from the non-controls part of the table, I now only pass experiments without controls, instead of experiments combined with controls.

emro (approved):

PR #4039, opened 2021-08-02
ENCD-6055-FCC-series-search-update

khzlin:


forresttanaka (dismissed):

src/encoded/static/components/search.js
```diff
@@ -705,6 +705,7 @@ const SeriesComponent = ({ context: result, auditDetail, auditIndicators }, reac
     const differentiationSeries = result['@type'].indexOf('DifferentiationSeries') >= 0;
 
     const biosampleTerm = result.biosample_ontology ? result.biosample_ontology[0].term_name : '';
+    const biosampleclassifications = result.biosample_ontology ? result.biosample_ontology[0].classification : '';
```
forresttanaka, line 708: You don’t need to change this in this ticket, but with the deprecation of IE11, we’re now writing lines such as this with this weird-looking shortcut:
```
const biosampleclassifications = result.biosample_ontology?.[0].classification || '';
```
…which means the same as:
```
if (result.biosample_ontology) {
    biosampleclassifications = result.biosample_ontology[0].classification;
} else {
    biosampleclassifications = '';
}
```
This is more common when dereferencing elements of a deep object, so something like: `replicate?.library?.biosample` means “if replicate exists, get its library, and if that exists, get its biosample. If either of those are not the case, return undefined”; whereas `replicate.library.biosample` would simply crash if either replicate or library was undefined.

src/encoded/static/components/search.js
```diff
@@ -867,7 +868,15 @@ const SeriesComponent = ({ context: result, auditDetail, auditIndicators }, reac
...
-                        {organism || lifeSpec.length > 0 ?
+                        {fccSeries && biosampleclassifications !== 'cell line' && (organism || lifeSpec.length > 0) ?
+                            <span>
+                                {' ('}
+                                {organism ? <i>{organism}</i> : null}
+                                {lifeSpec.length > 0 ? <span>{organism ? ', ' : ''}{lifeSpec.join(', ')}</span> : null}
+                                )
+                            </span>
+                        : null}
+                        {!fccSeries && (organism || lifeSpec.length > 0) ?
```
forresttanaka: This and the code below are the same, so you can combine these using this conditional instead of the two conditionals you have:
```
(!fccSeries && (organism || lifeSpec.length > 0)) || (biosampleclassifications !== 'cell line' && (organism || lifeSpec.length > 0))
```

forresttanaka (changes requested):
The first one’s optional, but I think we should reduce the code duplication in the second one — unless my replacement logic doesn't actually do the same thing, but I ran through it with different scenarios and don’t see a difference.

src/encoded/static/components/search.js
khzlin: noted for next time.

src/encoded/static/components/search.js
khzlin: added &&fccSeries to the conditon.

forresttanaka (approved):
🆒

PR #4040, opened 2021-08-02
ENCD-5984-fix-mixed-analysis-title

zoldello:


src/encoded/static/components/filegallery.js
```diff
@@ -200,16 +201,17 @@ const getSeriesFiles = (context) => {
...
     if (analyses && analyses.length > 0) {
         // Get all the analysis objects that qualify for inclusion in the Pipeline facet.
         // More than one lab OK, as long as none of them is `UNIFORM_PIPELINE_LAB` --
         // `UNIFORM_PIPELINE_LAB` is only valid if alone.
         const qualifyingAnalyses = analyses.filter((analysis) => (
             analysis.assembly
-            && analysis.assembly !== 'mixed'
-            && analysis.genome_annotation !== 'mixed'));
+            && ((allowMixed && analysis.assembly === 'mixed') || analysis.assembly !== 'mixed')
+            && ((!allowMixed && analysis.genome_annotation !== 'mixed') || allowMixed)
```
forresttanaka: This gets pretty complicated, but I think this could be replaced with:
```
&& (allowMixed || ((analysis.assembly !== 'mixed') && (analysis.genome_annotation !== 'mixed')))
```
…or less aggressively:
```
&& (allowMixed || analysis.assembly !== 'mixed')
&& (allowMixed || analysis.genome_annotation !== 'mixed')
```
I’ll pass this, as my suggestion is only an optimization, not a logical change. If you decide to use these, make sure the optimizations really _are_ equivalent to what you have.

forresttanaka (dismissed):
Make sure the PR name has “ENCD-5984” in it.

src/encoded/static/components/filegallery.js
zoldello: Great catch. Thanks.

forresttanaka (dismissed):
Cool

src/encoded/static/components/objectutils.js
```diff
@@ -918,9 +945,10 @@ export function filterForVisualizableFiles(fileList) {
         return [];
     }
 
-    return fileList.filter((file) => isFileVisualizable(file));
-}
+    const files = fileList.filter((file) => isFileVisualizable(file));
 
+    return filterForVisualizableAssembly(files);
```
zoldello, line 950: This is a system-wide change that also affects carts. I remove all files from the displaying in Valis if they don't have a specific genome_assembly.

src/encoded/static/components/objectutils.js
```diff
@@ -910,6 +910,33 @@ export const isFileVisualizable = (file) => {
...
+
+    return visualizableAssemblies.includes(assembly);
+}
+
+function filterForVisualizableAssembly(fileList) {
+    if (fileList?.length === 0) {
+        return [];
+    }
+
+    return fileList.filter((file) => isVisualizableAssembly(file.assembly));
```
forresttanaka, line 938: You could, at least conceptually, crash here since — if `fileList` isn’t defined — the first if fails, but now you call `filter()` on an undefined `fileList`. The version of this under `filterForVisualizableFiles` doesn’t have this issue.

forresttanaka (changes requested):
One very minor change to avoid a potential JS crash.

forresttanaka (approved):

PR #4042, opened 2021-08-03
ENCD-6069-add-redacted-transcriptome-alignments

khzlin:


gabdank (changes requested):
although it is addition of "import_only" I think you need at least add an insert for these file output type

gabdank (approved):

PR #4045, opened 2021-08-04
SNO-206-use-redis-cache-for-rnget-view

keenangraham:


keenangraham:
@pedrohr There's a little bit of weirdness here with binding the Redis client to the decorator. The issue is that the decorator and cache are bound at definition time but we need host and port settings from the registry in order to create the Redis client (if we hardcoded the host and port and defined the client inline it would work fine). 

As far as I know the only sort of okay way to access the registry settings outside of the request cycle is when Pyramid calls the `includeme` config functions, but this happens after the decorator and cache have already been defined. 

As a workaround I'm injecting the Redis client into the cache asynchronously. Works okay but don't think it's standard and maybe there's a better way?

The other thing to point out is that we're switching `FieldedResponse` with `FieldedInMemoryResponse` based on whether the results are going to be cached or not. By default `FieldedResponse` will try to stream most results but in order to store results in the cache we need to render them in memory.

cloud-config/run-scripts/ami-redis.sh
```diff
@@ -16,5 +16,9 @@ sudo cp /etc/redis/redis.conf /etc/redis/redis.conf.default
 sudo sed -i -e "s/port 6379/port $ENCD_REDIS_PORT/" /etc/redis/redis.conf
 # Allow remote connections
 sudo sed -i -e 's/bind 127\.0\.0\.1/bind 0\.0\.0\.0/' /etc/redis/redis.conf
+# Set maxmemory
+echo "maxmemory 5GB" | sudo tee -a /etc/redis/redis.conf
```
pedrohr, line 20: I think 5GB is too much... as it's an lru it will fill up the 5GB and take it as RAM until the server is restarted, which only occurs in releases. Idk how much RAM the servers usually have left, but it might be too much. The RNAGet page is still not very popular (except for bots :P) to lock this amount of RAM. What do you think?

src/encoded/search_views.py
```diff
@@ -830,11 +836,20 @@ def rnaget_autocomplete(context, request):
...
+    cache=get_redis_lru_cache(),
+    condition=should_cache_search_results,
+    key=partial(make_key_from_request, 'rnaget-search'),
+)
 def rnaget_search(context, request):
     # Note the order of rendering matters for some fields, e.g. AllResponseField and
     # NotificationResponseField depend on results from BasicSearchWithFacetsResponseField.
     rna_client = request.registry[RNA_CLIENT]
-    fr = FieldedResponse(
+    CachedFieldedResponse = cached_fielded_response_factory(
```
pedrohr, line 880: Just an idea, not a blocker, but would it make sense to move the caching logic that the decorator is performing inside of CachedFieldResponse as it's the only place being used?

src/encoded/searches/caches.py
```diff
@@ -0,0 +1,69 @@
...
+def get_redis_lru_cache():
+    return _redis_lru_cache
+
+
+def includeme(config):
+    # Handle to grab outside of the request cycle
+    # for configuring decorator.
+    global _redis_lru_cache
+    settings = config.registry.settings
+    client = StrictRedis(
```
pedrohr: we should move the client instantiation to the root __init__.py like we do with elasticsearch, it might solve the sync problem with pyramid: https://github.com/ENCODE-DCC/encoded/blob/dev/src/encoded/__init__.py#L246

src/encoded/search_views.py
keenangraham: Yeah definitely a good idea and probably avoids some of the weird decorator <-> cache binding issues. On the plus side now we have a conditional cache decorator that can be used on ~any view, not just FieldedResponse.

pedrohr (approved):

PR #4047, opened 2021-08-04
ENCD-5981-change-filegallery-defaults

emro:
Set default tab on file gallery to be the browser, as long as there is at least one visualizable file.

### DIFF
--- a/src/encoded/static/components/filegallery.js
+++ b/src/encoded/static/components/filegallery.js
@@ -2984,7 +2984,7 @@ class FileGalleryRendererComponent extends React.Component {
             /** Filters for files */
             fileFilters: {},
             /** Current tab: 'browser', 'graph', or 'tables' */
-            currentTab: 'tables',
+            currentTab: 'browser',
             /** Sorted compiled dataset analysis objects filtered by available assemblies */
             compiledAnalyses: compileAnalyses(props.analyses, datasetFiles, 'choose analysis'),
             /** Index of currently/last selected `compiledAnalyses`. */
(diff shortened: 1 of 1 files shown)


src/encoded/static/components/filegallery.js
```diff
@@ -3035,17 +3035,16 @@ class FileGalleryRendererComponent extends React.Component {
...
             this.setState({ selectedFilterValue: '0' });
         }
+
         // Determine how many visualizable files there are
-        // const tempFiles = filterForVisualizableFiles(this.state.files);
+        const vizFiles = filterForVisualizableFiles(this.props.context.files);
         // If the graph is hidden and there are no visualizable files, set default tab to be table and set default assembly to be 'All assemblies'
-        // if (this.props.hideGraph && tempFiles.length < 1) {
-        if (this.props.hideGraph) {
+        if (this.props.hideGraph && vizFiles.length < 1) {
```
zoldello: Why does the length of visualizable files matter if the Genome browser is hidden anyway?

zoldello (dismissed):
My comment is not significant.

src/encoded/static/components/filegallery.js
```diff
@@ -3035,17 +3035,19 @@ class FileGalleryRendererComponent extends React.Component {
...
         }
+
         // Determine how many visualizable files there are
-        // const tempFiles = filterForVisualizableFiles(this.state.files);
-        // If the graph is hidden and there are no visualizable files, set default tab to be table and set default assembly to be 'All assemblies'
-        // if (this.props.hideGraph && tempFiles.length < 1) {
-        if (this.props.hideGraph) {
-            this.setState({ currentTab: 'tables' }, () => {
-                this.filterFiles('All assemblies', 'assembly');
+        const vizFiles = filterForVisualizableFiles(this.props.context.files);
```
forresttanaka, line 3040: @zoldello I feel like this should use `this.props.data` or this test could be inaccurate. The files to render in the file gallery should come from `this.props.data` shouldn’t they?

src/encoded/static/components/filegallery.js
```diff
@@ -3035,17 +3035,19 @@ class FileGalleryRendererComponent extends React.Component {
...
+
+        if (!this.props.hideGraph && vizFiles.length > 0) {
+            this.setState({ currentTab: 'browser' }, () => {
+                // Determine available assemblies
+                const assemblyList = this.setAssemblyList(this.state.files);
```
forresttanaka: This part uses `this.state.files` while the `vizFiles` above uses `this.props.context.files`. Is this correct? I’m unclear when you should use one or the other.

src/encoded/static/components/filegallery.js
zoldello: @forresttanaka  I will take a look.

zoldello: @forresttanaka Let's talk about this on Friday. As things usually go with me... It's complicated. I also need a second opinion on doing this where it is always accurate but awkward to developers and users or it  not being 100% accurate but functional and less messy.

src/encoded/static/components/filegallery.js
```diff
@@ -3039,28 +3039,12 @@ class FileGalleryRendererComponent extends React.Component {
...
-                // We want to get the assembly with the highest assembly number (but not 'All assemblies')
-                const newAssembly = Object.keys(assemblyList).reduce((a, b) => (((assemblyList[a] > assemblyList[b]) && (a !== 'All assemblies')) ? a : b));
-                this.filterFiles(newAssembly, 'assembly');
-            });
+            this.setState({ currentTab: 'graph' });
         } else {
-            this.setState({ currentTab: 'tables' }, () => {
-                this.filterFiles('All assemblies', 'assembly');
-            });
+            this.setState({ currentTab: 'tables' });
```
forresttanaka, line 3047: You really don’t need to set the assembly anymore? I’m very glad if you don’t, because that’s a nice simplification!

forresttanaka (dismissed):

src/encoded/static/components/filegallery.js
zoldello: @forresttanaka It was not doing anything and was reset in other parts. So, I removed it.

zoldello: @keenangraham @forresttanaka Do either of you require any changes to this ticket?

forresttanaka: No.

forresttanaka (approved):
Doesn’t appear to have changed since the last time I reviewed.

PR #4049, opened 2021-08-05
ENCD-6050-loci-fcc-search-list-view

bonitalam:


src/encoded/static/components/search.js
```diff
@@ -405,6 +406,14 @@ const ExperimentComponent = (props, reactContext) => {
         )).map((replicate) => replicate.library.biosample.subcellular_fraction_term_name));
     }
 
+    // Collect examined loci
+    if (result.examined_loci && result.examined_loci.length > 0) {
```
zoldello, line 410: This code is fine. But you could also do - 
```
result.examined_loci?.length > 0
```

src/encoded/static/components/search.js
```diff
@@ -405,6 +406,14 @@ const ExperimentComponent = (props, reactContext) => {
...
+        result.examined_loci.forEach((loci) => {
+            examinedLoci.push(loci.gene.symbol);
+        });
+        examinedLoci = _.uniq(examinedLoci);
```
zoldello, line 414: This is code. But personally, I prefer to use third party libraries only when there is no in-built substitute. You could use Set.

src/encoded/static/components/search.js
```diff
@@ -435,12 +444,27 @@ const ExperimentComponent = (props, reactContext) => {
             <div className="result-item">
                 <div className="result-item__data">
                     <a href={result['@id']} className="result-item__link">
+                        {(isFunctionalExperiment && result.perturbation_type) ?
```
zoldello, line 447: This code is fine. Still, it is best not to have Ternary operator nested in a Ternary operator. There is no rule against it, it just gets a little harder to follow in my personal opinion.

zoldello (approved):
Code looks good, The comments are personal opinions and are not rule violations.

PR #4052, opened 2021-08-06
ENCD-6005-region-search-dserv-genome-browser

pedrohr:
* Using Data Services API for region search
* Adding Genome Browser view to search results

### DIFF
--- a/src/encoded/genomic_data_service.py
+++ b/src/encoded/genomic_data_service.py
@@ -284,19 +284,24 @@ class GenomicDataService():
         return response
 
 
-    def region_search(self, assembly, chromosome, start, end):
+    def region_search(self, assembly, query=None, chrom=None, start=None, end=None, expand_kb=0):
         params = {
             'assembly': assembly,
-            'chr': chromosome.replace('chr', ''),
-            'start': start,
-            'end': end,
+            'query': query,
+            'expand': expand_kb,
             'files_only': True
         }
+
+        if chrom and start and end:
+            params.pop('query')
+            params['chr']   = chrom
+            params['start'] = start
+            params['end']   = end
 
         query_params = '&'.join([f'{k}={params[k]}' for k in params.keys()])
 
         url = f'{self.path}{REGION_SEARCH}?{query_params}'
 
-        results = requests.get(url, timeout=3).json()
+        results = requests.get(url, timeout=10).json()
 
         return results
--- a/src/encoded/genomic_data_service.py
+++ b/src/encoded/genomic_data_service.py
@@ -289,7 +289,8 @@ class GenomicDataService():
             'assembly': assembly,
             'chr': chromosome.replace('chr', ''),
             'start': start,
-            'end': end
+            'end': end,
+            'files_only': True
         }
 
         query_params = '&'.join([f'{k}={params[k]}' for k in params.keys()])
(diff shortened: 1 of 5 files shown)


src/encoded/region_search.py
```diff
@@ -17,27 +19,19 @@
...
-        'title': 'Organism'
-    }),
     ('biosample_ontology.organ_slims', {'title': 'Organ'}),
     ('biosample_ontology.cell_slims', {'title': 'Cell'}),
-    ('assembly', {'title': 'Genome assembly'}),
-    ('files.file_type', {'title': 'Available data'})
+    ('assembly', {'title': 'Genome assembly'}) # necessary for aggregating data
+]
+
+_REMOVE_FACETS = [
```
hitz, line 28: seems odd that these have to be explicitly deleted.

hitz (dismissed):

src/encoded/region_search.py
pedrohr: yes, if we remove them before search, a few aggregations won't work, we have to filter them out after getting the results

src/encoded/region_search.py
```diff
@@ -343,6 +349,11 @@ def region_search(context, request):
             chr=chromosome, start=start, end=end
         )
 
+    if expand_2kb:
+        start = int(start) - 2000
```
hitz: I probably would have made this a parameter in case we want to change it some other number.

hitz (dismissed):

src/encoded/static/components/region_search.js
```diff
@@ -301,6 +302,7 @@ const RegionSearch = (props, context) => {
 
             window.sessionStorage.setItem(GV_COORDINATES_KEY, JSON.stringify(coordinatesObj));
             window.sessionStorage.setItem(GV_COORDINATES_ASSEMBLY, selectedAssembly === 'mm10' ? 'GRCm38' : selectedAssembly);
+            window.sessionStorage.setItem(GV_COORDINATES_ANNOTATION, 'V29');
```
hitz: Do we need to set the ANNOTATION version for hg19 and mm10 as well?  V29 is for GRCh38.  

hg19 -> V19
mm10 -> M21

src/encoded/genomic_data_service.py
```diff
@@ -284,19 +284,24 @@ def rna_get(self):
...
+            params['chr']   = chrom
+            params['start'] = start
+            params['end']   = end
+
         query_params = '&'.join([f'{k}={params[k]}' for k in params.keys()])
 
         url = f'{self.path}{REGION_SEARCH}?{query_params}'
 
-        results = requests.get(url, timeout=3).json()
+        results = requests.get(url, timeout=10).json()
```
hitz, line 305: ugh

hitz (dismissed):
conditionally approved, please check notes

src/encoded/genomic_data_service.py
pedrohr: is this for the high timeout value?

this is because of the ensembl API. We use the ensembl APIs for ENSG ids and as a failover when a symbol is not in our indexes. The ensembl API is incredibly slow and it takes a long time.

src/encoded/static/components/region_search.js
pedrohr: ohh good catch...

I will push the fix

src/encoded/genomic_data_service.py
hitz: ah OK 10s for a fail over is almost useless tho.

hitz (approved):

PR #4053, opened 2021-08-09
ENCD-6084-fix-mouse-development-matrix-crash

forresttanaka:
This matrix filters out “Control ChIP-seq” experiments, but one part of the code still expected to find all received assay types, and crashed when it didn’t find “Control ChIP-seq” from the filtered data.

It would probably be better to simply use the already-filtered data throughout the code instead of filtering twice as I do here, but that seemed like a large change at this stage.

“Control ChIP-seq” data was recently added, which caused this crash.


zoldello (approved):
Code looks good.

PR #4055, opened 2021-08-10
ENCD-6049 pooled clone sequencing dependency

gabdank:


src/encoded/schemas/changelogs/functional_characterization_experiment.md
```diff
@@ -1,5 +1,8 @@
 ## Changelog for functional_characterization_experiment.json
 
+### Schema version 10
+* The pooled clone sequencing assay term name require specification of the control_type
```
iyoungworth: This is a very minor edit, but I would say "The pooled clone sequencing assay term name requires* specification of the control_type"

src/encoded/tests/test_audit_functional_characterization_experiment.py
```diff
@@ -202,9 +202,8 @@ def test_audit_experiment_replicated(
         {'status': 'submitted', 'date_submitted': '2015-03-03'}
     )
     res = testapp.get(pooled_clone_sequencing['@id'] + '@@index-data')
-    assert any(
-        error['category'] == 'unreplicated experiment' and
-        error['level_name'] == 'WARNING'
```
iyoungworth, line 207: The reason it used to be WARNING is from these two lines: https://github.com/ENCODE-DCC/encoded/blob/dev/src/encoded/audit/experiment.py#L1535-#L1536

I know auditing of a PCS should no longer reach that section because it will exit at control_type, but those lines in audit/experiment.py should be removed for completeness

iyoungworth (changes requested):
Everything looks correct and functional, but I would like to see the two lines in audit/experiment.py cleaned up (removed)

src/encoded/schemas/changelogs/functional_characterization_experiment.md
gabdank: fixed

src/encoded/tests/test_audit_functional_characterization_experiment.py
gabdank: removed the lines in experiment audit

iyoungworth (approved):

PR #4057, opened 2021-08-10
ENCD-6076-update-gender-symbols

forresttanaka:


src/encoded/static/components/matrix_brain.js
```diff
@@ -336,7 +336,7 @@ const getDataTableData = (context) => {
         });
 
         const uniqueDiseases = Array.from(availableDiseases);
-        const genderSymbol = genderData === 'male' ? '♂' : '♀';
+        const genderSymbol = genderData === 'male' ? 'mars' : 'venus';
```
zoldello, line 339: "mars" and "venus" may not make sense to people who are familiar with Roman history

zoldello (approved):
Code looks good. My comment is not mandatory to address.

src/encoded/static/components/matrix_brain.js
forresttanaka: I originally called them “male” and “female” until I noticed fontawesome called the corresponding icons “mars” and “venus.” I figured they might have done some research on what to call them (though they certainly might not have).

PR #4070, opened 2021-08-19
encd-6108-controlled-cell-growth-tag

jenjou:


bonitalam (changes requested):
Looks good for the most part. Just missing updates to a few changelogs: dataset, experiment series, gene silencing series, project, publication data, and ucsc browser composite. I think that should be all, but feel free to check for any other objects that are subclasses of dataset. Also from @gabdank latest comment on the ticket it sounds like Mike Pazin wants the tag to be "DeeplyProfiled" instead?

bonitalam (dismissed):

bonitalam (approved):

PR #4072, opened 2021-08-20
ENCD-6077-perturbations

gabdank:


src/encoded/schemas/treatment.json
```diff
@@ -94,7 +94,9 @@
             "enum": [
                 "activation",
                 "deactivation",
-                "stimulation"
+                "stimulation",
```
bonitalam: Can these be added in alphabetical order?

bonitalam (changes requested):
Can you also please add the new enums to the treatment test inserts?

src/encoded/schemas/treatment.json
gabdank: done

bonitalam (approved):

PR #4074, opened 2021-08-24
ENCD-5893-cart-reclassify-non-raw-files-without-assembly-as-processed

forresttanaka:


forresttanaka:
Preview demo at ~https://encd-5893-cart-reclassify-raws-pv2-fytanaka.demo.encodedcc.org/~

src/encoded/static/components/batch_download/cart.js
```diff
@@ -80,8 +80,7 @@ export default class CartBatchDownloadController extends BatchDownloadController
      */
     formatDefaultFileQuery() {
         const query = this.buildBasicQuery();
-        if (this._assembly) {
-            // In rare cases, no selected files include an assembly.
+        if (this._assembly && this._assembly !== 'None') {
```
zoldello: Why not just do -
```
this._assembly !== 'None
```

zoldello (dismissed):
Looks fine.

src/encoded/static/components/batch_download/cart.js
forresttanaka: I want to treat assembly===undefined the same as assembly==='None'.

forresttanaka:
From feedback from Khine and Idan, I changed the way the download menu works on the cart-view page so that assembly===undefined and assembly==='None' get one kind of download menu, while other assemblies get a different download menu.

https://encd-5893-cart-reclassify-raws-pv2-fytanaka.demo.encodedcc.org/

src/encoded/static/components/batch_download/cart.js
```diff
@@ -20,7 +20,7 @@ export default class CartBatchDownloadController extends BatchDownloadController
         this._cartPath = cartPath;
         this._datasetType = datasetType;
         this._assembly = assembly;
-        this._downloadOptionsTemplate = [
+        this._downloadOptionsTemplate = assembly && assembly !== 'None' ? [
```
zoldello: Why not just do -
```
assembly !== 'None'
```

zoldello (dismissed):
Code looks good and see text changes requested in code.

src/encoded/types/file.py
```diff
@@ -842,6 +842,17 @@ def targets(self, request, dataset):
     def analyses(self, request, analyses):
         return paths_filtered_by_status(request, analyses)
 
+    @calculated_property(schema={
```
zoldello, line 845: I think a wranger needs to review this.

src/encoded/schemas/changelogs/file.md
```diff
@@ -1,6 +1,7 @@
 ## Changelog for file.json
 
 ### Minor changes since schema version 29
+* Added *processed* boolean as a calculated property to indicate that the file does not have the *output_category* of *raw data*
```
zoldello, line 4: "... output_category of raw data"? Seems like a diction crisis.

src/encoded/static/components/batch_download/base.js
```diff
@@ -228,7 +228,7 @@ const downloadOptionsTemplate = [
         id: 'raw-files',
         label: 'Download raw files',
         title: 'Raw files',
-        description: 'Downloads all files that don\u2019t have assemblies and without using any selected filters.',
+        description: 'Downloads all raw data files without using any selected filters.',
```
zoldello, line 231: "Downloads unfiltered all data files."?

src/encoded/static/components/batch_download/cart.js
```diff
@@ -20,7 +20,7 @@ export default class CartBatchDownloadController extends BatchDownloadController
         this._cartPath = cartPath;
         this._datasetType = datasetType;
         this._assembly = assembly;
-        this._downloadOptionsTemplate = [
+        this._downloadOptionsTemplate = assembly ? [
```
zoldello: The array after the '?' is long and so involves some digging to find the ':'. You may consider storing the array in a const and referring to it in the Ternary Operator to support KISS principle.

src/encoded/static/components/batch_download/cart.js
```diff
@@ -108,7 +125,11 @@ export default class CartBatchDownloadController extends BatchDownloadController
      */
     formatProcessedQuery() {
         const query = this.buildBasicQuery();
-        query.addKeyValue('files.assembly', this._assembly || '*');
+        if (this._assembly && this._assembly !== 'None') {
```
zoldello: Maybe - 
```
this?._assembly !== 'None'
```

src/encoded/static/components/cart/cart.js
```diff
@@ -115,40 +116,39 @@ const datasetTabTitles = {
...
+const CartBrowser = ({ files, assemblies, pageNumber, loading }) => {
+    if (loading) {
+        return <div className="cart__empty-message">Page currently loading&hellip;</div>;
     }
 
-    // Message for page loading.
-    if (loading) {
-        return <div className="nav result-table cart__empty-message">Page currently loading&hellip;</div>;
+    if (assemblies.length !== 1) {
+        return <div className="cart__empty-message">Select single assembly to view genome browser</div>;
```
zoldello, line 125: Maybe - 

Select **a** single assembly to view genome browser

src/encoded/static/components/cart/facet.js
```diff
@@ -304,31 +302,22 @@ SelectionItem.propTypes = {
...
-                    datasetTerms[term].map((selection) => (
-                        <SelectionItem key={selection} selection={selection} term={term} selectionType="dataset" clickHandler={datasetTermClickHandler} />
-                    ))
-                );
-            }
-            return null;
-        })}
-        {Object.keys(fileTerms).map((term) => {
+            datasetTerms[term].length > 0 && (
+                datasetTerms[term].map((selection) => (
```
zoldello: map function works on empty arrays - It just produces nothing. So I don't see the need to ensure the array is not empty.

src/encoded/static/components/cart/facet.js
```diff
@@ -614,78 +579,68 @@ FacetExpander.defaultProps = {
...
+    return (
+        <div className="facet">
+            {!options.supressTitle
+                ? (
+                    <FacetExpander
+                        title={facet.title}
+                        field={displayedFacetField.field}
+                        labelId={labelId}
+                        expanded={expanded}
+                        expanderEventHandler={displayedFacetField.nonCollapsable || !expanderClickHandler ? null : handleExpanderEvent}
```
zoldello, line 617: Maybe use Nullish coalescing operator - 
```
(displayedFacetField.nonCollapsable || !expanderClickHandler) ?? handleExpanderEvent
```

src/encoded/static/components/cart/facet.js
```diff
@@ -614,78 +579,68 @@ FacetExpander.defaultProps = {
...
+                        selectionCount={selectedFacetTerms.length}
+                        clearFacetSelections={clearFacetSelections}
+                    />
+                )
+            : null}
+            {expanded || displayedFacetField.nonCollapsable ?
```
zoldello, line 623: In my local VS Code, I used tools other than ESLint, to spot code that can be written better.  One tool is sonarlint.  It barks at me when I use nested tertiary operators. The you have an opinion about nested tertiary operators?

zoldello (dismissed):
The code looks fine. Most of my comments are not significant. But there are two I recommend you consider -

- I recommend a wrangler review the python

- I recommend you create a demo and list it. The one I saw in github was from August. So I assume it is invalid.

src/encoded/static/components/batch_download/cart.js
forresttanaka: Yeah, I agree.

src/encoded/static/components/batch_download/cart.js
forresttanaka: Actaully, 'None' isn’t a possible value anymore. Anything that used to have 'None' now has no value, so I removed this part.

src/encoded/schemas/changelogs/file.md
forresttanaka: Seems clear to me.

src/encoded/static/components/cart/cart.js
forresttanaka: Probably, “Select **a** single assembly to view **the** genome browser.” But I don’t think this needs to be a complete grammatical sentence since it’s just a warning.

src/encoded/static/components/cart/facet.js
forresttanaka: I wouldn’t call this a nested tertiary; I’m surprised sonarlint does. To me, a nested tertiary is:
```
a ? b : (c ? d : 0)
```
…which this isn’t — this is:
```
a ? b : 0
```
…where b is an independent expression, which in this case happens to contain a tertiary that’s unrelated to the outer tertiary — it’s only to compose a string within it.

My opinion on the first one (disallowed by AirBnB by default) is that I usually find them confusing, and occasionally find them the most clear, especially with React expressions.

forresttanaka:
The demo you called invalid is up to date, but updated from a long-ago demo (this ticket goes back a long way). I wasn’t going to launch a QA demo until Monday.

forresttanaka:
Updated demo: https://encd-5893-cart-reclassify-raws-pv3-fytanaka.demo.encodedcc.org/

zoldello (approved):
Looks fine.

PR #4076, opened 2021-08-25
ENCD-6109-origin-biosample-property

bonitalam:
- Added "origin_batch" biosample calculated property
- Added "origin_batches" file calculated property

### DIFF
--- a/src/encoded/types/biosample.py
+++ b/src/encoded/types/biosample.py
@@ -468,10 +468,10 @@ class Biosample(Item):
         part_of=None,
     ):
         if biosample_ontology:
-            ontology = request.embed(biosample_ontology, '@@object')
-            if ontology['classification'] in ['cell line', 'in vitro differentiated cells', 'primary cell']:
+            biosample_classification = request.embed(biosample_ontology, '@@object?skip_calculated=true').get('classification')
+            if biosample_classification in ['cell line', 'in vitro differentiated cells', 'primary cell']:
                 if part_of:
-                    part_of_object = request.embed(part_of, '@@object')
+                    part_of_object = request.embed(part_of, '@@object?skip_calculated=true')
                     if 'biosample_ontology' in part_of_object:
                         return is_part_of(request, accession, biosample_ontology, part_of_object)
                 else:
@@ -1243,5 +1243,5 @@ def is_part_of(request, accession, biosample_ontology, part_of_object):
         return part_of_object['accession']
     elif 'part_of' in part_of_object:
         accession = part_of_object['accession']
-        part_of_object = request.embed(part_of_object['part_of'], '@@object')
+        part_of_object = request.embed(part_of_object['part_of'], '@@object?skip_calculated=true')
         return is_part_of(request, accession, biosample_ontology, part_of_object)
--- a/src/encoded/types/biosample.py
+++ b/src/encoded/types/biosample.py
@@ -468,14 +468,12 @@ class Biosample(Item):
         part_of=None,
     ):
         if biosample_ontology:
-            biosample_type = request.embed(biosample_ontology, '@@object').get('classification')
             ontology = request.embed(biosample_ontology, '@@object')
-            if biosample_type in ['cell line', 'in vitro differentiated cells', 'primary cell']:
+            if ontology['classification'] in ['cell line', 'in vitro differentiated cells', 'primary cell']:
                 if part_of:
                     part_of_object = request.embed(part_of, '@@object')
                     if 'biosample_ontology' in part_of_object:
-                        part_of_biosample_ontology = request.embed(part_of_object['biosample_ontology'], '@@object')
-                        return is_part_of(request, accession, ontology, part_of_object, part_of_biosample_ontology)
+                        return is_part_of(request, accession, biosample_ontology, part_of_object)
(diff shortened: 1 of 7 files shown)


src/encoded/types/biosample.py
```diff
@@ -1210,3 +1234,14 @@ def construct_biosample_summary(phrases_dictionarys, sentence_parts):
     rep = dict((re.escape(k), v) for k, v in rep.items())
     pattern = re.compile("|".join(rep.keys()))
     return pattern.sub(lambda m: rep[re.escape(m.group(0))], sentence_to_return)
+
+
+def is_part_of(request, accession, biosample_ontology, part_of_object):
+    if biosample_ontology != part_of_object['biosample_ontology']:
+        return accession
+    elif ('part_of' not in part_of_object) and (biosample_ontology == part_of_object['biosample_ontology']):
```
gabdank: I think the comparison of biosample ontologies here is redundant after previous condition

src/encoded/types/biosample.py
```diff
@@ -1210,3 +1234,14 @@ def construct_biosample_summary(phrases_dictionarys, sentence_parts):
...
+        return part_of_object['accession']
+    elif 'part_of' in part_of_object:
```
gabdank: if you modify the previous condition by removal of ontologies comparison here it would become just `else`

src/encoded/types/biosample.py
```diff
@@ -1210,3 +1234,14 @@ def construct_biosample_summary(phrases_dictionarys, sentence_parts):
...
+        accession = part_of_object['accession']
+        part_of_object = request.embed(part_of_object['part_of'], '@@object?skip_calculated=true')
+        return is_part_of(request, accession, biosample_ontology, part_of_object)
```
gabdank: you can create a loop (bio A is part_of bio B that is part_of bio C that is part_of bio A) and your recursion will cause stack overflow? anyway maybe it is a good idea to stop if you encounter previously visited biosample? I know it will require you to aggregate a list of visited biosamples (or their IDs) but will prevent endless loops

src/encoded/types/biosample.py
```diff
@@ -453,6 +454,29 @@ def subcellular_fraction_term_id(self, request, subcellular_fraction_term_name):
...
+    ):
+        if biosample_ontology:
+            biosample_classification = request.embed(biosample_ontology, '@@object?skip_calculated=true').get('classification')
+            if biosample_classification in ['cell line', 'in vitro differentiated cells', 'primary cell']:
+                if part_of:
+                    part_of_object = request.embed(part_of, '@@object?skip_calculated=true')
+                    if 'biosample_ontology' in part_of_object:
+                        return is_part_of(request, accession, biosample_ontology, part_of_object)
+                else:
+                    return accession
```
gabdank: I think it would be better to stick with the @id instead of accession.

src/encoded/types/file.py
```diff
@@ -456,6 +456,67 @@ def donors(self, request, registry, dataset, root, replicate=None):
...
         return list(set(calculated_donors))
 
+    @calculated_property(schema={
+        "title": "Origin batches",
+        "description": "The origin batch biosample(s) associated with this file.",
+        "comment": "Do not submit. This field is calculated through the derived_from relationship back to the raw data.",
+        "type": "array",
+        "items": {
+            "title": "Origin batch",
+            "description": "Accession of the origin batch biosample",
```
gabdank: if switching to @id, should update the description to say @id instead of accession

src/encoded/types/biosample.py
```diff
@@ -453,6 +454,29 @@ def subcellular_fraction_term_id(self, request, subcellular_fraction_term_name):
         else:
             return 'Term ID unknown'
 
+    @calculated_property(schema={
+        "title": "Origin batch",
+        "description": "The origin batch biosample that the biosample was obtained from.",
```
gabdank: I like better "Biosample @id representing the origin batch the biosample was obtained from"

gabdank (changes requested):

bonitalam:
Addressed code review and confirmed the indentation is correct.

src/encoded/types/biosample.py
```diff
@@ -1210,3 +1235,18 @@ def construct_biosample_summary(phrases_dictionarys, sentence_parts):
...
+    elif 'part_of' not in part_of_object:
+        return part_of_object['@id']
+    else:
+        biosample_id = part_of_object['@id']
+        if biosample_id not in checked_biosamples:
+            checked_biosamples.append(biosample_id)
+            part_of_object = request.embed(part_of_object['part_of'], '@@object')
+            return is_part_of(request, biosample_id, biosample_ontology, part_of_object, checked_biosamples)
+        else:
+            return part_of_object['@id']
```
gabdank: I think here it should be `return biosample_id` rather than `return part_of_object['@id']`

gabdank (changes requested):
return variable rename?

src/encoded/types/biosample.py
bonitalam: Oh yes, thank you will change that now

gabdank (approved):

PR #4078, opened 2021-08-25
ENCD-5687-add-replication-count-to-experiments-and-fcc-experiments

forresttanaka:


src/encoded/types/shared_calculated_properties.py
```diff
@@ -516,6 +516,16 @@ def replication_type(self, request, replicates=None, assay_term_name=None):
         return 'anisogenic'
 
 
+class CalculatedReplicationCount:
+    @calculated_property(schema={
```
keenangraham, line 520: Just to make sure. We want `replication_count=0` in objects even if they don't have `replicates` defined? For example: https://www.encodeproject.org/search/?type=Experiment&replicates!=*

forresttanaka: That was my plan. Now that you mention it, I’m not sure if that’s a _good_ plan, but it was intentional.

src/encoded/types/shared_calculated_properties.py
```diff
@@ -516,6 +516,16 @@ def replication_type(self, request, replicates=None, assay_term_name=None):
...
+        "title": "Replication count",
+        "description": "Calculated field that indicates the number of replicates in a dataset",
+        "type": "integer",
+    })
+    def replication_count(self, replicates=[]):
```
keenangraham: Slight danger of using mutable default argument `replicates=[]`. I think usual idiom for something like this is `replicates=None`, then logic:
```python
if replicates is None:
    return 0 # or whatever you want
return len(replicates)
```

src/encoded/tests/test_types_experiment.py
```diff
@@ -505,6 +505,19 @@ def test_experiment_default_analysis(
 
     assert res.json['object']['default_analysis'] == analysis_2['@id']
 
+def test_experiment_replication_count_0(testapp, base_experiment):
+    res = testapp.get(base_experiment['@id'] + '@@index-data') 
+    assert res.json['object']['replication_count'] == 0
+
```
keenangraham, line 513: Two lines.

src/encoded/tests/test_types_experiment.py
```diff
@@ -505,6 +505,19 @@ def test_experiment_default_analysis(
...
+def test_experiment_replication_count_2(testapp, base_experiment, biosample_1, biosample_2, library_1, library_2, replicate_1_1, replicate_2_1):
+    testapp.patch_json(library_1['@id'], {'biosample': biosample_1['@id']})
+    testapp.patch_json(library_2['@id'], {'biosample': biosample_2['@id']})
+    testapp.patch_json(replicate_1_1['@id'], {'library': library_1['@id']})
+    testapp.patch_json(replicate_2_1['@id'], {'library': library_2['@id']})
+    testapp.patch_json(base_experiment['@id'], {'replicates': [replicate_1_1['@id'], replicate_2_1['@id']]})
+    res = testapp.get(base_experiment['@id'] + '@@index-data') 
+    assert res.json['object']['replication_count'] == 2
+
```
keenangraham, line 523: Two lines.

src/encoded/tests/test_types_functional_characterization_experiment.py
```diff
@@ -64,23 +64,36 @@ def test_fcc_crispr_assay_readout_method(testapp, functional_characterization_ex
     res = testapp.get(functional_characterization_experiment_disruption_screen['@id']+'@@index-data')
     assert 'crispr_screen_readout' not in res.json['object']
 
-def test_fcc_biosample_replicates_1_2(testapp, functional_characterization_experiment_disruption_screen, biosample_1, biosample_2, library_1, library_2, replicate_1_fce, replicate_2_fce, disruption_genetic_modification):
+
```
keenangraham, line 67: Two lines.

keenangraham (dismissed):
Looks good, a couple comments but no blockers.

src/encoded/types/shared_calculated_properties.py
forresttanaka: I see. I’ll make these changes later today and put it up for review again. I saw the previous function had only a one-line gap so I kept to that, but should have looked at more of the file.

keenangraham (dismissed):

src/encoded/types/shared_calculated_properties.py
```diff
@@ -516,6 +516,36 @@ def replication_type(self, request, replicates=None, assay_term_name=None):
         return 'anisogenic'
 
 
+class CalculatedReplicationCounts:
+    def _get_replicates(self, request, replicate_name, replicates=None):
+        replicate_numbers = set()
```
keenangraham: Made a comment in ticket. I think there's an issue with counting technical replicates with a set, since the numbers themselves can repeat within a biological replicate:
Screen Shot 2021-08-27 at 10 44 24 AM

I think the specification wasn't clear on how to actually count these, but the independent total number of each seems like the only thing that makes sense across diverse experiments.

keenangraham (changes requested):

src/encoded/types/shared_calculated_properties.py
forresttanaka: > but the independent total number of each seems like the only thing that makes sense across diverse experiments

I think that’s what this does, or do you mean something different? I mean, for example, the test case with two replicates, one with BR:1 and TR:1, and one with BR:2 and TR:1, it counts two biological and one technical replicate. Or do you mean the TR count should be 2 in this case?

keenangraham: Right, in ticket @hitz said https://www.encodeproject.org/experiments/ENCSR005UON/ should be 3 and 15, but I think it would only be 3 and 11 with the set.

src/encoded/types/shared_calculated_properties.py
hitz: I think replication_count=0 is fine.

src/encoded/types/shared_calculated_properties.py
forresttanaka: Got it, and confirmed that count is 11.

hitz: I think that it should not be a set because 3,1 is totally different from 2,1.  It's not symmetric.

src/encoded/types/shared_calculated_properties.py
```diff
@@ -516,6 +516,38 @@ def replication_type(self, request, replicates=None, assay_term_name=None):
         return 'anisogenic'
 
 
+class CalculatedReplicationCounts:
+    @calculated_property(schema={
+        "title": "Biological replicate count",
+        "description": "Calculated field that indicates the number of biological replicates in a dataset",
+        "type": "integer",
+    })
+    def bio_replicate_count(self, request, replicates):
```
keenangraham, line 525: Probably more of a wrangling question but I wonder if `bio` and `tech` should be fully spelled out to match `biological_replicate` and `technical_replicate`.

src/encoded/types/shared_calculated_properties.py
```diff
@@ -516,6 +516,38 @@ def replication_type(self, request, replicates=None, assay_term_name=None):
...
+        biological_replicate_numbers = set()
+        if replicates:
+            for rep in replicates:
+                replicateObject = request.embed(rep, '@@object')
```
keenangraham, line 529: Probably should be snake case.

keenangraham (dismissed):

src/encoded/types/shared_calculated_properties.py
forresttanaka: Unprompted, I would have. This is the way @hitz spelled it in the ticket, so that’s the only reason why I did it that way.

src/encoded/types/shared_calculated_properties.py
forresttanaka: Yeah, I wondered about that. It’s very common throughout this file to do it that way though, so I thought there was an unspoken rule about it.

forresttanaka:
@keenangraham Thanks for all your help on this.

PR #4083, opened 2021-08-27
ENCD-6096-remove-region-indexer

pedrohr:


paul-sud (approved):
Looks good. A test was failing, but it seems unrelated/spurious, I restarted it in CircleCI just to double check.

PR #4084, opened 2021-08-27
ENCD-6126--add-element-ref-to-file-details

zoldello:


src/encoded/static/components/filegallery.js
```diff
@@ -463,6 +463,18 @@ export class FileTable extends React.Component {
                 return analysisObjectsAccession || nonAnalysisObjectPrefix;
             });
 
+            const elementsReferences = context.elements_references;
+
+            if (elementsReferences) {
+                const elementsReferenceFiles = elementsReferences.map((elementsReference) => elementsReference.files).reduce((acc, val) => acc.concat(val), []);
```
forresttanaka: I see what you’re doing, though I think it’d be easier to understand in one step:
```
const elementsReferenceFiles = elementsReferences.reduce((accFiles, reference) => accFiles.concat(reference.files));
```

src/encoded/static/components/filegallery.js
```diff
@@ -2975,7 +2987,9 @@ class FileGalleryRendererComponent extends React.Component {
         this.experimentType = props.context['@type'][0];
 
         const seriesFiles = getSeriesFiles(props.context) || [];
-        const datasetFiles = [...props.data, ...seriesFiles];
+        const elementsReferenceFiles = props.context.elements_references?.map((elementsReference) => elementsReference.files).reduce((acc, val) => acc.concat(val), []) || [];
```
forresttanaka: Maybe move this into a utility function shared here and in your first change above? They do the same thing, and a well-named utility function would help documentation.

forresttanaka (dismissed):
OK, I think this works, basically adding FCC experiments’ reference files to the array of reference files. I have a couple suggestions in case you feel like taking them.

PR #4086, opened 2021-08-30
Encd 6130 file detail title not showing

zoldello:


src/encoded/static/components/filegallery.js
```diff
@@ -206,10 +206,7 @@ export const compileAnalyses = (analyses, files, dataFormat = null, allowMixed =
         // Get all the analysis objects that qualify for inclusion in the Pipeline facet.
         // More than one lab OK, as long as none of them is `UNIFORM_PIPELINE_LAB` --
         // `UNIFORM_PIPELINE_LAB` is only valid if alone.
-        const qualifyingAnalyses = analyses.filter((analysis) => (
-            analysis.assembly
-            && (allowMixed || (analysis.assembly !== 'mixed' && analysis.genome_annotation !== 'mixed'))
-        ));
+        const qualifyingAnalyses = analyses.filter((analysis) => (allowMixed || (analysis.assembly && (analysis.assembly !== 'mixed' && analysis.genome_annotation !== 'mixed'))));
```
zoldello, line 209: @forresttanaka let's have a discussion about this if you are open to it.

forresttanaka: I know not part of this ticket, but I don’t know when you’d use `allowMixed` and when you wouldn’t; no documentation.

forresttanaka (approved):
Overall I think it makes sense that you simply allow for cases without an assembly to be treated much like those cases that do.

src/encoded/static/components/filegallery.js
zoldello: I am making a new ticket for this. Here is a medal for pointing this out - 🏅

PR #4087, opened 2021-08-30
ENCD-5219-add-deleted-and-inserted-gene-details

bonitalam:


src/encoded/types/biosample.py
```diff
@@ -722,9 +723,13 @@ def summary_objects(request,
...
             if gm_object.get('modified_site_by_target_id'):
-                modification_dict['target'] = request.embed(
-                    gm_object.get('modified_site_by_target_id'),
-                                    '@@object').get('label')
+                target = request.embed(gm_object.get('modified_site_by_target_id'),'@@object')
+                if 'genes' in target:
+                    genes = target['genes']
+                    if len(genes) == 1:
+                        gene_object = request.embed(''.join(str(gene) for gene in genes), '@@object?skip_calculated=true')
+                        modification_dict['target'] = gene_object.get('symbol')
```
gabdank: i think this should be modification_dict['target_gene'] and not 'target' to reduce confusion. Another option that we might encounter is biosample with genetic modification linked to target with no gene? what is the solution there?

bonitalam: Changed to 'target_gene'. I also pushed changes to handle genes that are an empty array or have more than 1 item. For target with no gene and genes that are an empty array, I've implemented the original code of just obtaining the target directly from the the target label property. A test was also added to check that the summary is being generated correctly with a target with no 'genes' associated with it.

gabdank (approved):

PR #4088, opened 2021-08-31
ENCD-6138-fix-mixed-title-not-appearing

zoldello:


forresttanaka (dismissed):
Looks like it works. Would be nice to get some more documentation about the “mixed” concept where you use it; if I were to maintain this code, it would take a lot of tracing to see how it works.

forresttanaka (approved):

PR #4090, opened 2021-08-31
ENCD-6136-gkm-svm-annotation-type-file-output-type

bonitalam:


src/encoded/tests/data/inserts/file.json
```diff
@@ -4755,5 +4755,20 @@
...
         "aliases": ["hongkai-ji:human_elements_scored_file"]
+    },
+    {
+        "_test": "regulatory elements prediction model annotation",
+        "status": "in progress",
+        "uuid": "f4d07aba-8a45-49c9-9656-5379e439c16c",
+        "submitted_by": "facilisi.tristique@potenti.vivamus",
+        "md5sum": "8e5883dd35aaa5bce4721a4e5cc471dd",
+        "accession": "ENCFF318REM",
+        "file_format": "h5ad",
```
gabdank: will it work with TXT file? do we need to update dependencies?

bonitalam: Adding to the dependencies

gabdank (approved):

PR #4091, opened 2021-08-31
ENCD-6113 FCC search updates

gabdank:


src/encoded/schemas/functional_characterization_experiment.json
```diff
@@ -358,7 +358,7 @@
             "length": "long",
             "open_on_load": true
         },
-        "replicates.library.biosample.donor.organism.scientific_name": {
+        "organism.scientific_name": {
```
bonitalam: Do we also want to replace with this in the columns as well? I know this is for making the facets compatible, but in the future not sure if this will make a common report more easily compatible as well?

gabdank: updated columns

gabdank: reverting it back, it breaks search and I think we will handle columns in a separate ticket

src/encoded/types/dataset.py
```diff
@@ -1383,6 +1383,25 @@ class FunctionalCharacterizationSeries(Series):
     def elements_references(self, request, related_datasets):
         return request.select_distinct_values('elements_references', *related_datasets)
 
+    @calculated_property(schema={
```
bonitalam, line 1386: Should these new calculated properties also be added to the facets and columns for Functional Characterization Series?

bonitalam (dismissed):

src/encoded/types/dataset.py
cricketsloan: Idan has now added to the facet.  Adding to columns is bringing up issues with the columns that are larger than this ticket

cricketsloan (approved):
I would like to highlight that there is an unrelated bug fix for CRISPR perturbation_types in this ticket.  I think that it is good and needed, but this fix will be buried under FCC serach updates.

Also, we have scaled this back to simply adding assay_title in functional_characterization_series

src/encoded/types/dataset.py
bonitalam: Sounds good, I see assay_title added. assay_type is not needed as a facet correct? Only needs to exist as a property

bonitalam (approved):

PR #4096, opened 2021-09-09
ENCD-6159-updates-to-reference-data-table

zoldello:


src/encoded/static/components/batch_download/elementsReferences.js
```diff
@@ -0,0 +1,70 @@
...
+
+
+/**
+ * Batch-download controller for downloads in individual sections within a file table of the file
+ * gallery. It only offers two download options.
+ * @param {object} dataset Contains dataset with files to download
+ * @param {object} query QueryString of currently selected file facet terms
+ */
+export default class ElementsReferencesDownloadController extends BatchDownloadController {
+    // eslint-disable-next-line no-useless-constructor
```
forresttanaka: ESLint is correct — you literally don’t need this constructor here at all. The base class’s constructor does what you need.

src/encoded/static/components/batch_download/elementsReferences.js
```diff
@@ -0,0 +1,70 @@
+import BatchDownloadController, { buildDownloadOptionsIndex } from './base';
+
+
+/**
+ * Batch-download controller for downloads in individual sections within a file table of the file
+ * gallery. It only offers two download options.
```
forresttanaka: Looks like only one download option. I have that wrong in my reference file controller too.

forresttanaka (dismissed):
Nicely done. The only thing somewhat significant is the constructor for your new controller that isn’t needed, though it doesn’t hurt functionality either.

forresttanaka (dismissed):

src/encoded/static/components/filegallery.js
```diff
@@ -413,17 +413,9 @@ export class FileTable extends React.Component {
         const selectedAnnotation = null;
 
         const nonAnalysisObjectPrefix = 'Other';
-        const isReferenceDataType = context['@type'].includes('reference');
+        const isReferenceDataType = context['@type'].map((type) => type.toLowerCase()).includes('reference');
```
forresttanaka, line 416: Could this just use `includes('Reference')` without converting to lowercase?

forresttanaka (dismissed):

forresttanaka (dismissed):

forresttanaka:
_Is_ there a real CircleCI error here? I thought it was the usual Circle nonsense, but that’s twice in a row it’s failed on the render test.

forresttanaka (approved):
I believe I previously looked at all but the last commit, so I only reviewed that one.

PR #4097, opened 2021-09-10
ENCD-6149-add-hic-file-output-types

bonitalam:


src/encoded/tests/data/inserts/file.json
```diff
@@ -4770,5 +4770,41 @@
...
+    },
+    {
+        "accession": "ENCFF910GSC",
+        "aliases": [
+            "erez-lieberman:HIC036-genome-subcompartments"
+        ],
+        "assembly": "hg19",
+        "award": "U54HG006998",
+        "dataset": "ENCSR000AJK",
+        "file_format": "bigWig",
```
paul-sud: The SLICE subcompartment outputs are bed files, maybe we could change the format here to be more consistent with what will end up on the portal?
```suggestion
        "file_format": "bed",
        "file_format_type": "bed9",
```

paul-sud (changes requested):
One small change requested, looks good otherwise.

paul-sud (approved):

PR #4098, opened 2021-09-10
ENCD-6161-fix genome browser pinned files in sessionStorage

pedrohr:


src/encoded/static/components/region_search.js
```diff
@@ -324,10 +324,68 @@ const RegionSearch = (props, context) => {
...
+            const interval = coords.split(':')[1];
+
             const coordinatesObj = {
-                contig: coords.split(':')[0],
-                x0: coords.split(':')[1].split('-')[0],
-                x1: coords.split(':')[1].split('-')[1],
+                contig: `chr${contig.split('chr')[1].toUpperCase()}`,
+                x0: interval.split('-')[0],
+                x1: interval.split('-')[1],
+                pinnedFiles: pinnedFilesAssembly[selectedAssembly],
```
zoldello, line 388: `selectedAssembly` is not guaranteed to be a value in `pinnedFilesAssembly` based on line 321.

src/encoded/static/components/genome_browser.js
```diff
@@ -108,9 +108,9 @@ const getDefaultCoordinates = (assembly, annotation, ignoreCache = false) => {
 
     if (gVState && gVAssembly === assembly && gVAnnotation === annotation) {
         const savedState = gVState ? JSON.parse(gVState) : {};
-        ({ contig, x0, x1 } = savedState);
-    }
-    if (assembly === 'GRCh38') {
+        ({ contig } = savedState);
```
zoldello, line 111: Is this a good idea - 

`{x0, x1, pinnedFiles, contig } = savedState);`

src/encoded/static/components/genome_browser.js
```diff
@@ -108,9 +108,9 @@ const getDefaultCoordinates = (assembly, annotation, ignoreCache = false) => {
...
+        ({ x0, x1, pinnedFiles } = savedState);
```
zoldello, line 112: You are not using `pinnedFiles`. It gets overwritten every time.

zoldello (changes requested):
Let's have a conversation after you look at my comments.

src/encoded/static/components/genome_browser.js
zoldello: This is major. The other comments are minor/insignificant.

src/encoded/static/components/region_search.js
pedrohr: the interface only shows `mm10` and `GRCh38`

src/encoded/static/components/genome_browser.js
pedrohr: yes, this is def better. I'm just reverting the file back to what it was, we should add this next time we change this lib

src/encoded/static/components/genome_browser.js
pedrohr: which line are they getting overwritten?

zoldello: Never mind, I get it now.

zoldello (dismissed):
Looks fine. Let's test it.

src/encoded/region_search.py
```diff
@@ -215,7 +215,7 @@ def region_search(context, request):
         return result
 
     expand = True
-    if region and (re.match(r'chr\w:\d+-\d+', region) or re.match(r'rs\d+', region)):
+    if region and (re.match(r'chr\w+:\d+-\d+', region) or re.match(r'rs\d+', region)):
```
zoldello, line 218: Makes sense.

zoldello (approved):
Looks good. The CI test should pass as nothing affecting them changed.

PR #4104, opened 2021-09-16
ENCD-6093-facets-appear-in-groups-on-certain-search-result-pages

forresttanaka:
Preview demo: ~https://encd-6093-facet-groups-pv3-fytanaka.demo.encodedcc.org/~

* The framer-motion animation package adds `<motion.element>` to animate just about every kind of element, so you’ll see `<motion.div>` here. `<AnimatePresence>` lets you animate its child components when they mount and unmount. https://www.framer.com/motion/
* The change in `App.render` returns the ability to keep the page components mounted when navigating to a page with the same path. This lets the facet groups stay open when clicking facet terms. I also need this in another branch I’m working on.
* This last change also broke several pages, as they relied on navigation also causing a full reload. For those cases, I added a data-reload attribute to links, or the reload option for calls to `navigate` to specifically force reloads for those.
* I added a FacetContext React context to allow the `FacetList` component to pass options from callers to whomever wants them down the chain, without prop drilling.
* Added collapsibility to “exists” and “date-range” facets.
* Added a `forceDisplay` property to facet renderers to force them to be visible under all circumstances if they exist in a facet group.
* Removed several search BDD tests as they seemed to test basically the same thing over and over.


emro (commented):
The code looks good! Some comments below. I'm going to continue to look at the code but wanted to submit my code review for now so that you could start looking at it:

- If you visit the demo and select "perturbed", you arrive at the following link which is broken: https://encd-6093-facet-groups-pv3-fytanaka.demo.encodedcc.org/search/?type=Experiment&control_type!=*&status=released&perturbed=false&perturbed=true
- The body map pop-up loads less cleanly and I'm not sure exactly why. You can see more elements on the page flashing as the page reloads on the demo in a way you don't on production as you make selections:
-- https://encd-6093-facet-groups-pv3-fytanaka.demo.encodedcc.org/reference-epigenome-matrix/?type=Experiment&related_series.@type=ReferenceEpigenome&replicates.library.biosample.donor.organism.scientific_name=Homo+sapiens&biosample_ontology.organ_slims=large+intestine&biosample_ontology.organ_slims=colon&biosample_ontology.organ_slims=limb&biosample_ontology.organ_slims=kidney#openModal
-- https://www.encodeproject.org/reference-epigenome-matrix/?type=Experiment&related_series.@type=ReferenceEpigenome&replicates.library.biosample.donor.organism.scientific_name=Homo+sapiens&biosample_ontology.organ_slims=large+intestine&biosample_ontology.organ_slims=colon&biosample_ontology.organ_slims=limb&biosample_ontology.organ_slims=kidney#openModal
- It looks like there might be a slight spacing adjustment needed on the body map selections buttons also:
Screen Shot 2021-09-20 at 1 26 29 PM

- It appears that the de-selection buttons, "X large intestine" on the body map for instance, do not work at all. The modal to select an organism pops up and the page doesn't even reload properly once you make a selection. I think this is a component in another file, I think "matrix_reference_epigenomes.js" which needs to be updated.

forresttanaka:
I modified some pages to fully reload the page when they navigate, but that seems to cause that unclean update. But full page reloads on every navigation kills grouped facets and makes the series cart modal completely unusable.

What I’m going to try instead is to add a special case to `<App>` to force the full page reload at that level for certain paths, e.g. /reference-epigenome-matrix/. That’s super ugly, but as I haven’t yet looked into why a small subset of pages require full-page reloads when navigating, I think that might prove best (assuming that works).

src/encoded/static/components/facets/defaults.js
```diff
@@ -174,20 +187,47 @@ export const DefaultExistsBinaryFacet = ({ facet, relevantFilters, queryString,
...
+                    {isExpandable && <div className="facet-chevron">{svgIcon(isExpanded ? 'chevronUp' : 'chevronDown')}</div>}
+                </button>
+                <AnimatePresence>
+                    {isExpanded && (
+                        <motion.div
+                            className="facet-content"
+                            initial="close"
+                            animate="open"
+                            exit="close"
+                            transition={{ duration: 0.2, ease: 'easeInOut' }}
```
emro: I think maybe this 0.2 duration should be set as a variable at the top.

src/encoded/static/components/facets/facet_groups.js
```diff
@@ -0,0 +1,212 @@
...
+            <FacetGroupTitle group={group} hasSelections={hasSelections} isExpanded={isExpanded} isNameDisplayed={isNameDisplayed} expanderHandler={expanderHandler} />
+            <AnimatePresence>
+                {children &&
+                    <motion.div
+                        id={generateFacetGroupIdentifier(group)}
+                        className="facet-group__content"
+                        initial="collapsed"
+                        animate="open"
+                        exit="collapsed"
+                        transition={{ duration: 0.2, ease: 'easeInOut' }}
```
emro: This could import from defaults.js. Just in case we ever want to tweak the transition timing.

emro (commented):
The code looks good! There's a lot going on.

src/encoded/static/components/facets/defaults.js
forresttanaka: Oh, that's a good idea. I moved the whole object to the top so the ease can be tweaked too.

forresttanaka:
I removed the links that force a reload as well as the calls to context.navigate forcing a reload. I instead specifically look for paths that require a reload in `<App>`. I copied the styles from the facet selection buttons to the reference epigenome selection buttons. I also made a new standard transition object for the animations.

emro (approved):
Awesome!! The code looks great! Thank you for making those changes, everything works super well, it's very slick and well-organized!

PR #4106, opened 2021-09-17
ENCD-6144 tracks batch

gabdank:


src/encoded/static/components/genome_browser.js
```diff
@@ -597,6 +599,8 @@ const TrackLabel = ({ file, label, supplementalShortLabel, long }) => {
         ]).join(', ');
     }
 
+    const originBatchesList = file.origin_batches ? `batch${file.origin_batches.length > 1 ? 'es' : ''}: ${file.origin_batches.map((batch) => batch.split('/')[2]).join(', ')}` : '';
```
zoldello, line 602: This code is fine. Some feel though it makes code unnecessarily to have nested ternary operators.

zoldello (dismissed):
Code looks good.

src/encoded/static/components/genome_browser.js
gabdank: thanks.
I am not an expert in JS, but a one-liner looked better than 3 lines.

zoldello (approved):
Code looks good.

src/encoded/static/components/genome_browser.js
hitz: For future considerations I think 3 lines is more comprehensible unless you break it up with <returns>

PR #4109, opened 2021-09-20
ENCD-6134-add-elements-selection-method-enums

khzlin:


src/encoded/tests/test_upgrade_dataset.py
```diff
@@ -514,3 +514,8 @@ def test_upgrade_dataset_31_to_32(upgrader, experiment_35):
     value = upgrader.upgrade('experiment', experiment_35, current_version='35', target_version='36')
     assert value['schema_version'] == '36'
     assert value['assay_term_name'] == 'capture Hi-C'
+
+def test_upgrade_reference_20_to_21(upgrader, upgrade_20_21_reference):
+    value = upgrader.upgrade('reference', upgrade_20_21_reference, current_version='20', target_version='21')
+    assert value['schema_version'] == '21'
+    assert value['elements_selection_method'] == ['sequence variants', 'DNase hypersensitive sites']
```
bonitalam: Should add a newline to end of the file

src/encoded/upgrade/dataset.py
```diff
@@ -731,3 +731,49 @@ def experiment_35_36(value, system):
...
         value['assay_term_name'] = 'capture Hi-C'
+
+
+@upgrade_step('reference', '20', '21')
+def reference_20_21(value, system):
+    # https://encodedcc.atlassian.net/browse/ENCD-6134
+    new_elements_selection_method = []
+    if 'elements_selection_method' in value:
+        for element_selection_method in value['elements_selection_method']:
+            if re.match('ATAC-seq', element_selection_method):
```
bonitalam: Is there a reason you are using re.match? I think you can just simply test like so `if element_selection_method ==  'ATAC-seq':`

src/encoded/upgrade/dataset.py
```diff
@@ -731,3 +731,49 @@ def experiment_35_36(value, system):
...
+                new_elements_selection_method.append('accessible genome regions')
+                if 'notes' in value:
```
bonitalam: I think you need to define notes outside of the for loop like your new elements selection method array, because there could be multiple new notes added. Then at the end you only have to check if there is already a notes existing in the value once.

src/encoded/upgrade/dataset.py
```diff
@@ -731,3 +731,49 @@ def experiment_35_36(value, system):
...
+                    value['notes'] = f'{value.get("notes")}. This elements_selection_method was previously DNase-seq, but now renamed to be DNase hypersensitive sites.'
+                else:
+                    value['notes'] = 'This elements_selection_method was previously DNase-seq, but now renamed to be DNase hypersensitive sites.'
+            elif re.match('GRO-cap', element_selection_method):
+                new_elements_selection_method.append('transcription start sites')
+                if 'notes' in value:
+                    value['notes'] = f'{value.get("notes")}. This elements_selection_method was previously GRO-cap, but now renamed to be transcription start sites.'
+                else:
+                    value['notes'] = 'This elements_selection_method was previously GRO-cap, but now renamed to be transcription start sites.'
+            elif re.match('point mutations', element_selection_method):
```
bonitalam: You can combine checking for point mutations and SNPs since they are upgraded to the same thing

src/encoded/upgrade/dataset.py
```diff
@@ -731,3 +731,49 @@ def experiment_35_36(value, system):
...
+                    value['notes'] = f'{value.get("notes")}. This elements_selection_method was previously single nucleotide polymorphisms, but now renamed to be sequence variants.'
+                else:
+                    value['notes'] = 'This elements_selection_method was previously single nucleotide polymorphisms, but now renamed to be sequence variants.'
+            elif re.match('transcription factors', element_selection_method):
+                new_elements_selection_method.append('TF binding sites')
+                if 'notes' in value:
+                    value['notes'] = f'{value.get("notes")}. This elements_selection_method was previously transcription factors, but now renamed to be TF binding sites.'
+                else:
+                    value['notes'] = 'This elements_selection_method was previously transcription factors, but now renamed to be TF binding sites.'
+    value['elements_selection_method'] = new_elements_selection_method
```
bonitalam: Right now you are say that it should be upgraded even if the new elements selection method array is empty with no items. You should check for the length of the array.

src/encoded/upgrade/dataset.py
```diff
@@ -731,3 +731,49 @@ def experiment_35_36(value, system):
...
+    return
```
bonitalam: Should add newline to end of the file

bonitalam (changes requested):
In addition to the specific comments for the upgrade code, your upgrade needs to consider cases where there is a combination of enums in `elements_selection_method` where there are some that need to be upgraded and others that remain the same.

src/encoded/schemas/changelogs/reference.md
```diff
@@ -1,5 +1,10 @@
 ## Changelog for reference.json
 
+### Schema version 21
+* Renamed the *ATAC-seq*, *DNase-seq*, *transcription factors*, and *GRO-cap* enum of *elements_selection_method* to be *accessible genome regions*, *DNAse hypersensitive sites*, *TF binding sites* and *transcription start sites* respectively.
```
bonitalam: can *DNAse hypersensitive sites* be changed to *DNase hypersensitive sites*?

bonitalam (approved):

PR #4111, opened 2021-09-20
ENCD-6154-collection-series

bonitalam:


bonitalam:
Requesting @forresttanaka review as well for the JS changes

forresttanaka (dismissed):
The changes look very good. Once this goes into dev, I’ll add the cart toggle to the CollectionSeries pages and search results. Leaving the ticket in Code Review Ready for @gabdank.

src/encoded/schemas/replication_timing_series.json
```diff
@@ -44,29 +44,6 @@
...
-                        "properties": {
-                            "status": {
-                                "enum": ["released", "revoked"]
-                            }
-                        }
-                    }
-                }
-            ]
-        }
-    },
```
gabdank, line 69: why is this taken out?

gabdank: oghh it was duplicated...

gabdank (dismissed):
I think you have deleted dependency in

gabdank (dismissed):

src/encoded/static/components/dataset.js
```diff
@@ -1903,9 +1903,15 @@ function sortStage(a, b) {
...
-    if (a.replicates[0].library.biosample.age_display && b.replicates[0].library.biosample.age_display) {
+    if (a.replicates && b.replicates && a.replicates[0].library.biosample.age_display && b.replicates[0].library.biosample.age_display) {
         return (a.replicates[0].library.biosample.age_display.split(' ')[0] - b.replicates[0].library.biosample.age_display.split(' ')[0]);
     }
+    if (a.replicates && a.replicates[0].library.biosample.age_display) {
+        return -a.replicates[0].library.biosample.age_display.split(' ')[0];
+    }
+    if (b.replicates && b.replicates[0].library.biosample.age_display) {
+        return b.replicates[0].library.biosample.age_display.split(' ')[0];
+    }
```
forresttanaka, line 1914: For these you can also do the very weird-looking `a.replicates?.[0].library.biosample.age_display` to reduce the &&.

Not all replicates have libraries, and not all libraries have biosamples. Do you also need to check for their existence before dereferencing them, or are those already guaranteed by the time you get to this code?

src/encoded/static/components/dataset.js
```diff
@@ -3236,13 +3236,13 @@ const TreatmentConcentrationSeries = ({ context }, reactContext) => {
...
         // Collect any biosamples found in all related datasets.
-        const biosamples = relatedDataset.replicates.reduce((accBiosamples, replicate) => {
+        const biosamples = relatedDataset.replicates?.reduce((accBiosamples, replicate) => {
             const biosample = replicate.library && replicate.library.biosample;
             return biosample ? accBiosamples.concat(biosample) : accBiosamples;
         }, []);
 
         // Collect durations in the biosample treatments and compose them into displayable strings.
-        const collectedDurations = biosamples.reduce((accCollectedDurations, biosample) => {
+        const collectedDurations = biosamples?.reduce((accCollectedDurations, biosample) => {
```
forresttanaka, line 3251: If `biosamples` is undefined then `collectedDurations` becomes undefined here, which would cause the return value from this function to be an array containing an undefined entry. I think this would fix that:
```
return [...new Set(collectedDurations ? accTreatmentDurations.concat(collectedDurations) : accTreatmentDurations))];
```

forresttanaka (changes requested):
I think that `treatmentDurations` loop presents a slight danger of displaying “undefined” in the resulting display.

src/encoded/static/components/dataset.js
bonitalam: I'm not sure about this, maybe @emro would have a better idea. However, I did try checking for library and biosample on my branch and it somehow ends up breaking the UI again. e.g. `b.replicates && b.replicates[0].library?.biosample?.age_display`. Perhaps I did the check incorrectly though.

src/encoded/static/components/dataset.js
bonitalam: Will implement this! Thanks @forresttanaka

forresttanaka (dismissed):
I doubt this helps with the problem @khzlin found. That’s still a mystery to be discovered.

gabdank (approved):

forresttanaka (approved):

PR #4113, opened 2021-09-23
ENCD-6166-Filtering out unnecessary Region Search fields

pedrohr:


src/encoded/region_search.py
```diff
@@ -711,6 +725,19 @@ def format_results(request, hits, result=None):
...
+                filtered_file = {}
+                for field in REGION_SEARCH_FILE_FIELDS:
+                    if field in file_:
+                        filtered_file[field] = file_[field]
+                filtered_files.append(filtered_file)
+            item['files'] = filtered_files
+
+            for field in REGION_SEARCH_UNNECESSARY_FIELDS:
+                item.pop(field)
+
```
pedrohr, line 741: Pulling specific fields from the `embedded` doc from Elasticsearch turns out to be slower than just getting everything and filtering out in the backend.

hitz (dismissed):

hitz (approved):

PR #4115, opened 2021-09-27
ENCD-6141-update-fc-title-search

khzlin:


src/encoded/static/components/search.js
```diff
@@ -393,17 +396,30 @@ const ExperimentComponent = (props, reactContext) => {
...
+                    });
+                }
+                if (er.elements_selection_method && er.elements_selection_method.length > 0) {
+                    elementsSelectionMethod.push(...er.elements_selection_method);
+                }
             });
-            tilingModality = _.uniq(tilingModality);
         }
+        tilingModality = _.uniq(tilingModality);
+        referenceLoci = _.uniq(referenceLoci);
```
forresttanaka: `uniq` here only detects if multiple of these array entries point to the same loci object in memory, not that you have two separate copies of the exact same object, which search results would give you. You can give `uniq` an iteratee function to tell it what to compare, like:
```
referenceLoci = _.uniq(referenceLoci, (locus) => locus['@id']);
```
That way, instead of comparing their memory references (which won’t ever match), `uniq` compares their @ids which will match even if you have two separate copies of the same locus. You don’t need these for the other two `_.uniq()` calls because those compare strings which work correctly.

src/encoded/static/components/search.js
```diff
@@ -393,17 +396,30 @@ const ExperimentComponent = (props, reactContext) => {
...
         if (result.elements_references && result.elements_references.length > 0) {
             result.elements_references.forEach((er) => {
                 if (er.crispr_screen_tiling) {
                     tilingModality.push(er.crispr_screen_tiling);
                 }
+                if (er.examined_loci && er.examined_loci.length > 0) {
+                    er.examined_loci.forEach((gene) => {
+                        referenceLoci.push(gene);
+                    });
+                }
```
forresttanaka:
```
referenceLoci = referenceLoci.concat(er.examined_loci || []);
```
…so that referenceLoci references the `examined_loci` array, or [] if `examined_loci` doesn’t exist.

You could actually put the entire `if (result.elements_references…` into one loop, but that would look fairly complex. Might as well stick with this.

src/encoded/static/components/search.js
```diff
@@ -393,17 +396,30 @@ const ExperimentComponent = (props, reactContext) => {
...
     // Collect crispr_screen_tiling for FunctionalCharacterizationExperiment
     let tilingModality = [];
+    let elementsSelectionMethod = [];
+    let referenceLoci = [];
     if (isFunctionalExperiment) {
         if (result.elements_references && result.elements_references.length > 0) {
             result.elements_references.forEach((er) => {
                 if (er.crispr_screen_tiling) {
                     tilingModality.push(er.crispr_screen_tiling);
                 }
```
forresttanaka: You can replace this `if` with:
```
tilingModality = tilingModality.concat(er.crispr_screen_tiling || []);
```
For one line instead of three, and also below:

src/encoded/static/components/search.js
```diff
@@ -393,17 +396,30 @@ const ExperimentComponent = (props, reactContext) => {
...
+                if (er.examined_loci && er.examined_loci.length > 0) {
+                    er.examined_loci.forEach((gene) => {
+                        referenceLoci.push(gene);
+                    });
+                }
+                if (er.elements_selection_method && er.elements_selection_method.length > 0) {
+                    elementsSelectionMethod.push(...er.elements_selection_method);
+                }
```
forresttanaka: …and a similar thing here:
```
elementsSelectionMethod = elementsSelectionMethod.push(er.elements_selection_method);
```
Though `elements_selection_method` is an array, `[].concat()` breaks it open to individual items before adding them to the given array. `[].push()` does not break given arrays open, and so you would need the spread operator (“...”) for that.

src/encoded/schemas/reference.json
```diff
@@ -287,6 +287,14 @@
             "title": "Project",
             "type": "string"
         },
+        "elements_selection_method": {
+            "title": "Elements selection method",
+            "type": "array"
+        },
+        "examined_loci.gene.symbol": {
```
forresttanaka: I think this is just `examined_loci.symbol`, and it would be of type “string” (`examined_loci` is an array, but `examined_loci[x].symbol` is a string). Actually, I’m not sure why this `columns` has types when most (but not all) other schema’s `columns` properties have no type.

src/encoded/schemas/functional_characterization_experiment.json
```diff
@@ -448,6 +448,12 @@
         "elements_references.crispr_screen_tiling": {
             "title": "CRISPR screen tiling"
         },
+        "elements_references.elements_selection_method": {
+            "title": "Elements selection method"
+        },
+        "elements_references.examined_loci": {
+            "title": "Reference loci"
+        },
```
forresttanaka, line 459: You’re bringing in an entire elements_references array objects here but looks like you only use `symbol` and maybe `@id`. To reduce the size of the search, I would just specify `elements_references.examined_loci.symbol` and a separate `elements_references.examined_loci.@id`.

forresttanaka (changes requested):
I think this looks good. I have a few related optimizations, and a couple corrections.

src/encoded/schemas/reference.json
khzlin: ah thanks. yep, I think it should be examined_loci.symbol instead of examined_loci.gene.symbol.

src/encoded/schemas/functional_characterization_experiment.json
khzlin: ok

src/encoded/static/components/search.js
khzlin: ok

src/encoded/static/components/search.js
khzlin: talked to forrest and went with:
`referenceLoci = er.examined_loci ? referenceLoci.concat(er.examined_loci) : referenceLoci;`

src/encoded/static/components/search.js
khzlin: talked to forrest and went with:
`tilingModality = er.crispr_screen_tiling ? tilingModality.concat(er.crispr_screen_tiling) : tilingModality;`

src/encoded/static/components/search.js
khzlin: talked to forrest and went with:
`elementsSelectionMethod = er.elements_selection_method ? elementsSelectionMethod.concat(er.elements_selection_method) : elementsSelectionMethod;`

src/encoded/static/components/search.js
```diff
@@ -509,16 +521,39 @@ const ExperimentComponent = (props, reactContext) => {
                                 {synchronizations && synchronizations.length > 0 ?
                                     <div><span className="result-item__property-title">Synchronization timepoint: </span>{synchronizations.join(', ')}</div>
                                 : null}
-
+                                {elementsSelectionMethod && elementsSelectionMethod.length > 0 ?
+                                    <div><span className="result-item__property-title">Elements selection method: </span>{elementsSelectionMethod.join(', ')}</div>
+                                : null}
+                                {referenceLoci && referenceLoci.length > 0 ?
```
khzlin, line 527: To be consistent with other places in the repo, I changed this part to not include commas in the links.

forresttanaka (approved):

PR #4117, opened 2021-09-28
ENCD-6148 exclude fcc datapoints

gabdank:


src/encoded/tests/test_types_functional_characterization_experiment.py
```diff
@@ -75,3 +75,17 @@ def test_fcc_replication_count_2(testapp, functional_characterization_experiment
     testapp.patch_json(replicate_2_fce['@id'], {'library': library_2['@id']})
     res = testapp.get(functional_characterization_experiment_disruption_screen['@id'] + '@@index-data') 
     assert res.json['object']['bio_replicate_count'] == 2 and res.json['object']['tech_replicate_count'] == 2
+
+
+def test_fcc_datapoint(testapp, fcc_posted_CRISPR_screen, base_functional_characterization_series):
+    res = testapp.get(base_functional_characterization_series['@id'] + '@@index-data') 
+    print (res.json['object'])
+    print ("sss")
```
cricketsloan: These print statements don't look like they belong

src/encoded/types/functional_characterization_experiment.py
```diff
@@ -215,3 +216,21 @@ def crispr_screen_readout(self, request, assay_term_name, examined_loci=None, co
...
+        if not related_series:
+            return False
+        for series in related_series:
+            properties = {'series': series}
+            path = Path('series', include=["@type"])
+            path.expand(request, properties)
+            types = properties.get('series', {}).get('@type', [])
+            if "FunctionalCharacterizationSeries" in types:
+                return True 
+        return False
```
cricketsloan: Side note:  there is a EOF character on the final line, this would not pass most linters

cricketsloan: Question 1:  Are we going to attempt to exclude controls?  I think we shouldn't for now, but we may end up getting pushback.  

Question 2: Since we are going to likely add different types of series into this list, should we consider making it a for loop of 1 to prepare for future additions?

cricketsloan (commented):
Generally, this looks right.  I wonder if the style we have been using would make this property "is_datapoint" or if "datapoint" works fine.  There are a few other inline questions.

src/encoded/tests/test_types_functional_characterization_experiment.py
gabdank: fixed

src/encoded/types/functional_characterization_experiment.py
gabdank: A1. controls are excluded in the URL (explicitly control_type!=*) and if they on top of that belongs to series - they will be datapoints
A2. I think we will make it a loop when we have more series. I don't think we should write the code in loops just to support potential future development

gabdank: and I added an empty line 237

gabdank: as for the "datapoint vs is_datapoint" I have no strong feelings about it. I don't think we have been consistent about it: file.restricted vs biosample.is_part_of but at the same time biosample.perturbed

I think fcc.datapoint will work fine for now

cricketsloan (dismissed):

cricketsloan (dismissed):

src/encoded/types/dataset.py
```diff
@@ -1402,6 +1402,16 @@ def assay_slims(self, request, related_datasets):
     })
     def assay_title(self, request, related_datasets):
         return request.select_distinct_values('assay_title', *related_datasets)
+    
+    @calculated_property(schema={
+        "title": "Datapoint",
+        "description": "A flag to indicate whether the FC Series is a datapoint that should not be displayed on it's own.",
```
iyoungworth, line 1407: Very minor typo, not worth fixing if there aren't other code changes- should be "its" instead of "it's"

iyoungworth (dismissed):
1 minor typo in a description, no other issues I can see. I don't think it's serious enough to go back and fix.

iyoungworth (dismissed):
Code passes. Should retry CircleCI again tomorrow, or possibly run them locally.

hitz (approved):

PR #4122, opened 2021-10-05
ENCD-6157-allow-multiple-types-for-report-view-button

forresttanaka:
* The view control buttons now have a very simple plug-in mechanism that uses a new entry in `viewTypeMap` to determine which component should render each of the three types of view control buttons (search, report, matrix).
* Each of the three types of buttons have their own renderers now so they can easily have more custom actions without cramming them into a single component and doing type checking.

~https://encd-6157-mixed-reports-pv1-fytanaka.demo.encodedcc.org/~


src/encoded/static/components/__tests__/search-controls.js
```diff
@@ -37,4 +38,43 @@ describe('View controls for search with Experiments', () => {
             expect(buttons.at(2).text()).toEqual('{ ; }');
         });
     });
+
+    describe('Mutliple searched types', () => {
```
emro: typo: "Multiple"

src/encoded/static/scss/encoded/modules/_search.scss
```diff
@@ -370,6 +370,40 @@ $form-bg: #f3f3f3;
...
+        @extend .btn-info;
+        @extend .btn-sm;
+    }
+
+    .dropdown-button__content {
+        width: auto;
+        white-space: nowrap;
+
+        a {
+            font-family: Quicksand, sans-serif;
```
emro: I think you could use $button-font here, just in case we one day wanted to change all the instances of Quicksand - unlikely now but theoretically possible.

src/encoded/static/components/view_controls.js
```diff
@@ -1,24 +1,160 @@
...
+const ReportViewButton = ({ viewType, query }, reactContext) => {
+    const typeValues = query.getKeyValues('type');
+    if (typeValues.length === 1) {
+        // One "type=" in the query string so render a regular link to the report page.
+        return (
+            <ViewControlButton viewType={viewType} queryString={query.format()}>
+                {viewType.title}
+            </ViewControlButton>
+        );
+    }
```
emro, line 66: It looks like you could use the `<SearchViewButton>` here in lines 66-68 but I think you didn't do that because the code is easier to understand if you have one component for each type of view button. You could rename the SearchViewButton to be something more generic and use it here, but I also think it's fine as-is, it was just a thought I had.

emro (commented):
Looks good! Just a couple comments, but no changes needed.

src/encoded/static/components/view_controls.js
forresttanaka: Oh, that's a good idea.

forresttanaka:
I made your suggested modifications, and then noticed a one-pixel issue when the last (or only) button within .btn-attached has a dropdown because of the added `<ul>`, so I changed the :last-child into :last-of-type to target just the button style.

forresttanaka:
Updated demo at: ~https://encd-6157-mixed-reports-pv2-fytanaka.demo.encodedcc.org/~

emro (approved):
Looks good!

PR #4123, opened 2021-10-05
ENCD-6180-matrix-search

forresttanaka:
These search boxes have a `<div className="facet">` that seems to serve no purpose — no styles for the search box depend on being a child of `.facet`. For the grouped-facets change, .facet now has padding that offset these search boxes, so I removed the wrapper div from these search box input elements.

~https://encd-6180-matrix-search-pv1-fytanaka.demo.encodedcc.org/~

### DIFF
--- a/src/encoded/static/components/body_map.js
+++ b/src/encoded/static/components/body_map.js
@@ -1019,7 +1019,7 @@ export const BodyMapThumbnailAndModal = (props) => {
     };
 
     return (
-        <div className="facet body-map-thumbnail-and-modal">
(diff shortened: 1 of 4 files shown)


emro (dismissed):
Looks good!

forresttanaka:
Update: https://encd-6180-matrix-search-pv2-fytanaka.demo.encodedcc.org/

Adjusted body map position on reference epigenome page to its original position instead of with padding.

emro (approved):

PR #4124, opened 2021-10-05
ENCD-5869-add-missing-control_type-in-url

zoldello:


forresttanaka (dismissed):
Looks good. On the Human Body Map page, clicking “Clear all selections” clears “control_type!=*” from the query string. I _think_ that’s the desired behavior, but wanted to call that out in case it isn’t.

zoldello:
@forresttanaka  Good catch and I fixed it.

> Looks good. On the Human Body Map page, clicking “Clear all selections” clears “control_type!=*” from the query string. I _think_ that’s the desired behavior, but wanted to call that out in case it isn’t.

Good catch and I fixed it.

forresttanaka (approved):

PR #4128, opened 2021-10-08
ENCD-6190-html-editor-crash

forresttanaka:


src/encoded/static/libs/ui/modal.js
```diff
@@ -206,7 +206,7 @@ export const ModalFooter = ({
             submitBtn();
         }
         if (!dontClose) {
-            internalCloseModal();
+            internalCloseModal(e);
```
zoldello: e need to be set but not set in other places, so it can still crash. Does this ticket need to cover these too?

zoldello (dismissed):
Looks good. I searched for "internalCloseModal" and saw other places where e is not set, though.

src/encoded/static/libs/ui/modal.js
forresttanaka: Yeah, you’re absolutely right. I decided to reverse adding that stopPropagation call, as I didn’t actually need it anymore because I changed the way I was using modals from the cart page. I had kept the stopPropagation call, thinking it was a general enhancement. But I need to think it through more if we need it in the future.

zoldello (approved):

PR #4129, opened 2021-10-08
ENCD-6145-more-facet-headers

jenjou:


src/encoded/schemas/annotation.json
```diff
@@ -271,6 +271,62 @@
             "linkTo": "Experiment"
         }
     },
+    "facet_groups": [
+        {
+            "title": "Annotation",
+            "facet_fields": [
```
gabdank, line 277: I think we need to include here experimental_input

gabdank: and biochemical inputs

src/encoded/schemas/biosample.json
```diff
@@ -922,6 +922,66 @@
             ]
         }
     },
+    "facet_groups": [
+        {
+            "title": "Biosample",
+            "facet_fields": [
+                "biosample_ontology.classification",
+                "biosample_ontology.organ_slims",
```
gabdank, line 930: I think we need to include age_display

src/encoded/schemas/biosample.json
```diff
@@ -922,6 +922,66 @@
...
+        {
+            "title": "Dataset",
+            "facet_fields": [
+                "assay_title",
+                "annotation_type",
+                "annotation_subtype",
+                "platform.term_name",
+                "target.label",
+                "targets.label",
+                "encyclopedia_version"
```
gabdank: I am confused. why is it here?

jenjou: I think your review is highlighting multiple lines, can you clarify which facet(s) you are referring to that should be moved? Is it "encyclopedia_version"?

src/encoded/schemas/annotation.json
gabdank: do not add experimental inputs, it is not a good idea

gabdank (dismissed):

jenjou:
No code change, just rebased on the hotfix

gabdank (approved):

PR #4133, opened 2021-10-12
ENCD-6117-cell-line-matrix

zoldello:


zoldello:
@forresttanaka I did a hack of commenting out part of user.feature. Please, let me know if that is acceptable.

forresttanaka:
Is that the only change from the last code review? I think commenting that out for now is fine. We need to actively figure out what’s going on, if possible.

zoldello:
> Is that the only change from the last code review? I think commenting that out for now is fine. We need to actively figure out what’s going on, if possible.

No, I removed window.location and used reactContext.navigate. I also added more bdd test cases.

forresttanaka (dismissed):
For the record, I’m fine with commenting out the user test. I might know the reason; I’m testing my hypothesis.

forresttanaka:
I got the full user BDD test to work with this branch. I still don’t understand why this happens; some weird chromedriver thing maybe.

Right now we render the drop-down menus into a display:none div. The fix is to not render the menu at all while it’s not dropped down. You can do that in navbar.js under `NavItem`. Change:
```
{props.children}
```
into
```
{dropdownOpen && props.children}
```
You can remove the "open" CSS class in the `<li>` too.

zoldello:
> I got the full user BDD test to work with this branch. I still don’t understand why this happens; some weird chromedriver thing maybe.
> 
> Right now we render the drop-down menus into a display:none div. The fix is to not render the menu at all while it’s not dropped down. You can do that in navbar.js under `NavItem`. Change:
> 
> ```
> {props.children}
> ```
> 
> into
> 
> ```
> {dropdownOpen && props.children}
> ```
> 
> You can remove the "open" CSS class in the `<li>` too.

This will be addressed in a new ticket.

src/encoded/static/components/matrix_deeply_profiled_uniform_batch.js
```diff
@@ -0,0 +1,718 @@
...
+
+/** Page description */
+const matrixDescription = 'Cell line samples that were deeply profiled using a set of diverse biochemical approaches, listed with the corresponding batch identifiers.';
+
+/** Determines if user is on All-page or not  */
+const isAllDeeplyMatrix = (url) => url.includes('deeply-profiled-matrix');
+
+/** Switch page */
+const switchDeeplyProfilePageType = (isAll, pageUrl, navigate) => {
+    const query = new QueryString(pageUrl);
```
forresttanaka: The QueryString module is really designed to accept query strings only, not entire URIs. We usually use the `url` module to extract that. I’ve thought about modifying QueryString to accept either full URIs or query strings, but never felt it was difficult to extract the query string. But that avoids needing the call to `split()` below — just pass `query.format()` to that part of the string to `navigate`.

forresttanaka (changes requested):
Just one change for QueryString so you only pass it a query string.

src/encoded/static/components/matrix_deeply_profiled_uniform_batch.js
```diff
@@ -67,17 +68,17 @@ const sortByDescending = (a, b) => {
...
+const isAllDeeplyMatrix = (pageUrl) => pageUrl.includes('deeply-profiled-matrix');
 
 /** Switch page */
 const switchDeeplyProfilePageType = (isAll, pageUrl, navigate) => {
-    const query = new QueryString(pageUrl);
-    query.deleteKeyValue(internalTagKey);
-    const queryStringKeyValues = query.format().split('/?')[1];
-    const deeplyProfileType = isAll ? 'deeply-profiled-matrix' : 'deeply-profiled-uniform-batch-matrix';
-    const internalTagQueryString = deeplyProfileType === 'deeply-profiled-uniform-batch-matrix' ? `${internalTagKey}=Deeply Profiled` : '';
+    const link = url.parse(pageUrl, true);
```
forresttanaka: Well, not what I expected. The point I was making about using url.parse was so you can pass `url.parse().search` to `new QueryString()`. Now there’s not much point to using `url.parse()` since you’re manipulating the query string manually without QueryString anyway.

src/encoded/static/components/globals.js
```diff
@@ -313,6 +313,33 @@ export const biosampleTypeList = [
...
+    'IMR-90',
+    'K562',
+    'MCF-7',
+    'MCF10A',
+    'OCI-LY7',
+    'PC-9',
+    'PC3',
+    'Panc1',
+    'WTC11',
+];
```
forresttanaka, line 339: Seems like a very specific thing to put into globals.js. Could deeply profiled cell-lines really be shared across disparate parts of the code base? I probably would have put this into matrix.js to share them between the two deeply profiled matrices.

forresttanaka (dismissed):

src/encoded/static/components/datacolors.js
```diff
@@ -46,7 +46,7 @@ const rootColorList = [
  * @return {object} { r, g, b } values
  */
 export const colorToTriple = (color) => {
-    const num = parseInt(color.slice(1), 16);
+    const num = parseInt((color?.slice(1) || 1), 16);
```
forresttanaka: Why “1” if you don’t pass in a color?

forresttanaka: I’m also not sure when you’d ever not pass a color in the parameter.

zoldello: @forresttanaka - I fixed this. The issue was that 22 colors are set. In test cases like trying to show all data (no query strings to limit it); the rows can exceed the 22-color limit; so color was set to undefined. I added a modulus to ensure the limit is never exceeded.

src/encoded/static/components/matrix_deeply_profiled_uniform_batch.js
```diff
@@ -209,7 +209,12 @@ const convertDeeplyProfileDatatToDataTable = (context, getRowCategories, getRowS
     let matrixRow = 1;
     const matrixDataTable = rowCategoryData.sort((a, b) => sortByDescending(a.key, b.key)).reduce((accumulatingTable, rowCategoryBucket, rowCategoryIndex) => {
         const subCategoryData = getRowSubCategories(rowCategoryBucket);
-        const rowCategoryColor = rowCategoryColors[rowCategoryIndex];
+
+        // ensure colorIndex does not exceed the size of rowCategoryColors.
+        const rowCategoryColorsLength = rowCategoryColors.length;
+        const colorIndex = rowCategoryIndex < rowCategoryColorsLength ? rowCategoryIndex : (rowCategoryIndex % rowCategoryColorsLength);
```
forresttanaka: To be pedantic, this is equivalent to:
```
const colorIndex = rowCategoryIndex % rowCategoryColorsLength;
```
In which case you _can_ slip that expression between the brackets of line 217. That’s a pretty common paradigm: `a = b[c % b.length]`

forresttanaka (dismissed):

forresttanaka (dismissed):

zoldello:
@forresttanaka I need to re-add color-matching.

forresttanaka (approved):
That’s truly strange. Do you know specifically _why_ not rendering the menu causes the alert not to appear immediately?

PR #4134, opened 2021-10-12
ENCD-6171-upgrade-node-to-version-12-for-circleci-tests

forresttanaka:
One file, 20200430-app-es-pg.yml, has a name that makes me wonder if this is an auto-generated file, and we have to use some other way to modify it?

### DIFF
--- a/cloud-config/assembled-templates/20200430-app-es-pg.yml
+++ b/cloud-config/assembled-templates/20200430-app-es-pg.yml
@@ -14,7 +14,7 @@ bootcmd:
 - cloud-init-per once aptkeyadd-elasticsearch apt-key add /etc/apt/es_gpg_key
 - cloud-init-per once aptkeyget-postgres wget -qO - https://www.postgresql.org/media/keys/ACCC4CF8.asc > /etc/apt/pg_gpg_key
 - cloud-init-per once aptkeyadd-postgres apt-key add /etc/apt/pg_gpg_key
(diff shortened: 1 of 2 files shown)


paul-sud:
You're correct that the file is auto generated. It represents a prebuilt cloud config after applying our templating in `deploy.py`. I think it's probably fine to just leave it as is, I don't think it's used anywhere.

cloud-config/assembled-templates/20200430-app-es-pg.yml
```diff
@@ -14,7 +14,7 @@ bootcmd:
 - cloud-init-per once aptkeyadd-elasticsearch apt-key add /etc/apt/es_gpg_key
 - cloud-init-per once aptkeyget-postgres wget -qO - https://www.postgresql.org/media/keys/ACCC4CF8.asc > /etc/apt/pg_gpg_key
 - cloud-init-per once aptkeyadd-postgres apt-key add /etc/apt/pg_gpg_key
-- cloud-init-per once curl-nodeinstall curl -sL https://deb.nodesource.com/setup_10.x | sudo -E bash -
+- cloud-init-per once curl-nodeinstall curl -sL https://deb.nodesource.com/setup_12.x | sudo -E bash -
```
paul-sud, line 17: I think you will want to coordinate with @ottojolanki to get Node updated in the AMIs.

paul-sud (approved):

PR #4137, opened 2021-10-14
ENCD-6197-devcontainer

lrowe:
This lets me run encoded in a dev container on Codespaces and locally with Docker.

To run on my M1 Mac: in Docker Desktop preferences, under Resources increased to 8GB of RAM  and under Experimental features enabled 'Use the new Virtualization framework' (this greatly reduces background cpu usage.)

Click the Green icon in the bottom left corner of the VS Code window and then select "Reopen in container" in the menu that appears.  Wait for docker image to build and the `make devcontainer` postCreateCommand command to complete, then in one VS Code terminal run:
```
dev-servers development.ini --app-name app --clear --init --load
```
Once the data is loaded open another terminal inside vscode and run:
```
pserve development.ini
```
Port 6543 is forwarded so you can open http://localhost:6543 in the browser.

Note the added requirements.txt which duplicates the version pins in setup.py.

It should be possible to use without vscode. Build the docker container from the .devcontainer/Dockerfile. Run the docker container with port forwarding and your checkout directory mounted inside it. From inside the container run `make devcontainer` to perform the final setup.


paul-sud (dismissed):
Looks good to me. It is a little odd to have the dependencies duplicated between `requirements.txt` and `setup.py` but there isn't really a good workaround AFAIK. We could migrate the app to only use requirements files in the future, but it's not a trivial change.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+        locales build-essential git curl sudo python3-dev python3-venv python3-pip ca-certificates unzip openssh-client \
+        default-jdk libpq-dev postgresql postgresql-client graphviz libmagic-dev ruby-dev ruby-rubygems nginx-light redis python2 \
+        libjpeg-dev libtiff-dev libwebp-dev liblcms2-dev \
+    && rm -rf /var/lib/apt/lists/* \
+    && localedef -i en_US -c -f UTF-8 -A /usr/share/locale/locale.alias en_US.UTF-8 \
+    && curl -O https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-5.6.16.deb \
+    && dpkg -i elasticsearch-5.6.16.deb \
+    && rm elasticsearch-5.6.16.deb \
+    && gem install compass \
+    && curl -o awscliv2.zip "https://awscli.amazonaws.com/awscli-exe-linux-$(uname -m)-2.0.30.zip" \
```
paul-sud, line 36: There's no guarantee that the output of `uname -m` will match the names used by AWS for a given architecture, I think it would be better to leave explicit as `x86_64`:
```suggestion
    && curl -o awscliv2.zip "https://awscli.amazonaws.com/awscli-exe-linux-x86_64-2.0.30.zip" \
```

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+    && echo '%sudo ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers \
+    && mkdir /app \
+    && chown encoded /app \
+    # Workaround for https://bugs.launchpad.net/ubuntu/+source/nginx/+bug/1899164
+    && chown -R encoded /var/log/nginx /var/lib/nginx
+
+WORKDIR /app
+USER encoded
+ENV LANG en_US.utf8
+ENV PATH=/app/venv/bin:/usr/share/elasticsearch/bin:/usr/lib/postgresql/13/bin:${PATH}
```
paul-sud, line 50: Sorry I didn't notice last review, but we are on Postgres 11, I think we need to be explicit about the version when installing it from apt.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM node:12-bullseye-slim
+RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
+    && apt-get -y install --no-install-recommends \
+        locales build-essential git curl sudo python3-dev python3-venv python3-pip ca-certificates unzip openssh-client \
+        default-jdk libpq-dev postgresql postgresql-client graphviz libmagic-dev ruby-dev ruby-rubygems nginx-light redis python2 \
+        libjpeg-dev libtiff-dev libwebp-dev liblcms2-dev \
```
paul-sud: I'd prefer to see on package per line, alphabetically sorted, so it's easier to review later if dependencies are changed:
```suggestion
    && apt-get -y install --no-install-recommends \
        build-essential \
        ca-certificates \
        curl \
        default-jdk \
        git \
        graphviz \
        libjpeg-dev \
        liblcms2-dev \
        libmagic-dev \
        libpq-dev \
        libtiff-dev \
        libwebp-dev \
        locales \
        nginx-light \
        openssh-client \
        postgresql \
        postgresql-client \
        python2 \
        python3-dev \
        python3-pip \
        python3-venv \
        redis \
        ruby-dev \
        ruby-rubygems \
        sudo \
        unzip \
```

paul-sud (changes requested):

.devcontainer/Dockerfile
lrowe: I've confirmed this works on both x86_64 (codespaces) and aarch64 (which I run.) With folks likely transitioning to Arm Macs as they get new laptops I'd rather keep it working on both architectures since the new Macs' fast Rosetta translation does not apply to VMs.

.devcontainer/Dockerfile
lrowe: Debian Bullseye on which this container is based ships with Postgres 13 so switching versions would require installing from other repositories. While possible, it doesn't seem worth the hassle for a development container given it works with this version. Same for running with Python 3.9 instead of 3.8.

.devcontainer/Dockerfile
paul-sud: If you tested it, then that's fine by me.

.devcontainer/Dockerfile
paul-sud: I'm just concerned about devs introducing code that uses backwards-incompatible features of newer PG/Python/whatever, which works in the dev environment but errors on demos/RC/prod. IMO the hassle of installing in the container is worth it to save developer effort trying to figure out why code is breaking elsewhere when it works locally.

Granted, almost nobody is touching the PG stuff, but for Python it is certainly easy to be tempted to use 3.9-only features. We do test against PG 11 and Python 3.8.5 in CircleCI, but test coverage alone is insufficient.

PR #4138, opened 2021-10-14
ENCD-6175-support-collection-series-and-transgenic-enhancer-experiments-in-cart

forresttanaka:
* I found a very small but non-zero window in which the user can add datasets to the cart before the cart initialized on page load, causing a JS crash. The two modifications to `initializeCart` prevent this by setting the cart to the in-progress state on page load, and initializing the `elements` property on savedCartObj — a frequent dependency for cart code. Normal cart startup operation includes turning off the in-progress state once the user can truly interact with the cart.
* Many changes in this ticket guard against future dataset types being added to the cart before someone fully integrates them into the cart.

~https://encd-6175-collection-series-cart-qa1-fytanaka.demo.encodedcc.org/~

### DIFF
--- a/src/encoded/static/components/cart/cart.js
+++ b/src/encoded/static/components/cart/cart.js
@@ -110,6 +110,7 @@ const datasetTabTitles = {
     Annotation: 'annotations',
     FunctionalCharacterizationExperiment: 'FCEs',
     SingleCellUnit: 'single-cell units',
+    TransgenicEnhancerExperiment: 'TEEs',
 };
 
 
@@ -422,11 +423,12 @@ const CartTools = ({
     visualizable,
     isFileViewOnly,
 }) => {
-    // Make a list of all the dataset types currently in the cart.
-    const usedDatasetTypes = elements.reduce((types, element) => {
-        const type = atIdToType(element['@id']);
-        return types.includes(type) ? types : types.concat(type);
-    }, []);
+    // Make a list of all the allowed dataset types currently in the cart.
+    const usedDatasetTypes = new Set(
+        elements
+            .map((element) => atIdToType(element['@id']))
+            .filter((type) => allowedDatasetTypes[type])
+    );
 
     return (
         <div className="cart-tools">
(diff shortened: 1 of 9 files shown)


src/encoded/static/components/cart/cart.js
```diff
@@ -422,11 +423,12 @@ const CartTools = ({
     visualizable,
     isFileViewOnly,
 }) => {
-    // Make a list of all the dataset types currently in the cart.
-    const usedDatasetTypes = elements.reduce((types, element) => {
-        const type = atIdToType(element['@id']);
-        return types.includes(type) ? types : types.concat(type);
-    }, []);
+    // Make a list of all the allowed dataset types currently in the cart.
+    const usedDatasetTypes = new Set(
```
zoldello, line 427: It might be a few nano seconds faster to do the filter first and then map. As-is, map operates on what is later filtered out.

zoldello (dismissed):
Code looks fine. I was able to add a collection cart on the UI. My comment is insignificant.

src/encoded/static/components/cart/cart.js
forresttanaka: I didn’t time a comparison, but I didn’t want to filter first because I’d have to search `allowedDatasetTypes` to find the matching type. By using map first, I can just do a direct lookup of `allowedDatasetTypes`.

forresttanaka:
Updated for handling removing series and related datasets on individual pages, as described in Bonita’s QA.

Preview demo: https://encd-6175-collection-series-cart-pv2-fytanaka.demo.encodedcc.org/

zoldello (approved):
Code looks fine and has functionality to address issue Bonita brought up. I am curious, how do you rebase? I see like five commits rather than one.

forresttanaka:
I just rebased without squashing so it’s easier to tell which parts you’ve reviewed and which you haven’t.

zoldello (approved):
Let's have Bonita give this a go. Code looks fine.

PR #4141, opened 2021-10-15
ENCD-6199-fixed-deep-biosample-crash

forresttanaka:
We’ve followed biosample chains through their `parent_of` links to other biosamples, but at a certain point `parent_of` stops being embedded, so we have to request them from the server. To avoid loading entire large biosamples, possibly with their own embedded `parent_of` chains, we only retrieve the specific fields we need from those biosamples.

At a certain point, `BiosampleTable` started using the organism.scientific_name property of these downstream biosamples, but we don’t get that property.

This branch now retrieves that property, and attempts to make it more obvious that the `collectChildren` function needs its retrieved fields updated by putting those fields into an array outside the function.

Preview demo: ~https://encd-6199-deep-biosample-crash-pv1-fytanaka.demo.encodedcc.org/~

### DIFF
--- a/src/encoded/static/components/biosample.js
+++ b/src/encoded/static/components/biosample.js
@@ -25,7 +25,6 @@ const biosampleFieldsRetrieved = [
     'accession',
     'biosample_ontology.classification',
     'biosample_ontology.term_name',
-    'biosample_ontology.term_name',
     'organism.scientific_name',
     'parent_of',
     'status',
--- a/src/encoded/static/components/biosample.js
+++ b/src/encoded/static/components/biosample.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import _ from 'underscore';
+import QueryString from '../libs/query_string';
 import { Panel, PanelBody } from '../libs/ui/panel';
 import { auditDecor } from './audit';
 import { DbxrefList } from './dbxref';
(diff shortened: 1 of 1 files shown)


src/encoded/static/components/biosample.js
```diff
@@ -16,6 +17,21 @@ import { BiosampleSummaryDisplay, GeneticModificationOrganismNames, CollectBiosa
...
 
+/**
+ * Properties to retrieve from downstream biosamples. If you subsequently use any properties of
+ * these biosamples, make sure you include those properties in this array.
+ */
+const biosampleFieldsRetrieved = [
+    'accession',
+    'biosample_ontology.classification',
+    'biosample_ontology.term_name',
+    'biosample_ontology.term_name',
```
emro: Is `'biosample_ontology.term_name',` repeated intentionally?

emro (commented):
Looks good and solves the problem but one question about a repeated line.

src/encoded/static/components/biosample.js
forresttanaka: No, it’s not. I’ll fix that.

forresttanaka:
Updated preview demo: https://encd-6199-deep-biosample-crash-pv2-fytanaka.demo.encodedcc.org/

emro (approved):
Looks good!

PR #4142, opened 2021-10-15
ENCD-6168-update-fc-series-search-title

khzlin:


src/encoded/static/components/search.js
```diff
@@ -802,20 +803,25 @@ const SeriesComponent = ({ context: result, cartControls, removeConfirmation, au
...
             });
-            tilingModality = _.uniq(tilingModality);
         }
+        tilingModality = _.uniq(tilingModality);
+        referenceLoci = _.uniq(referenceLoci, (locus) => locus['@id']);
+        elementsSelectionMethod = _.uniq(elementsSelectionMethod);
     }
 
     // Dig through the biosample life stages and ages
+    // khine collect related_datasets.examined_lcoi
```
forresttanaka: Is this comment supposed to be here?

src/encoded/static/components/search.js
```diff
@@ -865,9 +871,12 @@ const SeriesComponent = ({ context: result, cartControls, removeConfirmation, au
                         if (lifeStage) {
                             lifeStages.push(lifeStage);
                         }
-                        if (biosample.treatments?.length > 0 && (differentiationSeries || fccSeries)) {
+                        if (biosample.treatments?.length > 0 && differentiationSeries) {
                             treatmentTerm = [...treatmentTerm, ...biosample.treatments.filter((t) => t.treatment_term_name).map((t) => t.treatment_term_name)];
                         }
+                        if (biosample.treatments?.length > 0 && fccSeries) {
+                            perturbationTreatments = biosample.treatments.filter((a) => a?.purpose === 'perturbation').map((t) => t.treatment_term_name);
```
forresttanaka: Can probably go without the question mark in `a?.purpose`, unless `biosample.treatments` could have null or undefined entries within the array, which doesn’t sound right to me.

src/encoded/static/components/search.js
```diff
@@ -1036,9 +1055,30 @@ const SeriesComponent = ({ context: result, cartControls, removeConfirmation, au
                         {constructionMethods.length > 0 ?
                             <div><span className="result-item__property-title">Construction method{constructionMethods.length > 1 ? 's' : ''}: </span>{constructionMethods.join(', ')}</div>
                         : null}
+                        {elementsSelectionMethod && elementsSelectionMethod.length > 0 ?
+                                     <div><span className="result-item__property-title">Elements selection method: </span>{elementsSelectionMethod.join(', ')}</div>
```
forresttanaka: Indented way too far to the right.

src/encoded/static/components/search.js
```diff
@@ -1036,9 +1055,30 @@ const SeriesComponent = ({ context: result, cartControls, removeConfirmation, au
...
+                        : null}
+                        {referenceLoci && referenceLoci.length > 0 ?
+                            <div><span className="result-item__property-title">Loci: </span>{referenceLoci.map((locus, i) => (
```
forresttanaka, line 1061: The indenting isn’t illegal, but unusual and kind of confusing. I wondered why you were using `key` in the following span when you didn’t have a loop, and then I finally found the `.map()` way over to the right.

src/encoded/static/components/search.js
```diff
@@ -1036,9 +1055,30 @@ const SeriesComponent = ({ context: result, cartControls, removeConfirmation, au
...
+                                        {locus.symbol}
+                                    </a>
+                                </span>
+                            ))}
+                            </div>
+                        : null}
                         {tilingModality.length > 0 ?
                             <div><span className="result-item__property-title">Tiling modality: </span>{tilingModality.join(', ')}</div>
                         : null}
+                        {perturbationTreatments.length > 0 && fccSeries ?
```
forresttanaka: Pretty common to test `fccSeries` before doing the comparison with zero, so that the JS parser doesn’t do that comparison, only to find it didn’t have to because `fccSeries` is false. But the time penalty is negligible.

forresttanaka (dismissed):
The only issues I found were with formatting, so change or ignore as you prefer. Question, do you know why some schema columns include a type while others don’t?

src/encoded/static/components/search.js
khzlin: nope. I'll delete this.

src/encoded/static/components/search.js
khzlin: ok.

forresttanaka (approved):

PR #4144, opened 2021-10-18
ENCD-4287 update d3

emro:
- updates d3 for node graph but keeps old version installed for dagre-d3 graphs
- creates component for node graph that can be used for sescc page as well as immune cells page
- adds clickable +/- icons to allow user to expand / collapse graph (will be useful for larger graphs)


src/encoded/static/components/matrix_sescc_stem_cell.js
```diff
@@ -580,53 +359,93 @@ class MatrixPresentation extends React.Component {
...
-     */
-    togglePebblesGroupVisibilty(visible) {
-        const cells = this.d3.selectAll('.js-cell')[0];
-
-        cells.forEach((cell) => {
-            const data = cell.__data__;
-            cell.style.fill = data[visible ? 'selectedColor' : 'deselectedColor'];
+    setSelectedNodes(newNode) {
+        this.setState((prevState) => {
+            const matrixSelection = newNode.replace(/\s/g, '_').toLowerCase();
```
zoldello: It may be a good idea to extract this into it's own function as it is called multiple times - 

`
replace(/\s/g, '_').toLowerCase()
`

src/encoded/static/libs/ui/node_graph.js
```diff
@@ -0,0 +1,254 @@
...
+// Creates a curved (diagonal) path from parent to the child nodes
+function diagonal(d, s) {
+    const path = `M${d.x},${d.y}C${d.x},${(d.y + s.y) / 2} ${s.x},${(d.y + s.y) / 2} ${s.x},${s.y}`;
+    return path;
+}
+
+const drawTree = (d3, targetDiv, data, fullWidth, fullHeight, margin, selectedNodes, setSelectedNodes) => {
+    let textWrapWidth = 115;
+    let characterLimitForWrap = 18;
+    const isMobile = fullWidth < mobileLimit;
```
zoldello: With mobile-first design in mind, it may be better to check if page in fullWidth (desktop) rather than check if mobile. Plus, I assume fullWidth covers table in horizontal and vertical rotations.

If you decide to stay with fullWidth-first design mode, have you considered something like - 
https://stackoverflow.com/a/10364620/178550
https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia

where the css is read to determine what is mobile rather than Javascript having determining it

src/encoded/static/libs/ui/node_graph.js
```diff
@@ -0,0 +1,254 @@
...
+        .attr('width', fullWidth)
+        .attr('height', fullHeight)
+        .attr('viewBox', `0 0 ${fullWidth} ${fullHeight}`)
+        .attr('preserveAspectRatio', 'xMidYMin meet')
+        .append('g')
+        .attr('transform', `translate(${margin.left},${margin.top})`);
+
+    const treemap = d3.tree().size([width, height]);
+    const root = d3.hierarchy(data, (d) => d.children);
+    root.x0 = width / 2;
```
zoldello, line 65: it may be nice to comment why this is done; for further reader. '2' is a magic number and division by 2 is a magic operation.

src/encoded/static/libs/ui/node_graph.js
```diff
@@ -0,0 +1,254 @@
...
+        text.each(function textWrap() {
+            text = d3.select(this);
+            const words = text.text().split(/\s+/);
+            let line = [];
+            let lineNumber = 0;
+            const lineHeight = 1.1;
+            const y = text.attr('y');
+            const dy = parseFloat(text.attr('dy'));
+            let newDy = '';
+            let tspan = text.text(null).append('tspan').attr('x', 0).attr('y', y);
```
zoldello, line 78: text set to null or ''?

src/encoded/static/libs/ui/node_graph.js
```diff
@@ -0,0 +1,254 @@
...
+                setSelectedNodes(d.data.name);
+                const clickedName = nodeKeyName(d.data.name);
+                if (internalSelectedNodes.indexOf(clickedName) > -1 && internalSelectedNodes.length > 1) {
+                    internalSelectedNodes = internalSelectedNodes.filter((s) => s !== clickedName);
+                } else if (internalSelectedNodes.indexOf(clickedName) > -1) {
+                    internalSelectedNodes = [];
+                } else {
+                    internalSelectedNodes = [...internalSelectedNodes, clickedName];
+                }
+                const clickedCell = d3.select(`.js-cell-${nodeKeyName(d.data.name)}`);
```
zoldello, line 139: Are you too a fan of prefixing css not meant for styling but for Javascript to identify specific DOM nodes, with 'js-'?

src/encoded/static/libs/ui/node_graph.js
```diff
@@ -0,0 +1,254 @@
...
+                d3.select(`.js-cell-${nodeKeyName(d.data.name)}`).selectAll('ellipse').style('transform', 'scale(1)');
+                d3.select(`.js-cell-${nodeKeyName(d.data.name)}`).selectAll('ellipse').attr('class', (d.data.class ? d.data.class : 'default'));
+                // remove highlight from corresponding matrix row
+                d3.select(`.${d.data.name.replace(/\s/g, '_').toLowerCase()}`).selectAll('th').attr('class', '');
+            });
+
+        ellipseSettings.forEach((ellipseSetting) => {
+            nodeGroup.append('ellipse')
+                .attr('cx', ellipseSetting.cx)
+                .attr('cy', ellipseSetting.cy)
```
zoldello, line 164: This code is feel. I just got a feeling it can be done with JSON/object rather than adding .attr one at a time.

src/encoded/static/libs/ui/node_graph.js
```diff
@@ -0,0 +1,254 @@
...
+        nodeEnter.selectAll('text')
+            .call(wrap, textWrapWidth);
+
+        nodeEnter.append('text')
+            .attr('class', (d) => `node-expander-collapser clicker-${nodeKeyName(d.data.name)}`)
+            .text((d) => (d._children ? circlePlus : d.children ? circleMinus : ''))
+            .attr('y', '-10px')
+            .attr('x', '-10px')
+            .style('font-size', '28px')
+            .on('click', function clickFunction(e, d) {
```
zoldello, line 188: After the sprint, if interested, let's have a philosophical discussion on this. In what you did here, the function context (this)  is not the "this" outside clickFunction; which new developers who grew up on arrow function may not realize.

src/encoded/static/libs/ui/node_graph.js
```diff
@@ -0,0 +1,254 @@
...
+        nodeUpdate.select('g.js-cell')
+            .attr('class', (d) => `js-cell-${nodeKeyName(d.data.name)} js-cell ${internalSelectedNodes.indexOf(nodeKeyName(d.data.name)) > -1 ? 'active-cell' : ''} ${d._children ? 'parent-cell' : ''}`);
+
+        const nodeExit = node.exit().transition()
+            .duration(animationDuration)
+            .attr('transform', `translate(${source.x},${source.y})`)
+            .remove();
+
+        nodeExit.select('text')
+            .style('fill-opacity', 1e-6);
```
zoldello, line 214: Cool. I have never seen anyone use this kind of notation before.

src/encoded/static/scss/encoded/modules/_immune_cells.scss
```diff
@@ -0,0 +1,238 @@
+$bubbleSize: 13px;
```
zoldello, line 1: This is another philosophical discussion. Should we use css var or stick to SCSS/SASS variables-style. I prefer SCSS/SASS; but one could argue to always take inbuilt over third party.

src/encoded/static/scss/encoded/modules/_immune_cells.scss
```diff
@@ -0,0 +1,238 @@
...
+        }
+    }
+
+    ellipse {
+        &:hover {
+            cursor: pointer;
+        }
+    }
+
+    .js-cell {
```
zoldello, line 64: When I see "js-cell"; that tells me this is meant only for Javascript to use to identify code and not for CSS/SCSS to use for styling. Am I correct?

src/encoded/static/scss/encoded/modules/_immune_cells.scss
```diff
@@ -0,0 +1,238 @@
...
+        }
+    }
+}
+
+.hide {
+    display: none;
+}
+
+// sescc matrix-specific colors and styles
+.matrix__row-data {
```
zoldello, line 85: While 
`
matrix__row-data
`
is valid; there is the matrix pages that uses this same name. Which may cause confusion for developers trying to modify a matrix and see this in the file and no the matrix scss file.

zoldello (dismissed):
Code looks good. The UI looks fine and looks better than it was before. My comments do not need to be addressed.

src/encoded/static/components/matrix_sescc_stem_cell.js
emro: Good idea!

src/encoded/static/libs/ui/node_graph.js
emro: You're right, this should be mobile first.

src/encoded/static/libs/ui/node_graph.js
emro: I think I copied this from the original ticket

zoldello (dismissed):
Code looks fine.

zoldello (approved):
I see the corrections to the issues Forrest brought up.

PR #4146, opened 2021-10-19
ENCD-6178-library-property-updates-Yu-lab

iyoungworth:
added-size-selection-enum

changelog-update


src/encoded/schemas/changelogs/library.md
```diff
@@ -1,5 +1,9 @@
 ## Changelog for library.json
 
+### Schema version 20
+* Added *construction_method* enum *capped RNA enrichment*, *uncapped RNA enrichment*, and *PCR-free*.
```
gabdank: I think you need to say that the property changed to be an array

gabdank (changes requested):

src/encoded/schemas/changelogs/library.md
iyoungworth: Thanks for pointing this out, I added "The property is now also submittable as an array."

gabdank (approved):

PR #4147, opened 2021-10-19
ENCD-6201-add-purpose-treatment-UI

bonitalam:


forresttanaka (approved):
Just to be sure, Experiment and GeneticModification pages can also display treatments, but on one line instead of three. I assume those don’t need to show the purpose (they don’t show the type either).

bonitalam:
> Just to be sure, Experiment and GeneticModification pages can also display treatments, but on one line instead of three. I assume those don’t need to show the purpose (they don’t show the type either).

Thanks for checking @forresttanaka. Yup, we are just adding it to the Treatment and Biosample object pages.

PR #4167, opened 2021-10-26
ENCD-6070-human-donor-matrix

zoldello:


src/encoded/static/components/matrix_human_donor.js
```diff
@@ -0,0 +1,744 @@
...
+
+
+/** General collection of attributes to exclude in some scenarios */
+const excludedAttributes = ['', 'n/a', 'unknown', undefined, null];
+
+/** life stages to display on UI */
+const includedLifeStage = ['child', 'embryonic', 'newborn'];
+
+/** Roman mythhology equivalent represening gender */
+const getGenderMythology = (genderType) => {
```
forresttanaka, line 24: This might be the best function name in all of encode.

src/encoded/static/components/matrix_human_donor.js
```diff
@@ -0,0 +1,744 @@
...
+    }
+
+    return '';
+};
+
+/** Query keys to remove from context's urls */
+const removedQueryKeys = ['@id', 'search_base'];
+
+/**
+ *  Key/value pairs mapping dseases to css classes
```
forresttanaka: “dseases” typo.

src/encoded/static/components/matrix_human_donor.js
```diff
@@ -0,0 +1,744 @@
...
+ *
+ * @param {object} data from a fetch call
+ */
+const addExtraRowHeaderDataToMatrix = (data, matrixUpdateCallBack) => {
+    const graph = data['@graph'].map((g) => {
+        let age = '';
+
+        if (g.age) {
+            const ageText = `${[g.age].filter((a) => !excludedAttributes.includes(a)).join(', ')}`;
+            const formattedAge = `${(ageText || '').replace(' or above', '+')} ${ageText ? g.age_units : ''}${ageText === 1 || ageText === '' ? '' : 's'}`.trim('');
```
forresttanaka: Haven’t seen `trim()` with an empty string before. It’s not defined to take parameters, so I guess it just ignores it.

src/encoded/static/components/matrix_human_donor.js
```diff
@@ -0,0 +1,744 @@
...
+            const lifeStageText = g.lifeStage ? ` (${g.lifeStage})` : '';
+
+            age = `${formattedAge}${lifeStageText}`;
+        }
+
+        return {
+            accession: g['@id']?.replace('/human-donors/', '')?.replace('/', ''),
```
forresttanaka: globals.js has a `atIdToAccession` function.

src/encoded/static/components/matrix_human_donor.js
```diff
@@ -0,0 +1,744 @@
...
+    const xLength = x.header.length;
+    const y = [];
+
+    matrix.y['replicates.library.biosample.donor.accession'].buckets.forEach((accessionBucket) => {
+        const accession = accessionBucket.key;
+
+        // In row, first item is accession. The next are assay number based on the assay in variable - x
+        const row = {
+            accession,
+            assayTotals: [...Array(xLength)].map(() => 0),
```
forresttanaka, line 101: Oh, that’s a good way to do that.

src/encoded/static/components/matrix_human_donor.js
```diff
@@ -0,0 +1,744 @@
...
+ */
+const getDataTableData = (context) => {
+    const { matrix } = context;
+    const x = {
+        header: matrix.x.assay_title.buckets.map((assayTitle) => assayTitle.key),
+    };
+    const xLength = x.header.length;
+    const y = [];
+
+    matrix.y['replicates.library.biosample.donor.accession'].buckets.forEach((accessionBucket) => {
```
forresttanaka, line 95: Technically you should retrieve `replicates.library.biosample.donor.accession` from matrix.y.group_by[0] so that this code automatically adjusts if you change experiment.py. At this stage, though, this seems unlikely to change.

src/encoded/static/components/matrix_human_donor.js
```diff
@@ -0,0 +1,744 @@
...
+        return response.json();
+    }).then((data) => {
+        addExtraRowHeaderDataToMatrix(data, matrixUpdateCallBack);
+    }).catch((e) => {
+        console.error('OBJECT LOAD ERROR: %s', e);
+    });
+};
+
+/**
+ * This provides a subtrate for building a matrix using css Grid. It receives an array of array where each subarray is a row
```
forresttanaka: “substrate” maybe?

src/encoded/static/components/matrix_human_donor.js
```diff
@@ -0,0 +1,744 @@
...
+ *              { content: ['a title'], itemKey: 'a-title-key' }
+ *              { content: ['another title'], itemKey: 'b-title-key' }
+ *          ]
+ *     },
+ *     {
+ *          type: 'matrixData',
+ *          itemKey: 'data-key',
+ *          items: [
+ *              { content: ['a data point'], itemKey: 'a-data-key' }
+ *              { content: ['another daya point'], itemKey: 'yet-another-key' }
```
forresttanaka: “data”

src/encoded/static/components/matrix_human_donor.js
```diff
@@ -0,0 +1,744 @@
...
+import { BodyMapThumbnailAndModal } from './body_map';
+
+
+/** General collection of attributes to exclude in some scenarios */
+const excludedAttributes = ['', 'n/a', 'unknown', undefined, null];
+
+/** life stages to display on UI */
+const includedLifeStage = ['child', 'embryonic', 'newborn'];
+
+/** Roman mythhology equivalent represening gender */
```
forresttanaka: “mythology” and “representing”

src/encoded/static/components/matrix_human_donor.js
```diff
@@ -0,0 +1,744 @@
...
+        const accessionTextColor = isLight(rowColor) ? '#000' : '#fff';
+        const { accession, assayTotals } = yEntry;
+        const maxAssayTotal = Math.max(...assayTotals);
+        const minAssayTotal = Math.min(...assayTotals);
+        const logBase = Math.log(1 + maxAssayTotal + minAssayTotal);
+
+        // change grid matrix default values
+        const rowHeaderFirstColStyle = {
+            color: `${accessionTextColor}`,
+            backgroundColor: `${rowColor}`,
```
forresttanaka: Don’t need the string templates here since you don’t add anything to these variables, so you can just use the variables alone.

src/encoded/static/components/app.js
```diff
@@ -277,6 +278,7 @@ const fullPageReloadPaths = [
     '/summary/',
     '/encyclopedia/',
     '/immune-cells/',
+    /human-donor-matrix/,
```
forresttanaka: I was hoping to never add to this array, and instead fix whatever bug causes a required reload. But as I don’t understand why these pages require a full-page reload, I guess we have to do this. Not sure how this works without the single quotes though? Does it interpret this as a regular expression or something?

forresttanaka (dismissed):
None of my comments require changes.

src/encoded/static/components/matrix_human_donor.js
zoldello: Thanks

src/encoded/static/components/matrix_human_donor.js
zoldello: Fixed

src/encoded/static/components/matrix_human_donor.js
zoldello: Opps. Fixed.

src/encoded/static/components/matrix_human_donor.js
zoldello: Are you okay with me not changing this? If there is a change to experiment.py. the code needs to be updated to reflect it and being explicit like using "replicates.library.biosample.donor.accession", to me, forces me to remember that.

src/encoded/static/components/app.js
zoldello: Removed as it is not needed. I was fixing an issue and throwing mud to try to figure it out. I should of had removed this after I figured it out. If curious, the issue was I had three radio buttons with two of them both having checked={true} based on a certain click-pattern and it was causing one button not to work.

forresttanaka (dismissed):

forresttanaka (dismissed):

src/encoded/static/components/matrix_human_donor.js
```diff
@@ -165,11 +166,14 @@ const convertToMatrixData = (dataTableData, baseUrl, searchBaseUrl, graph) => {
             ),
             itemKey: 'matrix-header-key-titles',
         }].concat(dataTableData.x.header.map((xData) => {
-            const assayTitleText = `assay_title=${xData}`;
-            const assayTitle = baseUrl.includes(assayTitleText) ? `&${assayTitleText}` : '';
+            const baseUrlQuery = new QueryString(baseUrl);
+            baseUrlQuery.deleteKeyValue('assay_title');
+            baseUrlQuery.addKeyValue('assay_title', xData);
```
forresttanaka, line 171: There’s also a `replaceKeyValue`, though all it does is the same two steps you have here.

src/encoded/static/components/matrix_human_donor.js
```diff
@@ -233,13 +237,18 @@ const convertToMatrixData = (dataTableData, baseUrl, searchBaseUrl, graph) => {
                     const itemKey = `${accession}-${assayCountIndex}`;
                     const assayTitle = x.header[assayCountIndex];
 
+                    const searchBaseUrlQuery = new QueryString(searchBaseUrl);
```
forresttanaka, line 240: Really supposed to separate out the query string from the rest of the URL and pass that to `new QueryString`, but in most (not all) cases it works without doing that. I feel like I need to add some code to the QueryString class to optionally accept a whole URI and split the query string off itself, because people keep passing whole URLs to it.

forresttanaka (approved):

src/encoded/static/components/matrix_human_donor.js
zoldello: Good point.  I used deleteKeyValue because there could be multiple instances of assay_title query string and I need to delete them all. So, replaceKeyValue will lead to a bug if used.

src/encoded/static/components/matrix_human_donor.js
zoldello: Another good point.  There could be multiple assay_titles, so, it would require fancy javascript to get them all out given they could be in any location. The approach I took is a little janky but simpler. Is it acceptable?

PR #4169, opened 2021-10-27
ENCD-6217-fix-crash-when-navigating-from-report-to-search-with-fields

forresttanaka:
The issue happens when you specify columns within a report view and then click the button to go to the List page. A Javascript crash happens because I don’t remove the `field=` elements of the query string.

This also applies to navigating to the matrix, though it’s harder to reproduce the crash there.

Preview demo: https://encd-6217-report-list-crash-pv1-fytanaka.demo.encodedcc.org/

QA demo indexing currently.

### DIFF
--- a/src/encoded/static/components/view_controls.js
+++ b/src/encoded/static/components/view_controls.js
@@ -38,11 +38,17 @@ ViewControlButton.propTypes = {
 /**
  * Renders the view type button for the search page.
  */
-const SearchViewButton = ({ viewType, query }) => (
-    <ViewControlButton viewType={viewType} queryString={query.format()}>
-        {viewType.title}
-    </ViewControlButton>
-);
+const SearchViewButton = ({ viewType, query }) => {
+    // In case the user goes to the search page from the report page with columns specified, strip
+    // the "field" query-string elements.
+    const searchQuery = query.clone();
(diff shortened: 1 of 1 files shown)


forresttanaka:
The not-bdd-non-indexing test currently fails for everyone.

src/encoded/static/components/view_controls.js
```diff
@@ -38,11 +38,17 @@ ViewControlButton.propTypes = {
...
  */
-const SearchViewButton = ({ viewType, query }) => (
-    <ViewControlButton viewType={viewType} queryString={query.format()}>
-        {viewType.title}
-    </ViewControlButton>
-);
+const SearchViewButton = ({ viewType, query }) => {
+    // In case the user goes to the search page from the report page with columns specified, strip
+    // the "field" query-string elements.
+    const searchQuery = query.clone();
```
zoldello, line 44: Do you need to clone?

zoldello (approved):
Let's have Bonita test this out when the demo comes up

src/encoded/static/components/view_controls.js
forresttanaka: Yes, to avoid having the query get modified as an unexpected side effect for the caller.

PR #4176, opened 2021-11-02
ENCD-6226-remove-git-protocol

paul-sud:


keenangraham:
@paul-sud I think you want to be slightly careful about letting `package-lock.json` update unrelated dependencies. I think it does this automatically but probably should avoid in this case. I manually cleaned to what seemed relevant here for example: https://github.com/ENCODE-DCC/encoded/compare/set-https-in-package-json

paul-sud:
From what I can gather `package-lock.json` shouldn't be manually edited. Maybe there is some NPM option to restrict the dependency solver though.

ottojolanki:
Looks good to me, assuming we are not breaking JS deps?

keenangraham:
Maybe @forresttanaka understands how to avoid the unrelated dependency upgrades in `package-lock.json`. I `git add --patch` the relevant sections and then reset the rest to head.

Pretty sure it's not appropriate to bump e.g. `@types/react` here.

forresttanaka:
I’m not aware of any ways to restrict unrelated dependencies from package-lock.json (@paul-sud is correct — this file isn’t intended to be hand edited — only updated through npm upgrades and installs).

New npm installs are pretty straightforward, but updates aren’t, and for the most part I let package-lock.json get changed however npm wants. The dependency trees for certain components get incredibly complicated.

Looking over these package-lock.json changes, I don’t see anything very scary, even the React one is just a minor version number change. I would mostly make sure various file graphs still look and work correctly, as does Valis, though Valis didn’t actually change versions so I wouldn’t expect any issues with that.

@keenangraham’s branch uses the “github:” prefixing while Paul’s uses nothing. I don’t know the difference there.

paul-sud:
@forresttanaka I just specified the GitHub URLs as indicated in the docs: https://docs.npmjs.com/cli/v6/configuring-npm/package-json#github-urls , it seems to work the same regardless of the presence of the `github:` prefix.

forresttanaka:
@paul-sud I see. That’s good to know.

keenangraham:
The `github:` prefix got added automatically for me by doing e.g. `npm install ENCODE-DCC/valis-hpgv.git#1.8.0 ENCODE-DCC/graphlib.git#v1.0.8`.

I suppose this is fine if @forresttanaka doesn't have any problem with the other dependency upgrades. But it seems weird that we are changing versions in a PR where all we really want to do is update the reference to the same version of a git branch. Unless the package-lock was somehow out of date before, it seems like all of the other versions should remain the same?

forresttanaka:
@keenangraham It does seem weird, but just based on past performance (granted, not an indicator of future incidents) we haven’t had any issues with collateral upgrades that I recall. Sometimes we end up with a merge conflict in package-lock.json (and I think we’ll have at least one in this sprint), and there’s little to do but to rebuild package-lock.json from scratch, which we do probably a few times a year typically.

keenangraham:
Yeah, also don't love it because it pushes an update that could be QAed by making sure the tests pass (and maybe double checking the right version of the package is installed) into the realm of QAing a variety of small version upgrades.

We know the current versions work in production. It seems reasonable to be able to change the method of getting the same version of a package (i.e. `git://github.com/ENCODE-DCC/graphlib.git` -> `github:ENCODE-DCC/graphlib`) without changing anything else. (Wonder if this is a case for manual editing?)

You are probably right about it not causing any breaking changes, but don't like that it suddenly bundles multiple changes into a PR where nothing (functional) should change.

paul-sud:
It is a known critique of how npm manages dependencies, see https://npm.community/t/impossible-to-update-single-package-without-updating-its-dependencies/1156.html and the corresponding RFC https://github.com/npm/rfcs/pull/21 (which to my knowledge hasn't been implemented).

In any case, I've pushed changes that I made via manually editing the package-lock.json, I do agree that it's not ideal to have a bunch of dependencies being bumped here given the nature of the ticket. Here it's actually tractable to edit it by hand but in general it's not a good idea, better to leave it to the npm resolver to update.

ottojolanki (dismissed):
LGTM

keenangraham:
@paul-sud Can you rebase this over dev and then we can merge if tests pass?

PR #4181, opened 2021-11-04
Encd 6225 valis track not refreshed

zoldello:


src/encoded/static/components/app.js
```diff
@@ -322,8 +321,10 @@ class App extends React.Component {
             logout: 'triggerLogout',
         };
 
+        this.href = props.href;
```
emro, line 323: Since we are using `this.href` here, do we need `this.state.href` set in line 301?

emro (dismissed):
Great solution! Just have one comment - I think we might be able to get rid of a line where this.state.href is set since we're not updating it and since we're hopefully using this.href everywhere.

emro (dismissed):
Looks great!

zoldello:
Added @forresttanaka to code review the new commit as he has the most knowledge about the issue I looked at here

forresttanaka (dismissed):
Glad you got it working. Do you know what this.contextRequest is for? I still don’t even know.

src/encoded/static/components/form.js
```diff
@@ -890,6 +891,13 @@ export class Form extends React.Component {
         return this.state.isDirty && this.state.isValid && !this.state.editor_error && !this.communicating;
     }
 
+    cancel() {
+        const link = url.parse(this.context.location_href, true);
+
+        // the last '/' is a hack to reload the page. In app.js, fallbackNavigate() did not without it.
+        this.context.navigate(`${link.pathname}/`, { reload: true });
```
forresttanaka, line 899: Thanks for that comment. It looks wrong to add the slash without it.

src/encoded/static/components/form.js
```diff
@@ -890,6 +891,13 @@ export class Form extends React.Component {
...
+    }
```
forresttanaka, line 900: Thought you had to bind `this` for `cancel()` but now see you’re calling it in an arrow function.

forresttanaka (dismissed):
Hopefully that’s all the special cases. That `<App>` change was pretty impactful, but I think we’re much better for it.

src/encoded/static/components/form.js
zoldello: In fallbackNavigate() (in app.js), window.location.assign would not redirect properly without it. I have no idea why. That is why I did this. I also did not want to add too much internal details about app.js in form.js and other files.

src/encoded/static/components/form.js
zoldello: Opps. You are correct. I will correct it. I have been dealing hooks for too long.

forresttanaka (approved):

PR #4184, opened 2021-11-05
ENCD-6191-listed-carts

forresttanaka:
* Calling the cart menu “status” got pretty confusing when cart status as a property of a cart has become so important, so I moved much of status.js into status_listing.js, and use status.js specifically for things relating directly to cart status.
* I also moved a lot of functionality out of index.js into store.js, as index.js’s job is mostly as an interface to other modules; not to implement functionality.
* Much of the other changes to seemly unrelated cart components have to do with released carts which can’t be altered by anyone, including their owners.

https://encd-6191-cart-sharing-pv1-fytanaka.demo.encodedcc.org/


forresttanaka:
I won’t be starting a QA demo now. I’ll start it Monday morning.

src/encoded/static/components/cart/util.js
```diff
@@ -205,3 +207,23 @@ export const getSeriesDatasets = (series, allDatasetsInCart) => (
...
+/**
+ * Returns various aspects about the given cart that determines whether the cart can be modified
+ * by their submitters.
+ * @param {object} cart Cart from database or context
+ * @returns {object} shape: {
+ *     locked - cart is locked by the submitter to prevent unintended modifications
+ *     released - cart is released by DCC
+ *     any - any of the above reasons apply
+ */
+export const getReadOnlyState = (cart) => {
```
zoldello, line 221: Can this be simplified? All you are doing it creating three states - isLocked, isReleased, isCartStateSet. The name of the method had me thinking you were doing much more.

src/encoded/static/components/cart/add_multiple.js
```diff
@@ -91,7 +93,7 @@ const CartAddAllSearchComponent = ({
                 {savedCartObj.name ? <div className="cart-toggle__name">{truncateString(savedCartObj.name, 22)}</div> : null}
                 <button
                     type="button"
-                    disabled={inProgress || savedCartObj.locked}
+                    disabled={inProgress || readOnlyState.any}
```
zoldello, line 96: Why is this button disabled if the state is released (readOnlyState.any is truthy if release state is set to anything)?

src/encoded/static/components/cart/cart.js
```diff
@@ -325,32 +331,19 @@ const requestDatasets = (elements, fetch, session) => {
...
-        // Shouldn't happen but just in case.
-        return null;
-    }
-
+const CartDatasetReport = ({ cart, usedDatasetTypes }) => {
     // Only display the Dataset Report button if we have at least one experiment in the cart. This
     // button drops down a menu allowing the user to select the data type to view which links to
     // that report view.
-    if (linkedCart.current && linkedCart.current.elements && linkedCart.current.elements.length > 0) {
+    if (cart && cart.elements && cart.elements.length > 0) {
```
zoldello: Can also be written as - 

if (cart?.elements?.length > 0) {

src/encoded/static/components/cart/cart.js
```diff
@@ -557,6 +541,8 @@ const CartSearchResultsControls = ({
 };
 
 CartSearchResultsControls.propTypes = {
+    /** Cart being displayed */
```
zoldello: Is the comment correct?

src/encoded/static/components/cart/cart.js
```diff
@@ -730,26 +716,30 @@ CounterTab.defaultProps = {
...
  * @return {object} -
  * {
+ *      {object} cart - Cart object from Redux for active carts or from context for object carts
  *      {string} cartType - Cart type: OBJECT, ACTIVE; '' if undetermined
  *      {string} cartName - Name of cart
  *      {array} cartDatasets - @ids of all datasets in cart
  * }
  */
 const getCartInfo = (context, savedCartObj) => {
+    let cart = null;
```
zoldello, line 726: This is me nitpicking. When I see something set to null, it tells me you need the value to be null for use in the code. On the flip side, if the value is set later but no setting is needed at declaration, I have seen people not set it, which sets it to undefined by default (undefined tells me no value is needed at declaration).

src/encoded/static/components/cart/clear.js
```diff
@@ -89,67 +89,61 @@ CartClearModal.contextTypes = {
...
+const CartClearButtonComponent = ({ elements, inProgress, isCartReadOnly }) => {
+    const [modalOpen, setModalOpen] = React.useState(false);
 
     /**
      * Handle a click in the Clear Cart button by showing the confirmation modal.
      */
-    handleClearCartClick() {
-        this.setState({ modalOpen: true });
-    }
+    const handleClearCartClick = () => {
```
zoldello, line 98: In the event name, doesn't "Click" imply "handle"? Since you are in the cart file-folder, is it safe to assume the method supports a cart operation?

src/encoded/static/components/cart/clear.js
```diff
@@ -89,67 +89,61 @@ CartClearModal.contextTypes = {
...
-                </div>
-            );
-        }
-        return null;
+    if (elements.length > 0) {
+        return (
+            <div className="cart-tools-extras__button">
+                <button
+                    type="button"
+                    disabled={inProgress || isCartReadOnly}
```
zoldello, line 114: Do you need a disabled-setting for a system failure?

src/encoded/static/components/cart/description.js
```diff
@@ -0,0 +1,328 @@
...
+const MAX_DESCRIPTION_LENGTH = 1000;
+
+
+/**
+ * Sanitize the given description to strip it of anything that could cause an XSS attack, as well
+ * as preceding and trailing spaces as required by the cart schema.
+ * @param {string} description Description entered by the user
+ * @returns {string} `description` but sanitized and with preceding/trailing spaces removed
+ */
+export const sanitizeDescription = (description) => sanitize(description).trim();
```
zoldello, line 28: Do you want to sanitize (as in remove DOM-like string) or ENCODE (prevent XSS attack by changing anything like DOM to a safe format)? This turns null to empty string, truthy to 'true', falsey to empty string. Are you okay with these?

src/encoded/static/components/cart/description.js
```diff
@@ -0,0 +1,328 @@
...
+
+
+/**
+ * Custom React hook to calculate the enabled/disabled state of the button to save a description
+ * and the label that should appear in the button given the current saved state of the description.
+ * @param {object} cart Saved cart object
+ * @param {string} editedDescription Description as entered by the user before sanitizing
+ * @returns See return statement
+ */
+export const useCartDescription = (cart, editedDescription) => {
```
zoldello: Is this a custom hook or just a regular function? Custom hooks I have seen typically wraps at least one of react inbuilt hooks.

src/encoded/static/components/cart/description.js
```diff
@@ -0,0 +1,328 @@
...
+/**
+ * Displays the formatted cart description.
+ */
+export const CartDescriptionDisplay = ({ description }) => {
+    /** Description cleaned of any dangerous HTML/Javascript */
+    const [sanitizedDescription, setSanitizedDescription] = React.useState(description);
+
+    React.useEffect(() => {
+        // Must have a DOM to sanitize the description.
+        setSanitizedDescription(marked(sanitizeDescription(description || '')));
```
zoldello, line 114: sanitize takes care of falsey states for you.

src/encoded/static/components/cart/description.js
```diff
@@ -0,0 +1,328 @@
...
+    }, [description]);
+
+    if (sanitizedDescription) {
+        return (
+            <div className="cart-description">
+                <div dangerouslySetInnerHTML={{ __html: sanitizedDescription }} />
```
zoldello, line 120: If you took the trouble to remove security-risk markup, why now use dangerouslySetInnerHTML?

src/encoded/static/components/cart/description.js
```diff
@@ -0,0 +1,328 @@
...
+        setEditedDescription(content);
+    };
+
+    return (
+        <Modal labelId="description-editor" descriptionId="description-editor-description" focusId="text-editing-area" closeModal={onClose} widthClass="sm">
+            <ModalHeader labelId="description-editor" closeModal={onClose} title={`Edit cart description: ${cart.name}`} />
+            <ModalBody addCss="cart-description-editor">
+                <p id="description-editor-description">
+                    This description gets saved with this cart, and appears along with this cart
+                    when you list it on the public <strong>Listed carts</strong> page. You can
```
zoldello, line 160: <strong> is valid; but isn't it mixing markup with styling ..?

src/encoded/static/components/cart/description.js
```diff
@@ -0,0 +1,328 @@
...
+                    format the text with Markdown syntax.
+                </p>
+                {cart.status === 'listed' &&
```
zoldello, line 163: This is always true if the LHS is true... Right?

src/encoded/static/components/cart/manager.js
```diff
@@ -472,89 +463,62 @@ const NameCartButton = connect(NameCartButtonComponent.mapStateToProps, NameCart
...
-        this.state = {
-            /** True if delete warning modal is open */
-            modalOpen: false,
-        };
-        this.handleSubmitClick = this.handleSubmitClick.bind(this);
-        this.handleDeleteClick = this.handleDeleteClick.bind(this);
-        this.handleCloseClick = this.handleCloseClick.bind(this);
-    }
+const DeleteCartButtonComponent = ({ cart, setInProgress, inProgress, fetch, updateCartManager }) => {
+    /** True if modal to confirm cart deletion open */
```
zoldello: I think you mean - True if modal to confirm cart deletion is opened

src/encoded/static/components/cart/menu.js
```diff
@@ -0,0 +1,215 @@
...
+import Status from '../status';
+import { CartClearModal } from './clear';
+import CartShare from './share';
+
+
+/**
+ * Renders the cart icon menu and count or in-progress spinner in the nav bar.
+ */
+const CartNavTitle = ({ elements, inProgress, locked }) => {
+    let status;
```
zoldello, line 21: Does status === undefined mean no status is set?

src/encoded/static/components/cart/menu.js
```diff
@@ -0,0 +1,215 @@
...
+    /**
+     * Called when the Clear Cart modal close buttons are clicked.
+     */
+    const closeClearCart = () => {
+        setClearModalOpen(false);
+    };
+
+    // Define the menu items for the Cart Status menu.
+    const viewCartItem = <a key="view" href="/cart-view/">View cart</a>;
+    const clearCartItem = !savedCartObj.locked && <button type="button" key="clear" onClick={clearCartClick}>Clear cart</button>;
```
zoldello, line 98: Would **nullish coalescing** be a good option in lieu of this pattern?

src/encoded/static/components/cart/menu.js
```diff
@@ -0,0 +1,215 @@
...
+    const menuItems = [];
+    if (loggedIn) {
+        // Build the disabled cart name and lock status.
+        const cartName = (savedCartObj && savedCartObj.name) && truncateString(savedCartObj.name, 22);
```
zoldello: Why magic number 22?

src/encoded/static/components/cart/search_listing.js
```diff
@@ -0,0 +1,141 @@
...
+ */
+const descriptionVariants = {
+    expanded: { height: 'auto' },
+    closed: { height: `${DESCRIPTION_COLLAPSED_HEIGHT}px` },
+};
+
+
+/**
+ * Display the description of a cart in the cart search results. Include a control to collapse and
+ * expand the description when the description has more text than would fit in the allowed vertial
```
zoldello: "vertical". It would be cool if we had the power like Shakespeare to invent new words rather than limiting ourselves to a dictionary.

src/encoded/static/components/cart/share.js
```diff
@@ -1,20 +1,30 @@
 /**
  * Components to display the Share Cart modal.
  */
+
+// node_modules
```
zoldello, line 5: These should be obvious or am I mistaken...?

src/encoded/static/components/cart/store.js
```diff
@@ -0,0 +1,215 @@
...
+/**
+ * Generates a new fileViews array with the given files removed. `fileViews` does not get mutated.
+ * @param {array} fileViews Array of fileViews
+ * @param {string} title Title of fileView to update
+ * @param {array} filePaths Array of file @ids to remove
+ * @returns {array} New fileViews array with updated fileView entry; null if no update possible
+ */
+const generateFileViewsMinusFiles = (fileViews, title, filePaths) => {
+    // Find file view with the given title.
+    const indexOfFileViewWithMatchingTitle = fileViews.findIndex((view) => view.title === title);
```
zoldello, line 38: Maybe - fileViewIndexWithMatchingTitle

src/encoded/static/components/objectutils.js
```diff
@@ -1178,3 +1178,26 @@ export const collectDatasetBiosamples = (dataset) => {
...
     return [];
 };
+
+
+/**
+ * Display the current and maximum character counter for a <textarea>.
+ */
+export const TextAreaCounter = ({ value, maxLength, css }) => (
+    <div className={`textarea-counter${css && ` ${css}`}`}>
+        {value.length} / {maxLength}
```
zoldello, line 1188: Do you need to Math. {floor/ceil} this so you always get a whole number?

src/encoded/static/scss/encoded/modules/_cart.scss
```diff
@@ -1660,9 +1671,229 @@ $file-view-checkbox-size: 14px;
...
+            margin-top: 0;
+        }
+
+        &:last-child {
+            margin-bottom: 0;
+        }
+    }
+
+    ul, ol {
+        margin: 0.3em;
```
zoldello, line 1803: should we be using "em"?

zoldello (dismissed):
None of my comment are must-have. A few are nitpicking.

src/encoded/static/components/cart/util.js
forresttanaka: These are the potential reasons a cart might be locked, but there are unusual cases where a locked cart gets treated differently from a released cart, so they need distinguishing. For example, the Unlock button only gets disabled for a released cart, not a locked one.

src/encoded/static/components/cart/add_multiple.js
forresttanaka: The button gets disabled if the cart is released or locked, so “any.” readOnlyState.any is truth if the cart status is released or the cart is locked.

src/encoded/static/components/cart/cart.js
forresttanaka: I noticed that just before code review; I had only converted existing code directly. I’ll change this if I decide to update the PR.

src/encoded/static/components/cart/cart.js
forresttanaka: No; more like, “The current cart.”

src/encoded/static/components/cart/description.js
forresttanaka: Yeah, really just a regular function.

src/encoded/static/components/cart/description.js
forresttanaka: Is there another way?

src/encoded/static/components/cart/description.js
forresttanaka: ?

src/encoded/static/components/cart/description.js
forresttanaka: No real LHS here, as there’s no assignment. Do you mean the left of the `&&`? The `<p>` gets called if and only if `cart.status` has the value “listed.”

src/encoded/static/components/cart/menu.js
forresttanaka: I originally called the little icon a cart “status” which now is confusing since the cart’s `status` property has new meaning. `status === undefined` means no little icon to display.

src/encoded/static/components/cart/menu.js
forresttanaka: Not sure how that would help. If the cart isn’t locked, then it displays the “Clear cart” button.

src/encoded/static/components/cart/cart.js
forresttanaka: I can see that argument. I’ll leave it since this is old code.

src/encoded/static/components/cart/clear.js
forresttanaka: I keep changing how I name event handlers, but I think “ClearCart” helps here instead of just “Clear” which might still leave people wondering “Clear what? The cart? A field?”

src/encoded/static/components/cart/description.js
forresttanaka: It strips anything within a script tag, so it seems unlikely to ever come across anything truthy or falsy AFAIK.

src/encoded/static/components/cart/description.js
forresttanaka: I think it’s likely it was named `dangerouslySetInnerHTML` because you _could_ pass it executable code, not that it conjures executable code in markup that has been stripped of it.

src/encoded/static/components/cart/share.js
forresttanaka: I’ve started to do this to make it more obvious that some imports are out of order, not that being out of order is a technical issue, but I just prefer not to have library imports below local ones, yet I was doing that sometimes. These comments help me prevent that.

src/encoded/static/components/objectutils.js
forresttanaka: This is a literal slash; not division, e.g. “50 / 1000.”

src/encoded/static/scss/encoded/modules/_cart.scss
forresttanaka: I copied this from a post with a URL I forgot to include that was for resetting list styles. So, I’m not sure in this case; I presume this had been tested, and I tested it as well. Not sure if there are cases this wouldn’t work.

src/encoded/static/components/cart/clear.js
zoldello: Got point.

src/encoded/static/components/cart/description.js
zoldello: It is called "dangerouslySetInnerHTML" because react designers believe an usual function name will force developer to put in some thought before using it. That is what I read about it. Now,  on my comment, my understanding is that sanitized function already took out the security-risk code. If that is true, do you need to use "dangerouslySetInnerHTML" to again sanitize the code. It also does not hurt to double secure security-risk code.

src/encoded/static/components/cart/description.js
zoldello: Wow. That came out wrong.  "strong tag" (\<strong\>) mixed html with css. Is that good practice?

src/encoded/static/components/cart/share.js
zoldello: Let's talk about this off-line sometimes. Imports cannot be out of order as package management handles that for you. Developers should get an error if they try to do something like call an internal package before an external one.

src/encoded/static/components/objectutils.js
zoldello: Opps. Thanks.

zoldello (dismissed):
Code looks fine. My comments are philosophical and don't matter. The last two changes are cosmetic; so nothing significant changed.

forresttanaka:
The cart status used the current cart object instead of the normalized `cart` object so it could display the wrong thing on shared-cart displays. It now always shows the normalized `cart` object status. I also changed the “Listing” button to say “List cart” for unlisted carts, or “Unlist cart” for listed carts.

forresttanaka:
Preview demo with the two new changes: https://encd-6191-cart-sharing-pv2-fytanaka.demo.encodedcc.org/

src/encoded/tests/features/cart.feature
```diff
@@ -82,7 +82,7 @@ Feature: Cart
         Then I should not see an element with id "description-editor"
 
     Scenario: Cart listing
-        When I press "Listing"
```
zoldello: Do you need to test for unlisted?

zoldello (approved):
Code looks good. My comment is not significant.

PR #4186, opened 2021-11-09
ENCD-6232 assay title

gabdank:


iyoungworth (changes requested):
The code looks correct for changing the assay_title calculation, test looks good.
I think you have to make similar changes in types/dataset.py though, since it looks like it's iterating through the assay_title from each item in related_datasets-
https://github.com/ENCODE-DCC/encoded/blob/ENCD-6232-assay-title/src/encoded/types/dataset.py#L1309
As well as search.js- https://github.com/ENCODE-DCC/encoded/blob/ENCD-6232-assay-title/src/encoded/static/components/search.js#L945

gabdank:
@iyoungworth I don't think we need it in related_datasets in general, this is done mostly to make FC series "closer" to datasets (to not confuse with Controls just because they contain controls (or pooled clone sequencing)).

And in regard to the UI, would not that rely in the calculated property of FC series?

iyoungworth (approved):
New change looks good to me, no alterations to previous assay_title code changes.

PR #4189, opened 2021-11-17
ENCD-5890-adjust-pulse-chase-time-series-page

iyoungworth:


gabdank (dismissed):

src/encoded/static/components/dataset.js
```diff
@@ -2201,6 +2201,67 @@ const functionalCharacterizationSeriesTableColumns = {
...
 
+
+/**
+ * Generate a comma-separated string of pulse-chase time durations from the given experiment's biosamples.
+ * @param {object} experiment - The experiment to generate the pulse-chase durations from.
+ * @returns {string} - A comma-separated string of pulse-chase durations.
+ */
+const computePulseDurations = (dataset) => {
+    const biosamples = collectDatasetBiosamples(dataset);
+    let pulseDurations = '';
```
forresttanaka: pulseDurations winds up being an array, so should initialize this to an empty array:
```
let pulseDurations = [];
```

src/encoded/static/components/dataset.js
```diff
@@ -2201,6 +2201,67 @@ const functionalCharacterizationSeriesTableColumns = {
...
+    if (biosamples.length > 0) {
+        biosamples.forEach((biosample) => {
+            if (biosample.pulse_chase_time) {
+                pulseDurations = `${biosample.pulse_chase_time} ${biosample.pulse_chase_time_units}${biosample.pulse_chase_time > 1 ? 's' : ''}`;
```
forresttanaka: This line reassigns a new string to the same `pulseDurations` string over and over, but I think you mean to add it to an array of strings, like:
```
pulseDurations.push(`${biosample.pulse_chase_time} ${biosample.pulse_chase_time_units}${biosample.pulse_chase_time > 1 ? 's' : ''}`);
```

src/encoded/static/components/dataset.js
```diff
@@ -2201,6 +2201,67 @@ const functionalCharacterizationSeriesTableColumns = {
...
+            }
+        });
+    }
+    pulseDurations = _.uniq(pulseDurations);
+    return pulseDurations;
+};
```
forresttanaka, line 2295: It’d be more efficient to make an array of `biosample.pulse_chase_time` values that contain no duplicates and then compose strings from those so that you don’t compose all these strings only to potentially throw some of them away with the `_.uniq()` call. But this is low-cost enough that we probably wouldn’t even be able to measure the time difference. So if you prefer doing it this way, it’s not a problem; just for future ideas.

src/encoded/static/components/dataset.js
```diff
@@ -2201,6 +2201,67 @@ const functionalCharacterizationSeriesTableColumns = {
...
+        getValue: (experiment) => computePulseDurations(experiment),
+    },
+
+    biosample_summary: {
+        title: 'Biosample summary',
+    },
+
+    lab: {
+        title: 'Lab',
+        getValue: (experiment) => (experiment.lab ? experiment.lab.title : null),
```
forresttanaka: With the demise of Internet Explorer, you can use this construction for expressions like these:
```
getValue: (experiment) => experiment?.lab.title
```
Which means, “Grab the `lab` property but only if `experiment` exists. If `experiment` doesn’t exist, return `undefined`.” React treats `undefined` the same as `null` these days (didn’t used to). `title` is a required property of `lab` so that doesn’t need the same treatment. I do notice `lab` is a required property of `experiment`, so does this need that check at all? In some cases the answer is “yes” depending on how the object gets retrieved.

src/encoded/static/components/dataset.js
```diff
@@ -3348,32 +3409,38 @@ globals.contentViews.register(TreatmentTimeSeries, 'TreatmentTimeSeries');
...
-    const treatmentAmounts = collectTreatmentAmounts(context.related_datasets);
-    const options = {};
-    if (treatmentAmounts.length > 0) {
-        options.Treatments = <>{treatmentAmounts.join(', ')}</>;
+    const durations = computePulseDurations(context.related_datasets);
+    const options = {
+        getSupplementalShortLabel: (dataset) => computePulseDurations(dataset),
+    };
+    if (durations.length > 0) {
+        options.Durations = <>{durations.join(', ')}</>;
```
forresttanaka, line 3526: After building this `options` object, it looks like you don’t use it, passing a new, different object to the `options` property in the `Series` call.

forresttanaka (changes requested):
Thanks for commenting the new function!

src/encoded/static/components/dataset.js
iyoungworth: Done

src/encoded/static/components/dataset.js
iyoungworth: Ah yes, thank you! Used your suggested line instead

src/encoded/static/components/dataset.js
iyoungworth: I think the value in keeping it this way is to make sure uniqueness checks all combinations of pulse_chase_time **and** pulse_chase_time_units (so that 2 minutes and 2 hours don't have the '2' removed?)
But there might be a way to do it by getting the values as tuples and then checking for uniqueness before making that a string? If we're not worried about efficiency, I'm happy leaving it as-is

forresttanaka: Yes, you’re right; I didn’t think about that.

src/encoded/static/components/dataset.js
iyoungworth: Ah I see, I added my 'suppressDonorDiversity' line to the earlier declaration of `options`

src/encoded/static/components/dataset.js
iyoungworth: Fancy! I adjusted to your suggested syntax, and kept the line- it might be that the check is needed in this context, since many other SeriesColumns definitions in this file have that line

iyoungworth:
I rebased to get the changes from 6088, and additionally:
- updated dataset.js to have comma-separation with different values within a dataset (the biosamples in a dataset should be the same, but just in case they aren't...)
- reworked the uniqueness check on search.js to start from time + units (now the display is slightly less attractive if the same units are used more than once and will say "2 hours, 6 hours" but will also correctly say "2 minutes, 2 hours"). Please let me know if you have thoughts on how to improve that piece, but this at least will avoid incorrect content in the appearance on search with mixed units

forresttanaka (dismissed):

gabdank (dismissed):

iyoungworth:
The new commit is a 1-line change to fix the breadcrumbs for this Series type (previously it was leading to /series-search/)

forresttanaka (approved):

PR #4191, opened 2021-11-18
ENCD-6152-change-mechanism-for-setting-current-cart

forresttanaka:
Preview demo: ~https://encd-6152-cart-current-pv1-fytanaka.demo.encodedcc.org/~


zoldello (approved):
Code looks good. It looks a bit awkward to replace the link with a button, but maybe that is just me.

PR #4192, opened 2021-11-19
ENCD-6242 For AWS_* env vars to work use None instead of 'default'

lrowe:
Running deploy on codespaces currently errors with:
```
botocore.exceptions.ProfileNotFound: The config profile (default) could not be found
```

This happens because AWS credentials are configured as secrets on GitHub Codespaces rather than as a profile in ~/.aws/credentials.

It can be fixed by specifying profile_name=None instead of 'default' in deploy.py. The default profile is still used if environment variables are not present.

Tested by launching a demo from both codespaces with AWS_* environment variables and locally with my ~/.aws/credentials.

### DIFF
--- a/src/encoded/commands/deploy.py
+++ b/src/encoded/commands/deploy.py
@@ -60,7 +60,7 @@ def _tag_ec2_instance(
         cluster_name,
         image_id,
         role='demo',
-        profile_name='default',
+        profile_name=None,
         dry_run=False,
         arm_arch=False,
         cluster_master=False,
@@ -92,7 +92,7 @@ def _tag_ec2_instance(
         'detailB': 'notused',
         'detailC': 'notused',
     }
-    if profile_name == 'default': 
+    if profile_name != "production": 
         if role == 'rc':
             tags_dict['Role'] += '-new-rc'
             tags_dict['section'] += '-rc'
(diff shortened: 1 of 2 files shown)


paul-sud (dismissed):

src/encoded/commands/deploy.py
```diff
@@ -1153,6 +1153,9 @@ def hostname(value):
     # arm arch only available on demo
     if not args.role == 'demo' and args.arm_image_id:
         raise ValueError('Arm architecture is only available on demos')
+    # For AWS_* env vars to work use None instead of 'default'
+    if args.profile_name == 'default':
```
keenangraham: Isn't logic like https://github.com/ENCODE-DCC/encoded/blob/ade188d6fd9dd181cf06c184fe6a6a64dc284195/src/encoded/commands/deploy.py#L95 affected by this?

keenangraham (changes requested):

src/encoded/commands/deploy.py
keenangraham: Updated comment.

paul-sud: Yeah you're right, I think that's the only spot that's affected by the change here. @lrowe maybe it would be better to set the profile name to `None` when creating the EC2 client here: https://github.com/ENCODE-DCC/encoded/blob/ade188d6fd9dd181cf06c184fe6a6a64dc284195/src/encoded/commands/deploy.py#L395

lrowe: I've adjusted conditions to use `profile_name != 'production'` and made profile_name None by default.

paul-sud (approved):

PR #4196, opened 2021-11-30
ENCD-6247 adding control back

gabdank:


src/encoded/types/functional_characterization_experiment.py
```diff
@@ -205,7 +205,7 @@ def perturbation_type(self, request, assay_term_name, replicates=None):
     })
     def crispr_screen_readout(self, request, assay_term_name, examined_loci=None, control_type=None):
         crispr_screen_readout = None
-        if assay_term_name in ['proliferation CRISPR screen', 'FACS CRISPR screen', 'Flow-FISH CRISPR screen']:
+        if assay_term_name in ['proliferation CRISPR screen', 'FACS CRISPR screen', 'Flow-FISH CRISPR screen', 'CRISPR screen']:
             # Return a specific CRISPR screen readout if there is only one method used in examined_loci, no examined_loci at all indicates a growth screen
             if control_type == 'control':
```
iyoungworth: L210-211 looks like it would need to be removed in order to calculate readout on controls

iyoungworth:
perturbation_type property looks like it should now include controls, but readout calculation removes controls in line 210-211

iyoungworth (changes requested):
perturbation_type property looks like it should now include controls, but readout calculation removes controls in line 210-211

iyoungworth (approved):
Code no longer additionally calculates 'proliferation' aspect of title (should already be specified in assay_term_name when appropriate); readout calculation only draws from specific expression_measurement_method entries collected on examined_loci

src/encoded/types/functional_characterization_experiment.py
cricketsloan: would it make sense to change this to be if "CRISPR Screen" in assay_slims?  Simpler code, but less specific in what you are doing.

PR #4199, opened 2021-12-02
ENCD-6234-add-description-fcs-search

khzlin:


forresttanaka (dismissed):

src/encoded/static/components/search.js
```diff
@@ -1024,6 +1043,11 @@ const SeriesComponent = ({ context: result, cartControls, removeConfirmation, au
                             </span>
                         : null}
                     </a>
+                    {fccSeries && (result.description || result.biosample_summary) ?
+                        <div className="result-item__highlight-row">
+                            {result.description ? <div>{result.description}</div> : result.biosample_summary ? <BiosampleSummaryDisplay summary={result.biosample_summary} organisms={organismName} /> : null}
```
forresttanaka: This nested conditional expression is OK, but generally not preferable. React now (probably as of quite a while ago, but I only found out recently) that you can use undefined instead of null to render nothing, so I _think_ this would work and would be easier (if it does):
```
{(result.description && <div>{result.description}</div>) || (result.biosample_summary && <BiosampleSummaryDisplay summary={result.biosample_summary} organisms={organismName} />)}
```
If `description` exists, then it renders the description and ignores the rest of the line. If `description` is undefined, that whole parenthized section evaluates to false and it goes to the next expression after the `||`. If `biosample_summary` is undefined then nothing at all gets rendered. If `biosample_summary` exists then BiosampleSummaryDisplay gets called.

I do write nested conditional expressions from time to time, but only for really simple cases with a couple booleans or something like that. I feel like this looks a bit too complex. You can ignore this because what you have is correct, but it goes beyond where I personally would draw the line.

src/encoded/static/components/search.js
```diff
@@ -961,6 +975,11 @@ const SeriesComponent = ({ context: result, cartControls, removeConfirmation, au
     }
     if (assays.length > 0) {
         assays = assays.filter((item) => item !== 'pooled clone sequencing');
+        if (perturbationType?.length === 0) {
```
forresttanaka: Technically, `perturbationType` always exists because it gets initialized to `[]` at the top of this function, so `perturbationType.length is fine, without the ?.

src/encoded/static/components/search.js
```diff
@@ -961,6 +975,11 @@ const SeriesComponent = ({ context: result, cartControls, removeConfirmation, au
...
+            assaysTitle = _.uniq(assays);
+        } else if (perturbationType?.length > 0) {
```
forresttanaka: You already know `perturbationType.length` is > 0 because you tested against 0 above. You can use just:
```
} else {
```

src/encoded/static/components/search.js
```diff
@@ -961,6 +975,11 @@ const SeriesComponent = ({ context: result, cartControls, removeConfirmation, au
...
+            assaysTitle = [...new Set(assaysTitle = assays?.map((a) => a.replace('CRISPR ', '')))];
```
forresttanaka: You don’t need the ? in `assays?.map` because you already know `assays` exists. I’m not clear on the assignment to `assaysTitle` within the `new Set()`. I think you can leave that out since you’re assigning to `assaysTitle` on the left of the `=` anyway. Unless Javascript does something I’m not aware of, you shouldn’t have an assignment within `new Set()`.

src/encoded/static/components/search.js
```diff
@@ -961,6 +975,11 @@ const SeriesComponent = ({ context: result, cartControls, removeConfirmation, au
     }
     if (assays.length > 0) {
         assays = assays.filter((item) => item !== 'pooled clone sequencing');
+        if (perturbationType?.length === 0) {
+            assaysTitle = _.uniq(assays);
```
forresttanaka: Just for symmetry with the other `if` case, you could use `[...new Set(assays)]` here instead of `_.uniq()`. You could even compress this into a single assignment without the `if` but I think with the `if` makes this clearer.

src/encoded/static/components/search.js
```diff
@@ -961,6 +975,11 @@ const SeriesComponent = ({ context: result, cartControls, removeConfirmation, au
...
+        } else if (perturbationType?.length > 0) {
+            assaysTitle = [...new Set(assaysTitle = assays?.map((a) => a.replace('CRISPR ', '')))];
+        }
         assays = _.uniq(assays);
```
forresttanaka, line 982: Do we still need this? The `Assays` data row that displays this seems nearly redundant now, displaying the same thing but without “CRISPR.”

src/encoded/static/components/search.js
```diff
@@ -506,25 +506,30 @@ const ExperimentComponent = (props, reactContext) => {
...
                     </a>
-                    {result.biosample_summary ?
+                    {!isFunctionalExperiment && result.biosample_summary ?
                         <div className="result-item__highlight-row">
                             {result.biosample_summary ? <div><BiosampleSummaryDisplay summary={result.biosample_summary} organisms={organismNames.concat(GeneticModificationOrganismNames(biosamples))} /> </div> : null}
                         </div>
                     : null}
+                    {isFunctionalExperiment && (result.description || result.biosample_summary) ?
+                        <div className="result-item__highlight-row">
+                            {result.description ? <div>{result.description}</div> : result.biosample_summary ? <div><BiosampleSummaryDisplay summary={result.biosample_summary} organisms={organismNames.concat(GeneticModificationOrganismNames(biosamples))} /> </div> : null}
```
forresttanaka: I have a comment below for Series that applies similarly to this line.

forresttanaka (changes requested):
Could you delete search.js line 812, the assignment to `biosampleclassifications`? This variable doesn’t get used anymore, so it shows up as an ESLint error.

src/encoded/static/components/search.js
khzlin: ok

src/encoded/static/components/search.js
khzlin: ok

src/encoded/static/components/search.js
khzlin: ok

src/encoded/static/components/search.js
khzlin: ok

src/encoded/static/components/search.js
khzlin: yeah we still do because L1055 still uses it. That's where the full assay titles with 'CRISPR' is displayed.

src/encoded/static/components/search.js
khzlin: changed as suggested. ok

src/encoded/static/components/search.js
khzlin: changed as suggested from comment below. ok

src/encoded/static/components/search.js
```diff
@@ -1043,9 +1042,9 @@ const SeriesComponent = ({ context: result, cartControls, removeConfirmation, au
                             </span>
                         : null}
                     </a>
-                    {fccSeries && (result.description || result.biosample_summary) ?
+                    {fccSeries ?
```
forresttanaka, line 1045: Do you not need `(result.description || result.biosample_summary)`? You might not; it depends whether the empty `<div className="result-item__highlight-row">` has an impact or not.

forresttanaka (approved):
Looks good! Just one question (though it applies to two lines), because I’m not sure what happens if you don’t have a description nor a summary.

PR #4201, opened 2021-12-04
ENCD-6250 audit readout

gabdank:


iyoungworth (approved):
The removal of proliferation and qPCR from auditing looks fine- 'proliferation' is already no longer calculated in crispr_screen_readout (so checking the title against it is not meanginful), and we have no 'qPCR' entries and don't expect to get any.
'qPCR' should be removed from the examined_loci.expression_measurement_method enum list, but this can be handled in a future ticket (any new qPCR submissions would fail to be audited here, but not cause a schema error).

cricketsloan (approved):
I think that we now need to remove crispr_readout as field, but if we will do that in the next release, that is fine.

PR #4203, opened 2021-12-07
Encd 6016 bad filters

mingjiecn:


forresttanaka (changes requested):
You have an ESLint error to fix before code review: https://app.circleci.com/pipelines/github/ENCODE-DCC/encoded/14837/workflows/a8bf2bbe-656e-4efb-8635-0f3a35e59423/jobs/72032

src/encoded/static/components/filegallery.js
```diff
@@ -3580,7 +3580,10 @@ class FileGalleryRendererComponent extends React.Component {
         // If filters other than assembly are chosen, we want to highlight the filtered files
         let highlightedFiles = [];
         if (Object.keys(this.state.fileFilters).length > 1) {
-            highlightedFiles = this.filterForInclusion(this.state.files);
+            highlightedFiles = this.filterForInclusion(this.props.data);
+            Object.keys(this.state.fileFilters).forEach((fileFilter) => {
+                highlightedFiles = filterItems(highlightedFiles, fileFilter, this.state.fileFilters[fileFilter]);
```
forresttanaka, line 3585: Initially looks odd assigning to `highlightedFiles` every time through the loop, but since that’s also a parameter, I guess it gradually gets more and more filtered, so that looks like it makes sense.

forresttanaka (dismissed):
I don’t see any issues either with how this looks or in some simple testing I did locally. A couple notes:

* The facet-term counts can appear wrong (e.g. ENCSR000AEN says five fastq files, but that’s the total for the experiment — not the number appearing in the graph). I’m assuming this isn’t part of the ticket.
* Facet terms don’t necessarily appear in graphed files, though I’m again assuming fixing that isn’t part of this ticket (e.g. ENCSR000AEN has one index read file because it exists in the experiment but not in the graph), again because they exist in the experiment but not in the graph. Notably, if you filter on “index reads” on the File Details tab of ENCSR000AEN, nothing appears even though it says “Displaying 1 of 1 files” which is a separate bug from this ticket’s.

PR #4211, opened 2021-12-09
ENCD-5946 immune cells rebased

emro:
Add immune cell page with clickable node graph
Mint/Histone columns are combined on results table


src/encoded/static/components/app.js
```diff
@@ -275,6 +276,7 @@ const fullPageReloadPaths = [
     '/single-cell/',
     '/summary/',
     '/encyclopedia/',
+    '/immune-cells/',
```
forresttanaka, line 279: I was hoping we don’t add more new pages to this list, but as I still don’t know why these pages require a full reload, I’m not sure if there’s a good reason for it.

src/encoded/static/components/immune_cells.js
```diff
@@ -0,0 +1,1073 @@
...
+
+    // If targetAssays only contains nulls, then just empty it so we can skip the targetAssay row.
+    if (!targetAssays.some((assay) => assay !== null)) {
+        targetAssays = [];
+    }
+
+    return { colMap, targetAssays };
+};
+
+const convertExperimentToDataTable = (context, expandedRowCategories, expanderClickHandler, clearClassifications) => {
```
forresttanaka: I guess they don’t really impact the operation, but I think on this page you don’t use `expandedRowCategories` and `expanderClickHandler` at all?

src/encoded/static/components/immune_cells.js
```diff
@@ -0,0 +1,1073 @@
...
+    const colSubcategory = context.matrix.x.group_by[1];
+    const rowCategory = context.matrix.y.group_by[0];
+    const rowSubcategory = context.matrix.y.group_by[1];
+    const rowKeys = [];
+    const headerRows = [];
+
+    const colQuery = new QueryString(context.search_base.replace('&config=immune', ''));
+    colQuery.deleteKeyValue('biosample_ontology.cell_slims');
+    colQuery.deleteKeyValue('biosample_ontology.classification');
+    colQuery.deleteKeyValue('biosample_ontology.system_slims');
```
forresttanaka: You can also chain these QueryString methods, just to save some typing.

src/encoded/static/components/immune_cells.js
```diff
@@ -0,0 +1,1073 @@
...
+    const colSearchBase = `${colQuery.format()}`;
+
+    const query = new QueryString(context.search_base.replace('&config=immune', ''));
```
forresttanaka: To save a _little_ compute and typing, you can make `query` right after you create `colQuery` by cloning `colQuery`:
```
const query = colQuery.clone()
```
…since both use the same URI. In addition, `QueryString` expects a clean query string like “a=1&b=2” but since you pass it an entire URI, the first query-string element in `colQuery` and `query` is `/search/?type` instead of just `type`. Now, it still works because you’re not doing a `deleteKeyValue('type')` or `getKeyValue('type')`, so `format()` happens to still work, but the _intent_ is to pass clean query strings to `new QueryString()`.

src/encoded/static/components/matrix_deeply_profiled.js
```diff
@@ -575,3 +575,5 @@ DeeplyProfiledMatrix.contextTypes = {
 };
 
 globals.contentViews.register(DeeplyProfiledMatrix, 'DeeplyProfiledMatrix');
+
+export default getNavbarHeight;
```
forresttanaka: I think makes sense to move `getNavbarHeight` to matrix.js or even globals.js so everyone can use it.

src/encoded/static/components/immune_cells.js
```diff
@@ -0,0 +1,1073 @@
...
+        });
+    }
+
+    toggleFacets() {
+        this.setState((prevState) => ({ facetsOpen: !prevState.facetsOpen }));
+    }
+
+    redirectClick(e) {
+        const clickedLink = e.target.closest('a');
+        const isCollapseButton = findElement(e.target, 'collapse-body-map', 'classList');
```
forresttanaka, line 889: Unless there’s an unusual case (might be), usually React refs can let you do this more easily.

src/encoded/static/components/search.js
```diff
@@ -1464,6 +1464,14 @@ export const FacetList = (props) => {
...
     }, [context.facet_groups, prevFacetGroups]);
 
+    if (props.context.title === 'Immune Cells' && facets.length === 0) {
+        return (
+            <>
+                {props.additionalFacet}
+            </>
+        );
+    }
+
```
forresttanaka: It sounded like you only needed this for local testing, so among my CR comments, I’d say removing this is the one required change. Nothing in these facet functions should specifically serve just one page’s facets — only a custom facet run through the facet registry should.

forresttanaka (changes requested):
Among all my CR comments, I would say the change in search.js is the only one requring a change. The page looks really cool though!

The home page ticket requires a change too all matrix page headers. Hopefully your ticket gets merged first, and I can make that change before putting the new home page in.

src/encoded/static/components/immune_cells.js
emro: I really should have cleaned this up; I got rid of a lot of extra code just now.

src/encoded/static/components/search.js
emro: Found a different solution for this.

src/encoded/static/components/matrix_deeply_profiled.js
emro: I actually didn't need this so I think the next time someone else needs this function then it should be moved to globals.js.

src/encoded/static/components/immune_cells.js
```diff
@@ -1039,6 +1039,8 @@ class ImmuneCells extends React.Component {
                                         }
                                         key={this.state.availableNodes.length}
                                     />
+                                : (this.state.facetsOpen) ?
+                                    <MatrixGraph context={this.state.pageContext} setFilters={this.setNodeFilters} availableNodes={this.state.availableNodes} key={this.state.nodes.length} selectedNodes={this.state.nodes} />
```
forresttanaka, line 972: Nice! Thanks for doing that. This can run afoul of the many who hate nested conditional expressions, but I’ve never had an issue with those as long as it’s clear, and you wrote it clearly here.

forresttanaka (dismissed):
Looks great!

src/encoded/static/libs/ui/node_graph.js
```diff
@@ -26,14 +26,16 @@ export const nodeKeyName = (name) => name.replace(/\s/g, '').replace(/[\W_]+/g,
...
+const nodeLabel = (name) => name.replaceAll('activated', 'ϟ')
+    .replaceAll('stimulated', '☆')
+    .replaceAll('alpha', 'α')
+    .replaceAll('beta', 'β')
+    .replaceAll('delta', 'δ')
+    .replaceAll('gamma', 'γ')
+    .replaceAll('-positive,', '+')
+    .replaceAll('-negative,', '-')
+    .replaceAll('-positive', '+')
+    .replaceAll('-negative', '-');
```
emro, line 38: I don't really expect there to be two instances of "stimulated" etc but it seemed like the overly cautious approach and that's why some "positive"s and "negative"s weren't getting replaced.

forresttanaka (dismissed):
OK, sounds reasonable.

forresttanaka (approved):
Looks like that should work

PR #4213, opened 2021-12-10
ENCD-6256-scatac

gabdank:


src/encoded/schemas/sc_atac_alignment_quality_metric.json
```diff
@@ -180,7 +180,8 @@
                     "title": "MIME type",
                     "type": "string",
                     "enum": [
-                        "text/plain"
+                        "text/plain",
```
iyoungworth, line 183: From the ticket description, should text/plain be removed completely? Do we want to allow both?

iyoungworth: Decided to allow both, approved.

iyoungworth (approved):

PR #4217, opened 2021-12-13
ENCD-6229 fix unnamed Valis tracks

emro:
Fix tracks that have no short label by adding output type as the label


forresttanaka (approved):
Looks good!

You can also use `&&` for components now. Not sure what version of React started treating undefined and null the same, and I didn’t know about it until recently, like:
```
{cartShortLabel && <li>{file.output_type}</li>}
```
Doesn't really make a difference. It just saves a little typing.

emro:
That's cool! I didn't know about this.

PR #4224, opened 2021-12-17
ENCD-6016 fix bad filters

mingjiecn:


forresttanaka (approved):
Looks fine, as before.

PR #4226, opened 2022-01-04
ENCD-6104-fix-series-table-accession-links

forresttanaka:


src/encoded/static/libs/user_roles.js
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * Returns true if the current user created their own account.
+     */
+    get isUnprivileged() {
+        return this._roles.length === 1 && this._roles[0] === 'unprivileged';
+    }
+
+    /**
+     * Returns true if the current user is a submitter or admin.
```
zoldello, line 54: technically, this only checks for submitter.

src/encoded/static/libs/user_roles.js
```diff
@@ -0,0 +1,73 @@
...
+     */
+    get isPrivileged() {
+        return this._roles.includes('submitter');
+    }
+
+    /**
+     * Returns true if the current user is a lab submitter.
+     */
```
zoldello, line 62: Can a submitter not also be an admin?

src/encoded/static/libs/user_roles.js
```diff
@@ -0,0 +1,73 @@
...
+    /**
+     * Returns true if the current user created their own account.
+     */
+    get isUnprivileged() {
+        return this._roles.length === 1 && this._roles[0] === 'unprivileged';
+    }
+
+    /**
+     * Returns true if the current user is a submitter or admin.
+     */
```
zoldello, line 55: Maybe it is just me - I could no tell what this does looking at the name. Like,  privilege to do what?

zoldello (approved):
Code looks good.  I recommend you look at my comments and decide if they can be dismissed or not.

src/encoded/static/libs/user_roles.js
forresttanaka: If you have 'admin' in the roles, then you also have 'submitter' in the roles, so checking only 'submitter' returns true for both submitters and admins.

src/encoded/static/libs/user_roles.js
forresttanaka: Just means submitter and admins, just in contrast to unprivileged (the only thing I’ve ever heard of logged-in users who have created their own accounts called) and not-logged-in users.

src/encoded/static/libs/user_roles.js
forresttanaka: Meaning specifically a submitter, not an admin. If you want both, then `isPrivileged`. If you want just admins, then `isAdmin`. If you want just submitters, then `isSubmitter`.

PR #4230, opened 2022-01-05
ENCD-6257 fastq dependency fix

mingjiecn:


hitz (dismissed):

mingjiecn:
added replicate requirement for output_type index reads, rejected reads and R2C2 subreads. Created test data to test locally. It should work now

hitz (approved):

PR #4231, opened 2022-01-07
ENCD-6266-add-immune-cells-and-human-donors-to-home-page

forresttanaka:
Mainly this ticket moves the “Immune cells” card for the new home page to a different section. It also adds a new “Human donors” card and adds its query used to highlight the card.

It also adds a new SVG icon that we hardcode in JSX.

The immune cells and human donor matrix pages also needed their headers fixed up as they used an older format that doesn’t work correctly anymore.

Because of the new cards, I had to update the .feature test.

Preview demo (a QA demo is coming up but won’t be ready until Monday):
~https://encd-6266-immune-cells-home-pv1-fytanaka.demo.encodedcc.org/~


mingjiecn (dismissed):

forresttanaka:
This update fixes a bug Idan found in which the query that generates the count on the “Functional Characterization” card has a different query string from the href for that card. I’ve now made them the same. I also corrected the query for the “Encyclopedia of Elements” card which doesn’t change the results, but prevents it from returning data that we don’t use. I also made one more query correction that makes no functional difference, but I think makes that code look confusing.

forresttanaka:
Updated demo: https://encd-6266-immune-cells-home-pv3-fytanaka.demo.encodedcc.org/

A fresh QA demo is coming up.

forresttanaka:
1. Correction to Functional Characterization query to remove TransgenicEnhancerExperiment. I’m also trying a regular search instead of a matrix search for deeply profiled cell lines
2. Not really a correction because I don’t use the `@id` property (inherited from site-wide search), but I think it looks confusing for that to have a different query string from `searchUrl`. Plus, I could use `@id` someday, unlikely though that is.
3. Restore the styling of the “HUMAN” and “MOUSE” buttons on the award pages. I didn’t realize these pages used styles from _home.scss
4. Restored the help icons on the ExperimentSeries pages by converting them from icon fonts to the same SVG I use on the home page, and adjusting the styles so that the home-page help icon appears darker, while the standard ones appear lighter.

Updated preview demo: ~https://encd-6266-home-fixes-pv4-fytanaka.demo.encodedcc.org/~

mingjiecn:
the help icon background for "Search the ENCODE Portal" is different than the help icon for "Search SCREEN". Should they be the same?
image
image

forresttanaka:
@mingjiecn They should appear the same way. I need to fix that.

forresttanaka:
Fresh QA demo: https://encd-6266-home-fixes-qa2-fytanaka.demo.encodedcc.org

mingjiecn (approved):

PR #4238, opened 2022-01-19
ENCD-6271-fix-award-labels

forresttanaka:
We had been using a now-deleted CSS class, which wasn’t the correct one to use to begin with.


forresttanaka:
Preview demo and an award that shows both problems:
~https://encd-6271-award-labels-pv1-fytanaka.demo.encodedcc.org/awards/UM1HG009390/~

mingjiecn (approved):

PR #4241, opened 2022-01-20
ENCD-6269 pin Valis summary page location

emro:
There are two main updates:
- in cases where `examined_loci` or `examined_regions` exist, those fields determine a default location for the browser
- files of output_type "guide locations" require additional track height

### DIFF
--- a/src/encoded/static/components/filegallery.js
+++ b/src/encoded/static/components/filegallery.js
@@ -3678,6 +3678,13 @@ class FileGalleryRendererComponent extends React.Component {
             meta = globals.graphDetail.lookup(this.state.infoNode)(this.state.infoNode, this.handleNodeClick, auditIndicators, auditDetail, this.context.session, this.context.sessionProperties);
         }
 
+        let defaultLocation = null;
+        if (context.examined_regions) {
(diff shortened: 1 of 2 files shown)


src/encoded/static/components/filegallery.js
```diff
@@ -3678,6 +3678,13 @@ class FileGalleryRendererComponent extends React.Component {
             meta = globals.graphDetail.lookup(this.state.infoNode)(this.state.infoNode, this.handleNodeClick, auditIndicators, auditDetail, this.context.session, this.context.sessionProperties);
         }
 
+        let defaultLocation = null;
+        if (context.examined_regions) {
+            defaultLocation = context.examined_regions[0];
+        } else if (context.examined_loci && context.examined_loci[0].gene && context.examined_loci[0].gene.locations) {
```
forresttanaka, line 3684: You could use the ?. syntax to simplify this, though I’d guess transpiling converts it to what you have.

forresttanaka (approved):
Cool. When they assigned it to me I had no idea what had to be done, so good to see it’s not real complicated.

PR #4242, opened 2022-01-20
ENCD-6264-display-annotation-trait

khzlin:


src/encoded/static/components/dataset.js
```diff
@@ -200,6 +200,13 @@ const AnnotationComponent = (props, reactContext) => {
                                 </div>
                             : null}
 
+                            {context.trait ?
+                                <div data-test="trait">
+                                    <dt>Trait</dt>
+                                    <dd>{context.trait}</dd>
+                                </div>
+                            : null}
+
```
forresttanaka, line 209: This is great, but FYI a more modern way of conditional JSX I’ve been doing lately is:
```
{context.trait &&
    <div>
        ...
    </div>
}
```
Meaning “If context.trait exists, render this. Otherwise, render “undefined” which is just like render “null” in more recent versions of React.

forresttanaka (approved):

PR #4243, opened 2022-01-20
ENCD-6278-qc-calculation

gabdank:


iyoungworth (dismissed):
Code looks like it should work as expected, I don't see specific corrections to make. Adding a test could be helpful but not strictly required.

iyoungworth (approved):

PR #4244, opened 2022-01-21
ENCD-6267-fcc-publication-filters-out-datapoint-experiments

forresttanaka:
The main complication with this ticket is that we now have to know which datasets within a publication qualify for display (based on their `datapoint` property) so that we know how many exist, but we don’t have those datasets to know whether they qualify until after the page has mounted.

To solve this, I do a request to the server for every dataset in the publication, but I only request the `datapoint` property for each, to reduce the size of the data we get from the server. We don’t use these objects for display — only to know which ones we _can_ display. We get the full objects for display as the user pages through the datasets.

### DIFF
--- a/src/encoded/schemas/functional_characterization_experiment.json
+++ b/src/encoded/schemas/functional_characterization_experiment.json
@@ -542,6 +542,9 @@
         "biosample_ontology.term_name": {
             "title": "Biosample term name"
         },
+        "datapoint": {
+            "title": "Datapoint"
+        },
         "description": {
             "title": "Description"
         },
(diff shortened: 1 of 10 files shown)


forresttanaka:
Here’s an example from a preview demo:
~https://encd-6267-fcc-publication-pv2-fytanaka.demo.encodedcc.org/publications/93103289-675a-489d-988a-e047f8348cc6/~

Compare with production:
https://www.encodeproject.org/publications/93103289-675a-489d-988a-e047f8348cc6/

The datasets with datapoint=true do not appear in this demo.

mingjiecn (dismissed):

forresttanaka:
Demo updated to handle Add All for series correctly.

https://encd-6267-fcc-publication-pv3-fytanaka.demo.encodedcc.org/

mingjiecn (approved):

PR #4246, opened 2022-01-25
ENCD-6147 biosample calc

gabdank:


src/encoded/tests/features/cart.feature
```diff
@@ -14,8 +14,7 @@ Feature: Cart
         Then I should see 3 elements with the css selector ".cart-toggle--in-cart"
         And I should see an element with the css selector ".cart__nav-button"
 
-        When I press "Data"
-        And I click the link with text that contains "High-throughput assays"
+        When I visit "/search/?type=FunctionalCharacterizationExperiment&type=FunctionalCharacterizationSeries&config=FunctionalCharacterization&datapoint=false&control_type!=*"
```
cricketsloan, line 17: By switching it to the link and not the navigation, are you skipping the testing of the navigation?  Come to think of it, now that there are two elements with the name "Experiment Search" will testing be a problem.

src/encoded/types/functional_characterization_experiment.py
```diff
@@ -234,6 +297,36 @@ def datapoint(self, request, related_series=None):
...
+                'replicates.library.biosample', 
+                include=[
+                    '@id',
+                    'library',
+                    'biosample',
+                ]
+            )
+            path.expand(request, properties)
+            
+            if properties['replicates']:
```
cricketsloan, line 323: This seems redundant as you have just assigned properties 2 actions above.  Maybe I am missing a logic flow.

src/encoded/types/functional_characterization_experiment.py
```diff
@@ -339,3 +432,4 @@ def biosample_summary(self, request, replicates=None, elements_references=None):
 
         if len(dictionaries_of_phrases) > 0:
             return construct_biosample_summary(dictionaries_of_phrases, sentence_parts)
+
```
cricketsloan: No newline. I am surprised the linter accepted that.

src/encoded/types/transgenic_enhancer_experiment.py
```diff
@@ -49,6 +49,7 @@ class TransgenicEnhancerExperiment(
         'biosamples.biosample_ontology',
         'biosamples.organism',
         'biosamples.characterizations',
+        'biosamples.treatments',
```
cricketsloan, line 52: Do we have treatments in TransgeneicEnhancerExperiments?

src/encoded/schemas/changelogs/functional_characterization_series.md
```diff
@@ -13,6 +13,7 @@
 * Added a calculated boolean property *datapoint* 
 * Modified *assay_title* calculated properties to not include Controls, unless the Series contains only Control experiments
 * Removed *control_type* calculated property
+* Added *biosamples* calculated property and removed *replicates* calculated property
```
cricketsloan: I think that this needs to be a new version.  By removing replicate we are breaking code the is in existence for someone.

cricketsloan (changes requested):
I left a few comments.  The only one I feel strongly on is that we need to version the series object.  Removing a field is not a small change.

src/encoded/tests/features/cart.feature
gabdank: yep, I am going around the fact that I can not uniquely press FCC Experiments vs regular experiments

src/encoded/types/functional_characterization_experiment.py
gabdank: just trying to avoid path.expand resulting in 'replicates' pointing to None, maybe it is not going to happen, but better safe than throwing exception trying to iterate over None.

src/encoded/types/functional_characterization_experiment.py
gabdank: fixed

src/encoded/types/transgenic_enhancer_experiment.py
gabdank: no we don't but I think theoretically they could be assigned, then I want them embedded

src/encoded/schemas/changelogs/functional_characterization_series.md
gabdank: upgraded to 4

cricketsloan (dismissed):

src/encoded/types/functional_characterization_experiment.py
```diff
@@ -234,6 +297,36 @@ def datapoint(self, request, related_series=None):
...
+    @calculated_property(schema={
+        "title": "Biosamples",
+        "type": "array",
+        "items": {
+            "type": 'string',
+            "linkTo": "Biosample",
+        },
+        "notSubmittable": True,
+    })
+    def biosamples(self, request, replicates=None):
```
cricketsloan, line 309: Is this function exactly the same for series and experiment?  Is there any way to share the function?

cricketsloan (approved):

PR #4249, opened 2022-01-27
ENCD-6279-add-output-type-accessibility

khzlin:


src/encoded/schemas/changelogs/file.md
```diff
@@ -20,7 +20,7 @@
 * Added *NcoI*, *MseI*, *MboI and MseI* to the enum list for *restriction_enzymes*
 * Added *kmer weights* and *training set* to the enum list for *output_type*
 * Added *functional conservation mapping* to the enum list for *output_type*
-* Added *3D structure* to the enum list for *output_type*, to be used with the new *file_format* *nucle3d*
+* Added *3D structure* and *DNA accessibility raw signal* to the enum list for *output_type*, to be used with the new *file_format* *nucle3d*
```
iyoungworth: The new output_type should probably be on a new line here, the way it's written makes it sound like it also has to be used with nucle3d file_format which is incorrect

iyoungworth (changes requested):
Minor change requested in File changelog, other code looks good

src/encoded/schemas/changelogs/file.md
khzlin: thanks for catching this

iyoungworth (approved):

PR #4257, opened 2022-02-05
upgrade-validator

keenangraham:


keenangraham:
Should update the frontend `profiles` code to use `$id` instead of `id`.

PR #4259, opened 2022-02-10
ENCD-6284 sescc bug fix

emro:
- Fixes styles on SESCC page which were getting over-written by immune cells page styles
- Fixes bugs noted on ticket (allows selection of nodes and loads table)
- Eliminates console error related to svgs
- Does not fix SESCC bug where report/list links don't preserve filters

### DIFF
--- a/src/encoded/static/components/immune_cells.js
+++ b/src/encoded/static/components/immune_cells.js
@@ -661,15 +661,13 @@ class MatrixGraph extends React.Component {
         });
     }
 
-    setSelectedNodes(newNode) {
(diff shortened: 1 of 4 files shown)


forresttanaka (dismissed):
I don’t see anything wrong. I couldn’t quite suss out what `activeBool` means, and wouldn’t know what to do with it if I needed to maintain this code. I think `textWrapWidth` as a parameter to `drawTree` would have let `drawTree` be more generic, but that’s already been established.

emro:
I don't know if we'll want anything like this for IGVF, but if we do having `textWrapWidth` as a parameter would be better.

forresttanaka (approved):
Thanks!

PR #4268, opened 2022-04-07
ENCM-25 add human body map to single cell

emro:
- Body maps for human and mouse tabs are displayed for the “High throughput” tab
- Organism facet is displayed for “Low throughput” and “Perturbed high throughput” tabs
- Status facet is added for all single-cell tabs, and status=released is added to each query
- Minor style updates are implemented for the body map facet for margin consistency across facets
- System slims are not displayed for the body map on this page only
- Tests are updated (note: I ended up deleting the "low throughput" test because I couldn't get it to pass but maybe Forrest has insight into how to fix it so it will pass?)

### DIFF
--- a/src/encoded/static/components/body_map.js
+++ b/src/encoded/static/components/body_map.js
@@ -794,12 +794,11 @@ class BodyMap extends React.Component {
                             </button>
                         </ul>
                     </div>
-                :
-                    <button type="button" className="clear-organs" onClick={this.clearOrgans}>
-                        <i className="icon icon-times-circle" />
-                        Clear body map selections
-                    </button>
-                }
+                : null}
+                <button type="button" className={`clear-organs ${this.props.displaySystems ? 'clear-organs-with-systems' : ''}`} onClick={this.clearOrgans}>
+                    <i className="icon icon-times-circle" />
+                    Clear body map selections
+                </button>
                 <div className="body-facet">
                     <div className="body-image-container">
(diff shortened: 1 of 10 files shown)


src/encoded/static/components/body_map.js
```diff
@@ -768,31 +768,38 @@ class BodyMap extends React.Component {
...
+                                Clear body map selections
+                            </button>
+                        </ul>
+                    </div>
+                :
+                    <button type="button" className="clear-organs" onClick={this.clearOrgans}>
+                        <i className="icon icon-times-circle" />
+                        Clear body map selections
+                    </button>
+                }
```
forresttanaka: This button is so similar to the button above, it’d be nice to make them one. At first I was under the false impression that with displaysSystems true, you wouldn't see a Clear button.

forresttanaka (changes requested):
If you check for the existence of `organismTabs`, it stops the crash when you visit the page locally:
```
{(selectedSeries === 'highThroughput' && organismTabs) ?
```

`organisms` === null when run locally.

I haven’t tried a local BDD test, but that should let the deleted tests run again. Better fixes likely exist, but this should do for now.

While I eventually figured it out, I did have difficulty understanding the role of `displaySystems`.

src/encoded/static/components/body_map.js
emro: I fixed this.

forresttanaka (approved):
Thanks

PR #4273, opened 2022-05-17
ENCM-40-dont-transform-limit-all

ottojolanki:


ottojolanki:
Not really sure what kind of tests would be good for this one. The existing ones don't give much clues.

keenangraham:
Code looks fine, have no idea what blast radius of this change is though. Demo will be important.

src/encoded/renderers.py
```diff
@@ -211,6 +212,10 @@ def canonical_redirect(event):
 
 
 def should_transform(request, response):
+    limit = request.params.get('limit')
+    if limit is not None and (limit == 'all' or int(limit) > MAX_NUMBER_OF_RENDERED_RESULTS):
```
keenangraham: I think this would trigger `ValueError` if you pass in string that isn't all? `limit=xyz`?

keenangraham: Maybe move these conditions into own function with name that says what it is doing? `if too_many_results_to_render():`?

keenangraham (changes requested):

src/encoded/renderers.py
ottojolanki: Fixed. Also added error handling for nonsensical and multiple limit values (these don't seem to reach this part of code in usual search cases, but just to be sure in case in some situation it would).

keenangraham (approved):

PR #4275, opened 2022-05-18
ENCM-39 replace view all links

emro:
I think there was only one place where `limit=all` was on a button (on the search page); I think all the other cases were cases where data was being fetched.

### DIFF
--- a/src/encoded/static/components/report.js
+++ b/src/encoded/static/components/report.js
@@ -6,6 +6,7 @@ import { Modal, ModalHeader, ModalBody, ModalFooter } from '../libs/ui/modal';
 import { encodedURIComponent } from '../libs/query_encoding';
 import * as Pager from '../libs/ui/pager';
 import QueryString from '../libs/query_string';
+import LimitSelector from '../libs/ui/limit-selector';
 import * as globals from './globals';
 import { TableItemCount } from './objectutils';
 import { FacetList } from './search';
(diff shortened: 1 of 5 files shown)


src/encoded/static/components/search.js
```diff
@@ -1903,51 +1902,48 @@ export const SearchControls = ({ context, visualizeDisabledTitle, showResultsTog
...
+    const query = React.useMemo(() => new QueryString(parsedUrl.query), [parsedUrl]);
+    // Get the current value of the "limit=x" query string parameter. No "limit=x" means the
+    // default value applies. The back end allows exactly zero or one "limit=x" parameter.
+    const pageLimit = React.useMemo(() => {
+        const limitValues = query.getKeyValues('limit');
+        return limitValues.length === 1 ? Number(limitValues[0]) || DEFAULT_PAGE_LIMIT : DEFAULT_PAGE_LIMIT;
+    }, [query]);
+    const pageLimitOptions = [25, 50, 100, 200];
+
     if (showResultsToggle) {
```
emro, line 1913: I also tried including the condition (context.total > results.length) so in that case if you filter down to 5 results, the box of results limits doesn't display at all. I decided that I didn't like having the results limits appear and disappear depending on selections. Similarly, I thought about not displaying the "200" if there were say 95 results total (having "100" be the max in that case), but thought it was better to just keep all the options for all situations. I can see the case for the other solution though.

forresttanaka: Yeah, seems reasonable to keep the controls there. If ENCODE was still in active development I might display it disabled, but this is probably fine as is. I doubt you’re getting any advantage from caching the query with `useMemo`. `context` only changes on user command, and the calculation for `parsedUrl` and `query` is pretty simple to do on every render. The caches will just take up memory, though they’re not very big either.

src/encoded/static/components/report.js
```diff
@@ -936,5 +952,7 @@ Report.contextTypes = {
     navigate: PropTypes.func,
 };
 
+export default PageLimitSelector;
```
forresttanaka: I usually think of the default export as the file’s purpose, but `PageLimitSelector` is more of a utility. I think putting it in its own file now that it’s shared would make sense, and default export it from there. Though with encoded in maintenance mode, this isn’t likely to cause future issues either.

forresttanaka (dismissed):
Looks good. I put in a couple of notes that I would have done differently, but I’m fine the way it is too — they don’t cause bugs. I saw a weird unidentifiable Javascript crash when running it locally and I kept clicking on the different display counts back and forth, but I only saw it in Chrome, and I never saw it in the demo even with Chrome. Probably nothing to worry about, but in case you see this happen some time, you’ll know I've seen it too.

forresttanaka (approved):
Yeah, looks really good

PR #4304, opened 2022-07-19
ENCM-65 indexing quality metrics

gabdank:


keenangraham (dismissed):

src/encoded/types/analysis.py
```diff
@@ -299,7 +301,26 @@ def superseded_by(self, request, superseded_by):
...
+                },
+                "files": {
+                    "type": "array",
+                    "items": {
+                        "type": "string",
+                        "linkTo": "File"
+                    }
+                },
+                "quality_metric": {
+                    "type": "string",
```
keenangraham, line 320: I think have to return `@id` of quality metric now instead of object (`qm_obj`).

keenangraham (changes requested):

src/encoded/types/analysis.py
gabdank: done

keenangraham (approved):

PR #4308, opened 2022-07-29
ENCM-70 Remove extra body button.

forresttanaka:
Made the two clear-body-map buttons exclusive, instead of one or both.


hitz (approved):
LGTM

PR #4335, opened 2022-11-08
ENCM-84 Make a matrix for Degron experiments

forresttanaka:
This is one of the simpler matrices in that it mostly follows what other matrices do. Even the basic Experiment matrix is more complicated because it has a nested Y and X axes where this one has only a nested X-axis. The important part is in matrix_degron.js. A lot of the other changed files generate data to test the Degron matrix locally.


dougli1sqrd (commented):
The first thing I'm noticing is when I very first go to the Degron page the horizontal items are all squished together: Home page -> Data drop down to Degron link -> squished.

Screen Shot 2022-11-09 at 1 45 03 PM

After a refresh on the page it behaves normally. But I can always reproduce by navigating away, and navigating back to the page (even with just forward and back buttons).
Screen Shot 2022-11-09 at 1 46 33 PM

forresttanaka:
Oh, that’s a Safari thing apparently, which I forgot to test. I’ll need to fix that. Hopefully it’s not difficult.

forresttanaka:
Looks like Firefox has a different issue.

dougli1sqrd:
Should I also see cells filled in at all in that view that I screenshotted?

forresttanaka:
Yes, like it appears in Chrome.

dougli1sqrd:
ah yeah, looks like something is up in firefox then there

forresttanaka:
I have fixes for Safari and Firefox ready, so I’ll push those and whatever else we need once you’re done with code review.

src/encoded/static/components/matrix_degron.js
```diff
@@ -0,0 +1,403 @@
...
+
+
+/**
+ * Compose the query string for a column header.
+ * @param {object} context Matrix search result object
+ * @param {object} colCategoryBucket Bucket for the current column
+ * @param {object} colSubcategoryBucket Bucket for the current subcolumn
+ * @returns {string} Query string for the column header
+ */
+const composeColCategoryQuery = (context, colCategoryBucket, colSubcategoryBucket) => {
```
dougli1sqrd, line 78: How come we say "context" for the search result object? Would another name that just describes the fact that it's the search results be more clear? Or is there a historical reason here?

src/encoded/static/components/matrix_degron.js
```diff
@@ -0,0 +1,403 @@
...
+    const { rowCategoryName } = getMatrixCategoryNames(context);
+    const rowCategoryTitle = rowCategoryBucket.key;
+    return `&${rowCategoryName}=${encodedURIComponent(rowCategoryTitle)}`;
+};
+
+
+/**
+ * Defines the sorting order of the assay/target table columns.
+ */
+const COL_CATEGORY_DATA_SORT_ORDER = [
```
dougli1sqrd, line 105: Why is this the sorting order? Maybe we should comment also how we know this so in case something has to change we know how we generated this order in the first place.

src/encoded/static/components/matrix_degron.js
forresttanaka: “context” refers to any object from the back end. It got that name before I started and by that time it had already become ubiquitous through the code base. Laurence felt some regret calling it that once React also got the concept of a “context.”

src/encoded/static/components/matrix_degron.js
forresttanaka: Ben defined the sorting order. I don’t know why it’s that order.

src/encoded/static/components/matrix_degron.js
dougli1sqrd: ahh yeah okay. Yeah I did think at first it could be related to the react context, but we have that also.

src/encoded/static/components/matrix_degron.js
dougli1sqrd: okay!

dougli1sqrd (approved):
Happy to take a look at your upcoming updates

forresttanaka:
Could you reject this PR ? Then I can put in the updates and re-request your review. _Edit: or “Request changes” rather_.

dougli1sqrd (changes requested):
Approve for what's currently here, awaiting updates from earlier convo

dougli1sqrd:
does the demo get updated? Or should I try running this locally? I guess that's what QA is for too though 🤷

forresttanaka:
I updated the demo and it’s rebuilding right now. I’ll wait for the tests to pass before putting it back into code review.

dougli1sqrd (dismissed):
@forresttanaka this worked!

dougli1sqrd (dismissed):

PR #4348, opened 2023-01-09
ENCM-120 hic-title-update

gabdank:


src/encoded/types/shared_calculated_properties.py
```diff
@@ -379,7 +379,7 @@ class CalculatedAssayTitle:
         "type": "string",
     })
     def assay_title(self, request, registry, assay_term_name,
-                    control_type=None, target=None, documents=None):
+                    control_type=None, replicates=None, target=None, documents=None):
```
ian-whaling, line 382: I think you could remove 'documents=None' here as it was added for the calculation by document but not necessary. Looks good otherwise.

ian-whaling (approved):

PR #4350, opened 2023-01-12
ENCM-103 Implement last requested changes to the mouse development page

forresttanaka:
## PR Notes
* Moved `sortMouseArray()` from matrix_mouse_development.js to dataset.js for the organism development series table which still needs it. matrix_mouse_development.js has a new sorting function that I didn’t want to test against the organism development series.
* The scroll indicators haven’t worked on matrix pages in a long time. Since no one has complained, I removed the code for that.

## Commit Log

* Sort months and years properly for donor ages.
* Correction to last commit.
* Try a new row-sorting method. Locally I now get no displayed data, so I’m not sure this is correct. But try this in the demo.
* Correct last commit. I was not converting the normalized age objects back to their original strings.
* Move `sortMouseArray` to dataset.js now that we don’t use it in the mouse development matrix. Remove the code to handle the scroll indicators as it hasn’t worked for a long time anyway.
* Add browser tooltips to each cell.
* Remove “Histone” from column titles in the browser tooltips you get when hovering over cells.
* Update formatting and comments.
* Try to get Chrome to install for CircleCI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-    browser-tools: circleci/browser-tools@1.0.1
+    browser-tools: circleci/browser-tools@1.4.0
 
 executors:
   encoded-executor:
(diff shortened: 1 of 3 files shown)


emro (approved):
Looks great

PR #4351, opened 2023-01-12
ENCM-111 fix nerve button on body map

emro:
Adding in a selected color for the nerves


forresttanaka (approved):
Looks like none of the other human and mouse tissues have the same issue.

PR #4354, opened 2023-01-14
ENCM-118 add ENSEMBL V87

gabdank:


jenjou (changes requested):
The File changelog should be updated with the new enum.

jenjou (approved):

PR #4355, opened 2023-01-18
ENCM-119 fix organism dev error

emro:
Eliminates an error for rows that are missing synchronization data and accounts for age stages with units of hours and ranges of hours.

### DIFF
--- a/src/encoded/static/components/dataset.js
+++ b/src/encoded/static/components/dataset.js
@@ -1940,7 +1940,8 @@ const differentiationTableColumnsWithoutTime = {
 // In the future there are likely to be additions to the data which will require updates to this function
 // For instance, ages measured by months will likely be added
 function computeAgeComparator(ageStage) {
-    const splitAge = ageStage.split(/[ ,]+/);
+    const matchRegex = /(.+) ([0-9.]+(?:-\d|.+)?) (.+)/;
+    const splitAge = ageStage.match(matchRegex);
     const age = `${splitAge[splitAge.length - 2]} ${splitAge[splitAge.length - 1]}`;
     let ageNumerical = 0;
     if (age.includes('hours') && age.includes('-')) {
--- a/src/encoded/static/components/dataset.js
+++ b/src/encoded/static/components/dataset.js
@@ -1940,9 +1940,14 @@ const differentiationTableColumnsWithoutTime = {
 // In the future there are likely to be additions to the data which will require updates to this function
 // For instance, ages measured by months will likely be added
 function computeAgeComparator(ageStage) {
-    const age = ageStage.split(/ (.+)/)[1];
+    const splitAge = ageStage.split(/[ ,]+/);
+    const age = `${splitAge[splitAge.length - 2]} ${splitAge[splitAge.length - 1]}`;
     let ageNumerical = 0;
-    if (age.includes('days')) {
+    if (age.includes('hours') && age.includes('-')) {
+        ageNumerical += +age.split('hours')[0].split('-')[1] / 24;
+    } else if (age.includes('hours')) {
+        ageNumerical += +age.split('hours')[0] / 24;
+    } else if (age.includes('days')) {
         ageNumerical += +age.split('days')[0];
     } else if (age.includes('weeks')) {
         ageNumerical += +age.split('weeks')[0] * 7;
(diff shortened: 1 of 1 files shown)


src/encoded/static/components/dataset.js
```diff
@@ -1940,9 +1940,14 @@ const differentiationTableColumnsWithoutTime = {
 // In the future there are likely to be additions to the data which will require updates to this function
 // For instance, ages measured by months will likely be added
 function computeAgeComparator(ageStage) {
-    const age = ageStage.split(/ (.+)/)[1];
+    const splitAge = ageStage.split(/[ ,]+/);
+    const age = `${splitAge[splitAge.length - 2]} ${splitAge[splitAge.length - 1]}`;
```
forresttanaka, line 1945: These splits are hard to decode and maintain IMO. This regex seems to work for everything I could find in the existing data using string.match():

(.+) ([0-9\.]+(?:-\d|.+)?) (.+)

Then you always get the life stage in the first group, the age or age range in the second, and the age units in the third. I tried it on the weirder cases I could find:

```
“adult 61 years”: adult, 61, years
“L4 larva 49 hours”: L4 larva, 49, hours
“L4/young adult 51.0 hours”: L4/young adult, 51.0, hours
“young adult 61 hours”: young adult, 61, hours
“adult 90 or above years”: adult, 90, or above years
“embryonic 0-24 hours”: embryonic, 0-24, hours
“adult 12.00 weeks”: adult, 12.00, weeks
“embryonic 0-1.5 hours”: embryonic, 0-1.5, hours
“late embryonic 2.5 hours”, late embryonic, 2.5, hours
“mixed stage (embryonic) 2 hours”: mixed stage (embryonic), 2, hours
“mixed stage (embryonic) 2.0 hours”: mixed stage (embryonic), 2.0, hours
“third instar larva 12 hours”: third instar larva, 12, hours
“prepupa unknown stages”: no match, so just sort at beginning or end
```
It’s not perfect in that it doesn’t necessarily reject completely crazy things (e.g. an age of “12.0.3”), but it doesn’t need to. I didn’t see any `life_stage_age` with a comma in it, so if those exist and I missed them, then this regex doesn’t work.

This means you might have to change how you process age ranges. Given that encoded is in maintenance mode, I don’t know if that’s worth it if this existing code works for some of the weird cases.

src/encoded/static/components/dataset.js
```diff
@@ -2209,7 +2226,12 @@ const organismDevelopmentSeriesWormFlyTableColumns = {
             const synch = computeSynch(experiment);
             return <span>{synch}</span>;
         },
-        objSorter: (a, b) => sortMouseAge(a.life_stage_age, b.life_stage_age),
+        objSorter: (a, b) => {
+            if (a.life_stage_age && b.life_stage_age) {
+                return sortMouseAge(a.life_stage_age, b.life_stage_age);
+            }
+            return null;
+        },
```
forresttanaka, line 2237: Should be able to hide these columns without _too_ much. In this case, adding this property to the `synch` column probably works, depending on what `computeSync` returns if no synchronization information exists:
```
hide: (experiments) => (
    experiments.some((experiment) => computeSynch(experiment))
),
```
The `postsynctime` column is more difficult, but putting the evaluation portion of its `display` callback into a function and calling that in the `hide` callback wouldn’t be too bad. I don’t remember from the meeting how high priority hiding these columns was though, so it might not be worth it anyway;

forresttanaka (changes requested):
Requesting changes, but the first comment causes changes a little outside your current changes, and the second depends on how important hiding columns is. So it could be argued that both comments cause more risk than we want right now.

src/encoded/static/components/dataset.js
emro: Ok, thank you! I'll look at this.

I actually meant to comment, this addresses another bug that I noticed while I was doing this ticket, but it's a different bug than the original ticket. The synchronization times aren't sorted correctly here: https://www.encodeproject.org/organism-development-series/ENCSR085DYI/ but they are here: https://encm-119-8ce1b9e88-emma.demo.encodedcc.org/organism-development-series/ENCSR085DYI/. I'll make a note on Jira as well.

emro: This is a great solution! I implemented it and I think it's working well.

src/encoded/static/components/dataset.js
emro: Thank you for suggesting a solution! I tried the fix that you suggested and also tried updating computeSynch() to see if having it return null instead of '' would hide the synchronization column, and I'm not sure why not but it didn't work. Ben said that this is 0 priority though, so I didn't try to track down another solution.

forresttanaka (dismissed):
I was thinking of keeping the age and age units separate so that you don’t have to use things like .includes(), but that would cause quite a few changes to long-existing code below there, so makes sense to keep what you have. This function would be a good candidate for a Jest test so you can throw all kinds of crazy stuff at it, though no real reason to at this stage of encoded.

emro:
That's a great idea, we should do it like that in future projects for sure.

forresttanaka (approved):
Not sure how to find potential other units that might have ranges, but hopefully we don’t have any.

PR #4357, opened 2023-01-20
ENCM-114 mixed genome annotation crashes

emro:
Prevents crash on mixed annotations and sorts mixed annotations between numbered annotations and "-minimal" annotations

### DIFF
--- a/src/encoded/static/components/objectutils.js
+++ b/src/encoded/static/components/objectutils.js
@@ -925,12 +925,13 @@ export function computeAssemblyAnnotationValue(assembly, annotation) {
     // First level of sorting: most recent assemblies are ordered first (represented by numerical component of assembly)
     // Second level of sorting: assemblies without '-minimal' are sorted before assemblies with '-minimal' at the end (represented by tenths place value which is 5 if there is no '-minimal')
     // Third level of sorting: Annotations within an assembly are ordered with most recent first, with more recent annotations having a higher annotation number (with the exception of "ENSEMBL V65") (represented by the annotation number divided by 10,000, or, the three decimal places after the tenths place)
+    // If an annotation is mixed, it will be ordered between annotations with annotation numbers and '-minimal' annotations
     let assemblyNumber = /\d/.test(assembly) ? +assembly.match(/[0-9]+/g)[0] : 0;
     if (assembly?.indexOf('minimal') === -1) {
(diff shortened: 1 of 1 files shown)


src/encoded/static/components/objectutils.js
```diff
@@ -925,12 +925,13 @@ export function computeAssemblyAnnotationValue(assembly, annotation) {
...
     // Second level of sorting: assemblies without '-minimal' are sorted before assemblies with '-minimal' at the end (represented by tenths place value which is 5 if there is no '-minimal')
     // Third level of sorting: Annotations within an assembly are ordered with most recent first, with more recent annotations having a higher annotation number (with the exception of "ENSEMBL V65") (represented by the annotation number divided by 10,000, or, the three decimal places after the tenths place)
+    // If an annotation is mixed, it will be ordered between annotations with annotation numbers and '-minimal' annotations
     let assemblyNumber = /\d/.test(assembly) ? +assembly.match(/[0-9]+/g)[0] : 0;
     if (assembly?.indexOf('minimal') === -1) {
         // If there is no '-minimal', add 0.5 which will order this assembly ahead of any assembly with '-minimal' and the same numerical component
         assemblyNumber += 0.5;
     }
-    if (annotation) {
+    if (annotation && annotation !== 'mixed') {
```
forresttanaka, line 934: I looked through all possible annotations and “mixed” is the only one without digits. But another way to handle this is to pass every non-empty annotation to `annotation.match(/[0-9]+/g)` and bail if it returns null. But if it returns something real _then_ get element [0] and put it through the algorithm you have. That way if another annotation without digits occurs in the future, this code will still handle it, not just “mixed.” But I’m betting the chances of that are quite close to zero.

forresttanaka (approved):
One possible change, but I doubt it would ever become relevant.

forresttanaka:
Oh, maybe you’re still making changes.

src/encoded/static/components/objectutils.js
emro: That is a great alternative solution! It would handle other cases elegantly. I checked with Meenakshi and she thinks it's very doubtful that we'd have new annotations without digits so I think I will leave this as is for now, since Meenakshi already QAed.

emro:
I added a comment and got rid of a linting error (extra trailing space) and squashed those commits so I think you've code reviewed the final changes.

PR #4361, opened 2023-01-26
ENCM-88 fix crispr screens browser locations

emro:
Added in a function to check related_datasets for locations

### DIFF
--- a/src/encoded/static/components/objectutils.js
+++ b/src/encoded/static/components/objectutils.js
@@ -40,6 +40,22 @@ DisplayAsJson.contextTypes = {
     location_href: PropTypes.string,
 };
 
+export function findExaminedLociLocation(context, assembly) {
+    let examinedLociLocation = null;
+    if (context.examined_loci.length > 0) {
+        context.examined_loci.some((examinedLoci) => {
+            if (examinedLoci.gene?.locations?.length > 0) {
+                examinedLociLocation = examinedLoci.gene.locations.find((elem) => elem.assembly === assembly);
+                if (examinedLociLocation) {
+                    return examinedLociLocation;
+                }
+            }
+            return examinedLociLocation;
+        });
+    }
+    return examinedLociLocation;
+}
+
 export function findRelatedLocation(context, assembly) {
     let relatedDatasetsLocation = null;
     if (context.related_datasets.length > 0) {
--- a/src/encoded/static/components/objectutils.js
+++ b/src/encoded/static/components/objectutils.js
@@ -40,6 +40,27 @@ DisplayAsJson.contextTypes = {
     location_href: PropTypes.string,
 };
 
+export function findRelatedLocation(context, assembly) {
+    let relatedDatasetsLocation = null;
+    if (context.related_datasets) {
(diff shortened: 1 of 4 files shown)


src/encoded/static/components/objectutils.js
```diff
@@ -40,6 +40,27 @@ DisplayAsJson.contextTypes = {
...
+                                return relatedDatasetsLocation;
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+    return null;
+}
```
forresttanaka, line 85: I feel like this isn’t easy to maintain, and it potentially executes the same `.find()` loop twice. I didn’t test whether this works, but I think this is a lot easier and has a lot less duplication:
```
export function findRelatedLocation(context, assembly) {
    if (context.related_datasets.length > 0) {
        for (const relatedDataset of context.related_datasets) {
            if (related_dataset.elements_references?.length > 0) {
                for (const elementsReference of relatedDataset.elements_references) {
                    if (elementsReference.examined_loci?.length > 0) {
                        for (const examinedLoci of elementsReference.examined_loci) {
                            if (examinedLoci.locations?.length > 0) {
                                const relatedDatasetsLocation = examinedLoci.find(
                                    (elem) => elem.assembly === assembly
                                );
                                if (relatedDatasetsLocation) {
                                    return relatedDatasetsLocation;
                                }
                            }
                        }
                    }
                }
            }
        }
    }
    return null;
}
```
What’s your rule for deciding whether to use a function declaration vs a function expression? I think Phil asked you that once but I didn’t pay attention to the conversation after that.

forresttanaka (dismissed):
I included an alternative way to write `findRelatedLocation` but I didn’t test whether it functions exactly the same as yours. It just seems easier to maintain. This is a good case for a Jest test since it’s not a simple function.

src/encoded/static/components/objectutils.js
emro: This is a great way to write the function, thank you for suggesting it! I agree that executing the find() just once is much better and I like the briefer syntax. I wrote this function the way that most of other functions were written in objectutils but if the other way would be better I can change it!

emro: I got an error when I tried this approach:

> iterators/generators require regenerator-runtime, which is too heavyweight for this guide to allow them. Separately, loops should be avoided in favor of array iterations  no-restricted-syntax

So I tried using .some() because maybe that is better than the regular for loop.
Let me know what you think!

forresttanaka: Interesting. It sounds like that error is from an AirBnB rule that people [debate](https://github.com/airbnb/javascript/issues/1271) the value of. But if what you have works, might as well stick with that. Maybe AirBnB has the right idea for reasons I’m not aware of.

forresttanaka (dismissed):

emro:
I realized I hadn't added a test and that was a good idea so I'm moving back to code review.

forresttanaka (dismissed):
OK, thanks for adding the test.

forresttanaka (dismissed):

forresttanaka (approved):

PR #4364, opened 2023-02-09
ENCM-115 Add functional characterization matrix

forresttanaka:
* The main addition exists in the matrix_functional_characterization.js file which mostly follows the pattern of the simpler matrix-rendering files.
* I changed `<MatrixAddCart>` to only request the `@id` and `@type` of the related datasets to reduce the size of the results, in some cases very radically. Those two properties are the only ones that the cart code uses.
* I changed `<SearchFilter>` to allow matrix pages to optionally pass the specific title they want for the search-box label.


forresttanaka:
@emro Ugh. I left an old URL in the menu. It should be identical to the URL in the FC home-page button. I’ll push an update to that momentarily. That’s the only change.

emro (dismissed):
Looks awesome! I like the colors and it works well.

src/encoded/search_views.py
```diff
@@ -38,6 +38,7 @@
 from snosearch.fields import AllResponseField
 from snosearch.fields import BasicMatrixWithFacetsResponseField
 from snosearch.fields import MissingMatrixWithFacetsResponseField
+from snosearch.fields import MultitypeMissingMatrixWithFacetsResponseField
```
emro, line 41: This is almost not worth noting but I might have done the cases like this: MultiTypeMissingMatrixWithFacetsResponseField

emro (approved):
Looks great!

PR #4365, opened 2023-02-10
ENCM-125-google-analytics-4

mingjiecn:


forresttanaka (approved):
I think an alternative would be to put the Javascript that’s inside the dangerouslySetInnerHTML into the `componentDidMount()` method, as straight Javascript goes in here on page mount. But if this works, that’s fine. As an aside, we don’t use `componentDidMount()` in new code, but the `useEffect()` hook instead. You’ll find that in newer encoded and all igvf-ui code.

PR #4367, opened 2023-02-15
ENCM-116 Add disease series tab to the functional genomics series page

forresttanaka:


emro (commented):
This looks great! I love the new icon! One minor comment: I think that $series-wrap-width should be changed, line 1 on _tabs.scss. The tabs wrap at a new width now and you can get it into a funny state (though I think this is a non-standard width and no one is actually likely to see it on mobile or desktop or tablet).

Screen Shot 2023-02-15 at 10 11 12 AM

forresttanaka:
OK, changing `$series-wrap-width` made it behave strangely, but changing the flex basis worked better. A shame that the Disease button wraps, but now with an odd number of tabs, I don’t know what else we can do. This change affects several other pages too, but none of them seemed to be damaged by this.

forresttanaka:
I updated the demo too.

emro (approved):
It looks great! Thank you!

PR #4371, opened 2023-02-19
ENCM-130-reg-internal-tag

meenakshikagda:


src/encoded/schemas/changelogs/experiment.md
```diff
@@ -1,5 +1,8 @@
 ## Changelog for experiment.json
 
+### Minor changes since schema version 37
```
jenjou: Since Experiment is still on version 36, I think this entry can be moved below "Minor changes since schema version 36" (after "added 'degron'...)

jenjou (changes requested):
Minor change in changelog for Experiments

src/encoded/schemas/changelogs/experiment.md
meenakshikagda: Ah thanks for catching that :) I missed it totally

jenjou (approved):

PR #4375, opened 2023-02-28
ENCM-92-organoid-search-links

khzlin:


src/encoded/tests/test_search_views.py
```diff
@@ -281,7 +281,8 @@ def test_search_views_search_view_columns_from_configs(index_workbook, testapp,
     search_registry.add(config1)
     search_registry.add(config2)
     r = testapp.get('/search/?type=Experiment')
-    assert 30 < len(r.json['columns']) < 40
+    print(len(r.json['columns']))
```
jenjou: Extra line

src/encoded/types/biosample.py
```diff
@@ -1040,6 +1042,19 @@ def generate_summary_dictionary(
             phrase += ' (' + dict_of_phrases['age_display'] + ')'
         dict_of_phrases['sex_stage_age'] = phrase
 
+    if ('sample_type' in dict_of_phrases and
+        dict_of_phrases['sample_type'] == 'organoid'):
+        phrase = ''
+        if 'sample_term_name' in dict_of_phrases:
+            phrase += dict_of_phrases['sample_term_name']
+
+        if 'sample_type' in dict_of_phrases:
```
jenjou: I think you could omit this check since on lines 1045-1046 there is already a check for `sample_type` -- so it will always be true that dict_of_phrase['sample_type'] exists and == 'organoid'

jenjou (changes requested):
Some minor comments but overall looks OK.

src/encoded/tests/test_search_views.py
khzlin: ok

jenjou (approved):

PR #4394, opened 2023-07-19
ENCM-160-snrna-seq-assay-title

jenjou:


src/encoded/types/shared_calculated_properties.py
```diff
@@ -387,6 +387,19 @@ def assay_title(self, request, registry, assay_term_name,
                                                                      assay_term_name)
```
gabdank, line 387: can you please add test for the calculation?

gabdank (changes requested):
please add test

gabdank (approved):

PR #4410, opened 2023-11-20
ENCM-152-publication-regex

jenjou:


ian-whaling (changes requested):
Could you add an insert to test that ":" is now accepted.

ian-whaling (approved):

